PR #27, opened 2019-12-30
Polish topic quick infos a litte bit

holgeradam:
A user of this lovely tool suggested this, so I created this PR to improve topic quick infos a little bit.

A topic can have one of three possible cleanup policies: delete, compact and a combination of both.
Depending on the policy some topic configuration entries are important while others are irrelevant.

This PR aims at displaying the relevant configs for the different policies in a defined order. Adding or removing favorites should still be working.

Let me know what you think. I might have broken some conventions or made React rookie mistakes. ;)

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -11,8 +11,6 @@ require (
 	github.com/go-chi/chi v4.0.2+incompatible
 	github.com/gorilla/schema v1.1.0
 	github.com/kafka-owl/common v0.1.0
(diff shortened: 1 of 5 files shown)


weeco:
Makes sense to me. I requested the technical review from @rikimaru0345

holgeradam:
Thanks. :)

I tested all variants of policies with adding and removing other favorites. It works as expected, as far as I can tell.

Issue #30, opened 2020-01-19
Rework "Copy" Button


rikimaru0345:
## The copy button sucks, look at it:
![copy button](https://i.imgur.com/0pfodzI.png)

## Issues with it

- You don't really know what you're copying...
  There should be multiple options:
  - Key
  - Value (as nicely formatted JSON)
  - Value (raw data)
  - Timestamp (in UTC seconds) 

- It takes up too much space!
  A whole column (even if its small) makes it more confusing to look at.
  - Maybe a context (rightclick) menu? Nahhh, that'd be the only use of a ctx menu, thus confusing UX
  - Maybe a button that only appears when the cursor is inside the row? Would make it visually less cluttered, but we'd still have to reserve the horizontal space for where the button will appear. (unless we want to re-layout, which is insanely bad lol)
  - Maybe make the copy button itself smaller (like the "expand message '+' icon"), and then put it inline as well (but where? into the 'value' column, together with the expand plus??)

rikimaru0345:
right now, the copy button has already been improved a little bit (no more text, only icon now, and only shown when hovering over the line).

but clicking it still copies just that one thing (msg content as json).

- instead of immediately copying something, it could show a popup/context menu when clicked

- ((not sure about this one)) however that'd turn copying something into 2 clicks, so maybe add a "dropdown arrow" next to the copy icon as well to change what it copies?  not sure how important it is to keep copying a single click from an UX standpoint.  One could probably argue that it even hurts usability because now there are two "buttons" the user can click :/

PR #77, opened 2020-07-25
#30 Adding more copy options on the row

rjmasikome:
So I added the more options on copy. I think from the UX capability it's probably fine to do it in two clicks. As it is also not clear what they want to copy just from one click in my opinion. I changed the icon to triple dots, as for the usability perspective it should mean there should be more options after clicking. But if you guys are not up for it, just changed for another icon.

Anyway if anything to change lemme know.

Cheers,
RJ

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages.tsx
@@ -1,7 +1,7 @@
 import { Component, ReactNode } from "react";
 import React from "react";
 import { TopicDetail, TopicMessage } from "../../../state/restInterfaces";
-import { Table, Tooltip, Row, Popover, Button, Select, Input, Typography, message, Tag, Alert, Empty, ConfigProvider, Modal, AutoComplete, Space, Switch } from "antd";
+import { Table, Tooltip, Row, Popover, Button, Select, Input, Typography, message, Tag, Alert, Empty, ConfigProvider, Modal, AutoComplete, Space, Switch, Menu, Dropdown } from "antd";
 import { observer } from "mobx-react";
 import { api } from "../../../state/backendApi";
 import { PreviewTag, TopicOffsetOrigin, FilterEntry, ColumnList } from "../../../state/ui";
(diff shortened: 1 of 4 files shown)


weeco:
Fixes #30

frontend/src/components/pages/topics/Tab.Messages.tsx
```diff
@@ -405,14 +424,19 @@ export class TopicMessageView extends Component<{ topic: TopicDetail }> {
                 onFilterDropdownVisibleChange: (_) => this.showColumnSettings = true,
                 filterIcon: (_) => {
                     return <Tooltip title='Column Settings' mouseEnterDelay={0.1}>
-                        <SettingFilled />
+                        <SettingFilled style={IsColumnSettingsEnabled ? { color: '#1890ff' } : { color : '#a092a0' }}/>
                     </Tooltip>
                 },
                 render: (text, record) => !record.isValueNull && (
                     <span>
                         <ZeroSizeWrapper width={32} height={0}>
```
rikimaru0345, line 432: 1. Icon is not vertically centered (but should be):
   


2. Clickable region of the button should fill the whole table-cell; remove padding in those (msgTableActionColumn) cells

    green: currently clickable
    red: should be clickable

rikimaru0345: I'd probably do something like this:
- remove usage of `<ZeroSizeWrapper>` here
- `.ant-table.ant-table-middle .ant-table-tbody>tr>td.msgTableActionColumn { padding: 0 }`
- ensure button fills the whole parent
- if the alignment of the icon inside the button is an issue, you can try making the button a flex container: something like `<Button style={{display:'flex', placeContent:'center'}} ...` or so

frontend/src/index.css
```diff
@@ -663,18 +667,25 @@ tr:hover button.ant-btn-link {
...
 .ant-table-middle>.ant-table-content>.ant-table-body>table>.ant-table-tbody>tr>td.msgTableActionColumn {
     padding-left: 0;
 }
 
-tr>td.msgTableActionColumn {
+.ant-table.ant-table-middle .ant-table-tbody>tr>td.msgTableActionColumn {
+    padding-bottom: 0;
+}
+
+tr>td {
```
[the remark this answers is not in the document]
rikimaru0345: actually, I just remembered that neither solution is necessary 😄 😅 
we can probably just set `align:'right'` in the column definitions (where column name, data index etc is set)!!

PR #82, opened 2020-07-26
Add live tail support

weeco:


weeco:
Affected by https://github.com/cloudhut/kowl/issues/81

PR #120, opened 2020-08-23
Encode the filter code with hex encoding before sending to backend

rjmasikome:
Hello everyone,

As per title. I encoded the special character on the message filter with hex encoding, before it was pushed to the backend.
Lemme know if anything needed to be changed.

Cheers,
RJ


rikimaru0345:
Did you test it with a filter that uses `öäüß` ?

rjmasikome:
> Did you test it with a filter that uses `öäüß` ?

Yes. I tried the filter and it works. For example, this is the filter that I tried that works.
```
if (value.payload.title === "Hähnchen und Gemüse") return true
```

PR #127, opened 2020-09-06
Add endpoints for schema registry UI

### LINKED ISSUE #126: Add UI for schema registry
Add an user interface to list all registered schemas and inspect how the schema looks like.

weeco:
This PR closes #126


frontend/src/components/pages/schemas/Schema.List.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { api } from '../../../state/backendApi';
+import { Row, Statistic, Table } from 'antd';
+import Card from '../../misc/Card';
+import { appGlobal } from '../../../state/appGlobal';
+
+@observer
+class SchemaList extends PageComponent<{}> {
+    initPage(p: PageInitHelper): void {
+        p.title = 'Scheme Registry';
+        p.addBreadcrumb('Schema Registry', '/schemas');
```
rikimaru0345, line 13: Shouldn't the path be the same as in routes.tsx:194 ?

```
MakeRoute<{}>('/schema-registry', SchemaList, 'Schema Registry'),
```

rikimaru0345 (commented):
👍 so far 😄

frontend/src/components/misc/SearchBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    render() {
+        return <div style={{ marginBottom: '.5rem', padding: '0', whiteSpace: 'nowrap', display: 'flex', alignItems: 'center' }}>
+            {/* <AutoComplete placeholder='Quick Search' size='large'
+                style={{ width: 'auto', padding: '0' }}
+                onChange={v => this.filteredSource.filterText = String(v)}
+                dataSource={['battle-logs', 'customer', 'asdfg', 'kafka', 'some word']}
+            > */}
+            <Input allowClear={true} placeholder='Quick Search' size='large' style={{ width: '350px' }}
+                onChange={e => this.filteredSource.filterText = uiSettings.topicList.quickSearch = e.target.value}
```
rikimaru0345, line 42: Careful with this. This is a remnant that needs to be cleaned up first.
Every instance of a SearchBar component should obviously have its own place where it saves its current value 😅  
Currently the variable is hardcoded (uiSettings.topicList.quickSearch).

The most obvious idiomatic solution would be to make the SearchBar "controlled" (in React lingo). Meaning that it would expose `value` and `onChange` (just like `<Input .../>` does), and then the caller (TopicList etc) would wire those two things up to the correct variable.

frontend/src/components/pages/schemas/Schema.List.tsx
```diff
@@ -29,19 +32,31 @@ function renderRequestErrors(requestErrors?: SchemaOverviewRequestError[]) {
 
 @observer
 class SchemaList extends PageComponent<{}> {
+    paginationConfig = makePaginationConfig(uiSettings.schemaList.pageSize);
+    @observable searchBar: RefObject<SearchBar<SchemaSubject>> = React.createRef();
```
rikimaru0345, line 36: Oh no, shame on me 😹 😹 😹, I think I recognize where that was copied from... 
 

### Background
If I remember correctly this code is from 'Topic.List' (I think?). I wrote this at a time where I wasn't aware of the best practices of passing data between components in react.

The issue here is that we're using the SearchBar as an "object" which exposes some observable list it procudes. This goes counter to the idea of react (hierarchy of components that pass data (and callback functions) down).
The approach obviously works, but it is not idiomatic, and requires some ugly code (such as needing to create a ref...).

### Solution 1
Make it so we pass in (1) the source data to be filtered and (2) a function that the searchbar calls once it has a new "output" (produced after a debounced reaction to the user typing stuff in the input field).
The handler we'd pass in would then save the filtered data to a local field (in class SchemaList), which is marked as observable.


### Solution 2
This is known as "function as children". It's an approach that I try to avoid when possible / or unless its an exceptional situation (like maybe routing definitions) but I'll include it anyway since I've seen it used in quite a few other projects already.  
This solution would make it so the SearchBar would take a function as its `children`.
The usage would then look something like this: 

```tsx
<SearchBar sourceData={...}  ...>
{ filteredData =>
    <Table ... rows={filteredData} .../>
}
</SearchBar>
```
I really don't like this solution, but it is at least a bit more idiomatic than the "createRef" thing I wrote a few months ago. 😅😅😅 

---

I think it's worth fixing this (I'd prefer solution 1). 
Currently some (most?) other pages each have their own implementation of the search bars (because of this issue, and the "where the value is saved" thing I mentioned earlier in a different comment), which is obviously not a good practice at all.  

I think right now, in this PR, we shouldn't start a larger refactoring, like replacing all the custom SearchBars with the shared SearchBar component. 

My suggestion would be that we should fix the API of the SearchBar (and of course update Schema.List and Topic.List accordingly); and at some later point come back to all the other SearchBars and update those as well (making those spots use the shared SearchBar component, and removing the custom implementation).

---

But most importantly, what do **you** think about this? 😄 
Do you agree / not agree? Is that change worth it? Would you go further (so updating all the locations where the SearchBar *should* be used)? Or maybe not change anything at all?

nilsroehrig: > If I remember correctly this code is from 'Topic.List' (I think?)
Yes, that's the one.

I agree with you in regards to fixing the shared componen now. A larger refactoring could be done well enough at a future date, after we have a more appropriate solution in place. I think thats a more pragmatic approach in order to deliver the new feature sooner. That's why I haven't gone further searching for other occurences of the search bar componen
From your two propositions I also agree that #1 is the better one. The second one to me - if understood correctly - seems to break separation of concerns in regards to the presentation of the filtered result which, in my opinion,  actually does not have anything to do with the core functionality of the search bar. They should be separated as clean as possible.

frontend/src/components/misc/SearchBar.tsx
nilsroehrig: Yes, you're obviously right. I missed that one while extracting. Gonna go fix it.

nilsroehrig: Please have another look, I moved on with your suggestion. I think I did not miss anything this time.

frontend/src/components/pages/schemas/Schema.List.tsx
nilsroehrig: @rikimaru0345 Care to take a look into this together? I tried a few things but did not come very far due to my infamiliarity with mobx I guess. I'd love to see how you'd approach this and maybe learn a thing or two.

Issue #132, opened 2020-09-27
Topics details can't be shown if one or more brokers are unavailable


weeco:
If one broker is not available (e.g. due to a restart) the topic details view is not available because Kowl can't fetch the watermarks from the respective broker. Reason is most likely the outdated cache with the metadata which holds information about what Broker is responsible for each topic.

Issue #102, opened 2020-08-04
Describing Consumer Groups fails after new leader election


weeco:
I noticed that after a broker in my cluster has failed Kowl wasn't able to describe the consumer groups. Reason for this was a wrong cached coordinator id for a consumer group id. Sarama caches the coordinators for each consumer group and at first glance it looks like these cache entries do not expire. Thus it seems like it is our responsibility to handle this by calling `RefreshCoordinator(groupID)` once we know the Coordinator is no longer up to date for a consumer group.

This has to happen before the requests are grouped by coordinator id here: https://github.com/cloudhut/kowl/blob/master/backend/pkg/kafka/describe_consumer_groups.go#L18 or requests must be reissued on outdated coordinator errors.

weeco:
Besides refreshing outdated group coordinators I identified another problem: We sent a request to all brokers to ask for their list of consumer groups. Currently Kowl handles the request so that if a single broker response fails it will also return an error. For larger clusters this can be annoying as broker restarts/failures are fairly common.

Therefore I propose to return a list of errors during that request and still present the consumer group list on a best effort basis then. Usually the group coordinator will be reelected quickly so that the rest of the cluster can actually still return the full list of the consumer groups.

PR #133, opened 2020-10-03
Fix consumer group requests

weeco:
This PR addresses issue #132 and #102 

Especially #102 is resolved on a best effort here as some requests may need to be triggered a second time. Reason for this is that Sarama utilizes several caches which we purge after a request has failed. If this is the case we do not retry the whole request chain as it would likely timeout anyways.

I intend to address this by potentially migrating to a library that exposes more core level functionality. https://github.com/twmb/kafka-go seems to be a good condidate to evaluate for this.

### DIFF
--- a/backend/pkg/kafka/connection_helper.go
+++ b/backend/pkg/kafka/connection_helper.go
@@ -23,8 +23,8 @@ func NewSaramaConfig(cfg *Config) (*sarama.Config, error) {
 	}
 	sConfig.ClientID = cfg.ClientID
 	sConfig.Version = version
-	sConfig.Net.KeepAlive = 30 * time.Second
-	sConfig.Net.DialTimeout = 10 * time.Second
+	sConfig.Net.KeepAlive = 15 * time.Second
+	sConfig.Net.DialTimeout = 15 * time.Second
 	sConfig.Net.ReadTimeout = 15 * time.Second
 	sConfig.Net.WriteTimeout = 15 * time.Second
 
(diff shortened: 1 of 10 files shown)


backend/pkg/kafka/connection_helper.go
```diff
@@ -23,8 +23,8 @@ func NewSaramaConfig(cfg *Config) (*sarama.Config, error) {
 	}
 	sConfig.ClientID = cfg.ClientID
 	sConfig.Version = version
-	sConfig.Net.KeepAlive = 30 * time.Second
-	sConfig.Net.DialTimeout = 10 * time.Second
+	sConfig.Net.KeepAlive = 15 * time.Second
```
rikimaru0345, line 26: KeepAlive will eventually end up here: https://godoc.org/net#Dialer.KeepAlive
Does it make sense to have the same value for KeepAlive as we have for Read/Write Timeout?
Shouldn't we send KeepAlive more often? Or am I misinterpreting how it works?

rikimaru0345 (approved):

backend/pkg/kafka/connection_helper.go
weeco: The KeepAlive is used for the TCP connection while the Dial, Read and Write timeout are not relevant for keeping the TCP connection alive, but they are used in Sarama for timeouts. Hence they can be all the same. I lowered them to 15s because sarama will wait up to ReadTimeout if a broker is dead (and the cached metadata is not yet updated).

PR #135, opened 2020-10-10
Resolves #23 / Adds endpoint to list ACL resources

### LINKED ISSUE #23: ACL overview
Add a page which shows all ACL rules for each resource type and principal

weeco:
This PR closes #23


weeco:
I changed the endpoint to accept all filter requests sent to Kafka. You can provide the following request parameters now:

```go
type getAclsOverviewRequest struct {
	// The resource type.
	ResourceType int `schema:"resourceType"`

	// The resource name, or null to match any resource name.
	ResourceName *string `schema:"resourceName"`

	// The resource pattern to match.
	ResourcePatternTypeFilter int `schema:"resourcePatternTypeFilter"`

	// The principal to match, or null to match any principal.
	Principal *string `schema:"principal"`

	// The host to match, or null to match any host.
	Host *string `schema:"host"`

	// The operation to match.
	Operation int `schema:"operation"`

	// The permission type to match.
	PermissionType int `schema:"permissionType"`
}
```

Each filter represents an iota, see: https://github.com/Shopify/sarama/blob/master/acl_types.go

Accordingly this request lists all available ACLs: `/api/acls?resourceType=1&resourcePatternTypeFilter=1&operation=1&permissionType=1` . If there's no authorizer configured or no ACLs are stored an empty array will be returned

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -0,0 +1,206 @@
...
+                        {OptionValues.Operations.map(v => <Select.Option key={v.value} value={v.value} >{v.name}</Select.Option>)}
+                    </Select>
+                </Label>
+
+                <Label text='Permission'>
+                    <Select<AclPermissionType> value={state.permissionType} onChange={x => state.permissionType = x} style={{ width: '11em' }} size='middle'>
+                        {OptionValues.Permissions.map(v => <Select.Option key={v.value} value={v.value} >{v.name}</Select.Option>)}
+                    </Select>
+                </Label>
+                */}
```
weeco, line 166: Should this be removed before merging?

rikimaru0345: no

PR #148, opened 2020-11-25
Moved the directory to increase readability

suhyunjeon:
I moved the directory to increase readability. Improved because I thought the novice would be confused.
And updated README.md 

```
1) docker-compose : /docs/local/ => /docker
2) config : /docs/config/ => /config
```


weeco:
I think all of this belongs to the category "documentation" hence we put it into the `/docs` subdirectory. Moving the docker compose files into a `docker` category could be missleading too as I'd expect the actual docker file for building Kowl there.

Potentially we'll be moving all the documentation to a dedicated website (something like GitBooks) where we can have versioned docs etc. Thus we'd like to keep the structure like it is currently. Can you eloborate what exactly confused you since everything is linked? Could we maybe improve on this without moving stuff out of the `/docs` directory?

suhyunjeon:
@weeco ```/docs``` directory is generally a collection of documents and is separated by judging that it is different from the ```docker``` or ```config``` personality.
I don't care if I do not move it, but I thought that separating it by personality conveys a clear meaning to beginners.

weeco:
> /docs directory is generally a collection of documents and is separated by judging that it is different from the docker or config personality.

I agree. All the provided docker-compose count imho towards documentation. The /docs folder does not contain anything about the actually built docker images for Kowl. These files are just meant to help users to get started with Kowl. I understand that this is very opinionated and that you (or others) may prefer a different structure. For us this currently feels right and we don't want to change this as of now.

Therefore I'll go ahead and close this PR. Thanks for your efforts though!

PR #162, opened 2021-01-07
fixed kafka config and some tests

snakeice:


weeco:
Thanks for finding and fixing it along with all the tests and syntax issues! Sorry for that.

PR #169, opened 2021-01-22
Use multiple go routines for decoding messages

weeco:


backend/pkg/kafka/consumer_worker.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/cloudhut/kowl/backend/pkg/proto"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"go.uber.org/zap"
+)
+
+func (s *Service) startMessageWorker(ctx context.Context, isMessageOK isMessageOkFunc, jobs <-chan *kgo.Record, resultsCh chan<- *TopicMessage) {
+	for record := range jobs {
```
twmb, line 12: This can be to be
```go
for {
  var record *kgo.Record
  var ok bool
  select {
  case <-ctx.Done():
    return
  case record, ok = <-jobs:
    if !ok {
      return
    }
  }
  ...
}
```
(or `record, ok :=`, and then move all logic for the loop into the case)
and then the select / default at the end of the loop can be removed.

backend/pkg/kafka/consumer.go
```diff
@@ -80,12 +85,22 @@ func (s *Service) FetchMessages(ctx context.Context, progress IListMessagesProgr
...
-	// 2. Setup JavaScript interpreter
-	isMessageOK, err := s.setupInterpreter(consumeRequest.FilterInterpreterCode)
-	if err != nil {
-		s.Logger.Error("failed to setup interpreter", zap.Error(err))
-		progress.OnError(fmt.Sprintf("failed to setup interpreter: %v", err.Error()))
-		return err
+	// 2. Create consumer workers
+	jobs := make(chan *kgo.Record, 100)
+	resultsCh := make(chan *TopicMessage, 100)
+	defer close(jobs)
```
twmb, line 91: This can cause a panic in `consumeKafkaMessages`. The context is canceled immediately before jobs is closed; if goroutines are scheduled in a pathological enough order, then the select condition in `consumeKafkaMessages` may try sending to a closed channel.

It'd be easier to move the `defer close(jobs)` to the top of `consumeKafkaMessages`.

backend/pkg/kafka/consumer.go
```diff
@@ -99,100 +114,84 @@ func (s *Service) FetchMessages(ctx context.Context, progress IListMessagesProgr
...
+			return
 		default:
 			fetches := client.PollFetches(ctx)
 			errors := fetches.Errors()
 			if len(errors) > 0 {
-				return fmt.Errorf("'%v' errors while fetching records: %w", len(errors), errors[0].Err)
+				s.Logger.Error("errors while fetching records",
+					zap.String("topic_name", errors[0].Topic),
+					zap.Error(errors[0].Err))
+				continue
```
twmb, line 163: There can be errors and _also_ be records to process, so this continue should not be here. Also, there can be more than one error, but only the first error is printed -- is that intentional?

backend/pkg/kafka/consumer.go
```diff
@@ -99,100 +114,84 @@ func (s *Service) FetchMessages(ctx context.Context, progress IListMessagesProgr
...
+		partitionReq := consumeRequest.Partitions[msg.PartitionID]
+		if msg.Offset >= partitionReq.EndOffset {
+			remainingPartitionRequests--
+		}
+
+		// Do we need more messages to satisfy the user request? Return if request is satisfied
+		isRequestSatisfied := messageCount == consumeRequest.MaxMessageCount || remainingPartitionRequests == 0
+		if isRequestSatisfied {
+			// This will stop all the launched go routines for this consume request
+			cancel()
```
twmb, line 143: This may as well move up to be a `defer cancel()` underneath where the cancel is declared.

backend/pkg/kafka/consumer.go
```diff
@@ -80,12 +85,22 @@ func (s *Service) FetchMessages(ctx context.Context, progress IListMessagesProgr
...
-		return err
+	// 2. Create consumer workers
+	jobs := make(chan *kgo.Record, 100)
+	resultsCh := make(chan *TopicMessage, 100)
+	defer close(jobs)
+	workerCtx, cancel := context.WithCancel(ctx)
+
+	for i := 0; i < 4; i++ {
+		// Setup JavaScript interpreter
+		isMessageOK, err := s.setupInterpreter(consumeRequest.FilterInterpreterCode)
```
twmb, line 96: I noticed that `setupInterpreter` compiles the Javascript and then runs it for every record, yet the compilation can be saved and instead the code can be switched to `vm.RunProgram(<saved compiled script>)`. That may be worth pursuing to cut down on some latency / cpu.

backend/pkg/kafka/consumer.go
```diff
@@ -99,100 +114,84 @@ func (s *Service) FetchMessages(ctx context.Context, progress IListMessagesProgr
...
-	// 4. Consume messages and check if filter code accepts the given records
+	// 4. Start go routine that consumes messages from Kafka and produces these records on the jobs channel so that these
+	// can be decoded by our workers.
+	go s.consumeKafkaMessages(workerCtx, client, consumeRequest, jobs)
+
+	// 5. Receive decoded messages until our request is satisfied. Once that's the case we will cancel the context
+	// that propagate to all the launched go routines.
 	messageCount := 0
 	remainingPartitionRequests := len(consumeRequest.Partitions)
+	for msg := range resultsCh {
```
twmb, line 125: It's possible for this to hang here permanently. The `workerCtx` wraps a parent context, and the parent context could be canceled. If the parent context is canceled, the message workers will return, but nothing closes `resultsCh`, and this will hang here.

One way to solve this would be to have a WaitGroup underneath the workerCtx and use that for each started worker like so:
```go
go func() {
  defer wg.Done()
  s.startMessageWorker(workerCtx, isMessageOK, jobs, resultsCh)
}()
```
and then underneath the loop, another goroutine
```go
go func() {
  wg.Wait()
  close(resultsCh)
}()
```
This will ensure that `resultsCh` is closed, which ensures that this loop will stop and allow the function to return.

PR #198, opened 2021-04-02
feat: Add msgpack deserializer

### LINKED ISSUE #197: Support for msgpack deserialization

joeirimpan:
Closes #197


joeirimpan:
For the following struct which was send to kafka
```go
       type Inner struct {
                Field1 string
        }

        type Item struct {
                Foo   string
                Inner Inner
        }

        b, err := msgpack.Marshal(&Item{Foo: "bar", Inner: Inner{"data1"}})
        if err != nil {
                panic(err)
        }
```
image

weeco (approved):
LGTM, thanks for the PR!

weeco:
@joeirimpan One more thing which came to my mind right after merging: Is it possible that a messagepack serialized message is also UTF-8 valid and therefore the deserializer would return the UTF-8 message (step 5) instead of the deserialized msgpack message?

backend/pkg/kafka/deserializer.go
```diff
@@ -200,6 +203,19 @@ func (d *deserializer) deserializePayload(payload []byte, topicName string, reco
 		}, Object: string(payload), RecognizedEncoding: messageEncodingText, Size: len(payload)}
 	}
 
+	// 6. Test for msgp
```
weeco, line 206: Would Step 5 (UTF-8 compatible messagee) potentially return a valid msgpack message? Should it be switched with Step 5?

joeirimpan: UTF-8 encoded msgpack fails at utf8.Valid check, it is successfully getting msgpack unmarshalled. But frontend code fails with below message when message is not json.
image

rikimaru0345:
@joeirimpan I fixed the display of msgpack objects when they encode just some primitive type (string, number, ...) instead of an object.

Please take a look and let me know if it works for you.

weeco:
@joeirimpan I just had a quick discussion with @rikimaru0345 . We both came to the conclusion that it would be wise to add some config properties to explicitly enable msgpack decoding as there's a high potential of conflicts with messages being deserialized as UTF-8 - especially for small messages.

So my suggestion is to add a MessagePack config in the Kafka config (https://github.com/cloudhut/kowl/blob/master/backend/pkg/kafka/config.go#L11-L24):

```go
type Config struct {
  // ....

  // Schema Registry
  Schema   schema.Config `yaml:"schemaRegistry"`
  Protobuf proto.Config  `yaml:"protobuf"`
  MessagePack messagePackCfg `yaml:"messagePack"`

  // ....
}

type MessagePackCfg struct {
  Enabled bool `yaml:"enabled"` // Defaults to false

  // TopicNames is a list of topic names that shall be considered for messagepack decoding.
  // These names can be provided as regex string (e. g. "/.*/" or "/prefix-.*/") or as plain topic name
  // such as "frontend-activities".
  // This defaults to `/.*/`
  TopicNames []string `yaml:"topicNames"`
}

// Validate() and SetDefaults() methods to be added and called from the respective parent config
```



We did something similar with the topic regexes in KMinion: https://github.com/cloudhut/kminion/blob/4f4b8c5f2d3fe7c5a4277558fbc9157a0f7273b5/minion/utils.go#L45-L62
https://github.com/cloudhut/kminion/blob/master/minion/config_topic_config.go#L15-L16

Does that make sense to you?

joeirimpan:
@rikimaru0345 Yes this works.
@weeco I have added this as configurable deserializer as you suggested. Can you take a look?

backend/pkg/msgpack/config.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// Config represents the message pack config.
+type Config struct {
+	Enabled bool `json:"enabled"`
+
+	// TopicNames is a list of topic names that shall be considered for messagepack decoding.
+	// These names can be provided as regex string (e. g. "/.*/" or "/prefix-.*/") or as plain topic name
+	// such as "frontend-activities".
+	// This defaults to `/.*/`
```
weeco, line 14: Either we have to remove this comment line (because right now the default is that enabled msgpack will not be used for any topic names), or we have to add a SetDefaults() method which sets `/.*/` as default. If you choose to use `/.*/` as default the reference config also has to be adapted accordingly; as well as the Validate() method

backend/pkg/msgpack/config.go
```diff
@@ -0,0 +1,37 @@
+package msgpack
+
+import (
+	"fmt"
+)
+
+// Config represents the message pack config.
+type Config struct {
+	Enabled bool `json:"enabled"`
```
weeco, line 9: We use the yaml struct tags everywhere

backend/pkg/msgpack/config.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// TopicNames is a list of topic names that shall be considered for messagepack decoding.
+	// These names can be provided as regex string (e. g. "/.*/" or "/prefix-.*/") or as plain topic name
+	// such as "frontend-activities".
+	// This defaults to `/.*/`
+	TopicNames []string `koanf:"topicNames"`
```
weeco, line 15: We use the yaml struct tags everywhere

weeco (commented):
LGTM, just a few notes

backend/pkg/msgpack/config.go
joeirimpan: ah this was left out. fixed in the new commit

backend/pkg/msgpack/config.go
joeirimpan: Alright. Added the default /.*/ to match all.

weeco (approved):

weeco:
Thanks for submitting this PR 🚀

joeirimpan:
@weeco Thank you for reviewing and accepting and @rikimaru0345 thanks for the quick fix. Appreciate the efforts behind this project 👏

PR #203, opened 2021-04-23
Add schemaregistry TLS authentication

julcollas:
Our SchemaRegistry need TLS authentication. This PR make the ability to send TLS certificate when connecting to it.

### DIFF
--- a/backend/pkg/schema/client.go
+++ b/backend/pkg/schema/client.go
@@ -3,9 +3,11 @@ package schema
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"time"
+	 "net/http"
 
 	"github.com/go-resty/resty/v2"
 )
@@ -42,6 +44,46 @@ func newClient(cfg Config) (*Client, error) {
 	}
 	if cfg.BearerToken != "" {
 		client = client.SetAuthToken(cfg.BearerToken)
+	}
+
+	// Configure Client Certificat transport
+	if cfg.TLS.Enabled {
+
+		cert, err := ioutil.ReadFile(cfg.TLS.CertFilepath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read cert file for schema registry client: %w", err)
+		}
+
+		privateKey, err := ioutil.ReadFile(cfg.TLS.KeyFilepath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read key file for schema registry client: %w", err)
+		}
+
+		pemBlock, _ := pem.Decode(privateKey)
+		if pemBlock == nil {
+			return nil, fmt.Errorf("no valid private key found")
+		}
+		if x509.IsEncryptedPEMBlock(pemBlock) {
+			decryptedKey, err := x509.DecryptPEMBlock(pemBlock, []byte(cfg.TLS.Passphrase))
+			if err != nil {
+				return nil, fmt.Errorf("private key is encrypted, but could not decrypt it: %s", err)
+			}
+			// If private key was encrypted we can overwrite the original contents now with the decrypted version
+			privateKey = pem.EncodeToMemory(&pem.Block{Type: pemBlock.Type, Bytes: decryptedKey})
+		}
+
+		tlsCert, err := tls.X509KeyPair(cert, privateKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load certificate pair for schema registry client: %w", err)
+		}
+
+		tlsConfig := &tls.Config{
+			Certificates: []tls.Certificate{tlsCert},
+		}
+		tlsConfig.BuildNameToCertificate()
+		transport := &http.Transport{TLSClientConfig: tlsConfig}
+
+		client.SetTransport(transport)
 	}
 
 	// Use custom root ca if desired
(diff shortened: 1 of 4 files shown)


backend/pkg/schema/client.go
```diff
@@ -44,6 +46,46 @@ func newClient(cfg Config) (*Client, error) {
...
+			return nil, fmt.Errorf("no valid private key found")
+		}
+		if x509.IsEncryptedPEMBlock(pemBlock) {
+			decryptedKey, err := x509.DecryptPEMBlock(pemBlock, []byte(cfg.TLS.Passphrase))
+			if err != nil {
+				return nil, fmt.Errorf("private key is encrypted, but could not decrypt it: %s", err)
+			}
+			// If private key was encrypted we can overwrite the original contents now with the decrypted version
+			privateKey = pem.EncodeToMemory(&pem.Block{Type: pemBlock.Type, Bytes: decryptedKey})
+		}
```
weeco, line 73: These methods are marked as deprecated in Go because the PEM encryption itself is insecure by design, see: https://pkg.go.dev/crypto/x509#IsEncryptedPEMBlock

I think we should not add that functionality if it's due to be removed

backend/pkg/schema/client.go
```diff
@@ -44,6 +46,46 @@ func newClient(cfg Config) (*Client, error) {
...
+
+		tlsCert, err := tls.X509KeyPair(cert, privateKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load certificate pair for schema registry client: %w", err)
+		}
+
+		tlsConfig := &tls.Config{
+			Certificates: []tls.Certificate{tlsCert},
+		}
+		tlsConfig.BuildNameToCertificate()
```
weeco, line 83: Also deprecated and not needed I think?

backend/pkg/schema/client.go
```diff
@@ -44,6 +46,46 @@ func newClient(cfg Config) (*Client, error) {
 		client = client.SetAuthToken(cfg.BearerToken)
 	}
 
+	// Configure Client Certificat transport
+	if cfg.TLS.Enabled {
+
```
weeco, line 51: Custom CA cert handling is missing

weeco:
I pushed two commits into your branch:

1. I removed the support for encrypted keys with passphrases
2. A bit refactoring how the tls client config is constructed
3. I added the configuration for `InsecureSkipTLSVerify`

What do you think? Do you have a chance to verify that it's working as intended?

backend/pkg/schema/client.go
julcollas: I took this from https://github.com/cloudhut/kowl/blob/e0413820058c13a9f8ebff41450604b20d6ca52c/backend/pkg/kafka/config_helper.go#L152 :)
My PEM is not encryted so I do not need this

julcollas:
I just tested, and SchemaRegistry connection works.
I just get a weird popup, but I do not think it come from this modifications ?

> Kowl has been updated
> The page must be reloaded to apply the newest version of the frontend.

I will update reference configs in my next commit

weeco:
Thanks, what kind of popup do you get? Can you post a screenshot?

julcollas:
This kind of popup. 
popup

weeco:
No that's just the update check from the frontend which is supposed to check if your local SPA should be updated. Every time you use a new docker image the backend is supposed to provide an updated version of your Single Page application and this popup is expected.

One the docs update is pushed it's ready to merge I think

backend/pkg/schema/client.go
weeco: I assumed that. Back when I implemented these methods were not deprecated and I was not aware that this should no longer be used. I believe it should not be added now that this is known.

julcollas:
I'm not able to make it disappear, but it's probably on my side.
Docs have been updated

rikimaru0345:
> I'm not able to make it disappear, but it's probably on my side.


Open the chrome developer tools (CTRL+SHIFT+I) and check the console.
You should see a message that says `frontend update available`.
The message can be expanded to reveal more fields like serverTimestamp, localTimestamp, ...

Copy paste the full message (including all those expanded fields) so I can take a look.

Also, are you running multiple kowl instances behind a load-balancer?

julcollas:
```
{
    "serverTimestamp": 12346,
    "serverDate": "1970-01-01T03:25:46.000Z",
    "serverVersionInfo": {
        "ts": "12346",
        "branch": "123456",
        "sha": "12345"
    },
    "localTimestamp": 12345,
    "localDate": "1970-01-01T03:25:45.000Z",
    "localVersion": {
        "NODE_ENV": "production",
        "REACT_APP_KOWL_GIT_SHA": "1234",
        "REACT_APP_KOWL_GIT_REF": "1234",
        "REACT_APP_KOWL_TIMESTAMP": "12345",
        "REACT_APP_KOWL_BUSINESS_GIT_SHA": "",
        "REACT_APP_KOWL_BUSINESS_GIT_REF": "",
        "REACT_APP_BUILT_FROM_PUSH": "12345",
        "REACT_APP_BUSINESS": "",
        "REACT_APP_DEV_HINT": ""
    }
}

> Also, are you running multiple kowl instances behind a load-balancer?
no

```

rikimaru0345:
Seems like the the browser is caching the frontend.
The popup should disappear when you do a hard reload (open dev tools, click and hold the reload button, select "hard reload").

I'll make sure to take another look at the update dialog later 👍

weeco (approved):

weeco:
Thanks for the PR @julcollas  and welcome as contributor 🚀

PR #223, opened 2021-05-31
Fix AWS_MSK_IAM configuration handling

rtimush:
There seems to be a typo that breaks AWS_MSK_IAM authentication.

### DIFF
--- a/backend/pkg/kafka/config_helper.go
+++ b/backend/pkg/kafka/config_helper.go
@@ -123,7 +123,7 @@ func NewKgoConfig(cfg *Config, logger *zap.Logger, hooks kgo.Hook) ([]kgo.Opt, e
 		}
 
 		// AWS MSK IAM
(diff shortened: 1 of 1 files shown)


weeco:
Hey @rtimush ,
good find! Unfortunately I never had the chance to actually test it. Do you have a chance to give it a shot and possibly report back whether it works?

rtimush:
Hey @weeco. I tried quay.io/cloudhut/kowl:master-99480a7e and unfortunately it still didn't work for me with an `Access denied: SASL_AUTHENTICATION_FAILED: SASL Authentication failed.` error.

twmb:
@rtimush how are you trying to access the cluster? from a local ec2 instance, or from your laptop? what is your IAM policy, as well? I know when I tested this, the default policies are `kafka:*`, but MSK wants `kafka-cluster:*`, which took a bit to find ...

rtimush:
I indeed made a mistake in the resources section of the IAM policy. Everything works, thanks.

PR #230, opened 2021-07-06
fix(schama_detail): get default compatibility if err on get for subject

snakeice:


backend/pkg/owl/schema_details.go
```diff
@@ -31,7 +33,11 @@ func (s *Service) GetSchemaDetails(_ context.Context, subject string, version st
 
 	cfgRes, err := s.kafkaSvc.SchemaService.GetSubjectConfig(subject)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get compatibility for given subject: %w", err)
+		s.logger.Error("failed to get compatibility for given subject: %w", zap.Error(err))
```
weeco, line 36:
```suggestion
		s.logger.Error("failed to get compatibility for given subject", zap.Error(err))
```

weeco (commented):
The changes look good to me, thank you for your contribution.

Can you eloborate your reasoning behind this pr? Is it in certain scenarios (or registry versions) expected that the compatibility can't returned per subject or why do we want to fallback to the default compatibility upon errors?

snakeice:
In several cases we don't have a configuration per subject, in this case it returned 40408 by schema registry api, this return front breaks with a "Backend API Error"... I see two alternatives, show the default (since it's the current configuration of the subject) or display empty (which particularly I think it's ugly)

weeco:
Makes sense, thank you Rodrigo 👍

PR #231, opened 2021-07-07
feat: Add json schema support

bakjos:


backend/pkg/kafka/deserializer.go
```diff
@@ -136,7 +136,25 @@ func (d *deserializer) deserializePayload(payload []byte, topicName string, reco
 		}
 	}
 
-	// 2. Test for valid XML
+	// 2. Test for json schema
+	if payload[0] == byte(0) {
+		schemaID := binary.BigEndian.Uint32(payload[1:5])
```
weeco, line 141: I think this can end up in panics because a value might be a single zero byte. Line 141 would try to read byte 1-5 which would be out of bounds. For the avro check we use something like that before checking the magic byte. 

```go
if d.SchemaService != nil && len(payload) > 5
```

frontend/src/state/restInterfaces.ts
```diff
@@ -626,7 +626,8 @@ export interface SchemaDetails {
     schemaId: number;
     version: number;
     compatibility: string;
-    schema: Schema;
+    type: string;
```
weeco, line 629: Let's use an enum here, so that we can compare against it?

frontend/src/state/backendApi.ts
```diff
@@ -741,7 +741,12 @@ const apiStore = {
         const rq = cachedApiRequest(`./api/schemas/subjects/${subjectName}/versions/${version}`, force) as Promise<SchemaDetailsResponse>;
 
         return rq
-            .then(({ schemaDetails }) => (this.schemaDetails = schemaDetails))
+            .then(({ schemaDetails }) => { 
+              if ( schemaDetails && typeof schemaDetails.schema === "string" && schemaDetails.type !== "PROTOBUF") {
```
weeco, line 745: I think the typeof check is not really needed as we rely on the correctness of our typescript types, but @rikimaru0345 might chime in here as he's mostly responsible for the frontend code.

weeco (commented):
Hey,
thanks for tackling this feature. I left a few comments.

Have you tested the changes against a schema registry / topic that uses a JSON schema?

frontend/src/state/restInterfaces.ts
bakjos: done

backend/pkg/kafka/deserializer.go
bakjos: done

bakjos:
@weeco Thanks for all the comments, Yes I tested it against the three types of schemas (JSON, PROTOBUF, and AVRO)

frontend/src/state/backendApi.ts
bakjos: @weeco The main reason I did this was that even for Avro schemas it was not showing the detail in the schema details page, and the issue was because the schema object is sent as string from the backend, adding this validation/conversion solved the issue and showed the schema correctly,

rikimaru0345:
Code looks good to me.
@bakjos I'll check it out and test it later today or tomorrow. 
If I read it correctly this PR will make it so the schema details page will now show the schema for Avro types, right?

bakjos:
@rikimaru0345 yes, it should show the fields for both Avro and Json schemas

backend/pkg/kafka/deserializer.go
```diff
@@ -136,7 +136,25 @@ func (d *deserializer) deserializePayload(payload []byte, topicName string, reco
 		}
 	}
 
-	// 2. Test for valid XML
+	// 2. Test for json schema
+	if d.SchemaService != nil && len(payload) > 5 && payload[0] == byte(0) {
+		schemaID := binary.BigEndian.Uint32(payload[1:5])
+		trimmed := bytes.TrimLeft(payload[5:], " \t\r\n")
+		startsWithJSON := trimmed[0] == '[' || trimmed[0] == '{'
+		if startsWithJSON {
```
rikimaru0345, line 144: @weeco from my comment in the PR thread: see here

rikimaru0345:
@weeco 
Tested the PR, works fine for me.
As far as frontend changes go, I looked through all of them and they look good to me! 😄 

I've also looked through the backend changes and the only thing I noticed is this: https://github.com/cloudhut/kowl/pull/231/files?diff=split&w=1#r671009018

Do you think that is "safe" enough for the detection?
I guess a payload starting with a zero byte + schema ID + json should be pretty obvious and robust, right?

I mean, I'd even go as far as saying that - compared to our other "auto detect payload type" checks - this is one of the more robust checks we have in there 😂. So I doubt there's much (anything really) that can be improved there.

In any case, the other backend changes seem fine to me as well (though I didn't check them as deeply as the frontend).

From my side this PR looks perfectly fine! 👍  
So when you've looked at the backend stuff feel free to press that merge button!

rikimaru0345 (approved):

frontend/src/state/backendApi.ts
rikimaru0345: no, the `typeof` check is fine.
yes, it is assumed that the backend currently sends `schema` as a string and not anything else, but defensive programming is always a good idea.

PR #254, opened 2021-09-08
Support Protobuf Imports

CEikermann:
Fixes https://github.com/cloudhut/kowl/issues/237

Fixed by removing the "/" at the beginning of all file paths and added "/" as import path to the proto parser

### DIFF
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -412,15 +412,12 @@ func (s *Service) createProtoRegistry() error {
 // protoFileToDescriptorWithBinary parses a .proto file and compiles it to a descriptor using the protoc binary. Protoc must
 // be available as command or this will fail.
 // Imported dependencies (such as Protobuf timestamp) are included so that the descriptors are self-contained.
-//
-// ProtoPath is the path that contains all .proto files. This directory will be searched for imports.
-// Filename is the .proto file within the protoPath that shall be parsed.
 func (s *Service) protoFileToDescriptor(files map[string]filesystem.File) ([]*desc.FileDescriptor, error) {
 	filesStr := make(map[string]string, len(files))
 	filePaths := make([]string, 0, len(filesStr))
 	for _, file := range files {
 		filesStr[file.Path] = string(file.Payload)
-		filePaths = append(filePaths, file.Path)
+		filePaths = append(filePaths, string(file.Path[1:len(file.Path)]))
 	}
 
 	errorReporter := func(err protoparse.ErrorWithPos) error {
@@ -434,6 +431,7 @@ func (s *Service) protoFileToDescriptor(files map[string]filesystem.File) ([]*de
 
 	parser := protoparse.Parser{
 		Accessor:              protoparse.FileContentsFromMap(filesStr),
+		ImportPaths:	       []string{"/"},
 		InferImportPaths:      true,
 		ValidateUnlinkedFiles: true,
 		IncludeSourceCodeInfo: true,
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -417,7 +417,7 @@ func (s *Service) protoFileToDescriptor(files map[string]filesystem.File) ([]*de
 	filePaths := make([]string, 0, len(filesStr))
 	for _, file := range files {
 		filesStr[file.Path] = string(file.Payload)
-		filePaths = append(filePaths, string(file.Path[1:len(file.Path)]))
+		filePaths = append(filePaths, file.Path[1:])
 	}
 
 	errorReporter := func(err protoparse.ErrorWithPos) error {


backend/pkg/proto/service.go
```diff
@@ -412,15 +412,13 @@ func (s *Service) createProtoRegistry() error {
...
-//
-// ProtoPath is the path that contains all .proto files. This directory will be searched for imports.
-// Filename is the .proto file within the protoPath that shall be parsed.
 func (s *Service) protoFileToDescriptor(files map[string]filesystem.File) ([]*desc.FileDescriptor, error) {
 	filesStr := make(map[string]string, len(files))
 	filePaths := make([]string, 0, len(filesStr))
 	for _, file := range files {
-		filesStr[file.Path] = string(file.Payload)
-		filePaths = append(filePaths, file.Path)
+		filePath := string(file.Path)
```
weeco, line 419: What's the reason for this variable? `file.Path` already is a string I think?

weeco (commented):
Can you explain why the default (currentDirectory) was an issue and `/` (root directory) works apparently? I'm unsure if I understand the actual circumstances causing the issue.

CEikermann:
Without the removal of the "/" at the beginning of the filePath, the parser is failing.

Error Log:
```
{"level":"warn","ts":"2021-09-08T16:05:25.599Z","msg":"failed to parse proto file to descriptor","file":"test_b.proto","line":5,"error":"test_b.proto:5:1: duplicate symbol test.b.protobuf.TestB: already defined as message in \"/test_b.proto\""}
{"level":"warn","ts":"2021-09-08T16:05:25.610Z","msg":"failed to parse proto file to descriptor","file":"test_b.proto","line":6,"error":"test_b.proto:6:3: duplicate symbol test.b.protobuf.TestB.test: already defined as field in \"/test_b.proto\""}
{"level":"fatal","ts":"2021-09-08T16:05:25.611Z","msg":"failed to start kafka service","error":"failed to create proto registry: failed to compile proto files to descriptors: failed to parse proto files to descriptors: parse failed: invalid proto source"}
```

My guess the files are parsed twice, once via input "/test_b.proto" and once via import "test_b.proto"
When we are removing the "/" at the beginning, seems the proto parser detects that the same file has been parsed already and skip it.

backend/pkg/proto/service.go
CEikermann: Oh right, sorry - this is actually my first Go development at all ;)

Fixed

CEikermann: Also wondering if there is not an easier way to create a sub string from it, but could not found any simpler solution

backend/pkg/proto/service.go
```diff
@@ -412,15 +412,12 @@ func (s *Service) createProtoRegistry() error {
...
-//
-// ProtoPath is the path that contains all .proto files. This directory will be searched for imports.
-// Filename is the .proto file within the protoPath that shall be parsed.
 func (s *Service) protoFileToDescriptor(files map[string]filesystem.File) ([]*desc.FileDescriptor, error) {
 	filesStr := make(map[string]string, len(files))
 	filePaths := make([]string, 0, len(filesStr))
 	for _, file := range files {
 		filesStr[file.Path] = string(file.Payload)
-		filePaths = append(filePaths, file.Path)
+		filePaths = append(filePaths, string(file.Path[1:len(file.Path)]))
```
weeco, line 420:
```suggestion
		filePaths = append(filePaths, file.Path[1:])
```

That should be fine :)

CEikermann: Nice, this looks much better, applied this

weeco (approved):

weeco:
Thanks for taking a look and finding this!

CEikermann:
Thanks for merging! :)

weeco:
Just FYI @CEikermann this broke protobuf support if the Git provider (instead of file provider) is used. Can you post what your protobuf configuration looked like, so that I can figure out a solution that works for both?

CEikermann:
> Just FYI @CEikermann this broke protobuf support if the Git provider (instead of file provider) is used. Can you post what your protobuf configuration looked like, so that I can figure out a solution that works for both?

Sorry this was not intended. 

https://github.com/cloudhut/kowl/issues/237#issuecomment-915312622 is this enough for you to reproduce?

weeco:
@CEikermann I just tested this locally and I was not able to reproduce this. It might be because I'm running on windows at the moment, but I could probably also test on Mac. I tend to revert your changes made here because of that.

In my case an import like `address.proto` ends up in `ddress.proto` and that's of course not existent in my in memory filesystem. Hence just removing the first character can not be the solution after all :/

weeco:
I just pushed a version that removes the prefix `/` on all filepaths. Maybe that helps with the filepath mismatch. Could you try that? Otherwise I'd like to figure out how to reproduce your issue so that I can resolve them in a more methodical way.

CEikermann:
@weeco I have tested with master 11afdfa9c9d834df7ac625045edfb27e2fb45f53

I do get an error when launching the kowl docker container
```
{"level":"fatal","ts":"2021-09-13T18:54:17.086Z","msg":"failed to start kafka service","error":"failed to create proto registry: failed to compile proto files to descriptors: failed to parse proto files to descriptors: test_a.proto: file does not exist"}
````

**Comment by CEikermann** [comment]
As addition to https://github.com/cloudhut/kowl/issues/237#issuecomment-915312622

docker-compose.yml
```
  kafka-ui:
    image: quay.io/cloudhut/kowl:master-11afdfa
    restart: unless-stopped
    volumes:
      - ./docker/kafka-ui/config.yml:/etc/kowl/config.yml
      - /tmp/protobuf/:/etc/protos/
    ports:
      - 9093:8080
    entrypoint: ./kowl --config.filepath=/etc/kowl/config.yml
    depends_on:
      - kafka
    networks:
      - main
  ```

Docker Desktop 4.0.0 @ MacOS 11.4

You can not reproduce this on your env with the given files?

CEikermann:
@weeco are you able to reproduce the issue on your env?

weeco:
@CEikermann Yes, I can reproduce it using Docker, not within my IDE unfortunately. So I'm not entirely sure what's going on there, but I will try to figure it out as soon as I have some time for it

CEikermann:
Have created an follow up issue of this: https://github.com/cloudhut/kowl/issues/265

PR #255, opened 2021-09-09
Add support for PKCS8 encrypted TLS keys

kernle32dll:
At the moment, kowl only supports TKS private keys, which have RFC 1423 encrypted PEM blocks.

A much better variant are PKCS8 encrypted keys. Especially considering that recent Zookeeper and Kafka versions support PKCS8 (v1) encrypted keys to replace key- and truststores, making for a streamlined TLS configuration.

However, kowl does not support PKCS8 encrypted keys - yet. This PR aims to remedy this, by adding support for PKCS8 (v2) encrypted keys. The `tryPKCS8Decryption` function is taken from [one of my MIT licensed projects](https://github.com/kernle32dll/keybox-go), but I opted to add the single function instead of the whole dependency.

Attached, find a small self-contained docker-compose example, to show how (and that) it works.
[example.zip](https://github.com/cloudhut/kowl/files/7139919/example.zip)

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/twmb/franz-go/pkg/kmsg v0.0.0-20210901051457-3c197a133ddd
 	github.com/vmihailenco/msgpack/v5 v5.3.1
 	github.com/xanzy/ssh-agent v0.3.1 // indirect
+	github.com/youmark/pkcs8 v0.0.0-20201027041543-1326539a0a0a // indirect
 	go.uber.org/zap v1.16.0
 	golang.org/x/mod v0.4.1 // indirect
 	golang.org/x/net v0.0.0-20210903162142-ad29c8ab022f // indirect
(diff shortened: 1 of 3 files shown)


weeco:
Hey @kernle32dll ,
thanks for taking the time implementing this and also adding the example.

Everything looks fine to me, but I'm still thinking about two things:
1. Because the legacy PEM encryption is insecure by design we can probably remove this method. It's already marked as deprecated in Go and possibly this will be removed sooner or later anyways.
2. I wonder whether it makes sense to be more explicit about the used encryption method in our configuration settings, so that we can always print a descriptive error message in case the decryption fails. I prefer more verbose configuration and therefore more descriptive error messages - especially in the connection / tls part. Personally I think it's frustrating to debug TLS / connection issues otherwise.

What do you think about both points?

kernle32dll:
Hi!

> 1. Because the legacy PEM encryption is insecure by design we can probably remove this method. It's already marked as deprecated in Go and possibly this will be removed sooner or later anyways.

I cannot really speak on point 1, as its hard to judge who might be using it. But yeah, I would assume it could/should be removed in a version 2.X of kowl. Maybe add a deprecation notice as a log statement for now?

> 2. I wonder whether it makes sense to be more explicit about the used encryption method in our configuration settings, so that we can always print a descriptive error message in case the decryption fails. I prefer more verbose configuration and therefore more descriptive error messages - especially in the connection / tls part. Personally I think it's frustrating to debug TLS / connection issues otherwise.

As for point 2.... Good question. I am more inclined to do an automatic inferring of the (encryption) type, hence https://github.com/kernle32dll/keybox-go. However, I can understand your reasoning, too. I think I would go with a two-step approach. Have an option to set the encryption time explicitly, or try to "guess" it, if not set. Sounds good to you?

weeco:
> Have an option to set the encryption time explicitly, or try to "guess" it, if not set. Sounds good to you?

Sounds like a compromise, my assumption is that there aren't a lot people using encrypted pems anyways, but since we don't collect any stats I can be wrong of course :D . If you could add that, that would be nice. And as you suggested we will remove it with a v2 - I'll submit an issue for that, so that this won't get lost.

PR #278, opened 2021-11-09
Update build-image.yml

rsmnarts:
add config for support arm64 platform

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -65,3 +65,4 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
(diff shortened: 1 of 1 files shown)


weeco:
@rsmnarts What would you expect to happen upon the next build? I couldn't find it in the GHA documentation at a first glance. I'm wondering if that would push another docker image with a different tag? Is that even necessary?

I was under the impression that it doesn't matter for Docker because the docker engine abstracts away the platform differences. Some more background information would be nice as I don't have any experience with multi platform builds (nor do I have an ARM based device to test)

xzyfer:
The docker engine will use qemu emulation to run amd64 containers on arm64. However this comes with a significant runtime overhead.

weeco:
We have added ARM builds for docker images, hence closing.

PR #282, opened 2021-11-18
Add new github ssh host keys for ECDSA and Ed25519.

rd-michael-pasdziernik:
kowl fails to clone github repositories because of missing new ssh host keys in the image:

`{"level":"fatal","ts":"2021-11-18T07:08:58.014Z","msg":"failed to start owl service","error":"failed to clone git repo: ssh: handshake failed: knownhosts: key mismatch"}`

For reference:
https://github.blog/2021-09-01-improving-git-protocol-security-github/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,6 @@ COPY --from=frontendBuilder /app/build/ /app/build
 
 # Add github.com to known SSH hosts by default (required for pulling topic docs & proto files from a Git repo)
 RUN apk update && apk add --no-cache openssh
-RUN ssh-keyscan -t rsa github.com >> /etc/ssh/ssh_known_hosts
+RUN ssh-keyscan github.com >> /etc/ssh/ssh_known_hosts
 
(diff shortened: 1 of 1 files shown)


weeco:
Hey @rd-michael-pasdziernik ,
thanks for the PR 🔥 . I'm wondering whether we should omit the `-t` type parameter which then would default to `“rsa”, “ecdsa”, and “ed25519” ` (according to the ssh-keyscan docs) , or whether we should pin it to `ed25519` for instance. Do you have more insight whether the right type will be used automatically or not?

rd-michael-pasdziernik:
I think server and client negotiate the best algorithm that works for both sides. So it seems to depend on the lib that kowl uses to clone the repo. Perhaps pinning to ed25519 would work, but having all keys supported by github available seems to be less error prone?

weeco:
If negotiation is supported, I agree this would be the best option. Let's give it a try! Thanks again for the PR

PR #284, opened 2021-11-19
Feature/configure schema registry timeout

dcordonu:
We have noticed that when trying to determine if a message has to be deserialized using Avro, if schema fails to be fetched in the next 5 seconds an error is thrown and the deserializing logic thinks it is not an Avro message (when it really is).

This PR adds the possibility to configure it, using 5 as a default value if no one is given or the value provided is less than that.

### DIFF
--- a/backend/pkg/schema/config.go
+++ b/backend/pkg/schema/config.go
@@ -33,7 +33,7 @@ func (c *Config) Validate() error {
 		return nil
 	}
 
-	if c.Timeout == 0 {
+	if c.Timeout <= 5 {
 		c.Timeout = 5
 	}
 
--- a/backend/pkg/schema/config.go
+++ b/backend/pkg/schema/config.go
@@ -17,6 +17,9 @@ type Config struct {
 
 	// TLS / Custom CA
 	TLS TLSConfig `yaml:"tls"`
+
+	// Timeouts
+	Timeout int `yaml:"timeout"`
 }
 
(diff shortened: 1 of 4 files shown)


backend/pkg/schema/config.go
```diff
@@ -17,6 +17,9 @@ type Config struct {
 
 	// TLS / Custom CA
 	TLS TLSConfig `yaml:"tls"`
+
+	// Timeouts
+	Timeout int `yaml:"timeout"`
```
weeco, line 22: I'm wondering whether we can use `time.Duration` immediately here. Can you try that, or aren't you able to run Kowl locally (then I can try that as well). Regardless in Go there is the concept of time.Duration where you can specify durations with strings like `5s`, `1h` etc. I think we should use this (we already use that for the `server.timeouts` for instance

backend/pkg/schema/config.go
```diff
@@ -30,6 +33,10 @@ func (c *Config) Validate() error {
 		return nil
 	}
 
+	if c.Timeout <= 5 {
+		c.Timeout = 5
+	}
```
weeco, line 38: If you intend to set a default we should be consistent with the other config structs and introduce a `SetDefaults()` method, that is called by it's parent config struct and then will be executed before Validate() is called.

weeco (commented):
Hey, thanks for your PR. I left a few comments about using Go's `time.Duration` type for parsing the input :).

PR #290, opened 2021-12-10
Fix immer version from showing up as critical during npm audit

thepoppingone:
Using https://www.npmjs.com/package/npm-force-resolutions to fix immerjs npm audit finding

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Master / unreleased
+## 1.5.0 / 2021-11-10
 
 - [FEATURE] Add support for deleting topics
 - [FEATURE] Add support for deleting records within topics (only possible on topics with cleanup.policy=delete due to limitations in Kafka)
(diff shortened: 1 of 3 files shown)


rikimaru0345:
@thepoppingone 

Thanks for the PR.

Can you please revert package lock completely?
I'll update package lock myself when I have some time to test the PR. 

Otherwise seems fine 👍

thepoppingone:
The package lock is now reverted completely. Thanks for reviewing!

thepoppingone:
Another note though:

When running this locally at the current stage there will be no issues, however, when building the image using Docker at 
https://github.com/cloudhut/kowl/blob/master/Dockerfile#L28

`npm install` will fail due to the stage `preinstall` is not allowed. The current workaround I found is to include the flag  `--unsafe-perm` to `npm install`. 

If you need me to include this in the PR, do let me know, this is dependent on the container used. I am using `node:17-alpine3.14`

thepoppingone:
bump!

rikimaru0345:
Sorry for the delay!
I will take care of it in a few hours.

rikimaru0345:
Ok, I checked it, seems good!

PR #300, opened 2021-12-22
Add list quotas

weeco:


frontend/src/components/routes.tsx
```diff
@@ -267,6 +268,10 @@ export const APP_ROUTES: IRouteEntry[] = [
         routeVisibility(true, [], ['canListAcls'])
     ),
 
+    MakeRoute<{}>('/quotas', QuotasList, 'Quotas', <span className='menuIcon anticon'><ScaleIcon /></span>, true,
+        routeVisibility(true, [Feature.GetQuotas], ['canListAcls'])
```
rikimaru0345, line 272: this still has canListAcls

frontend/src/state/restInterfaces.ts
```diff
@@ -642,6 +643,21 @@ export interface AclRule {
     permissionType: string;
 }
 
+export interface QuotaResponse {
+    error?: string;
+    items: QuotaResponseItem[];
+}
+
+export interface QuotaResponseItem {
+    entityType: string; // client-id, user, ip
```
rikimaru0345, line 652: sounds like an enum, is it one? (like line 658)

frontend/src/state/restInterfaces.ts
```diff
@@ -642,6 +643,21 @@ export interface AclRule {
...
+    entityName?: string;
+    settings: QuotaResponseSetting[];
+}
+
+export interface QuotaResponseSetting {
+    key: 'producer_byte_rate' | 'consumer_byte_rate' | 'connection_creation_rate';
+    value: number;
+}
```
rikimaru0345, line 660: It would be really great if  the types of `QuotaResponseItem.settings`  and `QuotaResponseSetting` were defined using a mapped type definition. Like this:
```ts
// at root level
export const QuotaTypes = [ "producer_byte_rate", "consumer_byte_rate", "connection_creation_rate" ] as const;
export type QuotaType =  typeof quotaTypes[number];

// inside QuotaResponseItem
   settings: { [key in QuotaType ]: number|undefined;  }
```

That would have all the advantages of an enum and more:
- like full auto completion
- members of the union can be iterated over
- no repeated definitions
- easily extendable with new quota types, as well as being able to add `string` as an "escape hatch" without having to choose (defined stuff will be auto-completed and suggested, unknown props are still accessible, but ofc won't get intellisense)
- gets fully type erased

However you'd have to change the backend so it returns something like this instead:
```json

"items": [
    {
            "entityType": "asdafgd",
            "entityName": "nnnnnnnnnn",
            "settings": {
                "producer_byte_rate": 1234,
                "connection_creation_rate": 12345
             }
     }
]
```

But that's pretty easy, right? `settings` would just be a `map[string]int` (or probably better `map[string]float64`)

rikimaru0345 (changes requested):

PR #321, opened 2022-02-24
Fix kafka metrics

tony2001:
The hook functions were named incorrectly, which resulted in empty/non-functional metrics.
Use the correct funcs and fix the metrics.


weeco:
@tony2001 Thanks for this awesome PR. I didn't know about that possibility to ensure an interface is implemented. That would have helped me to spot the issue when the function names of that interface have been changed in franz-go. Glad I learned something :) 🚀

PR #348, opened 2022-04-21
Fix partition tab leader tracking and status visualization

liarokapisv:
The partitions tab in the topics page is meant to show the
current Leader as well as all the replica brokers of a given partition.

The responsible component so far always picked the first broker in
the replica set as the Leader.

This is usually correct, on partition creation the first broker in the replica
set is denoted as Leader, this broker is the "preferred Leader" for the
given partition and the cluster (depending on the configuration) tends to
re-elect the preferred Leader.

Unfortunately, the preferred Leader is not always available,
it can be the case that that the preferred broker is offline and another
broker is elected as Leader.

The current implementation does not properly track the actual leader,
instead it shows the preferred Leader. It also doesn't show which
brokers are currently online. These issues together can lead to very
confusing visualization scenarios where the offline preferred Leader is shown as
the online current Leader while a different broker is the actual Leader.

@rikimaru0345  on commit cc46b25673b507f19d8143bcb8d586f7ecdaf6af
added better tracking support in the
`/pages/reassign-partitions/components/BrokerList.tsx`
component which not only properly tracks the Leader but also
shows the online/offline status of the brokers containing
the replicas. Unfortunately it seems that the augmented `partition`
property enabling these semantics was never actually used.

This commit replaces the naive manual passing of the `leaderId` and `brokerIds` properties to
the `BrokerList` component with the passing of the augmented `partition` property.

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Partitions.tsx
+++ b/frontend/src/components/pages/topics/Tab.Partitions.tsx
@@ -70,7 +70,7 @@ export class TopicPartitions extends Component<{ topic: Topic }> {
                 },
                 {
                     title: 'Brokers',
-                    render: (v, r) => r.replicas && <BrokerList brokerIds={r.replicas} leaderId={r.replicas.length > 0 ? r.replicas[0] : -1} />
+                    render: (v, r) => <BrokerList partition={r} />
                 }
             ]} />
 


liarokapisv:
The main goal here is to provide better visualization for the Brokers part of the topics' partitions tab.
This commit takes the path of least-resistance by reusing @rikimaru0345 implementation. 
At least this fixes the problem of properly tracking Leader and off/online status.

For future reference, it would also be very useful to also provide support for the In-Sync Replica Set, it would be very useful for demonstrating Kafka's Isr behavior without having to resort to the command line interface.

weeco:
@liarokapisv GREAT observation and GREAT summary here 🔥 . We will try to review this as soon as possible and merge it! We will also see how we can improve the visualization in regards to the ISR without making it too overloaded. If you have any ideas or inspiration, happy to see them.

PR #350, opened 2022-04-26
Feature/publish records

rikimaru0345:


backend/pkg/kafka/produce_records.go
```diff
@@ -0,0 +1,120 @@
...
+			return nil, fmt.Errorf("unable to begin transaction: %w", err)
+		}
+	}
+
+	mu := sync.Mutex{}
+	recordResponses := make([]ProduceRecordResponse, 0)
+	for _, r := range records {
+		client.Produce(ctx, r, func(producedRecord *kgo.Record, err error) {
+			mu.Lock()
+			defer mu.Unlock()
```
twmb, line 76: Produce as of v1.4 is now serial, so this lock is not necessary

irux:
@rikimaru0345 does this support avro ?

weeco:
@irux It doesn't, we still want to add it. The serialization of avro/protobuf when used with schema registry is a little bit harder due to the binary encoding. We plan to add support for this though!

PR #363, opened 2022-05-12
Use config vars instead of hardcoded timeouts

tony2001:
6 seconds is not enough when Kafka Connect has A LOT of connectors

### DIFF
--- a/backend/pkg/connect/config.go
+++ b/backend/pkg/connect/config.go
@@ -22,9 +22,9 @@ const (
 )
 
 type Config struct {
-	ConnectTimeout time.Duration   `yaml:"connect_timeout"` // used for connectivity test
-	ReadTimeout    time.Duration   `yaml:"read_timeout"`    // overall REST/HTTP read timeout
-	RequestTimeout time.Duration   `yaml:"request_timeout"` // timeout for REST requests to Kafka Connect
+	ConnectTimeout time.Duration   `yaml:"connectTimeout"` // used for connectivity test
+	ReadTimeout    time.Duration   `yaml:"readTimeout"`    // overall REST/HTTP read timeout
+	RequestTimeout time.Duration   `yaml:"requestTimeout"` // timeout for REST requests to Kafka Connect
 	Enabled        bool            `yaml:"enabled"`
 	Clusters       []ConfigCluster `yaml:"clusters"`
 }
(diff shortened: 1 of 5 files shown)


backend/pkg/connect/config.go
```diff
@@ -12,17 +12,30 @@ package connect
...
+	defaultReadTimeout    = 60 * time.Second
+	defaultRequestTimeout = 6 * time.Second
 )
 
 type Config struct {
-	Enabled  bool            `yaml:"enabled"`
-	Clusters []ConfigCluster `yaml:"clusters"`
+	ConnectTimeout time.Duration   `yaml:"connectTimeout"` // used for connectivity test
+	ReadTimeout    time.Duration   `yaml:"readTimeout"`    // overall REST/HTTP read timeout
+	RequestTimeout time.Duration   `yaml:"requestTimeout"` // timeout for REST requests to Kafka Connect
```
weeco, line 27: Makes sense to me, could you move them below the existing properties in that config struct (as they are more likely to be configured)

backend/pkg/connect/config.go
```diff
@@ -12,17 +12,30 @@ package connect
 import (
 	"flag"
 	"fmt"
+	"time"
+)
+
+const (
+	defaultConnectTimeout = 15 * time.Second
+	defaultReadTimeout    = 60 * time.Second
+	defaultRequestTimeout = 6 * time.Second
```
weeco, line 21: I think we don't need a extra const given that they are only used within SetDefaults(), but if you feel they are required here, it's also fine for me

backend/pkg/connect/config.go
tony2001: OK, no problem

backend/pkg/connect/config.go
tony2001: OK, removed

weeco:
Awesome thanks @tony2001 , one thing I forgot, could you document these new properties by adding them to https://github.com/redpanda-data/kowl/blob/master/docs/config/kowl.yaml and https://github.com/redpanda-data/kowl/blob/master/docs/config/kowl-business.yaml ?

Thanks for your PR

tony2001:
No problem, will do later today.

tony2001:
@weeco could check my latest commit? I've added some examples/descriptions there.

weeco (approved):
LGTM, thank you!

PR #376, opened 2022-06-03
Extend documentation for schema registry flags

dominikhuembeli:
Schema registry password and bearerToken can be set by flags. This is already implemented but not added to config documentation.


weeco:
Thanks @dominikhuembeli , I thought yoru issue was only about the helm repo variable. I didn't know it's not documented here.

Issue #390, opened 2022-06-16
Error retrieving Kafka Version


victorgawk:
After commit https://github.com/redpanda-data/console/commit/7f436c34697395e80fcb1fb777bc2f6c5c281dc4 when retrieving kafka version on `/brokers` endpoint causes the following error:

> "error":"failed to request api versions. Inner Kafka error: INVALID_REQUEST: This most likely occurs because of a request being malformed by the client library or the message was sent to an incompatible broker. See the broker logs for more details."

This error is caused by the presence of a space character in the `req.ClientSoftwareName` field with value "RP Console" (previouly was "Kowl").

Image below describe the problem (check file `backend/pkg/kafka/api_version.go`):
![image](https://user-images.githubusercontent.com/6736720/173976385-6452b524-fd9c-4384-9e32-7426a1de9282.png)

weeco:
Thanks for filing the issue + PR

PR #391, opened 2022-06-16
Fix Kafka GetAPIVersions

victorgawk:
Remove illegal space character from Kafka get API versions request (#390).


weeco (approved):
Thank you!

Issue #392, opened 2022-06-16
Bug: 'Delete Messages' button is displaying a fixed tooltip text when disabled


victorgawk:
Actual: display fixed string `Records on Topics with the 'compact' cleanup policy cannot be deleted.`
Expected: display content from `errorText` variable

https://github.com/redpanda-data/console/blob/cb51d7221e104293e0a6fce56566e5ca615b0c2f/frontend/src/components/pages/topics/Tab.Messages/index.tsx#L1437

PR #393, opened 2022-06-16
Fix 'Delete Records' title when disabled

victorgawk:
Fix 'Delete Records' tooltip title in order to display a proper message (#392).


rikimaru0345:
thanks!

Issue #212, opened 2021-05-09
Proper editor with Typescript support & auto completion for message filters


weeco:
When using search filters it can be annoying to write the Javascript filter code because not all the ES6+ features are supported. Additionally we don't have the usual IDE support in the form of autocompletion/intellisense as well as typesafety. We believe we can fix this by implementing the monaco-editor (https://microsoft.github.io/monaco-editor/) and by utilizing Typescript. The goal is to make it much easier, flexible (due to typescript that can be transpiled to ES5 compatible code for the JS engine) and fault tolerant (typesafety) to write filter queries.

This issue supersedes https://github.com/cloudhut/kowl/issues/92

sago2k8:
Hey @weeco, I'm currently working on this one, should I create a new Issue to track progress and link the PR ?

PR #395, opened 2022-06-17
feat(frontend): require monaco-editor from static files

sago2k8:
# Cover letter 

Require static bundle of `monaco-editor` instead of making a request to a CDN.

## points of interest:
- remove packages that seem to be unnecessary now
- use `@monaco-editor/react/loader` to use a different location for the `monaco-editor` source.
 
## Video reference
https://www.loom.com/share/4a1221c005824cbfb8dfb97995f84436


rikimaru0345:
> - remove packages that seem to be unnecessary now


This is a good point, but I'm not quite sure if *this* is the issue/PR that enables that, unless I'm misunderstanding.
What packages do you think can be removed now? 


Maybe we're thinking about different things! 😄 
I was thinking that, once we replace the old code-editor for message filters ( https://github.com/redpanda-data/console/issues/212 ), we can remove a lot of package that are causing us some problems. 
(react-simple-code-editor, react-syntax-highlighter, prismjs, maybe more)

sago2k8:
> 

Yes, I am not removing packages that are being used for the editor, just removing `monaco-editor` since it's not a dependency of `@monaco-editor/react` anymore.

sago2k8:
> I'll check out the PR later today.
> 
> When we want to update monaco, we need to pull the latest versions of those files, right? Like: `frontend/public/static/js/monaco-editor/min-maps/vs/base/worker/workerMain.js.map`
> 
> How do we actually get all those files? Is there some command to quickly do that?

I just added the script [fetch-monaco](https://github.com/redpanda-data/console/pull/395/files#diff-76cd8ac51c4e54fea172b7a8fbc14c2b1c7acf6626c197c3b0562f591fc0b3e0R1) to fetch the library from npm on prepare phase, please take a look at the script, it downloads and extract monaco-editor library and places it in the public folder.

| note: I had to update the Dockerfile to make sure it's aware of this step and the image is created correctly

rikimaru0345 (approved):

PR #399, opened 2022-06-23
Fix/previous master

sago2k8:
# Cover letter 

Fix the dependency tree and remove unnecessary packages.
This issue fixes an error in the build process of the application and removes some unnecessary dependencies related to babel, now we rely only in typescript to perform the transformations (decorators, private properties, esnext usage), This also fixes an issue brought by `@types/react` and `@types/react-dom`, using resolutions to force non-breaking version of those packages to be installed. 

## points of interest:
- usage of resolutions to force package versions
- remove unused babel-transform plugins
- app builds with no issues now
- remove some eslint alerts

> Notes: I think we should think of using [craco](https://www.npmjs.com/package/@craco/craco) for the CRA customisation; It's type-safe, well maintained and it works out of the box with the latest versions of CRA.
> We should keep an eye on module resolution to update the packages when the community has solved the breaking changes.
> I think having a CI is very necessary.


rikimaru0345 (approved):

rikimaru0345:
> Notes: I think we should think of using [craco](https://www.npmjs.com/package/@craco/craco) for the CRA customisation; It's type-safe, well maintained and it works out of the box with the latest versions of CRA.
> We should keep an eye on module resolution to update the packages when the community has solved the breaking changes.
I think having a CI is very necessary.

About craco: sounds good, I'll create an issue for it.

I'm not sure I understand the CI part, 
do you mean we should add a new step to check for when some packages gets updated?

sago2k8:
> I'm not sure I understand the CI part, do you mean we should add a new step to check for when some packages gets updated?

@rikimaru0345  Here I created a PR #404 with a pipeline for testing the frontend, may be easier to prove my point with a PR.

PR #404, opened 2022-06-27
ci(frontend): add action for linting frontend

sago2k8:
# Cover letter 
The motivation for this change is to make sure the code passes validations before merging it, improving code quality and asserting successful builds.

## References
we had to create a fix PR #399 because the build command was failing on master, this PR aims to diagnose possible issues earlier on.

## Points of interest 
- add proper triggers, action will only be executed when changes occur on the frontend folder
- add the new workflow for front-end test
- add a lint step to the workflow, when we set up proper tests there should be a test step too.
- add a step for building the app for production

> notes
> @weeco  I know we discussed only building the app when adding a tag and I have another PR with that trigger I just noticed that building it in the environment of the action, with the cache and using proper triggers it's actually quite efficient, so if you agree I would prefer this approach of always building (see actions execution time)


.github/workflows/frontent-ci-pipeline.yml
```diff
@@ -0,0 +1,36 @@
+name: "Frontend test - action"
```
weeco, line 1:
```suggestion
name: "Frontend Lint & Compile check"
```
Just to avoid confusion that this doesn't run any actual npm tests. Could we also update the filename to something like `frontend-lint-and-compile-check.yaml` ?

.github/workflows/frontent-ci-pipeline.yml
```diff
@@ -0,0 +1,36 @@
...
+          cd frontend
+          npm ci
+      - name: Run lint
+        run: |
+          cd frontend
+          npm run lint
+      - name: Run build frontend
+        run: |
+          cd frontend
+          CI=false
```
weeco, line 30: Why is it false? What effect does it have?

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -31,10 +31,9 @@ import { Link } from 'react-router-dom';
 
 @observer
 class KafkaConnectOverview extends PageComponent {
-    constructor(p: any) {
-        super(p);
-        // makeObservable(this);
-    }
+    // constructor(p: any) {
+    //     super(p);
+    // }
```
weeco, line 36: Does this belong to this PR? Why is the constructor commented?

.github/workflows/frontent-ci-pipeline.yml
sago2k8: This is to allowing a successful build when there are linting warnings, not necessary now.

sago2k8:
@weeco, @rikimaru0345 I updated the code, basically removed some unused variables, kept the ones that seemed to be related to mobx usage and commented lines that seemed to be relevant for reference. could you review it 🙏

PR #410, opened 2022-07-06
Feature/use monaco on topics view for filtering

sago2k8:
# Cover letter 
The motivation for this change is to unify the usage of the built-in code editor in the application and add extra support for types and modern typescript snippets on the filtering topics functionality.
The implementation is still a bit basic and it basically works exactly as the previous editor but it now supports types and typescript compilation.

## References
#212 

## Points of interest 
- created a new editor component for adding filter topic modal
- added support for typescript and made sure the compiled code is compatible with ES3 (@rikimaru0345 mentioned the interpreter might need old ES versions)
- Tested manually multiple combinations of filters and made sure it works ( hopefully I can add unit testing for this in the future).
- remove now unnecessary dependencies `primsjs`, `@types/pismjs` and `react-simple-code-editor`

## Video/screenshots

https://www.loom.com/share/cc15a91c8d634273a5a577c6095f5e6c

> notes
>  I understand we want to add typescript support for the messages and headers with better typing; reading the last message and generating typescript types from its content for instance, figure it would be better to just replicate current functionality and extend it when we are sure it works as expected in every case, wdyt? @weeco, @rikimaru0345.
> @rikimaru0345 I took into consideration your suggestion of using [constrained-editor-plugin](https://www.npmjs.com/package/constrained-editor-plugin) but same as above I wanted to replicate the current functionality, manual testing it broadly then move to include more features. 
We could create an issue to keep track of these feature requests.


weeco (approved):

frontend/src/state/ui.ts
```diff
@@ -74,30 +58,35 @@ export class FilterEntry {
...
+            return true;
+        case 'relative':
+            return true;
     }
     return false;
 }
 
-export enum TopicOffsetOrigin { EndMinusResults = -1, Start = -2, End = -3, Timestamp = -4, Custom = 0 }
-export type TopicMessageSearchSettings = TopicDetailsSettings['searchParams']
+export enum TopicOffsetOrigin {
```
weeco, line 82: Topics don't have offsets, but partitions have. I'd suggest to rename it to `PartitionOffsetOrigin`

PR #418, opened 2022-07-21
frontend: fix incorrect links

ronivay:
Fix incorrect links pointing to old kowl.yaml instead of console.yaml under docs


weeco (approved):
Thank you!

PR #425, opened 2022-08-10
proto: support protobuf any

chris-palmer-deltatre:
# Better Protobuf Any support

This commit adds better support for protobuf deserialization by allowing the console to deserialize custom proto messages that are supplied to fields with the `google.protobuf.Any` type.

Before this change, if your protobuf message contained a field of type `google.protobuf.Any` and the value supplied at runtime was not a standard proto message (e.g. `google.protobuf.Timestamp`), then `msg.MarshalJSON` (called from `deserializeProtobufMessageToJSON`) would yield the following error:

```
failed to marshal protobuf message to JSON:
unknown message type \"<type-url>\"
```

This error was emitted by the `github.com/jhump/protoreflect` because it would fail to resolve the 'type url' of the value supplied to the `google.protobuf.Any` field. This error would occur even if the type was available in the proto registry.

The error occurred, because the internal `AnyResolver` created by `github.com/jhump/protoreflect` only contained the file descriptor relating to the overall protobuf message being processed. Therefore if the type referred to by the 'google.protobuf.Any' field existed in some other file, the type could not be resolved.

This commit amends this behaviour by calling `msg.MarshalJSONPB` and passing a custom `AnyResolver` through, defined in `console/backend/pkg/proto/service.go`. The custom `AnyResolver` takes precedence over the internal `AnyResolver`, but the internal `AnyResolver` is still used by `github.com/jhump/protoreflect` if the custom `AnyResolver` cannot resolve a type. The custom `AnyResolver` provided, strips the "type url" of anything before the last '/' and then calls the proto registry to resolve the type. Removing the contents of the string before the slash is necessary because types begin with "type.googleapis.com/", but the proto registry does not contain this prefix.

The result of the change is that registry types can now be considered, which allows the console to support deserialization of custom proto message types that are supplied to a `google.protobuf.Any` field.

**WORKING EXAMPLE**
_config:_
```yaml
logger: 
  level: debug
serveFrontend: false
server:
  listenPort: 9090
kafka:
  brokers: ["localhost:9092"]
  protobuf:
    enabled: true
    mappings:
      - topicName: topic4
        valueProtoType: cloud.CloudEvent
    git:
      enabled: true
      refreshInterval: 5m
      repository:
        url: <path to proto repo>
        baseDirectory: proto
      basicAuth:
        enabled: true
        username: <>
        password: <>
```

_proto repo:_
```proto
// cloud/cloud.proto
syntax = "proto3";

package cloud;
option go_package = "./cloud";

import "google/protobuf/timestamp.proto";
import "google/protobuf/any.proto";

message CloudEvent {
    string spec_version = 1;
    string type = 2;
    string source = 3;
    string id = 4;
    google.protobuf.Timestamp time = 5;
    string data_content_type = 6;
    string event_version = 7;
    google.protobuf.Any data = 8;
}
```

```proto
// user/upload.proto
syntax = "proto3";
package user;
option go_package = "./user";

message UploadEvent {
    Properties properties = 1;
    string user_id                    = 2;
    repeated string tags        = 3;
}

message Properties {
   string name = 1;
   string email = 2;
   string age = 3;
}
```

_result:_
image

### DIFF
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/desc/protoparse"
 	"github.com/jhump/protoreflect/dynamic"
@@ -24,6 +25,7 @@ import (
 	"github.com/redpanda-data/console/backend/pkg/git"
 	"github.com/redpanda-data/console/backend/pkg/schema"
 	"go.uber.org/zap"
+	"google.golang.org/protobuf/runtime/protoiface"
 )
 
 type RecordPropertyType int
@@ -171,7 +173,9 @@ func (s *Service) deserializeProtobufMessageToJSON(payload []byte, md *desc.Mess
 		return nil, fmt.Errorf("failed to unmarshal payload into protobuf message: %w", err)
 	}
 
-	jsonBytes, err := msg.MarshalJSON()
+	jsonBytes, err := msg.MarshalJSONPB(&jsonpb.Marshaler{
+		AnyResolver: &anyResolver{s.registry},
+	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal protobuf message to JSON: %w", err)
 	}
@@ -472,4 +476,17 @@ func (s *Service) getFileDescriptorBySchemaID(schemaID int) (*desc.FileDescripto
 
 	desc, exists := s.fileDescriptorsBySchemaID[schemaID]
 	return desc, exists
+}
+
+type anyResolver struct {
+	mr *msgregistry.MessageRegistry
+}
+
+func (r *anyResolver) Resolve(typeURL string) (protoiface.MessageV1, error) {
+	mname := typeURL
+	if slash := strings.LastIndex(mname, "/"); slash >= 0 {
+		mname = mname[slash+1:]
+	}
+
+	return r.mr.Resolve(mname)
 }
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -478,11 +478,25 @@ func (s *Service) getFileDescriptorBySchemaID(schemaID int) (*desc.FileDescripto
 	return desc, exists
 }
 
+// AnyResolver is used to resolve the google.protobuf.Any type.
+// It takes a type URL, present in an Any message, and resolves 
+// it into an instance of the associated message.
+// 
+// This custom resolver is required because the built-in / default
+// any resolver in the protoreflect library, does not consider any
+// types that are used in referenced types that are not directly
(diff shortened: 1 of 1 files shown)


backend/pkg/proto/service.go
```diff
@@ -473,3 +477,16 @@ func (s *Service) getFileDescriptorBySchemaID(schemaID int) (*desc.FileDescripto
...
+
+type anyResolver struct {
+	mr *msgregistry.MessageRegistry
+}
+
+func (r *anyResolver) Resolve(typeURL string) (protoiface.MessageV1, error) {
+	mname := typeURL
+	if slash := strings.LastIndex(mname, "/"); slash >= 0 {
+		mname = mname[slash+1:]
+	}
```
weeco, line 489: @chris-palmer-deltatre What's the reason for stripping the last slash? When would that occur, how is that an issue. Could you add a comment for that?

weeco: @chris-palmer-deltatre Any chance that you remember this? Otherwise i'll try to reverse engineer it a bit

chris-palmer-deltatre: Hi there apologies! Yes I do remember this bit. So basically I pulled this code from the internals of jhump/protoreflect's internal anyResolver. The reason its there is that type URLs often have the form "type.googleapis.com/myprotopackage.MyProtoType". However they will be indexed in the registry in the format "myprotopackage.MyProtoType". Therefore the contents before the last slash is removed so that type resolution can occur properly against the registry.

If I am honest, I am not sure about it in every case, but it seemed to work for the types I was using in the `Any` field for our purposes.

chris-palmer-deltatre: So if you see the code here: https://github.dev/jhump/protoreflect/blob/master/dynamic/json.go#L1188

You will see that the internal `anyResolver` (which actually will invoke our custom `anyResolver` on line L1194 is strips the contents before the slash. However, when calling our custom `anyResolver` via `other.Resolve` on L1194, it passes the full typeURL which includes the contents before the slash. Therefore I copied the code from L1188 and put it in out custom `anyResolver` so I could replicate the same behavior.

chris-palmer-deltatre: I think it makes sense because the internal `anyResolver` uses `mname` to query the registry anyway: https://github.com/jhump/protoreflect/blob/master/dynamic/json.go#L1203

We are also trying to query the registry in our custom `anyResolver`, so I guess it makes sense that I needed to replicate this behavior in order to get it working.

weeco: Thanks @chris-palmer-deltatre , I tried to summarize that in a comment that I pushed into your branch for this PR

weeco:
🚀 Amazing PR, thank you!!

chris-palmer-deltatre:
Pleasure !

sashayakovtseva:
what if I have multiple repositories with proto files which Any can reference?

PR #431, opened 2022-08-15
helm: git credentials secret

chris-palmer-deltatre:
This commit adds support for a embedding a github basic auth password as a kubernetes `secret` in the helm chart.

Before this change, when setting up a git repository as a protobuf registry for
protobuf deserialization, you could provide basic auth credentials through the helm chart.
However you could not provide the secret to the helm chart without exposing it in a `configmap` resource.

After this change, you can now provide it to the helm chart via:

```yaml
secret:
    kafka:
        protobufGitBasicAuthPassword: <MYPASSWORD>
```

The change follows the pattern already present in the helm chart for console secrets:
1. add the value to the `secret` resource
2. map it the value from the `secret` to an environment variable using the `deployment` resource.
3. the environment variable name used is the config path that you would typically use in the yaml with the `.`s replaced by `_`s: i.e. `kafka.protobuf.git.basicAuth.Password`, becomes `KAFKA_PROTOBUF_GIT_BASICAUTH_PASSWORD`.

By doing this, its now possible for the user of the helm chart to embed the github basic auth password in the helm chart as a kubernetes `secret`.

### DIFF
--- a/helm/console/templates/deployment.yaml
+++ b/helm/console/templates/deployment.yaml
@@ -110,6 +110,13 @@ spec:
                   name: {{ include "console.fullname" . }}
                   key: kafka-sasl-password
             {{- end }}
+             {{- if .Values.secret.kafka.protobufGitBasicAuthPassword }}
+            - name: KAFKA_PROTOBUF_GIT_BASICAUTH_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "console.fullname" . }}
+                  key: kafka-protobuf-git-basicauth-password
+            {{- end }}
             {{- if .Values.secret.kafka.awsMskIamSecretKey }}
             - name: KAFKA_SASL_AWSMSKIAM_SECRETKEY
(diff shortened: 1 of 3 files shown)


helm/console/templates/deployment.yaml
```diff
@@ -109,6 +109,13 @@ spec:
                 secretKeyRef:
                   name: {{ include "console.fullname" . }}
                   key: kafka-sasl-password
+            {{- end }}
+             {{- if .Values.secret.kafka.protobufGitBasicAuthPassword }}
```
weeco, line 113: The indent is off by 1 space I believe

weeco (commented):
LGTM, only the indent is off by 1 space. Thanks!

helm/console/templates/deployment.yaml
chris-palmer-deltatre: no problem adjusted that now

weeco (approved):

PR #450, opened 2022-09-03
Add socks5 proxy support

stainboy:
This PR adds socks5 proxy support to redpanda/console (fka kowl). Refer [RFC 1928](https://rfc-editor.org/rfc/rfc1928.html) and [RFC 1929](https://rfc-editor.org/rfc/rfc1929.html).

Usage:
```
ssh -N -D localhost:1080 <bastion_server>
export all_proxy=socks5h://localhost:1080
export KAFKA_BROKERS=<brokers_behind_protected_network>
./console
```

### DIFF
--- a/backend/pkg/kafka/config_helper.go
+++ b/backend/pkg/kafka/config_helper.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/twmb/franz-go/pkg/sasl/aws"
 	"github.com/twmb/franz-go/pkg/sasl/oauth"
+	"golang.org/x/net/proxy"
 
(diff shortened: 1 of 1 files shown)


weeco:
Hey @stainboy ,
just FYI I looked at the PR but I'm a bit hesitant to merge. Generally all looks good to me, the only reason why I'm hesitant is that we usually provide all config options via YAML + Env via the explicit config options that Console offers.

If you want to get this merged, I'd like to ask you to add this to our usual configuration structs including validation (all config structs have a validate method). When I find time I would the same, but I have never heard about the need for this. Thus I would also be interested in your usecase and how this option helps you.

Thank you for filing the PR!

PR #460, opened 2022-09-20
feat(frontend): Migrate to react-router-dom v6 (WIP)

malinskibeniamin:
**In short:**
`v6.3.0` is a pain in the butt, but `v6.4.0` is even worse (!), because they are hiding the `history` object completely! That means you rely on abstractions and they handle everything internally within the library. Read more [here](https://reactrouter.com/en/main/start/overview#client-side-routing) if you are curious... or save yourself the headache and stay away (for now).

**Changes**
1. Removed `HistorySetter` in favour of `history.provider.ts` like we do in the Cloud UI
2. Used `unstable_HistoryRouter` to match Cloud UI
3. Used new hooks wherever possible
4. Used history provider for non-functional components
5. Fixed props problems for components (see below) where possible
6. Removed `Switch` in favour of `Routes`
7. Removed `exact` prop (this is no longer needed in v6, but nested routes should have `route/*` `*` wildcard)

**WIP**
`<Route>` component which accepts props needs to receive them either via query params with `useQueryParams` (only in hooks) or `useRouteMatch` (for any matching route, also only for functional components).
Did not fix any tests

**See links:**
https://reactrouter.com/en/6.4.0/upgrading/v5
https://reactrouter.com/en/v6.3.0/getting-started/concepts


frontend/src/components/routes.tsx
```diff
@@ -118,25 +119,29 @@ function EmitRouteViews(entries: IRouteEntry[]): JSX.Element[] {
...
-                    </Section>
-                )
-            }} />
 
-        </Switch>
+                )
+            }} /> */}
+            <Route element={
+                <Section title="404">
+                    <>{uiState.pageTitle = '404'}</>
```
malinskibeniamin, line 138: Nasty workaround since the route is no longer accepting a callback, just a component directly

weeco:
@malinskibeniamin Hey Ben, is this still WIP or can I close this PR?

PR #466, opened 2022-09-28
refactor: card to section

yougotashovel:
* Refactors Card to Section component for dark mode.
* Adds PageContent to abstract motion.div for each page.

Note: does not fix tables etc, that PR is coming separately.

image

image

### DIFF
--- a/frontend/src/components/pages/connect/CreateConnector.tsx
+++ b/frontend/src/components/pages/connect/CreateConnector.tsx
@@ -11,15 +11,12 @@
 
 import React, { useState } from 'react';
 import { PageComponent, PageInitHelper } from '../Page';
-import { animProps } from '../../../utils/animationProps';
-import { motion } from 'framer-motion';
 import {
     ApiOutlined,
     DatabaseOutlined,
     SearchOutlined,
 } from '@ant-design/icons';
 import { Wizard, WizardStep } from '../../misc/Wizard';
-import Card from '../../misc/Card';
 import { observer } from 'mobx-react';
 import { api } from '../../../state/backendApi';
 import { appGlobal } from '../../../state/appGlobal';
@@ -35,6 +32,8 @@ import KowlEditor from '../../misc/KowlEditor';
 import { useHistory } from 'react-router-dom';
 
 import styles from './CreateConnector.module.scss';
+import Section from '../../misc/Section';
+import PageContent from '../../misc/PageContent';
 
 const { Option } = Select;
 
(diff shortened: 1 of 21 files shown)


frontend/src/components/pages/connect/CreateConnector.tsx
```diff
@@ -100,11 +99,13 @@ class CreateConnector extends PageComponent {
...
-                    <ConnectorWizard connectClusters={clusters} />
-                </Card>
-            </motion.div>
+            <PageContent>
+                <Section>
+                    <div className={styles.wizardView}>
+                        <ConnectorWizard connectClusters={clusters} />
+                    </div>
+                </Section>
+            </PageContent>
```
rikimaru0345, line 108: The card previously had the wizard view style, now the div with the wizard view is nested inside the new component. Are you sure this will display the same as before?
If wizardView has margin for example, nesting might give a different result.

frontend/src/components/pages/consumers/Group.Details.tsx
```diff
@@ -128,11 +128,11 @@ class GroupDetails extends PageComponent<{ groupId: string }> {
...
 
         return (
-            <MotionDiv style={{ margin: '0 1rem' }} className="groupDetails">
+            <PageContent className="groupDetails">
                 {/* Statistics Card */}
                 {uiSettings.consumerGroupDetails.showStatisticsBar && (
-                    <Card className="statisticsBar">
-                        <Row>
+                    <Section py={4}>
+                        <Row className="statisticsBar">
```
rikimaru0345, line 135: statistics bar style was moved from the container to the row, will it have the same effect?

frontend/src/components/pages/schemas/Schema.List.tsx
```diff
@@ -27,25 +24,29 @@ import './Schema.List.scss';
...
+        </Section>
     );
 }
 
 function renderNotConfigured() {
     return (
-        <motion.div {...animProps} key={'b'} style={{ margin: '0 1rem' }}>
-            <Card style={{ padding: '2rem 2rem', paddingBottom: '3rem' }}>
+        <PageContent>
+            <Section>
```
rikimaru0345, line 49: old styles are not present in the new one, are they included indirectly?

rikimaru0345 (changes requested):
Generally I approve, just got a few questions

frontend/src/components/pages/schemas/Schema.List.tsx
yougotashovel: These styles aren't necessary anymore.  PageContent will space items correctly, and takes care of the entrance animation, so we can change it in one place if we need to in the future.

frontend/src/components/pages/connect/CreateConnector.tsx
yougotashovel: Here's the wizard view.  It still works the same.  Section doesn't have 'className' as a prop, so i added the div inside incase you were targeting the class for any reason.

image

frontend/src/components/pages/consumers/Group.Details.tsx
yougotashovel: I just pushed a change that should keep the fade effect you have on this working as it does for now.

image

PR #473, opened 2022-09-30
console: upgrade Node to v18 (upcoming LTS)

malinskibeniamin:
https://github.com/nodejs/release#release-schedule


yougotashovel:
@sago2k8 @rikimaru0345 I doubt this has an effect on our integration between cloud and console, but we should check that it still works after this.

malinskibeniamin:
I won't merge this now because I haven't checked it @yougotashovel.

If anyone has time to verify, please feel free to approve. Probably best to wait for 2 approvals just to be sure.

PR #477, opened 2022-10-04
frontend: Add Sentry

malinskibeniamin:
Added Sentry to `console-ui` project in the Redpanda team


rikimaru0345 (approved):

sago2k8 (changes requested):
This logic is not relevant for console, and it shouldn't go there. The sentry error handler will trigger even if the error happen on console, the real fix is to propagate the errors back to cloud-ui.

PR #499, opened 2022-10-25
add kics scanning 

wreet:
add kics IAC security scanning to the console project


weeco (approved):

wreet:
@weeco had to make some tweaks because act local testing is not quite the same as GH actions even though it is advertised as such haha. still looking good?

weeco:
LGTM, @wreet please go ahead :)

PR #505, opened 2022-10-27
[helm] Allow entering secret references not only plain text secrets

missingcharacter:


joejulian (approved):
LGTM! Thanks!

missingcharacter:
@joejulian will the new helm chart version be available after merge?

weeco:
Hey @missingcharacter ,
thanks for contributing.

I'm hesitant to merge, because I envisioned our users to use the more generic approach of injecting these kind of configuration via `extraEnv` etc., see: https://github.com/redpanda-data/console/blob/0655b5134ca08739eb101cb92813894439b80e90/helm/console/values.yaml#L95-L121

Console offers a lot of configuration options, and extending the chart so that you can always provide secrets via two options (as a value or secret ref) is not going to be maintainable in my opinion. We may need to do a better job by adding example configurations for that, but generally I believe that this is the better approach. What do you think @missingcharacter ?

missingcharacter:
@weeco I am ok following whatever pattern your team wants to follow, I just thought this would make it more obvious to people what secret goes where. I will try setting up the console using `extraEnv`, `extraEnvFrom` and `extraVolumes`

missingcharacter:
See related https://github.com/redpanda-data/console/pull/507

PR #507, opened 2022-10-28
Extras are lists and not strings

missingcharacter:
Without this change helm throws error:

```text
Error: template: console/templates/deployment.yaml:103:20: 
executing "console/templates/deployment.yaml" at <.>: 
wrong type for value; expected string; got []interface {}
```

Also I am following advice from https://github.com/redpanda-data/console/pull/505#issuecomment-1294657590
from closed PR https://github.com/redpanda-data/console/pull/505

### DIFF
--- a/helm/console/Chart.yaml
+++ b/helm/console/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.1
+version: 0.3.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/helm/console/values.yaml
+++ b/helm/console/values.yaml
@@ -93,17 +93,17 @@ console:
   # roleBindings:
 
 # Additional environment variables for the Console Deployment
-extraEnv: ""
+extraEnv: []
   # - name: KAFKA_RACKID
   #   value: "1"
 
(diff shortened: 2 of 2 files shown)


joejulian (dismissed):
lgtm

weeco:
Hey @missingcharacter ,
based on the error message I believe the values.yaml included a config block like this:

```yaml
extraEnv:
  - name: LOGIN_OIDC_CLIENTSECRET
    valueFrom:
      secretKeyRef:
        name: "secret"
        key: "key"
```

Because we are using Helm's tpl function, you'd need to provide a string actually. Try to add a | in the line after extraEnv:. Ref https://github.com/redpanda-data/console/blob/0655b5134ca08739eb101cb92813894439b80e90/helm/console/templates/deployment.yaml#L102-L104 . 

  A config block like this should work:

```
extraEnv: |
  - name: LOGIN_OIDC_CLIENTSECRET
    valueFrom:
      secretKeyRef:
        name: "secret"
        key: "key"
```

The tpl function offers more flexibility, so that you can reuse variables which turned out to be handy for some users. A typical usecase where this is handy is the topologySpreadConstraints, e.g.:

```
  topologySpreadConstraints: |
    - maxSkew: 1
      topologyKey: failure-domain.beta.kubernetes.io/zone
      whenUnsatisfiable: ScheduleAnyway
      labelSelector:
        matchLabels:
          {{- include "console.selectorLabels" . | nindent 6 }}
```

I hope this helps.

joejulian:
Thanks, @weeco. I didn't notice the template file interprets that string with tpl.

missingcharacter:
Thanks @weeco

PR #522, opened 2022-11-08
proto: allow multiple import paths

birdayz:
add new config parameter proto.importPaths.
if import paths are configured, these paths are all treated as roots. if not configured, the already existing behavior is preserved.

in addition, if git.basePath is empty / "." (default), the dot from file extensions is dropped, resuling in paths where the dot of the file extension is missing. the dot is now not stripped anymore.

### DIFF
--- a/backend/pkg/config/proto.go
+++ b/backend/pkg/config/proto.go
@@ -24,6 +24,11 @@ type Proto struct {
 
 	// Mappings define what proto types shall be used for each Kafka topic. If SchemaRegistry is used, no mappings are required.
 	Mappings []ProtoTopicMapping `json:"mappings"`
+
+	// Proto import paths. By default, the baseDir/root is used. Set import paths
+	// if multiple import paths relative to the baseDir shall be used. The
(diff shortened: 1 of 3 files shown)


backend/pkg/config/proto.go
```diff
@@ -24,6 +24,11 @@ type Proto struct {
 
 	// Mappings define what proto types shall be used for each Kafka topic. If SchemaRegistry is used, no mappings are required.
 	Mappings []ProtoTopicMapping `json:"mappings"`
+
+	// Proto import paths. By default, the baseDir/root is used. Set import paths
+	// if multiple import paths relative to the baseDir shall be used. The
+	// behavior is similar to the `-I` flag of protoc.
+	ImportPaths []string `json:"import_paths"`
```
weeco, line 31:
```suggestion
	ImportPaths []string `json:"importPaths"`
```

weeco (approved):

birdayz:
Tested with filesystem proto provider as well.

PR #529, opened 2022-11-10
backend: Add hook for protecting specific Kafka users

weeco:


birdayz (approved):
lgtm, just one minor comment about some text.

PR #536, opened 2022-11-13
Auto Refresh button

victorgawk:
Add an "Auto Refresh" button next to the "Force Refresh" button:

image

When pressed, the page will be automatically force refreshed every `X` seconds where `X` can be configured in the "User Preferences" panel:

image

When the auto refresh is active, the "Auto Refresh" and "Force Refresh" button will be displayed with pulsating and rotation animations respectively:

Animação

Pressing the button again disables the auto refresh.

### DIFF
--- a/frontend/src/components/misc/UserPreferences.tsx
+++ b/frontend/src/components/misc/UserPreferences.tsx
@@ -243,7 +243,7 @@ class AutoRefreshTab extends Component {
                                 uiSettings.autoRefreshIntervalSecs = e;
                             }
                         }}
-                        min={1} max={60}
+                        min={5} max={300}
                         style={{ maxWidth: '150px' }}
                     />
                 </Label>
--- a/frontend/src/components/misc/UserPreferences.tsx
+++ b/frontend/src/components/misc/UserPreferences.tsx
@@ -23,6 +23,7 @@ const settingsTabs: { name: string, component: () => ReactNode }[] = [
     { name: 'Statistics Bar', component: () => <StatsBarTab /> },
     { name: 'Json Viewer', component: () => <JsonViewerTab /> },
     { name: 'Import/Export', component: () => <ImportExportTab /> },
+    { name: 'Auto Refresh', component: () => <AutoRefreshTab /> },
 
     // pagination position
     // { name: "Message Search", component: () => <MessageSearchTab /> },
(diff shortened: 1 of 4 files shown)


weeco:
Hey @victorgawk ,
the auto refresh button is a great idea - thank you for coming up with this PR!

I have two thoughts and one question:

1. Some pages, such as the consumer groups page, are expected to load slow if you have a large number of consumer groups. That's because we have to issue a series of Kafka requests in order to collect all the showed information. Thus, we should never abort ongoing requests, and only start the interval timer after the previous request has been finished. We should also make this behaviour clear to the user, so that they are not surprised that they don't see up-to-date values in case the requests take a long time.
2. We should enforce a minimum auto refresh interval of 5seconds 

Question: On the topics page with messages, would this also retrigger the search over and over again? This may be undesired behaviour by a user.

victorgawk:
> Question: On the topics page with messages, would this also retrigger the search over and over again? This may be undesired behaviour by a user.

As I have observed, the message search is not called when `Force Refresh` button is pressed which is essentially the same action used by the `Auto Refresh` button.

weeco:
> about the consumer groups page, bye PR https://github.com/redpanda-data/console/pull/538 maybe?

I'm afraid this is the case, I'm sorry! I acknowledged the work you put into this. If you plan to commit more features we are also always available to discuss this pre implementation via GitHub or via the Redpanda slack.

> Trying to solve the problem, should I replace the behavior:

What you suggest there makes sense to me, yes please!

> What about the default and the maximum values?

Default: 10 sec
Minimum: 5 sec
Maximum: ? Should we limit the maximum? 60 secs or 5min sounds reasonable anyways.


Thanks!

victorgawk:
- Added a countdown span showing when the next refresh will happen.
  - For the countdown update, the interval function was fixed with a 250 milliseconds rate in order to display the countdown decreasing at almost real time.
- To avoid auto refresh trigger while another refresh is still running, the auto refresh timeout was changed to only trigger when no other API call is ongoing (`api.activeRequests.length == 0`).
- Updated auto refresh interval values:
  - default: 5 -> 10 seconds
  - minimum: 1 -> 5 seconds
  - maximum: 60 -> 300 seconds

Visual Preview:
Animação

rikimaru0345 (approved):

weeco:
Thanks @victorgawk , great first contribution :).

PR #537, opened 2022-11-13
On topic list, add the number of messages for each topic

victorgawk:
Add a column on topic list showing the current number of messages for each topic:

image


weeco:
The number of messages on compacted topics can only be estimated as there can be gaps between the low and high watermarks. So for any topic that is compacted we need to make that clear or entirely omit the number of records in these columns.

I'm a little concerned about the performance of the lookup again - but I would need to double check the impact on larger clusters with many partitions. This page should load very quickly and is already doing a bunch of requests that can fail.

If it turns out to be okay to add - we need to shorten the numbers. It's not unusual that topics have accumulated multi millions or even billions of records.

weeco:
Currently I'm leaning towards not merging this as the message count is only accurate in certain topics. I may reconsider it again in the future. 

The available space in that table is also limited and interferes with other information we may want to display in that table. I can imagine that we make this table customizable so that everyone can configure what they think is important to their setup.

PR #538, opened 2022-11-13
On topic list, add lag column for each topic

victorgawk:
Similar to PR #537, add a new column on the topic list to show the lag information for each topic.

image

Since lag is a consumer group related information and a topic can be in zero or more consumer groups. Some decisions was made about what to display on that column:

- if the topic has no consumer groups, the column value will be empty;
- otherwise, the column value will be the max lag from all consumer groups that this topic is part of (e.g., if a topic is part of two consumer groups, one with lag 10 and another with lag 20, then the column value will be 20).
- If the topic is part of a consumer group which has other topics inside, the lag from these other topics will not be considered for the topic lag.

In addition, when mouse hover the column value, it will show a list with all topic consumer groups and their respective lags:

Animação

### DIFF
--- a/backend/pkg/console/topic_overview.go
+++ b/backend/pkg/console/topic_overview.go
@@ -32,16 +32,27 @@ const (
 
 // TopicSummary is all information we get when listing Kafka topics
 type TopicSummary struct {
-	TopicName         string             `json:"topicName"`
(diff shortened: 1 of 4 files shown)


weeco:
I'm closing this PR due to the cost of the requests that would be required for this in larger clusters. Thanks for your work nevertheless!

PR #549, opened 2022-11-22
Helm: Set same key for login-oidc-client-secret in Secret and Deployment

JoschaLaubach:
Currently, in the Helm chart the name for the login-oidc-client-secret key in the Secret does not match the key reference in the Deployment. In the Deployment the key login-oidc-client-secret is referenced, but in the Secret it is named login-generic-oidc-client-secret. I would drop the "generic-" part in the Secret, because in the values file it is configured under secret.login.oidc.clientSecret, also without "generic".

Key in the Deployment: https://github.com/redpanda-data/console/blob/master/helm/console/templates/deployment.yaml#L207

### DIFF
--- a/helm/console/Chart.yaml
+++ b/helm/console/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.1
+version: 0.3.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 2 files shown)


joejulian:
Hey, @JoschaLaubach, would you please open this PR against [redpanda-data/helm-charts](https://github.com/redpanda-data/helm-charts)? We've moved the helm chart there.

JoschaLaubach:
> Currently, in the Helm chart the name for the login-oidc-client-secret key in the Secret does not match the key reference in the Deployment. In the Deployment the key login-oidc-client-secret is referenced, but in the Secret it is named login-generic-oidc-client-secret. I would drop the "generic-" part in the Secret, because in the values file it is configured under secret.login.oidc.clientSecret, also without "generic".
> 
> Key in the Deployment:

Done, see https://github.com/redpanda-data/helm-charts/pull/242

PR #583, opened 2022-12-21
backend: Fix panic when deserializing schema registry wrapper

weeco:
When decoding a binary message that by accident has a valid starting sequence we may ended up with a slice allocation that is too large. This commit will mitigate the problem by returning an error in all cases where we would want to allocate more than 128 message type ids.

### DIFF
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -22,12 +22,13 @@ import (
 	"github.com/jhump/protoreflect/desc/protoparse"
 	"github.com/jhump/protoreflect/dynamic"
 	"github.com/jhump/protoreflect/dynamic/msgregistry"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/runtime/protoiface"
+
 	"github.com/redpanda-data/console/backend/pkg/config"
 	"github.com/redpanda-data/console/backend/pkg/filesystem"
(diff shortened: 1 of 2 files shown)


backend/pkg/proto/service.go
```diff
@@ -310,6 +314,15 @@ func (s *Service) decodeConfluentBinaryWrapper(payload []byte) (*confluentEnvelo
...
+	// panic with: makeslice: len out of range. Therefore, we check this length
+	// before allocating the slice. We assume to not have more than 128 types in
+	// a single message
+	if arrLength > 128 {
+		return nil, fmt.Errorf("arrLength is higher than expected, unlikely a legit envelope")
+	}
+
 	msgTypeIDs := make([]int64, arrLength)
 	// If there is just one msgtype (default index - 0) the array won't be sent at all
 	if arrLength == 0 {
```
c4milo, line 328: would there be a chance for `arrLength` to be less than 0 for any reason?

weeco: Yes, good comment!

weeco: Fixed, added a check for negative length

c4milo (approved):
Seems good!

ivotron (approved):

Issue #405, opened 2022-06-29
Support regex in topic protobuf mappings


upils:
For the protobuf topic mapping, we currently need to set the exact topic name.

We have a use case where our topic names are generated dynamically so we cannot give the full list in the configuration. But they all contains the same kind of message, so we would like to deserialize them all the same way.

We would like a way to give a regex in the topicName so all of our topics would match.

What do you think about this feature?

The current implementation is filling a map with the topicName as the key. https://github.com/redpanda-data/console/blob/222b498ec0295acf0ea33888f20aa3a0440757f6/backend/pkg/proto/service.go#L96

Then this map is used to check if a mapping exists for a given topic: https://github.com/redpanda-data/console/blob/222b498ec0295acf0ea33888f20aa3a0440757f6/backend/pkg/proto/service.go#L238

We could consider the topicName as a regex, try to compile it when the config is loaded (and return an error if this is invalid) and keep using the given string as the key in the map.

Then after testing if the topic given in `getMessageDescriptor()` exists in the map, we would loop on the map to check if the topic match any regex. This would be a bit slower than the current implementation.

weeco:
Hey @upils ,
thanks for sharing your usecase. I think what you are suggesting makes sense to me. Do you want to work on this and submit a PR for this?

We should make sure that imports / nested types are supported. - regardless whether we use regexed strings or not. If I recall correctly we use protoreflect's proto registry to register all proto types in advance, but if your topics share the some proto schemas (with the same name) this should not be a problem I think.

upils:
Thanks for the reply.

Yes I might give it a try.

I am not sure to understand what you mean about the proto registry. It might be clearer if I dive in the code.

weeco:
I was referring to this part https://github.com/redpanda-data/console/blob/222b498ec0295acf0ea33888f20aa3a0440757f6/backend/pkg/proto/service.go#L370-L374 , but I think it's not relevant to your problem, because you are sharing the same protos across several topics and you only want to tell Console for what topics it should use the proto schemas.

upils:
OK. Yes, I understand it should not be an issue.

brycefisher:
I'd love to get this feature merged into master and shipped in a release. Are you still planning to work on this @upils or did you want another contributor to take over from here? No worries either way!

bojand:
This should be addressed now via #1291.

PR #584, opened 2022-12-21
Support regex in topic protobuf mappings

upils:
Fixing #405

This is still a work in progress (need more manual and unit tests) but If you find time you can validate you are OK with the direction I am heading.

I have concerns about performance if every message end up matching with the last regex in a long list. I may add a benchmark to get an idea of the difference between regex matching vs directly matching in the map (I expect a huge difference).

### DIFF
--- a/backend/pkg/config/proto_topic_mapping.go
+++ b/backend/pkg/config/proto_topic_mapping.go
@@ -17,9 +17,6 @@ type ProtoTopicMapping struct {
 	// TopicName is the name of the topic to apply these proto files. This supports regex.
 	TopicName string `yaml:"topicName"`
 
-	// IsRegex indicates if the topicName must be handled as a regex and match multiple topic names
-	IsRegex bool `yaml:"isRegex"`
-
 	// KeyProtoType is the proto's fully qualified name that shall be used for a Kafka record's key
 	KeyProtoType string `yaml:"keyProtoType"`
 
--- a/backend/pkg/config/proto_topic_mapping.go
+++ b/backend/pkg/config/proto_topic_mapping.go
@@ -14,7 +14,11 @@ package config
 // if the topics have been serialized without the schema registry being involved in the
 // serialization.
 type ProtoTopicMapping struct {
+	// TopicName is the name of the topic to apply these proto files. This supports regex.
 	TopicName string `yaml:"topicName"`
+
+	// IsRegex indicates if the topicName must be handled as a regex and match multiple topic names
+	IsRegex bool `yaml:"isRegex"`
 
 	// KeyProtoType is the proto's fully qualified name that shall be used for a Kafka record's key
 	KeyProtoType string `yaml:"keyProtoType"`
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -48,7 +48,7 @@ const (
 	RecordValue
 )
 
-type RegexProtoTopicMapping struct {
+type regexProtoTopicMapping struct {
 	config.ProtoTopicMapping
 	r *regexp.Regexp
 }
@@ -61,7 +61,7 @@ type Service struct {
 	logger *zap.Logger
 
 	strictMappingsByTopic map[string]config.ProtoTopicMapping
-	regexMappingsByTopic  map[string]RegexProtoTopicMapping
+	regexMappingsByTopic  map[string]regexProtoTopicMapping
 	gitSvc                *git.Service
 	fsSvc                 *filesystem.Service
 	schemaSvc             *schema.Service
@@ -139,9 +139,9 @@ func NewService(cfg config.Proto, logger *zap.Logger, schemaSvc *schema.Service)
 	}, nil
 }
 
-func setMappingsByTopic(mappings []config.ProtoTopicMapping) (strictMappingsByTopic map[string]config.ProtoTopicMapping, regexMappingsByTopic map[string]RegexProtoTopicMapping, err error) {
+func setMappingsByTopic(mappings []config.ProtoTopicMapping) (strictMappingsByTopic map[string]config.ProtoTopicMapping, regexMappingsByTopic map[string]regexProtoTopicMapping, err error) {
 	strictMappingsByTopic = make(map[string]config.ProtoTopicMapping)
-	regexMappingsByTopic = make(map[string]RegexProtoTopicMapping)
+	regexMappingsByTopic = make(map[string]regexProtoTopicMapping)
 
 	for _, mapping := range mappings {
 		if mapping.IsRegex {
@@ -150,7 +150,7 @@ func setMappingsByTopic(mappings []config.ProtoTopicMapping) (strictMappingsByTo
 				return nil, nil, fmt.Errorf("invalid regexp as a topic name: %w", err)
 			}
 
-			regexMappingsByTopic[mapping.TopicName] = RegexProtoTopicMapping{
+			regexMappingsByTopic[mapping.TopicName] = regexProtoTopicMapping{
 				ProtoTopicMapping: mapping,
 				r:                 r,
 			}
@@ -159,7 +159,7 @@ func setMappingsByTopic(mappings []config.ProtoTopicMapping) (strictMappingsByTo
 		strictMappingsByTopic[mapping.TopicName] = mapping
 	}
 
-	return
+	return strictMappingsByTopic, regexMappingsByTopic, err
 }
 
 // Start polling the prototypes from the configured provider (e.g. filesystem or Git) and sync these
--- a/backend/pkg/proto/service.go
+++ b/backend/pkg/proto/service.go
@@ -119,23 +119,9 @@ func NewService(cfg config.Proto, logger *zap.Logger, schemaSvc *schema.Service)
 		}
 	}
 
-	strictMappingsByTopic := make(map[string]config.ProtoTopicMapping)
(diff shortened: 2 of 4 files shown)


upils:
I fixed merge conflicts. @weeco / @rikimaru0345 can you take a look on this?

weeco:
Hey @upils ,
I share your concerns in regards to the performance. This is code executed in the hot-path if you search a topic's messages with search filters enabled. It wouldn't be uncommon to process ~100k records / seconds. I think there are solutions to this problem - such as caching the results for requested types etc.

upils:
@weeco thanks for your answer. I now add matching topics in the strictMappingsByTopic (see https://github.com/redpanda-data/console/pull/584/files#diff-cb0859875b7a3634d6b559a35d024fc14ba558a7960fabbd6287d716423246a7R296) so a given topic never have to be searched twice in the regexMappingsByTopic.

I wrote a benchmark with a variable proportion of strict/regex mappings.

Before the fix, performances were degrading following the ratio of regex mappings:

```
goos: linux
goarch: amd64
pkg: github.com/redpanda-data/console/backend/pkg/proto
cpu: Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
BenchmarkService_getMatchingMapping
BenchmarkService_getMatchingMapping/Only_strict_mappings
BenchmarkService_getMatchingMapping/Only_strict_mappings-8                  5998            186877 ns/op               8 B/op          0 allocs/op
BenchmarkService_getMatchingMapping/10%_regex_mappings
BenchmarkService_getMatchingMapping/10%_regex_mappings-8                     249           4464246 ns/op             451 B/op          2 allocs/op
BenchmarkService_getMatchingMapping/50%_regex_mappings
BenchmarkService_getMatchingMapping/50%_regex_mappings-8                      91          12583620 ns/op            2226 B/op         15 allocs/op
BenchmarkService_getMatchingMapping/90%_regex_mappings
BenchmarkService_getMatchingMapping/90%_regex_mappings-8                      58          17922738 ns/op            4674 B/op         41 allocs/op
PASS
ok      github.com/redpanda-data/console/backend/pkg/proto      5.692s
```

After the fix, it shows no significant difference:

```
goos: linux
goarch: amd64
pkg: github.com/redpanda-data/console/backend/pkg/proto
cpu: Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
BenchmarkService_getMatchingMapping
BenchmarkService_getMatchingMapping/Only_strict_mappings
BenchmarkService_getMatchingMapping/Only_strict_mappings-8                  5724            180517 ns/op               8 B/op          0 allocs/op
BenchmarkService_getMatchingMapping/10%_regex_mappings
BenchmarkService_getMatchingMapping/10%_regex_mappings-8                    7711            150304 ns/op              12 B/op          0 allocs/op
BenchmarkService_getMatchingMapping/50%_regex_mappings
BenchmarkService_getMatchingMapping/50%_regex_mappings-8                    7566            147093 ns/op              30 B/op          0 allocs/op
BenchmarkService_getMatchingMapping/90%_regex_mappings
BenchmarkService_getMatchingMapping/90%_regex_mappings-8                    6926            157498 ns/op              42 B/op          0 allocs/op
PASS
ok      github.com/redpanda-data/console/backend/pkg/proto      6.286s
```

I may add more use cases to the benchmark if I think of some corner cases.

backend/pkg/proto/service.go
```diff
@@ -251,12 +282,35 @@ func (s *Service) getMessageDescriptorFromConfluentMessage(wrapper *confluentEnv
...
 
+func (s *Service) getMatchingMapping(topicName string) (mapping config.ProtoTopicMapping, err error) {
+	mapping, strictExists := s.strictMappingsByTopic[topicName]
+	if strictExists {
+		return mapping, nil
+	}
+
+	var match bool
+	for _, rMapping := range s.regexMappingsByTopic {
+		match = rMapping.r.MatchString(topicName)
```
smatvienko-tb, line 293: This is a performance degradation for a non-matched topic.
Each time for non-matched topics, the algorithm iterate and match all known regex.

The maximum number of searches has to be estimated as total unique topic names O(N).

I suggest saving the `nil` result to the strict mapping `s.strictMappingsByTopic[topicName]` and checking on the `nil` output at the beginning of this function to avoid another loop searching by regex.

smatvienko-tb (changes requested):
Hey! Your feature is awesome and I really need it for my deployment!

Here you can find the solution how to avoid performance degradation to get a green light for this feature!

It also will be nice to have a small example in the configuration mapping in the same project. Because it is the only source of information right behind the mapping data.

Have a good day!

smatvienko-tb:
@upils , @weeco can we apprise this topic one final time?
It's almost done, guys!
I left my comment in the review on how to fix the complexity and keep the steam inside.
Can you help me?

brycefisher:
@upils were you able to sign the Contributor License Agreement? I honestly can't tell from the bot 😅 Looks like once that is sorted and you get an approval, this PR is ready for prime time!

backend/pkg/config/proto_topic_mapping.go
```diff
@@ -14,8 +14,12 @@ package config
 // if the topics have been serialized without the schema registry being involved in the
 // serialization.
 type ProtoTopicMapping struct {
+	// TopicName is the name of the topic to apply these proto files. This supports regex.
 	TopicName string `yaml:"topicName"`
 
+	// IsRegex indicates if the topicName must be handled as a regex and match multiple topic names
+	IsRegex bool `yaml:"isRegex"`
```
smatvienko-tb, line 21: After all suggestions... 
**Why** can we not approach any topic name as a regex? 
**Why** do we need to increase the complexity of the mapping?

Let's try to compile **every `topicName` as regex**. 
For the Kafka name policy, it is completely fine!

If the topic name can not be compiled, let's put a warning to the logs and recognize such a `topicName` as a strict match to not break any exotic deployments.

smatvienko-tb (changes requested):
Please, find my suggestions in the comments.

upils:
> @upils were you able to sign the Contributor License Agreement? I honestly can't tell from the bot 😅 Looks like once that is sorted and you get an approval, this PR is ready for prime time!

Yes, but I guess I messed up with the mail address I used in my commits. I will rewrite them and it should be good.

smatvienko-tb:
raising it up

twmb:
@upils ping on rewriting the commit emails.

upils:
Hey @smatvienko-tb @twmb @brycefisher . Sorry for this long hiatus. I fixed the commit author and started to apply the changes  suggested by @smatvienko-tb . I need to add test cases but let me know if this is what you had in mind.

smatvienko-tb (approved):
LGTM

weeco:
I left two comments. Overall I'd like to suggest to change this PR so that regex strings are recognized if the input is wrapped into two slashes (see my other PR comment). 

We can implement a custom type `RegexpOrLiteral` which implements the Stringer method and has a custom Unmarshaler, so that you can provide mappings like this:

```yaml
mappings:
  - topicName: normal-exact-match
    keyProtoType: "foo"
  - topicName: /some-topic-prefix-matc.*/
    keyProtoType: "bar"
```


You can do so by creating a type such as:

```go
type RegexpOrLiteral struct {
	literal string
	*regexp.Regexp
}

func (r *RegexpOrLiteral) String() string {
	if r.literal != "" {
		return r.literal
	}
	if r.Regexp != nil {
		return r.Regexp.String()
	}
	return ""
}

// UnmarshalText unmarshals json into a regexp.Regexp
func (r *RegexpOrLiteral) UnmarshalText(b []byte) error {
	regex, err := compileRegex(string(b))
	if err == nil {
		r.Regexp = regex
	}
	r.literal = string(b)
	return nil
}

// MarshalText marshals regexp.Regexp as string
func (r *RegexpOrLiteral) MarshalText() ([]byte, error) {
	if r.Regexp != nil {
		return []byte(r.Regexp.String()), nil
	}

	return []byte(r.literal), nil
}
```

Additionally koanf (our config management lib) needs to be configured so that it will try the TextUnmarshalling, by adding the `mapstructure.TextUnmarshallerHookFunc(),` option here: https://github.com/redpanda-data/console/blob/94ef19bcb9c8202040b3f8bf93ee1ead4fe94ae8/backend/pkg/config/config.go#L133

The `compileRegex` method already exists here and checks for the wrapped slashes: https://github.com/redpanda-data/console/blob/94ef19bcb9c8202040b3f8bf93ee1ead4fe94ae8/backend/pkg/config/util_regex.go#L19-L36 . By doing so you have the best of both worlds:

- Ability to provide regexes while being backwards compatible (a single string can be provided for the topic name)
- Regexes are compiled at unmarshal time so that we error early and users should clearly know whether regex is used or not

What do you think @upils @smatvienko-tb ?

brycefisher:
Hi @upils and @weeco! Thanks so much for all the hardwork on implementing and reviewing this feature so far. Is there anything I can do to help push this feature over the line?

twmb:
@brycefisher if you'd like to take this on, would you be able to fork and address suggestions from @weeco just above?

brycefisher:
Yes! Happy to put the finishing touches on this.

weeco:
@upils @smatvienko-tb @brycefisher ping

twmb:
We'll move this to unscheduled on our side, but please ping once y'all are able to uptake this and we can re-review.

PR #589, opened 2023-01-04
fix transaction

rikimaru0345:


frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
```diff
@@ -253,17 +256,19 @@ export const AclPrincipalGroupEditor = observer((p: {
...
-    const isConsumerGroup = !isCluster && !isTopic;
     const isAllSet = res.all == 'Allow' || res.all == 'Deny';
 
     let resourceName = 'Cluster';
-    if (isTopic) resourceName = 'Topic';
-    if (isConsumerGroup) resourceName = 'Consumer Group';
+    if (p.resourceType == 'Topic') resourceName = 'Topic';
+    if (p.resourceType == 'Group') resourceName = 'Consumer Group';
+    if (p.resourceType == 'TransactionalID') resourceName = 'Transactional ID';
+
```
yougotashovel, line 270: Can this be any other types?  Just wondering if there's a fallback.

yougotashovel (approved):

weeco (approved):

frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
rikimaru0345: acl type is:
``` 

export type AclStrResourceType =
    | 'Unknown'
    | 'Any'
    | 'Topic'
    | 'Group'
    | 'Cluster'
    | 'TransactionalID'
    | 'DelegationToken'
    ;

```

Unknown: should never be reported by the backend.
Any: is only for search.
DelegationToken: is not supported.

By default it shows Cluster as name.

PR #595, opened 2023-01-09
Add backend linting

weeco:


backend/pkg/api/websocket_client.go
```diff
@@ -44,8 +45,8 @@ func (wc *websocketClient) upgrade(w http.ResponseWriter, r *http.Request) *rest
 		if len(parts) != 2 {
 			continue
 		}
-		if parts[0] == "access_token" {
-			wc.access_token = parts[1]
```
rikimaru0345, line 48: should this really be renamed?

rikimaru0345 (approved):

PR #596, opened 2023-01-09
Making aws.s3.region a basic option for S3 sink connector

tomasz-sadura:
Making aws.s3.region a basic option for S3 sink connector, adding aws.s3.bucket.check definition.

### DIFF
--- a/backend/pkg/connect/guides/s3_sink_connector.json
+++ b/backend/pkg/connect/guides/s3_sink_connector.json
@@ -23,7 +23,7 @@
         "order": 2
       }
     },
-    
+
     {
       "nameSelector": "aws.s3.bucket.name",
       "definition": {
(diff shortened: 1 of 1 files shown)


backend/pkg/connect/guides/s3_sink_connector.json
```diff
@@ -57,26 +57,35 @@
...
       }
     },
+    {
+      "nameSelector": "aws.s3.bucket.check",
+      "definition": {
+        "display_name": "AWS S3 bucket check",
+        "group" : "Amazon S3",
+        "order": 5
+      }
+    },
```
weeco, line 71: @tomasz-sadura Is that an option that is provided by the targetted sink connector?

tomasz-sadura: yes

weeco (approved):

PR #598, opened 2023-01-17
backend: Add CORS handling to prevent CSRF attacks

weeco:
Add CORS handling into Console. Prior to this commit we default to a same-site policy for all HTTP endpoints and allowed all origins for initiating a websocket connection


d3k4z (approved):
approved! gj @weeco

PR #602, opened 2023-01-26
adds snyk scanning workflow

wreet:
this workflow will scan the `master` and `release-*` branches and send results to snyk.

### DIFF
--- a/.github/workflows/backend-snyk-scan.yml
+++ b/.github/workflows/backend-snyk-scan.yml
@@ -1,0 +1,24 @@
+name: snyk scanning
+
+on:
+  push:
+    branches: ["master", "release-*"]
+
+  workflow_dispatch:
+
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: install snyk
+        run: npm install -g snyk
+
+      - name: install deps and scan
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/backend-snyk-scan.yml
```diff
@@ -0,0 +1,24 @@
+name: snyk scanning
+
+on:
+  push:
+    branches: ["master", "release-*"]
+
+  workflow_dispatch:
```
weeco, line 7: What is this for?

wreet: Since the CLI scanning won't just be pulling your whole repo, we have to define which branches we want it to scan. In this case we're doing master and release branches but if you don't think that's right we can change it. `workflow_dispatch` just lets us run it manually from the UI if there is a need to, for example if a scan fails because of an API error or something.

weeco: @wreet Sorry I meant to only highlight the workflow dispatch. Can you explain how that works? Because there's no configuration attached to the `workflow_dispatch` property

wreet: Ah I see what you mean. That property does not require any additional configuration since we are not passing any inputs to the action.

weeco (approved):

PR #608, opened 2023-02-01
Secret management

sago2k8:
Pr for state handling on the connectors page, there are still a bunch of debug statements and logs that will be removed eventually, focus on frontend/src/state/connect/state.ts

### DIFF
--- a/frontend/src/components/pages/connect/CreateConnector.tsx
+++ b/frontend/src/components/pages/connect/CreateConnector.tsx
@@ -10,6 +10,7 @@
  */
 
 import React, { useEffect, useState } from 'react';
+import { comparer } from 'mobx';
 import { message } from 'antd';
 import { PageComponent, PageInitHelper } from '../Page';
 import { ApiOutlined, DatabaseOutlined, SearchOutlined } from '@ant-design/icons';
@@ -130,7 +131,10 @@ const ConnectorWizard = observer(({ connectClusters, activeCluster }: ConnectorW
     const [validationFailure, setValidationFailure] = useState<unknown>(null);
     const [creationFailure, setCreationFailure] = useState<unknown>(null);
     const [genericFailure, setGenericFailure] = useState<Error | null>(null);
-    const [stringifiedConfig, setStringifiedConfig] = useState<null | any>({});
+    const [stringifiedConfig, setStringifiedConfig] = useState<string>('');
+    const [parsedUpdatedConfig, setParsedUpdatedConfig] = useState<any | null>(null);
+    const [postCondition, setPostCondition] = useState<boolean>(false);
+    const [loading, setLoading] = useState<boolean>(false);
     const [connectClusterStore, setConnectClusterStore] = useState(ConnectClusterStore.getInstance(activeCluster));
 
     useEffect(() => {
@@ -143,6 +147,16 @@ const ConnectorWizard = observer(({ connectClusters, activeCluster }: ConnectorW
     useEffect(() => {
         setConnectClusterStore(ConnectClusterStore.getInstance(activeCluster));
     }, [activeCluster]);
+
+    useEffect(() => {
+        try {
+            setParsedUpdatedConfig(JSON.parse(stringifiedConfig));
+        } catch (e) {
+            setParsedUpdatedConfig(null);
+            setPostCondition(false);
+        }
+        setPostCondition(true);
+    }, [stringifiedConfig]);
 
     const clearErrors = () => {
         setCreationFailure(null);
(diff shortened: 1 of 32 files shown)


frontend/src/utils/utils.ts
```diff
@@ -753,3 +753,21 @@ export function setHeader(init: RequestInit, name: string, value: string) {
...
+// very simple retrier utility for allowing some retries, if we ended up using it more often we should consider making it more elaborate
+
+export const retrier = (
+    operation: any, { attempts = Infinity, delayTime = 100 }) => new Promise((resolve, reject) => {
+        return operation()
+            .then(resolve)
+            .catch((reason: unknown) => {
+                if (attempts > 0) {
+                    return delay(delayTime)
+                        .then(retrier.bind(null, operation, { attempts: attempts - 1, delayTime }))
```
malinskibeniamin, line 766: If `attempts` are `Infinity` this part should be redundant, you could do
```suggestion
                        .then(retrier.bind(null, operation, { attempts: Number.isFinite(attempts) ? attempts - 1 : attempts, delayTime }))
```

frontend/src/state/backendApi.ts
```diff
@@ -1291,6 +1292,45 @@ const apiStore = {
...
+                ['Content-Type', 'application/json']
+            ],
+            body: JSON.stringify({
+                secretData: secretValue            }),
+        });
+        return parseOrUnwrap<any>(response, null);
+    },
+
+
+    async deleteSecret(clusterName: string, secretId: string): Promise<any> {
```
malinskibeniamin, line 1327:
```suggestion
    async deleteSecret(clusterName: string, secretId: string): Promise<void> {
```

frontend/src/utils/utils.ts
```diff
@@ -753,3 +753,20 @@ export function setHeader(init: RequestInit, name: string, value: string) {
         (init.headers as Record<string, string>)[name] = value;
     }
 }
+
+// very simple retrier utility for allowing some retries, if we ended up using it more often we should consider making it more elaborate
+export const retrier = (
+    operation: any, { attempts = Infinity, delayTime = 100 }) => new Promise((resolve, reject) => {
```
rikimaru0345, line 759: Please convert this to `function ...` so that you can make it generic ( `function retrier<T>(...)` )
The `operation` argument should be typed as `Promise<T>`, or `()=>Promise<T>`,  or something.
No `any`.

frontend/src/utils/utils.ts
```diff
@@ -753,3 +753,20 @@ export function setHeader(init: RequestInit, name: string, value: string) {
         (init.headers as Record<string, string>)[name] = value;
     }
 }
+
+// very simple retrier utility for allowing some retries, if we ended up using it more often we should consider making it more elaborate
+export const retrier = (
```
rikimaru0345, line 758: Please convert this to `function ...` so that you can make it generic ( `function retrier<T>(...)` )
The `operation` argument should be typed as `Promise<T>`, or `()=>Promise<T>`,  or something.
No `any`.

frontend/src/state/backendApi.ts
```diff
@@ -1058,7 +1061,7 @@ const apiStore = {
     },
 
 
-    async refreshConnectClusters(force?: boolean): Promise<void> {
+    async refreshConnectClusters(force?: boolean): Promise<void | KafkaConnectors> {
```
rikimaru0345, line 1064: Isn't this endpoint supposed to *always* return ``KafkaConnectors`? 
If yes, then there should be no `| void` here.

If an error happens (can be that connection didn't work, or can be that the backend returned an error response), that is ok, then an error/exception will be thrown.

Generally there should *never* be a `T | void` type in any promise result

frontend/src/components/pages/connect/CreateConnector.tsx
```diff
@@ -107,30 +120,43 @@ class CreateConnector extends PageComponent {
...
+
+    useEffect(() => {
+        try {
+            setParsedUpdatedConfig(JSON.parse(stringifiedConfig));
+        } catch (e) {
+            setParsedUpdatedConfig(null);
+            setPostCondition(false);
+        }
+        setPostCondition(true);
+    }, [stringifiedConfig]);
```
rikimaru0345, line 159: Those are a lot of "useState" for local state.
Seeing as the component is an observer, are there any downsides to putting all of that state into a `useLocalObservable`? 

If it works, we don't have to change it, just asking out of curiosity.

frontend/src/components/pages/connect/CreateConnector.tsx
```diff
@@ -144,218 +170,287 @@ function ConnectorWizard({ connectClusters }: ConnectorWizardProps) {
...
-                onActiveClusterChange={setActiveCluster}
-                selectedPlugin={selectedPlugin}
-                onPluginSelectionChange={setSelectedPlugin}
-            />,
+            content: (
+                <ConnectorType
+                    connectClusters={connectClusters}
+                    activeCluster={activeCluster}
+                    onActiveClusterChange={(clusterName) => {
+                        uiState.pageBreadcrumbs[1] = { title: clusterName!, linkTo: `/connect-clusters/${clusterName}` };
```
rikimaru0345, line 178: Can we instead clear the breadcrumbs and push new entries instead modifying existing ones?
If its too hard, leave it as it is.
But I feel a bit uneasy about modifying array entries when there is no guarantee (array could be empty, or maybe there are additional entries or so...)

rikimaru0345 (changes requested):

frontend/src/utils/utils.ts
sago2k8: I don't see the value of making this more illegible when the performance impact is basically 0

frontend/src/components/pages/connect/CreateConnector.tsx
sago2k8: I see we could use local Observable here and it would be more legible, I just kept adding `useState` because I didn't want to refactor too much, silly me

frontend/src/state/backendApi.ts
sago2k8: I did need the value coming from this function, but I didn't want to rewrite the cachedResponsed logic, with the current implementation, is a bit hard and patchy to workaround the `T | void` type declaration, maybe we can keep this for a future change

sago2k8:
Thanks @rikimaru0345 Could you review this again and lemme know :)

frontend/src/utils/utils.ts
rikimaru0345: yea, leave it as is.

rikimaru0345 (approved):

PR #614, opened 2023-02-09
backend: Fix origin check func

weeco:
We used to compare the URL host against the configured allowed origin. This doesn't work because the host omits the scheme whereas the configured allowed origin
does not. Hence the result was that the requester's origin does not match any of the configured allowed origin, unless we also omit the scheme in the configuration.


sago2k8 (approved):
LGTM

PR #619, opened 2023-02-14
backend: Fallback to system cert pool if no ca configured

weeco:
The schema registry connectivity may fail with a
x509: certificate signed by unknown authority error because we used to initialize an empty cert pool that is passed to the TLS config. Instead of always initializing the cert pool, we should pass a nil cert pool if no CA has been specified so that the clients will implicitly load the system cert pool.


RafalKorepta (approved):
LGTM

PR #622, opened 2023-02-16
frontend/connectors: swap success message for Paused/Resumed

birdayz:
When clicking Pause, it says "Resumed". Apparently it's displaying exactly the opposite of what we expect.


sago2k8 (approved):

weeco:
@sago2k8 Will fix this as part of the secret management PR

PR #642, opened 2023-03-01
rework topic details statistics bar

rikimaru0345:
remove dynamic config entries, just show retention/segment ms/bytes based on cleanup.policy

### DIFF
--- a/frontend/src/components/pages/topics/QuickInfo.tsx
+++ b/frontend/src/components/pages/topics/QuickInfo.tsx
@@ -14,8 +14,6 @@ import { Row, Statistic } from 'antd';
 import { observer } from 'mobx-react';
 import { api } from '../../../state/backendApi';
 import '../../../utils/arrayExtensions';
-import { uiState } from '../../../state/uiState';
-import { FavoritePopover } from './Tab.Config';
 import { prettyBytesOrNA } from '../../../utils/utils';
 import { formatConfigValue } from '../../../utils/formatters/ConfigValueFormatter';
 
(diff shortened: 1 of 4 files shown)


rikimaru0345:
Changes:

1) no more dynamic (user defined) config properties; what config entries will be shown depends on cleanup policy
2) old table-style config display has been removed

image

frontend/src/components/pages/topics/QuickInfo.tsx
```diff
@@ -42,19 +40,30 @@ export const TopicQuickInfoStatistic = observer((p: { topic: Topic }) => {
...
+
+        if (cleanupPolicy.includes('delete')) {
+            dynamicEntries.push('retention.ms');
+            dynamicEntries.push('retention.bytes');
+        }
+        if (cleanupPolicy.includes('compact')) {
+            dynamicEntries.push('segment.ms');
+            dynamicEntries.push('segment.bytes');
+        }
+
```
sago2k8, line 57: [sand]: I see this and I notice we could reuse this logic in the future, not blocking at all, it's very easy to read and understand, but In my mind a helper function would be ideal, so you could potentially reuse it.

sago2k8 (approved):
LGTM, small comment not blocking

frontend/src/components/pages/topics/QuickInfo.tsx
rikimaru0345: Where would we reuse it?
What would the helper function do? 

Are you thinking about something like extracting name+value pairs of the config?
Or the logic where it switches what kinds of configs to display depending on cleanupPolicy?

PR #667, opened 2023-04-13
msgviewer: Support for utf8 control char payloads

### LINKED ISSUE #660: Differentiate null and empty string as keys in UI
### Version
```
Kowl CI
v1.5.0-598117c
(built November 10, 2021)
```
Unfortunately I cannot test it on the latest release, maybe it is already fixed, I don't know.

### Problem
Producing messages with `key=null` (no key) and with `key=""` leads to the same view in UI:
image

But `kafka-console-consumer` with `--property print.key=true` differentiates these two situations:
- for null key it is:
```
null    <msg1>
null    <msg2>
...
```
- for "" key it is:
```
        <msg1>
        <msg2>
...
```

This fix could help me a lot in debugging problems with assigning messages to partitions: messages with null key are usually uniformly distributed to partitions but messages with `""` key always go to one partition.

weeco:
Fixes #660

### DIFF
--- a/frontend/src/state/backendApi.ts
+++ b/frontend/src/state/backendApi.ts
@@ -382,17 +382,14 @@ const apiStore = {
                     m.keyJson = JSON.stringify(m.key.payload);
                     m.valueJson = JSON.stringify(m.value.payload);
 
-                    if (m.value.encoding == 'binary') {
-                        m.value.payload = decodeBase64(m.value.payload);
+                    if (m.key.encoding == 'binary' || m.key.encoding == 'utf8WithControlChars') {
+                        m.key.payload = decodeBase64(m.key.payload);
+                        m.keyBinHexPreview = this.base64ToHexString(m.key.payload);
+                    }
 
-                        const str = m.value.payload as string;
-                        let hex = '';
-                        for (let i = 0; i < str.length && i < 50; i++) {
-                            let n = str.charCodeAt(i).toString(16);
-                            if (n.length == 1) n = '0' + n;
-                            hex += n + ' ';
-                        }
-                        m.valueBinHexPreview = hex;
+                    if (m.value.encoding == 'binary' || m.key.encoding == 'utf8WithControlChars') {
+                        m.value.payload = decodeBase64(m.value.payload);
+                        m.valueBinHexPreview = this.base64ToHexString(m.value.payload);
                     }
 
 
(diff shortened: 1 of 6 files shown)


frontend/src/state/restInterfaces.ts
```diff
@@ -151,6 +151,7 @@ export enum CompressionType {
 
 export interface Payload {
     payload: any, // json obj
+    isPayloadNull: boolean,
```
rikimaru0345, line 154: is this actually getting populated by the backend?

weeco: Yes, backend code change is also part of this PR

frontend/src/state/restInterfaces.ts
```diff
@@ -171,12 +172,11 @@ export interface TopicMessage {
     key: Payload,
     value: Payload,
 
-    isValueNull: boolean, // todo: rename to isTombstone
-
     // Added by the frontend
     valueJson: string,
     valueBinHexPreview: string,
     keyJson: string,
+    keyBinHexPreview: string,
```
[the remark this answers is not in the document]
weeco: from here: https://github.com/redpanda-data/console/pull/667/files#diff-b0f2d81a7446ac88a7118694dc5b3935d60118d7b0322facf55f39046920ef61R387

rikimaru0345 (approved):

frontend/src/state/backendApi.ts
```diff
@@ -405,6 +402,16 @@ const apiStore = {
...
 
+    base64ToHexString(b64: string): string {
+        let hex = '';
+        for (let i = 0; i < b64.length && i < 50; i++) {
+            let n = b64.charCodeAt(i).toString(16);
+            if (n.length == 1) n = '0' + n;
+            hex += n + ' ';
+        }
+        return hex
+    },
```
rikimaru0345, line 413:
```suggestion
    function base64ToHexString(base64: string): string {
        const binary = atob(base64);
        const bytes = new Uint8Array(binary.length);
        for (let i = 0; i < binary.length; i++) {
            bytes[i] = binary.charCodeAt(i);
        }
        
        let hex = '';
        for (let i = 0; i < bytes.length; i++) {
            const b = bytes[i].toString(16);
            hex += b.length === 1 ? '0' + b : b;
            
            if (i < bytes.length - 1)
                hex += ' ';
        }
        
        return hex;
    },
```

PR #670, opened 2023-04-14
backend: Rely on franz-go to fetch grp offsets in bulk

weeco:
Franz-go nowadays can request group offsets in bulk (if supported by target kafka cluster). If it's not supported the client will split the request so that we no longer need to worry about it.

### DIFF
--- a/backend/pkg/console/consumer_group_offsets.go
+++ b/backend/pkg/console/consumer_group_offsets.go
@@ -13,8 +13,7 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/twmb/franz-go/pkg/kerr"
-	"github.com/twmb/franz-go/pkg/kmsg"
+	"github.com/twmb/franz-go/pkg/kadm"
 	"go.uber.org/zap"
 
 	"github.com/redpanda-data/console/backend/pkg/kafka"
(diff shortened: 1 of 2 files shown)


backend/pkg/console/consumer_group_offsets.go
```diff
@@ -63,13 +63,19 @@ func (s *Service) getConsumerGroupOffsets(ctx context.Context, groups []string)
...
-		return nil, fmt.Errorf("failed to list consumer group offsets in bulk")
+		s.logger.Error("failed to list consumer group offsets", zap.Error(err))
+		return nil, fmt.Errorf("failed to list consumer group offsets")
 	}
 
 	offsetsByGroup := make(map[string]map[string]partitionOffsets) // GroupID -> TopicName -> partitionOffsets
-	for group, offset := range offsets {
-		offsetsByGroup[group] = convertOffsets(offset)
+	for _, grpOffsetRes := range offsets.Groups {
+		if kerr.ErrorForCode(grpOffsetRes.ErrorCode) != nil {
```
bojand, line 72: should we assign a new error here for logging?

```go
if err := kerr.ErrorForCode(grpOffsetRes.ErrorCode); err != nil {
  s.logger.Warn("failed to fetch offset for consumer group",
    zap.String("group_id", grpOffsetRes.Group),
    zap.Error(err)) // logs the actual error from grpOffsetRes.ErrorCode() rather than the previous one
    continue
}
```

bojand (approved):

PR #700, opened 2023-05-12
connectors: improved description and documentations in debezium connectors

bochenekmartin:


backend/pkg/connector/patch/debezium_postgres_source.go
```diff
@@ -81,6 +81,9 @@ func (*ConfigPatchDebeziumPostgresSource) PatchDefinition(d model.ConfigDefiniti
 			ClearRecommendedValuesWithMetadata().
 			AddRecommendedValueWithMetadata("io.confluent.connect.avro.AvroConverter", "AVRO").
 			AddRecommendedValueWithMetadata("org.apache.kafka.connect.json.JsonConverter", "JSON")
+	case "database.tcpKeepAlive":
+		d.SetDocumentation("Enable to avoid dropping TCP connection. Disable otherwise.").
```
tomasz-sadura, line 85: I think we should consistently get rid of trailing dots

tomasz-sadura (approved):

PR #711, opened 2023-05-22
Backend get topics tests

bojand:
**Changes**

Adds tests for get topics route `GET /api/topics`. 

Minor refactoring of common testing hooks into `api_integration_test.go` and added ability to specify return values into the mocked hooks.

### DIFF
--- a/backend/pkg/api/handle_topics_integration_test.go
+++ b/backend/pkg/api/handle_topics_integration_test.go
@@ -86,14 +86,14 @@ func (s *APIIntegrationTestSuite) TestHandleGetTopics() {
 		topicName := testutil.TopicNameForTest("get_topics_1")
 
 		oldHooks := s.api.Hooks
-		newHooks := newAssertHooks(t, map[string]map[string]bool{
+		newHooks := newAssertHooks(t, map[string]map[string]assertCallReturnValue{
 			"CanSeeTopic": {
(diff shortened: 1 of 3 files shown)


backend/pkg/api/handle_topics_integration_test.go
```diff
@@ -0,0 +1,406 @@
...
+		oldHooks := s.api.Hooks
+		newHooks := newAssertHooks(t, map[string]map[string]assertCallReturnValue{
+			"CanSeeTopic": {
+				"any": assertCallReturnValue{BoolValue: true, Err: nil},
+			},
+			"AllowedTopicActions": {
+				testutil.TopicNameForTest("get_topics_0"): assertCallReturnValue{SliceValue: []string{}, Err: nil},
+				topicName: assertCallReturnValue{Err: &rest.Error{
+					Err:     fmt.Errorf("error from test"),
+					Status:  505,
```
weeco, line 144: Do we return status 505 somewhere? What is this?

What do you think about using the `http.StatusCode` consts instead of the status code number?

bojand: No normally this would return Unauthorized usually. I was using a different signal value since it's a mocked failure. But I have changed it to return Unauthorized to represent a real world scenario.

weeco (approved):

PR #718, opened 2023-06-01
connectors: added support for max.request.size UI property for MM2 and Debezium connectors

bochenekmartin:


tomasz-sadura (approved):
LGTM

PR #720, opened 2023-06-01
backend: if there are unassigned tasks on a running connector, set co…

bojand:
…nnector status to unassigned


bochenekmartin (approved):
LGTM

tomasz-sadura (approved):

PR #730, opened 2023-06-13
connectors: changed auto.offset.reset value to 'none'

bochenekmartin:
Issue: https://github.com/redpanda-data/cloudv2/issues/6816


backend/pkg/connector/interceptor/mirror_hook.go
```diff
@@ -331,7 +331,7 @@ func KafkaConnectToConsoleMirrorSourceHook(response model.ValidationResponse, _
 			},
 			Value: model.ConfigDefinitionValue{
 				Name:              "consumer.auto.offset.reset",
-				Value:             "earliest",
+				Value:             "none",
```
tomasz-sadura, line 334: wondering whether we should also set Definition.CustomDefaultValue...

tomasz-sadura (approved):

PR #748, opened 2023-06-28
Sj/fix topic config

sago2k8:
Fix issue when opening the modal for change topic configuration.

We needed to provide the Theme to the Portal Antd uses for showing the modals, easiest was to reuse the ChakraProvider.

I also installed some packages that are imported but weren't set in the package.json

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,10 +23,12 @@
     },
     "dependencies": {
         "@ant-design/icons": "^4.7",
+        "@chakra-ui/icons": "^2.0.19",
+        "@chakra-ui/react-use-disclosure": "^2.0.8",
         "@heroicons/react": "^1.0.6",
         "@monaco-editor/react": "^4.5",
         "@primer/octicons-react": "^17",
-        "@redpanda-data/ui": "^3.0.0",
+        "@redpanda-data/ui": "^3.2.0",
         "@textea/json-viewer": "^1.24.4",
         "antd": "^4.21",
         "array-move": "^4",
@@ -43,6 +45,7 @@
         "react-dom": "^18.2.0",
         "react-draggable": "^4.4.3",
         "react-highlight-words": "^0.18.0",
+        "react-icons": "^4.10.1",
         "react-markdown": "^8.0.7",
         "react-router-dom": "^5.2.0",
         "react-scripts": "^5.0.1",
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -23,10 +23,12 @@
     },
     "dependencies": {
         "@ant-design/icons": "^4.7",
+        "@chakra-ui/icons": "^2.0.19",
+        "@chakra-ui/react-use-disclosure": "^2.0.8",
```
malinskibeniamin, line 27: Is it possible to add them as peer dependencies? Ideally we should import everything from Redpanda UI as it re-exports the entire chakra UI

frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -65,47 +65,49 @@ export default class ConfigurationEditor extends Component<{
...
-                    </Radio.Group>
-
-                    {this.modalError && <Alert status="error" style={{ margin: '1em 0' }}>
-                        <AlertIcon />
-                        {this.modalError}
-                    </Alert>}
-
-                </div>
+                return(
+                    <ChakraProvider theme={redpandaTheme} disableGlobalStyle={true} disableEnvironment={true}>
```
malinskibeniamin, line 69: Nice one! I think as long as we add a follow up GH issue to remove the theme after using the chakra modal, this will work fine in the meantime

malinskibeniamin (approved):
Nice work! Thanks for pairing, it was good to know how console setup works

frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -65,47 +65,49 @@ export default class ConfigurationEditor extends Component<{
...
+                            <Radio.Group className="valueRadioGroup" value={this.modalValueType} onChange={e => this.modalValueType = e.target.value} >
+                                <Radio value="default">
+                                    <span>Default: </span>
+                                    <span style={{ fontWeight: 'bold' }}>{friendlyDefault}</span>
+                                    <div className="subText">Inherited from {defaultSource}</div>
+                                </Radio>
+                                <Radio value="custom">
+                                    <span>Custom</span>
+                                    <div className="subText">Set at topic configuration</div>
+                                    <div onClick={e => {
```
rikimaru0345, line 90:
```suggestion
                                    <div style={{ position: 'relative', zIndex: 2 }}  onClick={e => {
```

rikimaru0345 (changes requested):

rikimaru0345 (approved):

PR #750, opened 2023-06-29
Serverless UI

rikimaru0345:
Change how isServerless() is implemented; going from APP_FEATURES html/env to injected config props by cloud-ui


weeco (approved):

weeco (approved):

frontend/src/config.ts
```diff
@@ -130,6 +134,10 @@ export function isEmbedded() {
     return config.jwt != null;
 }
 
+export function isServerless() {
+    return config.isServerless;
+}
```
malinskibeniamin, line 139: [question] Is it possible that the config is being accessed prematurely by calling the `isServerless` function, causing MobX to throw an error?

Issue #699, opened 2023-05-12
OAUTHBEARER sasl mechanism


vfolgosa:
Hi guys, I have a question, to use the OAUTHBEARER sasl mechanism, do I have to generate the token outside the application and put it in the config kafka.sasl.oauth.token?

So if the token expires do I need to update the config and restart the console?

weeco:
Hey @vfolgosa ,
based on our Slack thread I assume you want to use short-lived tokens which need to be refreshed periodically. Currently this is not supported (neither in Console, nor the Kafka client we use). I'd need to look in more detail how this should work and possibly this should be implemented upstream in the franz-go client rather than Console. I'll leave this issue in Console open anyways, but you could create a separate one in https://github.com/twmb/franz-go too.

cthtrifork:
FYI

https://github.com/twmb/franz-go/issues/482

cthtrifork:
> The oauth package itself provides an `Oauth` function,
> 
> 
> 
> ```
> 
> func Oauth(authFn func(context.Context) (Auth, error)) sasl.Mechanism {
> 
> ```
> 
> 
> 
> You can use this to create a new Auth with a new short lived token:
> 
> 
> 
> ```
> 
> kgo.SASL(oauth.Oauth(func(ctx context.Context) (oauth.Auth, error) {
> 
>     shortToken, err := aquireToken(ctx)
> 
>     return oauth.Auth{Token: shortToken}, err
> 
> }))
> 
> ```
> 
> The `Auth.AsMechanism` function is only provided if you don't intend to change the token
> 
> 
> 
> note the aws_msk_iam example, which is similar in concept:
> 
> https://github.com/twmb/franz-go/blob/a61b2fa4407091ae3679cea5868dbd128f3a995b/examples/sasl/aws_msk_iam/main.go#L41

cthtrifork:
This should be closed 😄

PR #751, opened 2023-06-30
Added clientid, clientsecret and tokenendpoint to SASLMechanismOAuthBearer mechanism

cthtrifork:
See https://github.com/redpanda-data/console/issues/699

### DIFF
--- a/backend/pkg/config/kafka_sasl_oauth.go
+++ b/backend/pkg/config/kafka_sasl_oauth.go
@@ -10,25 +10,95 @@
 package config
 
 import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
 )
 
 // KafkaSASLOAuthBearer is the config struct for the SASL OAuthBearer mechanism
 type KafkaSASLOAuthBearer struct {
-	Token string `yaml:"token"`
+	Token         string `yaml:"token"`
+	ClientID      string `yaml:"clientId"`
+	ClientSecret  string `yaml:"clientSecret"`
+	TokenEndpoint string `yaml:"tokenEndpoint"`
+	Scope         string `yaml:"scope"`
 }
 
 // RegisterFlags registers all sensitive Kerberos settings as flag
 func (c *KafkaSASLOAuthBearer) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&c.Token, "kafka.sasl.oauth.token", "", "OAuth Bearer Token")
+	f.StringVar(&c.ClientID, "kafka.sasl.oauth.clientId", "", "OAuth Bearer Client Id")
+	f.StringVar(&c.ClientSecret, "kafka.sasl.oauth.clientSecret", "", "OAuth Bearer Client Secret")
+	f.StringVar(&c.TokenEndpoint, "kafka.sasl.oauth.tokenEndpoint", "", "OAuth Bearer Token Endpoint")
+	f.StringVar(&c.Scope, "kafka.sasl.oauth.scope", "", "OAuth Bearer Scope")
 }
 
 // Validate Kafka SASL OAuth configurations.
 func (c *KafkaSASLOAuthBearer) Validate() error {
-	if c.Token == "" {
+	if c.TokenEndpoint != "" {
+		if c.ClientID == "" || c.ClientSecret == "" {
+			return fmt.Errorf("OAuth Bearer client credentials must be set")
+		}
+	} else if c.Token == "" {
 		return fmt.Errorf("OAuth Bearer token must be set")
 	}
 
 	return nil
+}
+
+// AcquireToken is used to acquire a new access token using client credentials
+func (c *KafkaSASLOAuthBearer) AcquireToken(ctx context.Context) (string, error) {
+	authHeaderValue := base64.StdEncoding.EncodeToString([]byte(c.ClientID + ":" + c.ClientSecret))
+
+	queryParams := url.Values{}
+	queryParams.Set("grant_type", "client_credentials")
+	queryParams.Set("scope", c.Scope)
+
+	req, err := http.NewRequestWithContext(ctx, "POST", c.TokenEndpoint, strings.NewReader(queryParams.Encode()))
+	if err != nil {
+		return "", err
+	}
+
+	req.URL.RawQuery = queryParams.Encode()
+
+	req.Header.Set("Authorization", "Basic "+authHeaderValue)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	if err := resp.Body.Close(); err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("token request failed with status code %d", resp.StatusCode)
+	}
+
+	var tokenResponse map[string]interface{}
+	err = json.Unmarshal(body, &tokenResponse)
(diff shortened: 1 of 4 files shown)


backend/pkg/kafka/config_helper.go
```diff
@@ -92,9 +93,18 @@ func NewKgoConfig(cfg *config.Kafka, logger *zap.Logger, hooks kgo.Hook) ([]kgo.
 
 		// OAuth Bearer
 		if cfg.SASL.Mechanism == config.SASLMechanismOAuthBearer {
-			mechanism := oauth.Auth{
-				Token: cfg.SASL.OAUth.Token,
-			}.AsMechanism()
+			var mechanism sasl.Mechanism
+			if cfg.SASL.OAUth.TokenEndpoint != "" {
```
cthtrifork, line 97: Alternatively we could add another config parameter called `TokenType` with options `ShortLived` and `LongLived` but I personally like this better.

backend/pkg/config/kafka_sasl_oauth.go
```diff
@@ -10,25 +10,70 @@
...
+			return fmt.Errorf("OAuth Bearer client credentials must be set")
+		}
+	} else if c.Token == "" {
 		return fmt.Errorf("OAuth Bearer token must be set")
 	}
 
 	return nil
 }
+
+func (c *KafkaSASLOAuthBearer) AcquireToken(ctx context.Context) (string, error) {
```
cthtrifork, line 49: We could also use a standard library for this, but I am not sure what opinion you had on including new libraries

backend/pkg/config/kafka_sasl_oauth.go
```diff
@@ -10,25 +10,75 @@
...
+	"net/http"
 )
 
 // KafkaSASLOAuthBearer is the config struct for the SASL OAuthBearer mechanism
 type KafkaSASLOAuthBearer struct {
-	Token string `yaml:"token"`
+	Token         string `yaml:"token"`
+	ClientID      string `yaml:"clientid"`
+	ClientSecret  string `yaml:"clientsecret"`
+	TokenEndpoint string `yaml:"tokenendpoint"`
```
weeco, line 26:
```suggestion
	TokenEndpoint string `yaml:"tokenEndpoint"`
```

backend/pkg/config/kafka_sasl_oauth.go
```diff
@@ -10,25 +10,75 @@
...
 }
 
 // RegisterFlags registers all sensitive Kerberos settings as flag
 func (c *KafkaSASLOAuthBearer) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&c.Token, "kafka.sasl.oauth.token", "", "OAuth Bearer Token")
+	f.StringVar(&c.ClientID, "kafka.sasl.oauth.clientid", "", "OAuth Bearer Client Id")
+	f.StringVar(&c.ClientSecret, "kafka.sasl.oauth.clientsecret", "", "OAuth Bearer Client Secret")
+	f.StringVar(&c.TokenEndpoint, "kafka.sasl.oauth.tokenendpoint", "", "OAuth Bearer Token Endpoint")
```
weeco, line 34:
```suggestion
	f.StringVar(&c.TokenEndpoint, "kafka.sasl.oauth.token-endpoint", "", "OAuth Bearer Token Endpoint")
```

backend/pkg/config/kafka_sasl_oauth.go
weeco: What library do you have in mind? I'm mostly afraid of doing something wrong in the OAuth token process, but the code looks straight forward and given that the authority is the other side which decides whether the token is valid or not, I'm also not that afraid.

backend/pkg/kafka/config_helper.go
weeco: Yes seems reasonable 👍

weeco (commented):
Hey @cthtrifork ,
thank you for working on this! Generally LGTM, just have some NITs. Additionally, could you document the new properties in the reference config here: https://github.com/redpanda-data/console/blob/master/docs/config/console.yaml ?

cthtrifork:
@weeco Thanks for the review, as you seem to approve the general approach. I have done some integration testing with a few different OIDC providers and confirmed it now works successfully.

When do you think a release could be a possible?

backend/pkg/config/kafka_sasl_oauth.go
cthtrifork: I have reverted your suggestion, as I could not get it to work with ENV variable injection.
 `KAFKA_SASL_OAUTH_TOKEN-ENDPOINT` and `KAFKA_SASL_OAUTH_TOKEN_ENDPOINT` did not work.

Keeping it as `tokenendpoint` allowed me to set `KAFKA_SASL_OAUTH_TOKENENDPOINT=`

weeco: The flags are entirely separate from the env variables. The flag name should not have any impact on the env variable name, it's derived from the struct tags (`yaml`).

weeco:
@cthtrifork , we build a new docker image for each push to master, would that be good enough for you? We have an ongoing release branch (rc for v2.3) but since this is a new feature it wouldn't fit there and we only do one feature release per quarter.

cthtrifork:
@weeco That is perfect - we can use the specific version from a SHA

backend/pkg/config/kafka_sasl_oauth.go
```diff
@@ -10,25 +10,95 @@
...
+	"net/url"
+	"strings"
 )
 
 // KafkaSASLOAuthBearer is the config struct for the SASL OAuthBearer mechanism
 type KafkaSASLOAuthBearer struct {
-	Token string `yaml:"token"`
+	Token         string `yaml:"token"`
+	ClientID      string `yaml:"clientid"`
+	ClientSecret  string `yaml:"clientsecret"`
```
weeco, line 28:
```suggestion
	ClientID      string `yaml:"clientId"`
	ClientSecret  string `yaml:"clientSecret"`
```

For consistency, lower camel case

docs/config/console.yaml
```diff
@@ -46,7 +46,11 @@ kafka:
   #     realm:
   #     enableFast: true
   #   oauth:
-  #     token: # can be set via the --kafka.sasl.oauth.token flag as well
+  #     token: # can be set via the --kafka.sasl.oauth.token flag as well. Will be used as a long living token if no token-endpoint is specified
+  #     clientid: # can be set via the --kafka.sasl.oauth.clientid flag as well
+  #     clientsecret: # can be set via the --kafka.sasl.oauth.clientsecret flag as well
+  #     tokenendpoint: # can be set via the --kafka.sasl.oauth.token-endpoint flag as well. Will be used to fetch a short living token if specified
```
weeco, line 52:
```suggestion
  #     clientId: # can be set via the --kafka.sasl.oauth.clientid flag as well
  #     clientSecret: # can be set via the --kafka.sasl.oauth.clientsecret flag as well
  #     tokenEndpoint: # can be set via the --kafka.sasl.oauth.token-endpoint flag as well. Will be used to fetch a short living token if specified
```

weeco (commented):
Last changes and then we should be good to merge 👍

weeco (approved):
Thanks for implementing

weeco:
Use one of the latest tags from here to use this feature: https://hub.docker.com/r/redpandadata/console-unstable/tags

cthtrifork:
It is working perfectly on our clusters. Thanks for the collaboration!

> @cthtrifork , we build a new docker image for each push to master, would that be good enough for you? We have an ongoing release branch (rc for v2.3) but since this is a new feature it wouldn't fit there and we only do one feature release per quarter.

We will use this tag until a release is ready. Will the unstable image be available until then?
I would rather not save a copy in our docker registry.

weeco:
Yes, we will never delete any docker tags from the registry unless we believe it would be good for our users (e.g. serious security issues). Feel free to use it until the release. Thank you for reporting back too!

PR #754, opened 2023-06-30
fix(frontend): fix issue rendering serverless console

sago2k8:
- make available routes an observable
- make sure there are no dependencies referencing each other.

### DIFF
--- a/frontend/src/state/supportedFeatures.ts
+++ b/frontend/src/state/supportedFeatures.ts
@@ -17,8 +17,7 @@
 // That way we can easily check if (for example) "partition reassignment" should be visible/allowed.
 //
 
-import { computed, observable, when } from 'mobx';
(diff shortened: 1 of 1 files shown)


frontend/src/config.ts
```diff
@@ -12,12 +12,13 @@ import { loader } from '@monaco-editor/react';
 import { ConfigProvider } from 'antd';
 import { autorun, configure, observable, when } from 'mobx';
 import colors from './colors';
-import { embeddedAvailableRoutes } from './components/routes';
+// import { embeddedAvailableRoutes } from './components/routes';
```
malinskibeniamin, line 15: Let's remove the comment in the future. Not a blocker

malinskibeniamin (approved):

PR #764, opened 2023-07-13
Connect integration test

bojand:
This is an API integration test for creating connector. The test does the following:

1. Starts Redpanda container using testcontainers.
2. Starts [HTTP Bin](https://github.com/postmanlabs/httpbin) container using testcontainers. 
3. Starts Connectors container.
4. There is only 1 test right now. The test makes an HTTP API request to create a connector instance for HTTP input that makes requests to UUID endpoint from the HTTP Bin container. It lets the connector run momentarily. Then we use a Kafka client to verify that the data is in the specified topic.


backend/pkg/connect/create_connector_integration_test.go
```diff
@@ -0,0 +1,378 @@
...
+		for {
+			fetches := cl.PollFetches(ctx)
+			if fetches.IsClientClosed() {
+				break
+			}
+
+			errs := fetches.Errors()
+			if len(errs) == 1 && (errors.Is(errs[0].Err, context.DeadlineExceeded) || errors.Is(errs[0].Err, context.Canceled)) {
+				break
+			}
```
weeco, line 196: Can you explain why we want to break here? It seems we are checking for a timeout , so makes sense to not retry forever, but if we just break here we will continue in the test flow no and in the end we will see the assert len of 3 is not fulfilled

bojand: We use the context timeout set above to terminate the polling of the kafka records. So canceled context or deadline exceeded is the expected way we break out of the `for` loop in this case. I agree it's weird... open to some other suggestions.

bojand: Not sure I understand...? we haven't collected all the records yet. I think I am missing something?

backend/pkg/connect/create_connector_integration_test.go
```diff
@@ -0,0 +1,378 @@
...
+				"value.converter":                           "org.apache.kafka.connect.json.JsonConverter",
+				"value.converter.schemas.enable":            "false",
+			},
+		})
+
+		require.Empty(connectErr)
+		assert.NotNil(res)
+
+		timer := time.NewTimer(32 * time.Second)
+		<-timer.C
```
weeco, line 161: What is this for? 32s seems very long.

backend/pkg/connect/create_connector_integration_test.go
weeco: Maybe we should have a chat, may be quicker to clarify my questions. WDYT? Just hit me up when you are ready for it

backend/pkg/connect/create_connector_integration_test.go
bojand: We run the HTTP input connect instance to do requests every `10s`, and we wait `32s` for it to do 3 requests, that then we verify below. I have changed the settings, so that the input does a request every `3s` and we wait `10s`. LMK if there are better settings, maybe `1.5s` and wait for `5s` ?

weeco: I was wondering what the purpose of waiting was. Do you basically want to make sure the connector doesn't transition into an error state a few seconds after creating because it'll take a while until it gets to the running state? As far as Console is concerned I'd only check if the request was successful and possibly describe the created connector afterwards to see if it exist in the connect cluster no?

I don't want to fail if the connect cluster, conenctor plugin or the config itself has an issue because it wouldn't be addressable for us no?

bojand: Oh ok... yes this is attempting full e2e type of test where we are verifying that the connector did its work properly...

bojand:
I have moved the test to API integration test.

backend/pkg/api/handle_kafka_connect_integration_test.go
```diff
@@ -0,0 +1,402 @@
...
+group.id=connectors-cluster
+offset.storage.topic=_internal_connectors_offsets
+config.storage.topic=_internal_connectors_configs
+status.storage.topic=_internal_connectors_status
+config.storage.replication.factor=-1
+offset.storage.replication.factor=-1
+status.storage.replication.factor=-1
+offset.flush.interval.ms=1000
+producer.linger.ms=1
+producer.batch.size=131073
```
tomasz-sadura, line 258: I think these 3 config lines can be removed for clarity, should not make any difference

backend/pkg/api/handle_kafka_connect_integration_test.go
```diff
@@ -0,0 +1,402 @@
...
+		assert.NoError(err)
+
+		assert.Equal("http_connect_input", getConnectRes.Name)
+		assert.Equal("httpbin-input", getConnectRes.Config["kafka.topic"])
+		assert.Equal("1000", getConnectRes.Config["http.timer.interval.millis"])
+	})
+}
+
+const testConnectConfig = `key.converter=org.apache.kafka.connect.converters.ByteArrayConverter
+value.converter=org.apache.kafka.connect.converters.ByteArrayConverter
```
tomasz-sadura, line 248: if you specify key/value.converter in the connector config, it I think can be removed from here as well

tomasz-sadura (approved):

backend/pkg/api/handle_kafka_connect_integration_test.go
bojand: When I try to remove these I get the following error on startup:
```
{"ts":"2023-07-17 12:18:28,043","level":"ERROR","thread":"main","class":"org.apache.kafka.connect.cli.ConnectDistributed:89","msg":"Stopping due to error"}
org.apache.kafka.common.config.ConfigException: Missing required configuration "key.converter" which has no default value.
	at org.apache.kafka.common.config.ConfigDef.parseValue(ConfigDef.java:493)
	at org.apache.kafka.common.config.ConfigDef.parse(ConfigDef.java:483)
	at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:112)
	at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:132)
	at org.apache.kafka.connect.runtime.WorkerConfig.<init>(WorkerConfig.java:475)
	at org.apache.kafka.connect.runtime.distributed.DistributedConfig.<init>(DistributedConfig.java:560)
	at org.apache.kafka.connect.runtime.distributed.DistributedConfig.<init>(DistributedConfig.java:555)
	at org.apache.kafka.connect.cli.ConnectDistributed.startConnect(ConnectDistributed.java:98)
	at org.apache.kafka.connect.cli.ConnectDistributed.main(ConnectDistributed.java:83)
```

So leaving them for now.

backend/pkg/api/handle_kafka_connect_integration_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+	// save old
+	oldConnectSvc := s.api.ConnectSvc
+
+	// switch
+	s.api.ConnectSvc = newConnectSvc
+
+	// allow for connect service to establish connection to connect cluster
+	timer := time.NewTimer(200 * time.Millisecond)
+	<-timer.C
```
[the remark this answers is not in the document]
bojand: I have removed this timer since it's not needed as we do the connection check synchronously within the service. There is only one other timer in tests in this repo, and it's a timer to allow for connect instance to do some work and write data into the kafka topic.

PR #776, opened 2023-07-20
Backend deserializer tests

bojand:
### Changes

- Move `MatchesOrder` out of `testutil` package as it was causing cyclical imports and making the package unusable from within the `kafka` package. The `MatchesOrder` code is moved to api integration test file where it is used.
- Add `buf` installation to taskfile and config files.
- Add `testcontainers-go` and franz-go `sr` package as dependencies.
- Add test proto files: This includes a simple "Order" message proto file, and another full "Order" proto package taken from owl-shop code.
- Added deserializer tests. The test starts up Redpanda container with schema registry. Added tests for the following use cases:
    - Plain JSON. We write data to a topic. Get the record, and attempt to deserialize it.
    - Plain protobuf. We write a simple protobuf data to a topic with file mapping for the message type. Get the record, and attempt to deserialize it. Verify the data retrieved and deserialized matches what was created.
    - Plain protobuf with references. We write a complicated protobuf data with references to other types to a topic with file mapping for the message type. Get the record, and attempt to deserialize it. Verify the data retrieved and deserialized matches what was created.
    - Simple protobuf schema registry. We register a simple protobuf to the schema registry and serialize it to a topic. Get the record, and attempt to deserialize it. Verify the data retrieved and deserialized matches what was created.
    - Complex protobuf schema registry. We register a complex protobuf with references to the schema registry and serialize it to a topic. Get the record, and attempt to deserialize it. Verify the data retrieved and deserialized matches what was created.


taskfiles/backend.yaml
```diff
@@ -96,11 +111,21 @@ tasks:
...
       - task: 'fmt'
+      - |
+        export PATH=$PATH:{{ .BUILD_ROOT }}/bin/go
+        {{ .BUILD_ROOT }}/bin/buf format -w pkg/kafka/testdata/proto
+      - |
+        export PATH=$PATH:{{ .BUILD_ROOT }}/bin/go
+        {{ .BUILD_ROOT }}/bin/buf lint pkg/kafka/testdata/proto
+      - |
+        export PATH=$PATH:{{ .BUILD_ROOT }}/bin/go
+        {{ .BUILD_ROOT }}/bin/buf generate pkg/kafka/testdata/proto
```
weeco, line 128: Should we run the fmt task at the end? otherwise I imagine that the generated code will trigger linting errors

backend/pkg/kafka/deserializer_integration_test.go
```diff
@@ -0,0 +1,972 @@
...
+		defer func() {
+			_, err := s.kafkaAdminClient.DeleteTopics(ctx, testTopicName)
+			assert.NoError(err)
+		}()
+
+		cfg := s.createBaseConfig()
+		cfg.Kafka.Protobuf.Enabled = true
+		cfg.Kafka.Protobuf.SchemaRegistry.Enabled = true
+		cfg.Kafka.Schema.Enabled = true
+		cfg.Kafka.Schema.URLs = []string{"http://" + s.registryAddress}
```
weeco, line 151: Should this be part of the base config? These are not necessarily needed for the plain JSON case. Possibly another test case to check if JSON deserialization works with both schema registry enabled/disabled if that was intended I'd recommend a comment here.

backend/pkg/kafka/deserializer_integration_test.go
```diff
@@ -0,0 +1,972 @@
...
+
+		// register the protobuf schema
+		rcl, err := sr.NewClient(sr.URLs(registryURL))
+		require.NoError(err)
+
+		// address
+		protoFile, err := os.ReadFile("testdata/proto/shop/v2/address.proto")
+		require.NoError(err)
+
+		ssAddress, err := rcl.CreateSchema(context.Background(), testTopicName+"-address-value", sr.Schema{
```
weeco, line 652: We can register the subject under the same name as the import in the proto file. This is apparently the standard that the Java clients use, hence suggesting to match this. In the past we discovered an issue with slashes in the subject names for instance.

```suggestion
		ssAddress, err := rcl.CreateSchema(context.Background(), "shop/v2/address.proto", sr.Schema{
```

backend/pkg/kafka/deserializer_integration_test.go
```diff
@@ -0,0 +1,972 @@
...
+			Type:   sr.TypeProtobuf,
+		})
+		require.NoError(err)
+		require.NotNil(ssAddress)
+
+		// customer
+		protoFile, err = os.ReadFile("testdata/proto/shop/v2/customer.proto")
+		require.NoError(err)
+
+		ssCustomer, err := rcl.CreateSchema(context.Background(), testTopicName+"-customer-value", sr.Schema{
```
weeco, line 663:
```suggestion
		ssCustomer, err := rcl.CreateSchema(context.Background(), "shop/v2/customer.proto", sr.Schema{
```

weeco (approved):

PR #812, opened 2023-08-15
Added functionality to skip loading hidden files

### LINKED ISSUE #810: Skip hidden dirs when loading Protos from filesystem
Issue created by @smatvienko-tb:

              Hey, @weeco!
The proto-decoding feature is great!
The issue is still there. The files were imported twice in Kubernetes, and it was painful to map config map with many photos.

The problem and root cause are shown in the picture:
The **root cause** is **scanning hidden** folders.

image

Steps to reproduce on any Linux environment:

1. Create a hidden folder `/tmp/protos/..data`
2. Put a proto file to the folder `/tmp/protos/..data/tbmsg.proto`
3. Make a symlink `/tmp/protos/tbmsg.proto` to `/tmp/protos/..data/tbmsg.proto`

```yaml
kafka:
  protobuf:
    enabled: true
    fileSystem:
      enabled: true
      paths:
        - /tmp/protos
```

The fix is to **avoid** scanning **hidden** files and folders (same as `ls` command do).

Otherwise, the Kubernetes deployment became painful. Some users might give up using the proto-decoding feature because of such an issue.

I am preparing the Redpanda console deployment for the [ThingsBoard IoT framework](https://thingsboard.io/) and I don't really want to submit a non-perfect deployment for a wide community. 
Unfortunately,  I can not help you with a good PR because it is not my language at this moment.
Let me know if you need to create a new issue.
Will be appreciated if you can implement the fix required for Kubernetes users 🙏 .

@pon0marev FYI

image

_Originally posted by @smatvienko-tb in https://github.com/redpanda-data/console/issues/336#issuecomment-1677694488_

anishrajan25:
Fixes #810 

- Added functionality to offer ability to specify if the user wants to skip hidden files.
    - Added a boolean flag `SkipHiddenFiles` in filesystem config.
    - By default it's value would be `false` and no files would be skipped.
- While reading files, added check to avoid loading a hidden file based on following conditions:
    - The `SkipHiddenFiles` flag in filesystem config is set to true
    - The `filepath` refers to a file that is hidden, i.e.
        - either the file belongs to a hidden folder (tmp/..hiddenFolder/file)
        - or the file itself is a hidden file (tmp/data/.hiddenFile)


anishrajan25:
@weeco, @smatvienko-tb This is the fix PR for #812 . Could you please have a look here?

backend/pkg/filesystem/util.go
```diff
@@ -21,6 +21,13 @@ func isStringInSlice(item string, arr []string) bool {
 	return false
 }
 
+// isHiddenFile returns true if the filename refers to a file that is hidden, i.e.
+// - either the file belongs to a hidden folder (tmp/..hiddenFolder/file)
+// - or the file itself is a hidden file (tmp/data/.hiddenFile)
+func (c *Service) isHiddenFile(filename string) bool {
```
smatvienko-tb, line 27: Hey @anishrajan25 !
I guess this is not platform-independent.
I'm not a Golang guru, so I had a chat with GPT and after couple sentences I stay on this example:
It uses `filepath.SkipDir` to not walk into hidden folders and recognize the hidden attribute for Windows
```golang
package main

import (
	"fmt"
	"os"
	"path/filepath"
	"strings"
)

func isHidden(path string, info os.FileInfo) bool {
	// Check for UNIX-style hidden files/directories
	if strings.HasPrefix(filepath.Base(path), ".") {
		return true
	}

	// Check for Windows hidden attribute
	if info.Mode()&os.ModeHidden != 0 {
		return true
	}

	return false
}

func main() {
	root := "." // start from the current directory

	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}

		// Skip hidden files and directories
		if isHidden(path, info) {
			if info.IsDir() {
				return filepath.SkipDir // skip the entire directory if it's hidden
			}
			return nil // skip the hidden file and move to the next item
		}

		// Your logic to process the file or directory goes here
		fmt.Println(path)

		return nil
	})

	if err != nil {
		fmt.Printf("error walking the path: %v\n", err)
	}
}
```

smatvienko-tb (changes requested):
Hey, @anishrajan25 I hope my suggestion will help.

backend/pkg/filesystem/util.go
anishrajan25: @smatvienko-tb Thanks for pointing this out.
- It definitely makes more sense to skip a directory if it is hidden, rather than exploring each file inside it
- As for the platform-independent part,
    - I check the os package and it does not contain any constant as `ModeHidden`.
    - Hence I explored more and found out about conditional compilation and how to check if a file or directory is hidden on windows OS by using the `syscall` package. [ref](https://freshman.tech/snippets/go/detect-hidden-file/)
    - I will make the changes accordingly to have conditional compilation because the `syscall` package itself uses it and the `FILE_ATTRIBUTE_HIDDEN` constant that is required is only made available if the compilation is done on WINDOWS.

anishrajan25: I've updated the PR

smatvienko-tb: @anishrajan25, good progress!

I guess the new parameter would be nice to see in the config example: https://github.com/redpanda-data/console/blob/master/docs/config/console.yaml 

I recommend making some unit tests to be 100% sure and secure future updates.

smatvienko-tb (commented):
The PR looks good to me. Platforms, tests, and docs are here. Let's see what @weeco will say!

anishrajan25:
@smatvienko-tb Thank You for the review.

@weeco I was not able to run the test cases in `util_windows_test.go` because I do not have a windows system. Would you be able to confirm if the test cases are working fine?

bojand (approved):

backend/pkg/filesystem/util.go
```diff
@@ -41,3 +41,12 @@ func (c *Service) isValidFileExtension(filename string) (bool, string) {
...
 	return false, trimmedFilename
 }
+
+// shouldSkip returns true if the given file or directory needs to be skipped
+func (c *Service) shouldSkip(filename string) bool {
+	if c.Cfg.SkipHiddenFiles && isHidden(filename) {
+		return true
+	}
+
+	return false
```
weeco, line 51:
```suggestion
	if !c.Cfg.SkipHiddenFiles  {
		return false
	}

	return isHidden(filename)
```

weeco (approved):

weeco:
Thank you for your work and collaboration @anishrajan25 and @smatvienko-tb

PR #832, opened 2023-09-08
misc: Chakra progress bar

jvorcak:
Migrated progress bar to Chakra
Screenshot 2023-09-08 at 12 07 20

Screenshot 2023-09-08 at 12 07 21

Screenshot 2023-09-08 at 11 11 56

### DIFF
--- a/frontend/src/utils/tsxUtils.tsx
+++ b/frontend/src/utils/tsxUtils.tsx
@@ -396,7 +396,7 @@ export class StatusIndicator extends Component<StatusIndicatorProps> {
         const content =
             <ChakraProvider theme={redpandaTheme}>
                 <div style={{marginBottom: '0.2em'}}>
-                    <Progress value={this.props.fillFactor * 100} isIndeterminate={this.props.statusText === 'Connecting'}/>
+                    <Progress value={this.props.fillFactor * 100} isIndeterminate={this.props.statusText === 'Connecting'} colorScheme="brand" />
                     <div style={StatusIndicator.statusBarStyle}>
                         <div>{this.showWaitingText ? 'Kafka is waiting for new messages...' : this.props.statusText}</div>
                         <div style={StatusIndicator.progressTextStyle}>{this.props.progressText}</div>
--- a/frontend/src/utils/tsxUtils.tsx
+++ b/frontend/src/utils/tsxUtils.tsx
@@ -396,7 +396,11 @@ export class StatusIndicator extends Component<StatusIndicatorProps> {
         const content =
             <ChakraProvider theme={redpandaTheme}>
                 <div style={{marginBottom: '0.2em'}}>
-                    <Progress value={this.props.fillFactor * 100} isIndeterminate={this.props.statusText === 'Connecting'} colorScheme="brand" />
+                    <Progress
+                        value={this.props.fillFactor * 100}
+                        isIndeterminate={this.props.statusText === 'Connecting'}
+                        colorScheme="brand"
+                    />
                     <div style={StatusIndicator.statusBarStyle}>
                         <div>{this.showWaitingText ? 'Kafka is waiting for new messages...' : this.props.statusText}</div>
                         <div style={StatusIndicator.progressTextStyle}>{this.props.progressText}</div>
--- a/frontend/src/utils/tsxUtils.tsx
+++ b/frontend/src/utils/tsxUtils.tsx
@@ -400,12 +400,12 @@ export class StatusIndicator extends Component<StatusIndicatorProps> {
                             colorScheme="brand"
                         />
                     </Box>
-                    <Flex fontWeight="bold" fontSize="80%" fontFamily='"Open Sans", sans-serif'>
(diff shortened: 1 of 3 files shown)


frontend/src/utils/tsxUtils.tsx
```diff
@@ -394,25 +393,30 @@ export class StatusIndicator extends Component<StatusIndicatorProps> {
...
+                            </div>
+                            <div style={{display: 'flex', alignItems: 'center', marginLeft: 'auto'}}>
+                                <CopyOutlined style={{color: colors.brandOrange}}/>{this.props.messagesConsumed} messages
+                            </div>
+                        </div>
+                    }
                 </div>
-            }
-        </div>
+            </ChakraProvider>
```
malinskibeniamin, line 419: A few questions:

- When migrating this file to use Chakra UI, we have added another `ChakraProvider` component. Would it be possible to avoid it, given we would have several instances of Chakra provider, and it might be injecting CSS several times, I am a bit worried (I know this works, but maybe we don't need the nesting as much)
- Should the convention be to also migrate the remaining parts in this file, for example
```tsx
<div style={{display: 'flex', alignItems: 'center'}}>
```
becomes
```tsx
<Flex alignItems="center">
```

or
```tsx
<div style={StatusIndicator.statusBarStyle}>
```
becomes
```tsx
<Flex fontWeight="bold"> // The rest of the properties should be in a global config
```

We can also do it in separate PRs, but just wondering what the convention/strategy should be, to ensure that by the end after we migrated all components we are using Chakra in the designed way

frontend/src/utils/tsxUtils.tsx
```diff
@@ -330,7 +330,6 @@ interface StatusIndicatorProps {
 @observer
 export class StatusIndicator extends Component<StatusIndicatorProps> {
 
-    static readonly progressStyle: CSSProperties = { minWidth: '300px', lineHeight: 0 } as const;
     static readonly statusBarStyle: CSSProperties = { display: 'flex', fontFamily: '"Open Sans", sans-serif', fontWeight: 600, fontSize: '80%' } as const;
     static readonly progressTextStyle: CSSProperties = { marginLeft: 'auto', paddingLeft: '2em' } as const;
```
malinskibeniamin, line 334: [question] Shouldn't this be removed as part of the `Progress` refactor? We can just apply the styles to a `Text` or `Box` element or similar from Chakra, so that
```tsx
<div style={StatusIndicator.progressTextStyle}>{this.props.progressText}</div>
```
becomes
```tsx
<Text ml="auto" pl="2em">{StatusIndicator.progressTextStyle}</Text>
```

malinskibeniamin (commented):
Thanks for the change, but I think we may need to consider the 2 comments

frontend/src/utils/tsxUtils.tsx
jvorcak: @malinskibeniamin thanks for your comment and a note about the convention to convert styles. I'll do that from now on and I'll do that as part of this PR.

When it comes to the ChakraProvider, it's needed here because we're in a modal and thanks to `customRender` method the existing ChakraProvider doesn't work. I think this will no longer be an issue once we migrate modals to Chakra. Or do you have some other suggestion how to solve this "modal" problem?

frontend/src/utils/tsxUtils.tsx
jvorcak: You're right, I'll go through this and apply Chakra's Style Props to the whole PR (`progressStyle` , `progressTextStyle` and others). I'm a bit new to Chakra, but I get the point now.

malinskibeniamin: No worries, just making sure :) Happy to check again

frontend/src/utils/tsxUtils.tsx
malinskibeniamin: Good point, I think with modals it was an issue before, indeed once we have moved everything over including `Chakra` modals then hopefully the nested providers can go away.

jvorcak: @malinskibeniamin I'm a bit new to Chakra, so the fastest will be to ask. How to move the `fontFamily` and `fontSize` to global config? I guess these two specific props are relevant to this local component only right?


 `<Flex fontWeight="bold" fontSize="80%" fontFamily='"Open Sans", sans-serif'>`

malinskibeniamin: I don't think we should change the fontFamily only for this 1 particular component, why not keep using `Inter` like we do elsewhere?

You could use it in the theme.ts under `fonts`

jvorcak: @malinskibeniamin I wanted to reduce changes to the UI as much as possible. But I guess the consistency is more important. I'll change this once Progress component in merged in the UI

jvorcak: @malinskibeniamin I think we can merge

rikimaru0345 (approved):

PR #834, opened 2023-09-11
backend: add support for numerical record values

bojand:


backend/pkg/kafka/deserializer.go
```diff
@@ -315,7 +317,66 @@ func (d *deserializer) deserializePayload(payload []byte, topicName string, reco
 		}
 	}
 
-	// 8. Test for UTF-8 validity
+	// 8. Numerical values are tricky.
+	// If the payload is of specific length we can try to convert to a numerical value.
```
weeco, line 321: Maybe we should add another criteria - does not start with UTF-8 BOM?

weeco: Overall not sure if I like it, but we can revisit with the message viewer improvements. Let's discuss tomorrow or let me know what your thoughts are on this. Pros/Cons of doing this?

bojand: My initial thinking was to put this logic _after_ the utf8 section... That way basically the current state of things is unchanged in terms of existing behaviour with respect to textual or UTF8 data. This may be beneficial. It's a tough call..

For example if the payload is `[116 101 120 116]`, "text", which is integer value `1952807028`. 
In case of this PR we would interpret that as a uint32, and return `1952807028` as normalized value.

But it could be that the actual data was intended to be "text".

Conversely if the numerical logic is _after_ UTF8 logic, we would interpret that as "text". Which could be correct, or it could be that the actual data was intended to be numerical value of `1952807028`. 

I opted to place it before UTF8 because my thinking was that if payload data is actually textual data, in the normal real world, it seems unlikely to be specifically of lengths `1`, `2`, `4`, or `8`, but I suppose it is possible.

weeco (approved):

PR #837, opened 2023-09-12
WIP: Feature/migrate accordion component

jvorcak:


frontend/src/components/pages/consumers/Group.Details.tsx
```diff
@@ -29,10 +29,15 @@ import { EditOffsetsModal, GroupOffset, DeleteOffsetsModal, GroupDeletingMode }
 import { ShortNum } from '../../misc/ShortNum';
 import AclList from '../topics/Tab.Acl/AclList';
 import { SkipIcon } from '@primer/octicons-react';
-import { Flex, Section, Tabs, Tag, Tooltip, Popover } from '@redpanda-data/ui';
+import {Flex, Section, Tabs, Tag, Tooltip, Popover, Accordion} from '@redpanda-data/ui';
 import PageContent from '../../misc/PageContent';
 import { Features } from '../../../state/supportedFeatures';
 import { Statistic } from '../../misc/Statistic';
+
+function notNull<TValue>(value: TValue | null): value is TValue {
```
rikimaru0345, line 37: There is `filterNull` in `arrayExtensions.ts`

frontend/src/components/pages/connect/dynamic-ui/PropertyGroup.tsx
```diff
@@ -40,33 +39,27 @@ export const PropertyGroupComponent = observer((props: {
...
-                <div style={{ gridColumn: 'span 4', paddingLeft: '8px' }}>
-                    <Collapse ghost bordered={false}>
-                        {subGroups.map((subGroup) => (
-                            <Collapse.Panel
-                                className={subGroup.propertiesWithErrors.length > 0 ? 'hasErrors' : ''}
-                                key={subGroup.group.name ?? '<null>'}
-                                header={
-                                    <div style={{ display: 'flex', alignItems: 'center', gap: '1em' }}>
-                                        <span style={{ fontSize: '1.1em', fontWeight: 600, fontFamily: 'Open Sans' }}>
+                <div style={{gridColumn: 'span 4', paddingLeft: '8px'}}>
```
jvorcak, line 45: I haven't modified styles here because this code is currently not used (because of the backend) and I had no easy way to test it

weeco:
@jvorcak 
ss-nWD3WLVB

It looks a bit weird that one is blue while the other pills have red text. With the red text it may look a bit problematic, I think we should adopt a similar style as here on the consumer groups list page:

ss-G1yP3Ehx

jvorcak:
> @jvorcak ss-nWD3WLVB
> 
> It looks a bit weird that one is blue while the other pills have red text. With the red text it may look a bit problematic, I think we should adopt a similar style as here on the consumer groups list page:
> 
> ss-G1yP3Ehx

I totally agree, but isn't it a bit out of scope of this PR?

jvorcak:
@weeco following our discussion on Slack, I think the tags regression is there in master for quite some time. I'll be fixing in a different PR

rikimaru0345 (approved):

PR #846, opened 2023-09-17
feat: toast component migration

jvorcak:
This PR migrates the message and toast component to use Chakra's Toast implementation.

WIP: I'm in a middle of testing, but I think this PR can be already looked at.

- Chakra's Toast is implemented using a hook, so a functional component is required. 
  1. On a couple of places, I've migrated the component to a FC, this needs to be tested
  2. At the same time, I wanted to minimise the number of changes, so on some places I've left React as a component class and left a comment about a future migration. In these cases, I use Chakra's standalone toasts.
  3. A couple of messages are not migrated, because of the problem with Modals. Chakra's toast and Ant design modals doesn't play well together, because of ChakraProvider that kills the toasts on its removal. This will be addressed together with a modal migration.

Screenshot 2023-09-19 at 7 38 43

There are a couple of "special" places in the platform worth checking out:

-  Using Toast as a loader indicator
 Video:  https://www.loom.com/share/9749918a941b497688c8a0b56ea72258
 Code: https://github.com/redpanda-data/console/pull/846/files#diff-b8d5e9131eed247da5d345f825ecaccfed927cb04939825ae7a8335374bc7819
- Error screen
 Screen: 
Screenshot 2023-09-19 at 8 00 43

- Updating existing Toast
   https://www.loom.com/share/4127a1ac5c8144a6ad8462a35607af4a?from_recorder=1&focus_title=1

PR #847, opened 2023-09-20
Feature/chakra modals migration

jvorcak:
Migrates the usage of modals in a platform

Depends on https://github.com/redpanda-data/console/pull/846


frontend/src/components/misc/ErrorBoundary.tsx
```diff
@@ -219,6 +217,29 @@ function getStringFromInfo(info: InfoItem) {
     }
 }
 
+const CopyToClipboardButton: FC<{message: string, disabled: boolean; isLoading: boolean}> = ({message, disabled, isLoading}) => {
```
malinskibeniamin, line 220: [question] Why can't you just use `CopyButton`? Is it because it doesn't trigger a toast? We can add a UI library variant that would do that, to keep the logic in the library

https://redpanda-ui.netlify.app/?path=/story/components-copybutton--default

frontend/src/components/misc/HideStatisticsBarButton.tsx
```diff
@@ -9,26 +9,25 @@
...
-                    </div>
-                </div>
-            </Tooltip>
-        );
-    }
+export const HideStatisticsBarButton: FC<{ onClick: Function }> = ({onClick}) => {
+    const toast = useToast()
+    return (
+        <Tooltip label={<Text whiteSpace="nowrap">Hide statistics bar</Text>} placement="right" hasArrow>
+            <div className="hideStatsBarButton" onClick={() => {
```
malinskibeniamin, line 20: Consider using Box and removing classname so that styles are applied in the Chakra-preferred way

frontend/src/components/pages/reassign-partitions/ReassignPartitions.tsx
```diff
@@ -590,16 +634,29 @@ class ReassignPartitions extends PageComponent {
...
+                        status: 'success',
+                        description: baseText + ' -  Done',
+                        duration: 2500,
+                    })
                 }
                 else {
-                    msg.setError(baseText + ': ' + errors.length + ' errors');
+                    toast.update(toastId, {
+                        status: 'error',
+                        description: baseText + ': '  + errors.length + ' errors',
```
malinskibeniamin, line 657: Maybe string interpolation would be nicer here

jvorcak:
@malinskibeniamin thanks for the comments, I'll implement/consider them when doing a QA now. Today I've migrated all of the remaining modals, but there is still some details left, the following serves also as a note to myself:

- [ ] Making sure to remove inline styles in favour of Chakra
- [ ] Test everything properly
- [ ] Make sure our modals are consistent (gaps in a footer, icons next to the headers, etc)
- [ ] Fix comments by @malinskibeniamin 
- [ ] Fix z-index issues (Toasts vs Notification), right now Toasts are still open behind the ModalOverlay after the migration
- [ ] Get rid of the last 2 usages of antd's messages (`tsxUtils`, `utils/utils`)

frontend/src/components/misc/HideStatisticsBarButton.tsx
jvorcak: This seems like it's not working anyway. I'm a bit confused on what to do here. There is an eye icon when I investigate how it should work. But I think it doesn't and I personally find it a bit intrusive. Should we just get rid of this eye icon at all and keep it in the settings only?
Screenshot 2023-09-29 at 9 38 51

malinskibeniamin: Agreed, it does seem out of place. Let's hear from Moritz

rikimaru0345: Lets remove the entire "hide statistics bar" button.
It was an experiment that stuck around for a while, but it's not being used by people.

frontend/src/components/misc/ErrorBoundary.tsx
jvorcak: We can have a look at this, this was there from before but I think it's unrelated to Modals PR. It was here because the branch was based on top of toasts. I'll add it on my TODO list.

frontend/src/components/pages/topics/Topic.List.tsx
```diff
@@ -328,6 +326,8 @@ function ConfirmDeletionModal({ topicToDelete, onFinish, onCancel }: { topicToDe
     const [deletionPending, setDeletionPending] = useState(false);
     const [error, setError] = useState<string | Error | null>(null);
     const toast = useToast()
+    const cancelRef = React.useRef(null)
```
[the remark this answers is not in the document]
malinskibeniamin: Also, do you need to pass in the generic like `<HTMLButtonElement
 | null>`?

frontend/src/components/pages/reassign-partitions/components/ActiveReassignments.tsx
```diff
@@ -215,53 +224,56 @@ export const ThrottleDialog: FC<{ visible: boolean, lastKnownMinThrottle: number
...
+            onClose={onClose}
+        >
+            <ModalOverlay/>
+            <ModalContent minW="3xl">
+                <ModalHeader>Throttle Settings</ModalHeader>
+                <ModalBody>
+                    <Flex flexDirection="column" gap={4}>
+                        <Box mx={4}>
+                            <Text>Using throttling you can limit the network traffic for reassignments.</Text>
+                            <UnorderedList mt={2} px={6}>
```
malinskibeniamin, line 240: Thank you for migrating this, even though it wasn't part of the PR, it is great

frontend/src/components/pages/reassign-partitions/components/ActiveReassignments.tsx
```diff
@@ -10,22 +10,30 @@
...
 import { BrokerList } from '../../../misc/BrokerList';
-import { Button, Checkbox, Progress, ToastId, useToast } from '@redpanda-data/ui';
-
+import { Box, Button, Checkbox, createStandaloneToast, Flex, ListItem, Modal, ModalBody, ModalContent, ModalFooter, ModalHeader, ModalOverlay, Progress, redpandaToastOptions, Text, ToastId, UnorderedList, useToast } from '@redpanda-data/ui';
+
+// TODO - once ActiveReassignments is migrated to FC, we could should move this code to use useToast()
+const { ToastContainer, toast } = createStandaloneToast({
+    defaultOptions: {
+        ...redpandaToastOptions.defaultOptions,
+        isClosable: false,
```
malinskibeniamin, line 33: [question] Is it expected that the `isClosable` is set to false? Shouldn't we let the user dismiss the toast. Not sure how Console usually handles it.

frontend/src/components/pages/connect/helper.tsx
```diff
@@ -704,21 +705,29 @@ export const TaskState = observer((p: { observable: { state: ClusterConnectorTas
...
+                isOpen={err != null}
+                onClose={close}
+            >
+                <ModalOverlay />
+                <ModalContent minW="5xl">
+                    <ModalHeader>
+                        {
+                            task.taskId == null
+                                ? <>Error in Connector</>
+                                : <>{`Error trace of task ${task.taskId}`}</>
```
malinskibeniamin, line 719: I see we replaced `<h3>` by `<>`, maybe we can do `<Text as="h3">` or use the new variant? Or just revert back to `h3` and do the change later.

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -171,9 +171,7 @@ class AclList extends PageComponent {
         const m = createAutoModal({
             modalProps: {
                 title: 'Create User',
-                width: '80%',
-                style: { minWidth: '400px', maxWidth: '600px', top: '50px' },
-                bodyStyle: { paddingTop: '1em' },
+                style: { width: '80%', minWidth: '400px', maxWidth: '600px', top: '50px' },
```
malinskibeniamin, line 174: I think `paddingTop: '1em'` is missing, do we still need it?

frontend/src/components/misc/common.tsx
```diff
@@ -92,63 +91,60 @@ export function range(start: number, end: number): number[] {
...
+    if (serverTimestamp == curTimestamp)
+        return null; // version already matches
+
+    console.log('frontend update available', {
+        serverTimestamp: serverTimestamp,
+        serverDate: new Date(serverTimestamp * 1000),
+        localTimestamp: curTimestamp,
+        localDate: new Date(curTimestamp * 1000),
+        localVersion: clone(env),
+    });
```
malinskibeniamin, line 122: @rikimaru0345 do we want to keep this console log?

frontend/src/components/misc/common.tsx
```diff
@@ -92,63 +91,60 @@ export function range(start: number, end: number): number[] {
...
+                <ModalHeader>Redpanda Console has been updated</ModalHeader>
+                <ModalBody>The page must be reloaded to apply the newest version of the frontend.</ModalBody>
+                <ModalFooter gap={2}>
+                    <Button variant="outline" colorScheme="red" onClick={() => {
+                        setUpdateDialogOpen(false);
+                    }}>
+                        Cancel
+                    </Button>
+                    <Button variant="solid" onClick={() => {
+                        console.log('reloading frontend...');
```
malinskibeniamin, line 137: Same question, do we need this console log

frontend/src/components/misc/ErrorModal.tsx
```diff
@@ -32,59 +32,47 @@ class ErrorModal extends Component<ErrorModalProps> {
...
+                        {/* Header */}
+                        <Flex flexDirection="row" gap={2} pr={6}>
+
+                            {/* Icon */}
+                            <div style={{
+                                height: '70px', // height determines icon size
+                                alignSelf: 'center',
+                            }}>
+                                <XCircleIcon color="#F53649"/>
+                            </div>
```
malinskibeniamin, line 55: Don't worry about this for now, but we do have an `Icon` component too, so I guess this was just not available few months ago. We can migrate it later

frontend/src/components/pages/acls/Acl.List.tsx
jvorcak: It is intentional, it's not needed anymore

frontend/src/components/pages/connect/helper.tsx
jvorcak: I think this is handled by `<ModalHeader>`, but I transformed these fragments into strings

frontend/src/components/pages/reassign-partitions/components/ActiveReassignments.tsx
jvorcak: I tried to make "messages" non closable and toast closable.

jvorcak:
@malinskibeniamin did you also test in a cloud UI?

jvorcak:
So I've tested quite carefully today and I think the zIndices work as expected in cloud UI. If you guys want to check this again, feel free to do so, maybe you can test more scenarios. But I'd probably merge the UI dependency followed by this one. WDYT?

PR #849, opened 2023-09-22
feat(proto): secrets protos

sago2k8:
first version of proto for the the secrets endpoint


proto/redpanda/api/dataplane/v1alpha1/secret.proto
```diff
@@ -0,0 +1,230 @@
...
+    (buf.validate.field).map = {
+      values: {
+        string: {pattern: "^([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]*)$"}
+      },
+    },
+    (google.api.field_behavior) = IMMUTABLE
+  ];
+}
+
+message SecretInput {
```
sago2k8, line 26: created this due this limitation, https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/#:~:text=For%20IMMUTABLE%20and%20INPUT_ONLY%20fields%2C%20there%20is%20an%20open%20issue%20in%20the%20Open%20API%20specification%20for%20adding%20functionality%20for%20write%2Donce%20or%20immutable%20fields%20to%20the%20spec.

proto/redpanda/api/dataplane/v1alpha1/secret.proto
```diff
@@ -0,0 +1,227 @@
...
+
+// UpdateSecretResponse is the response of UpdateSecret
+message UpdateSecretResponse {
+  Secret secret = 1;
+}
+
+// DeleteSecretRequest is the request of DeleteSecret
+message DeleteSecretRequest {
+  // The id of the secret you want to delete
+  string id = 1;
```
bojand, line 94: Minor nit, maybe we should add some validation (ex min and max width?) for `id` and `name` fields in different properties? For example something like:
```proto
[
    (buf.validate.field).string.min_len = 1,
    (buf.validate.field).string.max_len = 128,
      min_length: 1,
      max_length: 128
    }
  ];
  ```

PR #874, opened 2023-10-11
Schema registry management

rikimaru0345:
As discussed with @bojand :
Even though "Producing typed messages" is part of the schema-registry improvements, we need to merge those improvements first, before we can start on "producing typed messages".

The frontend changes in this PR are complete (as complete as they can be), and can be deployed standalone.
There were some frontend merge issues that I fixed, and some backend merge issues that @bojand fixed.

For backend especially, it would be good if someone else could double-check if everything is as it should be.


backend/pkg/api/handle_schema_registry.go
```diff
@@ -0,0 +1,819 @@
...
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		rest.SendResponse(w, r, api.Logger, http.StatusOK, &response{IsConfigured: false})
+	}
+}
+
+func (api *API) handleGetSchemaRegistryMode() http.HandlerFunc {
+	if !api.Cfg.Kafka.Schema.Enabled {
+		return api.handleSchemaRegistryNotConfigured()
+	}
```
bojand, line 39: Not relevant to the current state of things, but it might be worth calling out... I think that if we ever implement hot reloading for `Kafka.Schema` where users can enable `Kafka.Schema` and set other properties and we reload, these APIs will then still return the "not configured" responses? Not sure if it's worth future-proofing somehow perhaps, or just commenting somewhere?

backend/pkg/api/handle_schema_registry_integration_test.go
```diff
@@ -0,0 +1,126 @@
...
+		require.Equal(200, res.StatusCode)
+
+		validationResponse := console.SchemaRegistrySchemaValidation{}
+		err := json.Unmarshal(body, &validationResponse)
+		require.NoError(err)
+
+		assert.False(validationResponse.IsValid)
+		assert.NotEmpty(validationResponse.ParsingError, "parsing error should be set")
+	})
+}
```
bojand, line 126: We should perhaps add a follow up task to expand these integration tests to cover more of the schema registry APIs?

backend/pkg/console/schema_registry.go
```diff
@@ -0,0 +1,569 @@
...
+// Use version = 'all' to retrieve all schemas for this subject.
+// Use version = 3 to retrieve the specific version for the given subject
+func (s *Service) GetSchemaRegistrySubjectDetails(ctx context.Context, subjectName, version string) (*SchemaRegistrySubjectDetails, error) {
+	// 1. Retrieve all schema versions registered for the given subject
+	versions, err := s.getSchemaRegistrySchemaVersions(ctx, subjectName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve subject versions: %w", err)
+	}
+	latestActiveVersion := -1
+	softDeletedVersions := make(map[int]bool)
```
bojand, line 167: minor performance nit, we can initialize capacity in this `make()` call.
```go
softDeletedVersions := make(map[int]bool, len(versions))
```

backend/pkg/console/schema_registry.go
```diff
@@ -0,0 +1,569 @@
...
+		versionsToRetrieve = make([]string, len(versions))
+		for i, v := range versions {
+			versionsToRetrieve[i] = strconv.Itoa(v.Version)
+		}
+	}
+
+	schemasCh := make(chan *SchemaRegistryVersionedSchema, len(versionsToRetrieve))
+	for _, v := range versionsToRetrieve {
+		copiedVersion := v
+		grp.Go(func() error {
```
bojand, line 206: Should we bound this concurrency? Probably unlikely to be an issue, but if there are a lot of versions it could result in a lot of concurrent requests to the SR API.

backend/pkg/console/schema_registry.go
```diff
@@ -0,0 +1,569 @@
...
+		})
+	}
+
+	if err := grp.Wait(); err != nil {
+		return nil, err
+	}
+	close(schemasCh)
+
+	// Send requests to retrieve schemas
+	schemas := make([]*SchemaRegistryVersionedSchema, 0)
```
bojand, line 226: Another minor perf nit, I think we can make with capacity of `len(versionsToRetrieve)`

backend/pkg/console/schema_registry.go
```diff
@@ -0,0 +1,569 @@
...
+
+	// 2. Get versions with soft-deleted
+	g.Go(func() error {
+		versionsRes, err := s.kafkaSvc.SchemaService.GetSubjectVersions(ctx, subjectName, true)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve subject versions (with soft-deleted): %w", err)
+		}
+
+		for _, v := range versionsRes.Versions {
+			versionsWithSoftDeleted[v] = struct{}{}
```
bojand, line 290: same concurrent write concern here

backend/pkg/console/schema_registry.go
```diff
@@ -0,0 +1,569 @@
...
+// subject-version. You can use -1 or 'latest' to check the latest version.
+func (s *Service) GetSchemaRegistrySchemaReferencedBy(ctx context.Context, subjectName, version string) ([]SchemaReference, error) {
+	schemaRefs, err := s.kafkaSvc.SchemaService.GetSchemaReferences(ctx, subjectName, version)
+	if err != nil {
+		return nil, err
+	}
+
+	ch := make(chan SchemaReference, len(schemaRefs.SchemaIDs))
+	grp, grpCtx := errgroup.WithContext(ctx)
+	for _, schemaID := range schemaRefs.SchemaIDs {
```
bojand, line 382: Similar as above, should we limit the amount of concurrency?

backend/pkg/api/handle_schema_registry.go
weeco: I'll leave this as is for now since we don't have any plans to introduce hot-reloading and I think this should immediately pop-up as an issue when testing this functionality.

backend/pkg/console/schema_registry.go
weeco: I actually thought about it when writing the code, but I wasn't sure if it's actually a performance improvement, because the number of soft deleted versions is not known or predictable. My hypothesis is that most of the time there's no soft deleted version in a subject.

backend/pkg/console/schema_registry.go
weeco: I wondered too! We are not really sure what rate limits we should apply though because the SR has never been benchmarked - the majority of the retrieved information seems to be served from memory. I think the larger factor that drives the concurrency here is the number of concurrent incoming requests (one request for each sr subject).

I think it's very rare to have a significant number of versions (e.g. more than 20). However, I added a limit of 10 concurrent Go routines, it shouldn't hurt as you said

weeco (approved):

PR #875, opened 2023-10-12
Migration of a breadcrumb

jvorcak:
This should be a 1:1 replacement. I've checked usages everywhere and tested.
We can always adjust this later with more functionality.

Screenshot 2023-10-12 at 11 03 14
Screenshot 2023-10-12 at 11 02 09


rikimaru0345:
Does the breadcrumb work in all pages? 
There may be some pages that might look strange and would need to be updated iirc.

Also have you tried this in embedded mode? (where the breadcrumbs are rendered by cloud-ui, not console; but console still renders the page title and the refresh buttons)

rikimaru0345:
>  I could + also tested cloudUI. 
Sounds good to me!
Should there be any inconsistencies beyond that, they'll be caught/fixed in the design review

rikimaru0345 (approved):

PR #879, opened 2023-10-16
Backend: Add Common protobuf types

bojand:
Adds Common Protobuf types [as included](https://github.com/redpanda-data/redpanda/tree/dev/src/v/pandaproxy/schema_registry/protobuf) in Redpanda to Console.

Adds a utility package to `schema` package that embeds the protobuf files and provides utility functions to access them.

In `proto` and `schema` packages we add the common protobuf types to file maps before doing protobuf parsing.


backend/pkg/proto/service.go
```diff
@@ -498,6 +499,18 @@ func (s *Service) protoFileToDescriptor(files map[string]filesystem.File) ([]*de
 		return nil
 	}
 
+	// Add common proto types
```
weeco, line 502: NIT: Can we add a link to a reference here, which tells what other types besides the Google WKT should be included? Just in case we are missing something in the future because something has been added, we can look it up from there ideally.

weeco (approved):
Nicely done 👍

PR #892, opened 2023-10-18
Migrates more Select components to Chakra

jvorcak:
Screenshot 2023-10-18 at 20 48 29


frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -10,8 +10,8 @@
  */
 
 import { ClockCircleOutlined, DeleteOutlined, DownloadOutlined, SettingFilled, SettingOutlined } from '@ant-design/icons';
-import { DownloadIcon, PlusIcon, SkipIcon, SyncIcon, XCircleIcon, KebabHorizontalIcon } from '@primer/octicons-react';
-import { ConfigProvider, DatePicker, Radio, Select, Table } from 'antd';
+import { DownloadIcon, KebabHorizontalIcon, PlusIcon, SkipIcon, SyncIcon, XCircleIcon } from '@primer/octicons-react';
```
malinskibeniamin, line 13: In the future, before the migration ends, probably we should do an audit and consider using 1 set of icons like discussed previously. Of course nothing to change for now.

frontend/src/components/pages/topics/PublishMessagesModal/PublishMessagesModal.tsx
```diff
@@ -60,59 +60,74 @@ export class PublishMessagesModalContent extends Component<Props> {
 
     render() {
         return (
-            <div className="publishMessagesModal">
-                <div style={{ display: 'flex', gap: '1em', flexWrap: 'wrap' }}>
+            <Flex gap={4} flexDirection="column">
```
malinskibeniamin, line 63: I think we lost the `flexWrap`? Maybe it's better to be verbose just in case

frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
```diff
@@ -311,24 +311,9 @@ const ResourceACLsEditor = observer((p: {
...
-                        flexDirection: 'column',
-                        rowGap: '8px',
-                        columnGap: '20px',
-                        flexWrap: 'wrap',
-                        width: '100%',
-                        maxHeight: '60px',
-                        alignItems: 'flex-start'
-                    }}
-                >
+            <Label text="Operations" style={{width: '100%'}}>
```
malinskibeniamin, line 314: Do we need the remaining styles from the diff?

frontend/src/components/pages/topics/Tab.Messages/index.tsx
jvorcak: @malinskibeniamin for sure, I'm 100% onboard with this, but hopefully in a different PR

frontend/src/components/pages/topics/PublishMessagesModal/PublishMessagesModal.tsx
jvorcak: @malinskibeniamin the relevant flexWrap moved to line 91

frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
jvorcak: Unfortunately yes, but I plan to remove these `Operations` label completely in the future iteration

jvorcak:
Screenshot 2023-10-30 at 12 16 32

One UX change that is contained in this PR is this one. In the previous version, if one item was contained in a select, it was disabled and users didn't get an explanation why, they couldn't even see the option in the select.

Right now, we just show it's one option and don't disable it. I think it's a better a11y convention that doesn't block user if we really don't have to. Also I like to disable states with explanations why they're disabled. Are you OK with it?

malinskibeniamin (approved):
Changes make sense, about disabling options within Select, let's ask @rikimaru0345 or @frankieshakes

frankieshakes:
> Changes make sense, about disabling options within Select, let's ask @rikimaru0345 or @frankieshakes

I _think_ there was a discussion with @ivpanda about this as well. I'm fine with us making the change @jvorcak describes, as I feel it's better overall for UX/usability.

ivpanda:
In that scenario, I prefer not disabling anything— because they _are_ able to make a selection, it just so happens that there is only one item to select. The only time I would disable anything is if: 

1. There are multiple options, and some are enabled and some are not— the ones that are unavailable are disabled (for reasons) 
2. The entire dropdown cannot be accessed (i.e. it _cannot_ be changed, even if there are options).

PR #895, opened 2023-10-20
Improve message viewer

weeco:


backend/pkg/serde/service.go
```diff
@@ -0,0 +1,282 @@
...
+	addTS := opts.Troubleshoot
+	if rp == nil || err != nil || (rp != nil && rp.Encoding == "") {
+		// Anything else is considered binary
+		rp = &RecordPayload{
+			Encoding: PayloadEncodingBinary,
+		}
+
+		// if we failed to find a successful serde always add troubleshooting info
+		addTS = true
+	}
```
weeco, line 138: Would it be cleaner to add a default serde at the very end of the serde list? Not sure, but the if condition feels a bit hard to understand / maintain.

backend/pkg/serde/service.go
```diff
@@ -0,0 +1,282 @@
...
+		found = true
+
+		bytes, serErr := serde.SerializeObject(ctx, input.Key.Payload, PayloadTypeKey, input.Key.Options...)
+		if serErr == nil {
+			err = nil
+			keySerResult.Encoding = serde.Name()
+			keySerResult.Payload = bytes
+			break
+		}
+		err = serErr
```
weeco, line 220:
```suggestion
		var bytes []byte // I assume?
		bytes, err = serde.SerializeObject(ctx, input.Key.Payload, PayloadTypeKey, input.Key.Options...)
		if err == nil {
			keySerResult.Encoding = serde.Name()
			keySerResult.Payload = bytes
			break
		}
```

weeco: Maybe also worth to refactor it so that you just find the serde in that for loop and proceed outside the for loop with the serialization of the found serde

backend/pkg/serde/service.go
```diff
@@ -0,0 +1,282 @@
...
+
+		bytes, serErr := serde.SerializeObject(ctx, input.Value.Payload, PayloadTypeValue, input.Value.Options...)
+		if serErr == nil {
+			err = nil
+			valueSerResult.Encoding = serde.Name()
+			valueSerResult.Payload = bytes
+			break
+		}
+
+		err = serErr
```
weeco, line 260: Would do the same refactoring here as suggested for the key

backend/pkg/kafka/produce_records.go
```diff
@@ -15,14 +15,17 @@ import (
...
-	compressionTypeSnappy // 2
-	compressionTypeLz4    // 3
-	compressionTypeZstd   // 4
+	// CompressionTypeNone means no compression.
+	CompressionTypeNone   = iota
+	CompressionTypeGzip   // CompressionTypeGzip represents Gzip compression.
+	CompressionTypeSnappy // CompressionTypeSnappy represents Snappy compression.
+	CompressionTypeLz4    // CompressionTypeLz4 represents LZ4 compression.
+	CompressionTypeZstd   // CompressionTypeZstd represents ZSTD compression.
 )
```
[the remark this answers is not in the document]
bojand: The current API `POST /api/topics-records` handled by `handlePublishTopicsRecords()` uses `CompressionType int8`, so to keep this around for now we need some support for this. I have changed all Publish and Produce functions in kafka and console packages to take `kgo.CompressionCodec` options, and moved the supporting utility function into `pkg/api/handle_topic_publish_records.go` for easier removal later on.

backend/pkg/serde/serde.go
```diff
@@ -0,0 +1,125 @@
...
+func WithSchemaPath(path string) SerdeOpt {
+	return serdeOpt{func(t *serdeCfg) { t.schemaPath = path }}
+}
+
+// WithTopic adds a topic name to serde options.
+func WithTopic(topic string) SerdeOpt {
+	return serdeOpt{func(t *serdeCfg) { t.topic = topic }}
+}
+
+// WithUintSize adds the uint size to use for serialization and deserialization of numeric payloads.
```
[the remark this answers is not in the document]
bojand: Maybe the name of the option function is confusing. This is an option function that configures the size of "numeric" payload for serializing. If for example we want to serialize a payload (key of value) as a numeric value (uint serde), we also need a way to specify the size preferred. This function helps accommodate the requirement. The input is already an enum:

```go
// UintSize is the size of uint to be used in uint serialization.
type UintSize uint

const (
	// Uint8 is the uint of size of 8 bits.
	Uint8 UintSize = iota
	// Uint16 is the uint of size of 18 bits.
	Uint16
	// Uint32 is the uint of size of 32 bits.
	Uint32
	// Uint64 is the uint of size of 64 bits.
	Uint64
)
```

backend/pkg/serde/service.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	troubleshooting := make([]TroubleshootingReport, 0, len(s.SerDes))
+
+	serdeEncoding := opts.KeyEncoding
+	if payloadType == PayloadTypeValue {
+		serdeEncoding = opts.ValueEncoding
+	}
+
+	doCustomEncoding := serdeEncoding != PayloadEncodingUnspecified && serdeEncoding != ""
```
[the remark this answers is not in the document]
bojand: When deserializing we have an option to specify the exact encoding we wish to use. This is a check to determine if the preferred encoding is set. If so we only need to loop to that specific serde and attempt deserializing.
I have changed the name of the variable to `doSpecificEncoding` and added a comment. Hopefully that clears it up.

rikimaru0345 (approved):

PR #906, opened 2023-10-26
Migrates remaining Alert components

jvorcak:
This migrates remaining occurrences of an AntD Alert.

The most untypical one is this one, that works exactly like before.

Screenshot 2023-10-26 at 21 05 10


frontend/src/components/pages/connect/CreateConnector.tsx
```diff
@@ -599,12 +608,13 @@ function getDataSource(validationResult: ConnectorValidationResult) {
...
-                <>
-                    <h3>Submitted configuration is invalid</h3>
-
+            status="warning"
+            variant="left-accent"
+            my={4}
+        >
+            <AlertDescription>
+                <Box>
+                    <Text as="h3" mb={4}>Submitted configuration is invalid</Text>
```
malinskibeniamin, line 617: @frankieshakes is there any typography text variant we should be using here, or is it fine as-is?

frontend/src/components/pages/connect/helper.tsx
```diff
@@ -594,11 +593,20 @@ export const ConfirmModal = observer(<T,>(props: ConfirmModalProps<T>) => {
...
+                {err && <Box mt={4}>
+                     <Alert
+                         status="error"
+                         variant="left-accent"
+                     >
+                         <AlertIcon />
+                         <AlertDescription>
+                             <Box>
+                                 <Text as="h3">{err.title}</Text>
+                                 <Text as="p">{err.content}</Text>
```
malinskibeniamin, line 605: I don't think you need the `as="p"`, if I remember correctly, by default `Text` renders a paragraph

frontend/src/components/pages/connect/CreateConnector.tsx
frankieshakes: The `as` is fine here if it _must_ be an H3 for semantic reasons. Otherwise, the default `<Text>` (with no variant) should suffice.

frontend/src/components/pages/connect/helper.tsx
frankieshakes: Yup. `<Text>` defaults to a `<p>` tag.

frontend/src/components/pages/connect/CreateConnector.tsx
jvorcak: I think semantically `h3` makes sense here, but we should probably create some styleguide what to use where in the future for the whole UI

jvorcak:
I'll wait with resolving conflicts after https://github.com/redpanda-data/console/pull/892 is merged

jvorcak:
@malinskibeniamin @frankieshakes this is ready for re-review once you have time

malinskibeniamin (approved):

PR #907, opened 2023-10-27
Migrates Skeleton component from AntD

jvorcak:
This PR Migrates `Skeleton` to Chakra's `SkeletonText`. I think the look'n'feel is very similar to the original.

Screenshot 2023-10-27 at 9 19 35
Screenshot 2023-10-27 at 9 19 16

### DIFF
--- a/frontend/src/components/pages/overview/Overview.tsx
+++ b/frontend/src/components/pages/overview/Overview.tsx
@@ -10,7 +10,6 @@
  */
 
 import { observer } from 'mobx-react';
-import { Skeleton } from 'antd';
 import { PageComponent, PageInitHelper } from '../Page';
 import { api } from '../../../state/backendApi';
 import { uiSettings } from '../../../state/ui';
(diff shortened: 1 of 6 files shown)


frontend/src/components/pages/overview/Overview.tsx
```diff
@@ -213,20 +212,23 @@ class Overview extends PageComponent {
                             </ul>
 
                             <ul className="resource-list">
-                                {news
-                                    ? news.map((x, i) => <li key={i}>
-                                        <a href={x.url} rel="noopener noreferrer" target="_blank" className="resource-link">
+                                <SkeletonText
+                                    isLoaded={Boolean(news)}
+                                    noOfLines={4}
```
malinskibeniamin, line 217: I see we had 3 lines previously, not sure if it's a big deal.

frontend/src/components/pages/overview/Overview.tsx
```diff
@@ -261,8 +263,10 @@ const ResourcesBadge = (p: { type?: string | undefined }) => {
 function ClusterDetails() {
     const overview = api.clusterOverview;
     const brokers = api.brokers;
-    if (!overview || !brokers)
-        return <Skeleton paragraph={{ rows: 16 }} />
+
+    if (!overview || !brokers) {
+        return <SkeletonText mt={5} noOfLines={13} spacing={5} skeletonHeight={4} speed={0} />
```
malinskibeniamin, line 268: 16 -> 13?
also is it intentional that here we have speed = 0

malinskibeniamin (approved):
Just some small questions, we are getting close!

frontend/src/components/pages/overview/Overview.tsx
jvorcak: I tried to check the UI when migrating and somewhere I tried to fine-tune some of the places. The original and this component are not exactly 1:1

frontend/src/components/pages/overview/Overview.tsx
jvorcak: For `Skeleton` you need to provide `isActive` or sth like that, for `SketetonText` it's the other way around, you need to disable animation with speed=0, so I think this is correct.

PR #909, opened 2023-10-30
Migrate Input Components to Chakra

jvorcak:
Screenshot 2023-10-30 at 21 00 42
Screenshot 2023-10-30 at 21 00 39

### DIFF
--- a/frontend/src/components/misc/KowlTable.tsx
+++ b/frontend/src/components/misc/KowlTable.tsx
@@ -10,7 +10,7 @@
  */
 
 import React, { Component } from 'react';
-import { Input, InputRef, Pagination, Table } from 'antd';
+import { Pagination, Table } from 'antd';
 import { ColumnType } from 'antd/lib/table';
 import styles from './KowlTable.module.scss';
 import { ColumnTitleProps, ExpandableConfig, FilterDropdownProps, FilterValue, SorterResult, TableCurrentDataSource, TablePaginationConfig } from 'antd/lib/table/interface';
@@ -21,7 +21,7 @@ import { clone } from '../../utils/jsonUtils';
 import { SearchOutlined } from '@ant-design/icons';
 import Highlighter from 'react-highlight-words';
 import { findPopupContainer } from '../../utils/tsxUtils';
-import { Checkbox } from '@redpanda-data/ui';
+import { Box, Checkbox, Input } from '@redpanda-data/ui';
 
 type EnumFilter = {
     type: 'enum',
@@ -539,7 +539,7 @@ export class SearchTitle extends Component<{
     observableSettings: { quickSearch: string },
 }>
 {
-    inputRef = React.createRef<InputRef>(); // reference to input, used to focus it
+    inputRef = React.createRef<HTMLInputElement>(); // reference to input, used to focus it
 
     constructor(p: any) {
         super(p);
@@ -560,15 +560,14 @@ export class SearchTitle extends Component<{
         setTimeout(this.focusInput);
 
         return <span>
-            <span >{this.props.title}</span>
-            <div className="tableInlineSearchBox"
+            {!props.observableFilterOpen.filterOpen && <span>{this.props.title}</span>}
+            <Box
                 onClick={e => e.stopPropagation()}
                 onMouseDown={e => e.stopPropagation()}
                 onMouseUp={e => e.stopPropagation()}
                 style={{
                     position: 'absolute', inset: '0px 0px 0px -8px',
-                    display: 'flex', placeContent: 'center', placeItems: 'center',
-                    padding: '4px 6px',
+                    display: 'flex', placeContent: 'center', placeItems: 'center'
                 }}
             >
                 <Input
@@ -586,17 +585,15 @@ export class SearchTitle extends Component<{
                         }
                     }}
                     onKeyDown={this.onKeyDown}
-                    allowClear={true}
                     placeholder="Enter search term/regex"
                     value={props.observableSettings.quickSearch}
 
                     onChange={e => {
                         props.observableSettings.quickSearch = e.target.value;
                     }}
-                    style={{ borderRadius: '3px' }}
                     spellCheck={false}
                 />
-            </div>
+            </Box>
         </span>
     }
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/pages/topics/CreateTopicModal/CreateTopicModal.tsx
```diff
@@ -590,15 +600,15 @@ export function RatioInput(p: {
             onChange={x => p.onChange(x / 100)}
             tooltip={{ formatter: null }}
         />
-        <InputNumber
+        <NumberInput
             min={0} max={100}
-            value={Math.round(p.value * 100)}
+            value={Math.round(p.value * 100) + '%'}
```
jvorcak, line 605: This one is a bit hard to test as I'm not getting `RATIO` config entry type cc @weeco . Is it still used?

I've tried to test by emulating this for other values and it seems to work.

frontend/src/index.scss
```diff
@@ -2804,8 +2775,6 @@ $easeInOutCirc: cubic-bezier(0.785, 0.135, 0.150, 0.860);
     background: rgba(200, 205, 210, 0.16);
 
     max-height: 300px;
-    overflow: hidden;
```
jvorcak, line 2807: Removing this lines fixes an overflow in a Preview Fields modal. Before it created a scroll element, now it relies on an extendability of a modal and plays much better with overlays.

frontend/src/components/pages/topics/CreateTopicModal/CreateTopicModal.tsx
rikimaru0345: So far, I have not seen a ratio type entry yet. So it should be ok for now; if there is an issue, we will find it later when there is an actually reproducible situation

rikimaru0345 (approved):

frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -133,15 +133,18 @@ const ConfigurationEditor: FC<ConfigurationEditorProps> = observer((props) => {
                                             {$state.modalValueType === 'custom' && <>
                                                 <Text fontWeight="bold">Set at topic configuration</Text>
                                                 <Box maxWidth={300}>
-                                                    <ConfigEntryEditor className={'configEntryEditor ' + (isCustom ? '' : 'disabled')} entry={configEntry}/>
+                                                    ....{configEntry.frontendFormat}
+                                                    <ConfigEntryEditor
+                                                        className={'configEntryEditor ' + (isCustom ? '' : 'disabled')}
```
malinskibeniamin, line 138: Can/should we use `clsx` or `classnames` in these kinds of situations? @rikimaru0345 this question might be better answered by you, it's just a style guide we want to have for conditional classnames. Ideally we don't use classnames at all, so maybe we can just disregard this for now and fix later.

frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -133,15 +133,18 @@ const ConfigurationEditor: FC<ConfigurationEditorProps> = observer((props) => {
                                             {$state.modalValueType === 'custom' && <>
                                                 <Text fontWeight="bold">Set at topic configuration</Text>
                                                 <Box maxWidth={300}>
-                                                    <ConfigEntryEditor className={'configEntryEditor ' + (isCustom ? '' : 'disabled')} entry={configEntry}/>
+                                                    ....{configEntry.frontendFormat}
```
malinskibeniamin, line 136: I think you meant to spread the object?
```...{configEntry.frontendFormat}```

File: frontend/src/components/misc/KowlTable.tsx [hunk 5]
```diff
@@ -560,15 +560,14 @@ export class SearchTitle extends Component<{
         setTimeout(this.focusInput);
 
         return <span>
-            <span >{this.props.title}</span>
-            <div className="tableInlineSearchBox"
+            {!props.observableFilterOpen.filterOpen && <span>{this.props.title}</span>}
```
malinskibeniamin, line 563: small suggestion, `<Text>` (not a big deal)

frontend/src/components/pages/topics/TopicConfiguration.tsx
jvorcak: @malinskibeniamin removed

frontend/src/components/pages/topics/TopicConfiguration.tsx
jvorcak: I'm all for using `clsx` or `classNames`, I just didn't migrate it here.

frontend/src/components/misc/KowlTable.tsx
jvorcak: I actually wanted to discuss this before. Are we sure about migrating stuff to Chakra that doesn't actually provide any value. I'd personally vote to sticking to standard HTML as much as possible.

We're now using Chakra and will use it for some time, but hypothetically if in a future it's decided to migrate away from it to something else that ever changing JS world will provide, I think it'll be easier to migrate away from using HTML rather than Charka. I know we're not migrating away from Chakra, but who knows what will be usable and valuable in 2-3 years timeframe.

Of course, if there is any value in using `Box` instead of `div`, `Text` instead of `p` that I'm not seeing, I'm all up to it.

WDYT?

frontend/src/components/misc/KowlTable.tsx
```diff
@@ -560,15 +560,14 @@ export class SearchTitle extends Component<{
...
+            <Box
```
malinskibeniamin, line 564: Small suggestion, can just use <Flex> instead of <Box display="flex">, also if it's a Chakra component you don't need `style` object, can just use the props instead.

frontend/src/components/pages/topics/DeleteRecordsModal/DeleteRecordsModal.tsx
```diff
@@ -277,7 +278,7 @@ const ManualOffsetContent = observer(
                     className={styles.slider}
                 />
                 <Input
-                    className={styles.sliderValue}
+                    maxWidth={124}
```
malinskibeniamin, line 281: Why specifically this value? Just curious, does it look better?

malinskibeniamin (approved):
A couple of small suggestions

frontend/src/components/pages/topics/DeleteRecordsModal/DeleteRecordsModal.tsx
jvorcak: https://github.com/redpanda-data/console/pull/909/files#diff-d349e3386d50bc0993600dd1ae2a8530c0ec6bd7005f2ade392a1257d25f3e76L18 I've migrated 1:1. We need however introduce some consistency globally using a styleguide.

PR #912, opened 2023-11-07
fix msgpack filtering

skarrok:
Currently msgpack-encoded message is passed as string into JS VM. So it does not allow filtering by object fields.

This little fix changes msgpack object to be the same as JSON, XML and other encoded messages.


weeco:
Hey @skarrok , is this supposed to be a draft PR or ready for review?

skarrok:
Hi!
I had trouble running tests locally, so can not confirm this doesn't break something.

But it's one line change, I guess you can review it as is.
Or I can figure out my issue with tests and mark pr as ready later.

weeco (approved):

PR #932, opened 2023-11-23
frontend: Fix console issues rerendering when in embedded mode

malinskibeniamin:
We want to make sure that ChakraUI SkeletonText component is not used anymore, because it is causing rerendering, which triggers unnecessary loading/unloading of Console when in embedded mode, breaking the connection with the parent.

In addition, we want to explicitly pass the information from the parent that it's now safe to proceed with rendering, so that it's not prematurely fetching data when it's not yet ready to show anything on the screen.

### DIFF
--- a/frontend/src/EmbeddedApp.tsx
+++ b/frontend/src/EmbeddedApp.tsx
@@ -33,18 +33,25 @@ import { observer } from 'mobx-react';
 import { ChakraProvider, redpandaTheme, redpandaToastOptions } from '@redpanda-data/ui';
 
 export interface EmbeddedProps extends SetConfigArguments {
-    // This is the base url that is used:
-    //   - when making api requests
-    //   - to setup the 'basename' in react-router
-    //
-    // In the simplest case this would be the exact url where the host is running,
-    // for example "http://localhost:3001/"
-    //
-    // When running in cloud-ui the base most likely need to include a few more
-    // things like cluster id, etc...
-    // So the base would probably be "https://cloud.redpanda.com/NAMESPACE/CLUSTER/"
-    //
+    /**
+     * This is the base url that is used:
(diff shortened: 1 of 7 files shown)


c4milo (approved):

frontend/src/EmbeddedApp.tsx
```diff
@@ -68,11 +75,15 @@ function EmbeddedApp({ basePath, ...p }: EmbeddedProps) {
 
     setup(p);
 
+    if (!p.isConsoleReadyToMount) {
```
sago2k8, line 78: We have to make sure we support this in every environment before we create a install pack with this console version

frontend/src/utils/tsxUtils.tsx
```diff
@@ -487,15 +487,15 @@ export class ZeroSizeWrapper extends Component<{ width?: string, height?: string
...
+const innerSkeleton = <Skeleton noOfLines={8} height={4} />
 export const DefaultSkeleton = (
     <motion.div {...animProps} key={'defaultSkeleton'} style={defaultSkeletonStyle}>
         {innerSkeleton}
     </motion.div>
 );
 
 export const InlineSkeleton = (p: { width: string | number }) => (
-    <SkeletonText noOfLines={1} skeletonHeight="2" width={p.width} display="flex" alignItems="center" />
+    <Skeleton noOfLines={1} height={2} width={p.width} display="flex" alignItems="center" />
```
yougotashovel, line 498: [question] why was SkeletonText causing the rerendering? Is it not just an animated dumb component?

frontend/src/EmbeddedApp.tsx
malinskibeniamin: Agreed

PR #933, opened 2023-11-23
frontend: Replace SkeletonText with Skeleton to fix rerenders

sago2k8:
This is a fix for the Chakra UI issue which causes rerendering at least twice, making Console unusable


malinskibeniamin:
Related to https://github.com/redpanda-data/console/pull/932, but we can merge #933 first and then rebase.

malinskibeniamin:
we already applied this change so we can leave it out for now.

PR #936, opened 2023-11-23
Change to render the number of partitions and replicas

### LINKED ISSUE #935: Topic partitions display shows replicas
Right now Console shows under "Total Partitions" the number of replicas, rather than the number of partitions. The finer detail is hidden in a hover-over:

image

Instead we should show Partitions and Replicas separately:

image

tmgstevens:
Fixes #935


malinskibeniamin (approved):
Can you post a screenshot of how it looks like now?

rikimaru0345 (approved):

Issue #946, opened 2023-11-30
containsControlChars treat LF(10) as a control character


cobolbaby:
https://github.com/redpanda-data/console/blob/1aa62628864b502a6f1502177a579790a2cfd1f5/backend/pkg/kafka/deserializer.go#L523-L530

This logic will also treat LF(10) as a control character, causing front-end UI problems.

![2023-11-30 15-28-45屏幕截图](https://github.com/redpanda-data/console/assets/11460848/4ea12575-98bd-4f16-af01-3c22b9e60d43)
![2023-11-30 15-29-02屏幕截图](https://github.com/redpanda-data/console/assets/11460848/a3cf0e2a-c4d8-49db-8eeb-73fb8a178be9)

The actual data:

[messages.json](https://github.com/redpanda-data/console/files/13509804/messages.json)

weeco:
Hey @cobolbaby ,
this is already fixed (among many other minor things) in the `improve-message-viewer` branch. If you want to verify this, you can test this with the following Docker image:

```
redpandadata/console-unstable:improve-message-viewer-b5a3f6b
```

Thanks for the report 👍

cobolbaby:
Why do so much in one branch? When is the bugfix expected to be merged into the main trunk?

weeco:
@cobolbaby Can you elaborate this? I only see a couple messages which were filtered because the value payload size is too large (the browser will crash if you try to open very large messages due to the resource constraints). This seems unrelated. Linebreaks are considered control characters but are handled in the frontend so that you can see there's a linebreak, but it's still being rendered as a linebreak. This has a couple reasons becasue we've got reports from users that saw "empty payloads" which in fact contained invisible whitespace / linebreaks. This way we get the best of both worlds (we still render linebreaks, while showing them that they exist).

cobolbaby:
@weeco I'll briefly describe the issue again. When a Kafka message consists of multi-line text, the webpage displays hexadecimal encoding instead of the actual text information. However, using the "Copy Value" option allows you to obtain the correct text information, as you can see in the attached content above. The reason for this discrepancy is likely due to the message being identified as `messageEncodingUtf8WithControlChars`. Therefore, I hope to correct the logic of `containsControlChars` so that multi-line text is recognized as `messageEncodingText`.

weeco:
Messages with encoding `utf8WithControlChars` should not be rendered as hex but UTF8 plus control chars handled slightly different. I made a note to check this.

[...]

PR #947, opened 2023-12-01
fix: containsControlChars treat LF(10) as a control character

cobolbaby:
Ref: https://github.com/redpanda-data/console/issues/946


weeco:
Fixed inside the message viewer branch, which we're planning to merge at the moment.

cobolbaby:
Yes, it works in redpandadata/console-unstable:master-81e6897.

PR #957, opened 2023-12-09
Migrates radio to Chakra

jvorcak:
Screenshot 2023-12-09 at 23 35 33

We need to merge https://github.com/redpanda-data/ui/pull/441 first + update @redpanda-data/ui

Note: we haven't updated DatePicker related selects for now. They will be migrated as part of DatePicker component migration


frontend/src/utils/tsxUtils.tsx
```diff
@@ -309,25 +295,29 @@ export class RadioOptionGroup<T> extends Component<{
...
-        );
-
-        return radioGroup;
+        return <RadioGroup
+            direction="column"
+            // TODO - we need to make the API more TS safe and make name optional
+            name=""
+            // @ts-ignore
+            value={p.value}
+            // @ts-ignore
```
malinskibeniamin, line 304: What are the errors you were getting here?

frontend/src/utils/tsxUtils.tsx
```diff
@@ -309,25 +295,29 @@ export class RadioOptionGroup<T> extends Component<{
...
+            options={p.options.map(kv => ({
+                value: kv.value,
+                disabled: p.disabled,
+                label: <Box p={3}>
+                    <Text fontWeight={500}>{kv.title}</Text>
+                    <Text color="gray.500">{kv.subTitle}</Text>
+                    {kv.content && (p.showContent == 'always' || p.value == kv.value) && (
+                        <Box key={String(kv.value)} style={{marginLeft: '27px', marginTop: '12px'}}>
+                            <div>{kv.content}</div>
```
malinskibeniamin, line 315: I don't think this div is necessary.

frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -672,10 +672,24 @@ class SaveMessagesDialog extends Component<{ messages: TopicMessage[] | null, on
...
+                                options={[
+                                    {
+                                        value: 'json',
+                                        label: 'JSON'
+                                    },
+                                    {
+                                        value: 'csv',
+                                        label: 'CSV',
+                                        disabled: true
+                                    }
```
malinskibeniamin, line 689: I know it was like this before, but is there a point in having an always disabled CSV radio group option? @weeco maybe you know why

frontend/src/utils/tsxUtils.tsx
jvorcak: `T` can be null, while `RadioGroup` accepts `undefined`. Solvable problem, but I'd leave it for later.

frontend/src/components/pages/topics/Tab.Messages/index.tsx
jvorcak: I know, I just didn't want to introduce any "breaking" changes just based on my decision. I try to do this migration 1:1

rikimaru0345: Leave it in for now.
We might have a solution to the CSV issue (that is, doing it the "simple" way, where only the root properties are different columns, and the payload/content stays as json).
Depends on what people actually expect from saving as CSV.
So overall, its not completely off the table that we might want to add it in the future if people want it.

PR #958, opened 2023-12-10
Migrates slider component to Chakra

jvorcak:
Dependent on https://github.com/redpanda-data/ui/pull/442

Screenshot 2023-12-10 at 16 57 58
Screenshot 2023-12-10 at 16 56 17


malinskibeniamin (approved):
@mo-redpanda @ivpanda if you remember, for a feature with cluster throughput tier selection, we originally planned on using a slider - we built the slider using Figma designs and never used it in the end. Do you want us to use the Slider from Chakra UI or the one we originally had in Figma? Just asking what is the style we want to go for.

mo-redpanda:
We answered this in the UX sync today. Will move forward with removing as this slider was never used

PR #962, opened 2023-12-13
Add delete acl endpoint

weeco:
This PR adds the implementation for DELETE `/v1alpha1/acls`


backend/pkg/api/connect/service/acl/service.go
```diff
@@ -156,10 +156,52 @@ func (s *Service) CreateACL(ctx context.Context, req *connect.Request[v1alpha1.C
...
+	if err != nil {
+		return nil, apierrors.NewConnectError(
+			connect.CodeInternal,
+			err,
+			apierrors.NewErrorInfo(v1alpha1.Reason_REASON_KAFKA_API_ERROR.String(), apierrors.KeyValsFromKafkaError(err)...),
+		)
+	}
+
+	if len(res.Results) != 1 {
+		// Should never happen since we only create one ACL, but if it happens we want to err early.
```
sago2k8, line 180: nit: Maybe here you meant 
```suggestion
		// Should never happen since we only delete one ACL, but if it happens we want to err early.
```

backend/pkg/api/connect/service/acl/service.go
```diff
@@ -156,10 +156,52 @@ func (s *Service) CreateACL(ctx context.Context, req *connect.Request[v1alpha1.C
...
+	res, err := s.consoleSvc.DeleteACLsKafka(ctx, kafkaReq)
+	if err != nil {
+		return nil, apierrors.NewConnectError(
+			connect.CodeInternal,
+			err,
+			apierrors.NewErrorInfo(v1alpha1.Reason_REASON_KAFKA_API_ERROR.String(), apierrors.KeyValsFromKafkaError(err)...),
+		)
+	}
+
+	if len(res.Results) != 1 {
```
sago2k8, line 179: quesiton: if this is always an operation over a single entity, shouldn't it be called `DeleteACL`?

sago2k8 (approved):
lgtm

backend/pkg/api/connect/service/acl/service.go
weeco: This is a good question and I'm not entirely sure. So with one such filter you can delete multiple ACLs actually (any ACLs that match the filter), and in fact you could construct one filter that matches all ACLs by using the rigth enums (ANY operation, ANY resource pattern type, etc). However we could offer a second endpoint that accepts multiple filters in one request, since this is what the Kafka API does under the hood as well.

@twmb @birdayz What are your opinions on this?

weeco: I'll proceed with DeleteACLs because it deletes multiple ACLs, but we could add another DeleteACLsBulk or something in the future which accepts more than one filter. The response handling is more complex (due to partial errors/success) though, so both methods would have their right to exist I believe

PR #968, opened 2023-12-14
frontend: allow selecting (overriding) key/value deserializer for topics

rikimaru0345:
image
image

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages/index.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages/index.tsx
@@ -312,6 +312,18 @@ export class TopicMessageView extends Component<TopicMessageViewProps> {
                         </div>
                     </Label>
 
+                    {/* Show warning if a deserializer is forced for key or value */}
+                    <Flex alignItems="center" height="32px">
+                        {((uiState.topicSettings.keyDeserializer != PayloadEncoding.UNSPECIFIED && uiState.topicSettings.keyDeserializer != null) ||
+                            (uiState.topicSettings.valueDeserializer != PayloadEncoding.UNSPECIFIED && uiState.topicSettings.valueDeserializer != null))
+                            && (
+                                <Text fontWeight="bold" fontSize="small" color="red.500" >
+                                    Key/Value deserializer override is active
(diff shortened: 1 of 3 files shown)


weeco:
Why is the key and value column empty in your given screenshot?

rikimaru0345:
Because I selected "none" for that screenshot in both key and value.

rikimaru0345:
I tested various deserializers, works properly from what I can see.
Only (maybe) strange thing is when, for example, selecting "message pack" for a json topic, it will just not give you any content.
I mean, fine, but maybe some sort of additional message would be good? Some thing like:

"You forced using "MESSAGE PACK" deserializer, but it failed, that's why the content of the messages is empty."

weeco (approved):

ivpanda:
Sharing thoughts here so they don't only live in Slack. 

## Deserializer change 

The thought is that we show it in the same area as filters: 

image

It's not ideal that filters are different to this/the two settings are in a different place, but the mechanism will do what we want: the cog will lead the user to the modal to change it, the 'x' will remove it.

It might be nice in future to combine filters and table settings somehow? 

## "Empty" error 

Is this an error or a warning? Either way, I'd like to only show it once even if many/all messages are empty. 

image

weeco:
@ivpanda Should we always show these then, so that there's always one for key and one for value deserializer but they are set to "auto" in most cases?

ivpanda:
@weeco  I'm not sure I understand. Do you mean that these are almost always changed?

weeco:
A key/value deserializer strategy must always be chosen. By default we always chose the strategy "Auto" and I wondered whether this should be shown there as well then, or only if either key/value deviate from the default

ivpanda:
Ah, right. I think this should only be shown if we deviate from the default.

rikimaru0345:
As requested it looks like a tag now:

https://github.com/redpanda-data/console/assets/28133019/eec35d99-00f8-4673-8181-3b621fbb5cc3

rikimaru0345:
Video above is outdated, this is how it looks like when the deserializers are set/active:

image

PR #973, opened 2023-12-19
Feature/chakra migrated cloud UI

jvorcak:


frontend/src/components/pages/topics/CreateTopicModal/CreateTopicModal.tsx
```diff
@@ -593,13 +597,13 @@ export function RatioInput(p: {
...
 
-    return <div className="ratioInput">
-        <Slider
-            min={0} max={100} step={1}
-            value={Math.round(p.value * 100)}
-            onChange={x => p.onChange(x / 100)}
-            tooltip={{ formatter: null }}
-        />
+    return <Flex alignItems="center" gap={2}>
+        <Slider min={0} max={100} step={1} onChange={x => p.onChange(x / 100)} value={Math.round(p.value * 100)}>
```
malinskibeniamin, line 601: 👏

frontend/src/components/pages/reassign-partitions/Step2.Brokers.tsx
```diff
@@ -58,37 +46,52 @@ export class StepSelectBrokers extends Component<{ selectedBrokerIds: number[],
...
+                                    }
+                                }}
+                            />;
+                        }),
+                        cell: observer(({row: {original: broker}}: { row: Row<Broker> }) => {
+                            const checked = selectedBrokers.includes(broker.brokerId)
+                            return (
+                                <Checkbox
+                                    isChecked={checked}
+                                    onChange={() => selectedBrokers.includes(broker.brokerId)
```
malinskibeniamin, line 82: I think you can just reuse the `checked` flag from line 78

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -226,55 +203,40 @@ class TabTasks extends Component {
...
+                        size: 300
                     },
                     {
-                        title: 'Task ID',
-                        dataIndex: 'taskId',
-                        sorter: sortField('taskId'),
-                        width: 50
+                        header: 'Task ID',
+                        accessorKey: 'taskId',
+                        // Include sorting logic if necessary
```
malinskibeniamin, line 223: Is sorting necessary? We had it before

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -72,148 +70,127 @@ class TabClusters extends Component {
...
+                        cell: ({row: {original}}) => original.clusterInfo.version,
+                    },
+                    {
+                        accessorKey: 'connectors',
+                        size: 150,
+                        header: 'Connectors',
+                        cell: ({row: {original}}) => <ConnectorsColumn observable={original}/>
+                    },
+                    {
+                        accessorKey: 'connectors',
```
malinskibeniamin, line 111: I don't think this is accurate, we should use different `accessorKey` for tasks vs connectors?

malinskibeniamin: I know we had this in the code previously but just double checking this actually makes sense

frontend/src/components/misc/KowlTimePicker.tsx
```diff
@@ -32,48 +31,16 @@ export class KowlTimePicker extends Component<{
...
-                onChange={(e) => {
-                    // console.log("date mode changed", { newValue: e.target.value, isLocalMode: this.isLocalTimeMode });
-                    this.isLocalTimeMode = e.target.value == 'local';
-                }}
-            >
-                <Radio value="local">Local</Radio>
-                <Radio value="utc">UTC</Radio>
-            </Radio.Group>
+            <Box maxW={300}>
+                <DateTimeInput
```
malinskibeniamin, line 36: Does the DateTimeInput contain the UTC radio?

frontend/src/components/misc/ConfigList.tsx
```diff
@@ -50,75 +51,70 @@ export function ConfigList({ configEntries, valueDisplay, renderTooltip }: { con
...
+                            <p>Resources can be configured at different levels. Example: A topic config may be inherited
+                                from the static broker config.</p>
+                            <p>Valid sources are: Dynamic Topic, Dynamic Broker, Default Broker, Static Broker, Dynamic
+                                Broker Logger and Default config.</p>
+                        </>
+                    }
+                    placement="left"
+                    hasArrow
+                >
+                    <InfoCircleFilled style={{color: '#bbbbbb'}}/>
```
malinskibeniamin, line 87: Maybe just use `colorScheme` instead of hex code

frontend/package.json
```diff
@@ -30,9 +30,8 @@
         "@heroicons/react": "^1.0.6",
         "@monaco-editor/react": "^4.6",
         "@primer/octicons-react": "^17",
-        "@redpanda-data/ui": "^3.31",
+        "@redpanda-data/ui": "^3.37.0-alpha.0",
         "@textea/json-viewer": "^1.24.4",
-        "antd": "^4.21",
```
malinskibeniamin, line 35: Awesome!

malinskibeniamin: It might be worth checking what were the dependencies that we had to bring in to the codebase because of `antd`, and remove those as well?

frontend/src/components/misc/KowlTimePicker.tsx
jvorcak: @malinskibeniamin it does

jvorcak:
@malinskibeniamin thanks for the review. I also have a question - how can I make the preview deployment link please?

malinskibeniamin:
I don't think this is currently set up for console, @rikimaru0345 or @sago2k8 would you know

sago2k8:
Yeap for preview in console you need a couple of things, I can help you deploy a cluster, ping me when available

frontend/package.json
jvorcak: I can check that, do you have some specific things in mind or it's more of a general comment?

frontend/src/components/pages/connect/Overview.tsx
jvorcak: Done

frontend/src/components/misc/ConfigList.tsx
```diff
@@ -50,75 +51,70 @@ export function ConfigList({ configEntries, valueDisplay, renderTooltip }: { con
...
-                    return false;
+    return (
+        <DataTable<ConfigEntry>
+            data={configEntries}
+            showPagination={false}
+            enableSorting={false}
+            defaultPageSize={10000}
+            size="md"
+            getRowCanExpand={row => (row.original.synonyms?.length ?? 0) > 0 }
+            // getSubRows={row => row.synonyms as ConfigEntry[]}
```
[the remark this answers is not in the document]
jvorcak: Done

frontend/package.json
malinskibeniamin: I just thought we could analyse the bundle or at least check what are the main packages that antd depends on, and if they still need to be brought into our build

frontend/src/utils/utils.ts
```diff
@@ -488,6 +488,31 @@ export const prettyBytesOrNA = function (n: number) {
...
+ * @returns {boolean} - Returns `true` if the sets are equal, otherwise returns `false`.
+ * @example
+ * // returns true
+ * eqSet(new Set([1, 2, 3]), new Set([3, 2, 1]));
+ *
+ * @example
+ * // returns false
+ * eqSet(new Set([1, 2, 3]), new Set([4, 5, 6]));
+ */
+export const eqSet = <T = string,>(xs: Set<T>, ys: Set<T>): boolean =>
```
rikimaru0345, line 512: This is a classic mistake, this fails for 

[1, 1, 2]  and [1, 2, 3]

Since size is just array size and not "number of unique elements", they have the same size and checking only the first one one by one, and checking if the second one contains all the first elements, they would be considered equal.

Proper set equality is a bit more difficult and (cpu)time consuming

rikimaru0345 (changes requested):

frontend/src/utils/utils.ts
jvorcak: @rikimaru0345 but this function is for `Set` not for arrays. Typescript is guarding that.

jvorcak: The problem would be if somebody silence Typescript and provides an array instead. We could be defensive and use `new Set([...xs])` in the implementation to handle both Set and Arrays and then compare sizes.

But that only happens if somebody silences TS and uses `eqSet` for arrays instead of Set.

rikimaru0345: I see, should be good then! :)

rikimaru0345 (approved):

PR #977, opened 2023-12-20
backend: add interceptor that logs connect erors

weeco:
Whenever a connect handler returns an error it is
currently not logged. However, usually this should be of interest. This PR adds an interceptor that
always logs returns errors by any connect handler.


sago2k8 (approved):
lgtm

bojand (approved):

bojand (approved):

PR #978, opened 2023-12-21
backend: add always enabled procedures

weeco:
The endpoint checker currently checks whether
the request on the given procedure matches
the configuration. If there's a misconfiguration
this can severely impact the Console UX, which
should never be desired. Thus, we are adding
an alwaysEnabledProcedures option here. This
ensures that endpoints that are used by the
frontend will always pass the endpoint check
interceptor.


sago2k8 (approved):
lgtm

PR #988, opened 2024-01-03
backend: pin proto tooling versions, make github action run verify ta…

bojand:
…sk, and update generate


.github/workflows/backend-lint-test.yml
```diff
@@ -32,7 +32,12 @@ jobs:
           version: v1.55.2
           args: --timeout=5m
           working-directory: backend
+      
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          version: 3.x
```
weeco, line 39:
```suggestion
          version: 3.x
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

From their docs:

> (Optional) GitHub access token used for GitHub API requests. Heavy usage of the action can result in workflow run failures caused by rate limiting. GitHub provides a more generous allowance for Authenticated API requests.
> 
> It will be convenient to use [${{ secrets.GITHUB_TOKEN }}](https://docs.github.com/en/actions/reference/authentication-in-a-workflow).

Probably not needed, but given it's a simple one liner I don't see a disadvantage of doing this

.github/workflows/backend-lint-test.yml
```diff
@@ -32,7 +32,12 @@ jobs:
           version: v1.55.2
```
weeco, line 32: Do we still need to install the go binary and linter? These are installed locally as part of the task command right?

weeco (approved):

PR #997, opened 2024-01-11
Add get topic config endpoint

weeco:
This PR adds the endpoint GET `/v1alpha1/topics/{topicName}/configurations`. Topic attributes such as partition count and replication factor are not part of the response, since these are not configurations. Because these are still of interest, I'd recommend adding a new endpoint GET `/v1alpha1/topics/{topicName}` that yields this information.


backend/pkg/api/connect/service/common/mapper.go
```diff
@@ -0,0 +1,84 @@
...
+
+// KafkaClientMapper is a utility to map common types between the Kafka Client and the
+// proto types that are used for requests and responses.
+type KafkaClientMapper struct{}
+
+// ConfigTypeToProto maps the kmsg.ConfigType enum to the proto enum.
+func (*KafkaClientMapper) ConfigTypeToProto(configType kmsg.ConfigType) (v1alpha1.ConfigType, error) {
+	switch configType {
+	case kmsg.ConfigTypeBoolean:
+		return v1alpha1.ConfigType_CONFIG_TYPE_DOUBLE, nil
```
bojand, line 30: just checking... Is this correct, the case is boolean, but we return double?

weeco: No not correct, good find. Thanks!

PR #999, opened 2024-01-11
backend: add secret ID validation options to secrets API proto

bojand:
This adds validation to the secret ID in secrets API proto validation.

For AWS secret manager, the name has the following rules:

```
// The name of the new secret.
//
// The secret name can contain ASCII letters, numbers, and the following characters:
// /_+=.@-
```

while for GCP the secret ID has the following rules:

```
// A secret ID is a string with a maximum length of 255 characters and can
// contain uppercase and lowercase letters, numerals, and the hyphen (`-`) and
// underscore (`_`) characters.
```

I have opted to be more inclusive rather than restrictive to common subset, and include the AWS allowed characters since it's reasonable for AWS clients would actually want to use some of those (ex `/`) with relative frequency.


proto/redpanda/api/dataplane/v1alpha1/secret.proto
```diff
@@ -24,7 +24,12 @@ message Secret {
 }
 
 message SecretInput {
-  string id = 1 [(google.api.field_behavior) = IMMUTABLE];
+  string id = 1 [
+    (google.api.field_behavior) = IMMUTABLE,
```
weeco, line 28: Could possibly remove this since this type is exclusively used for creating secrets and not reused for retrieving or updating right?

weeco (approved):

PR #1009, opened 2024-01-17
serverless: do not force hiding internal topics

nicolaferraro:
Serverless does not create internal topics under the hood and this settings prevented users to see topics beginning with underscore (_), even if they could create them.


sago2k8 (approved):
LGTM

PR #1011, opened 2024-01-18
Add WASM Transform Proto and Endpoints for Programmatic API

gene-redpanda:
Adds the following:
* ability to deploy WASM binaries to RP using the standard api endpoints with multipart form
* grpc defined list/get/delete endpoints for managing transforms on the RP cluster
* tests for the above
* enabled admin api in integration test
* [added support for enabling transforms ](https://github.com/testcontainers/testcontainers-go/pull/2170) to testcontainers-go
* openapi docs integration


backend/pkg/api/connect/service/transform/mapper.go
```diff
@@ -0,0 +1,36 @@
...
+// by the Apache License, Version 2.0
+
+package transform
+
+import (
+	"fmt"
+	v1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/dataplane/v1alpha1"
+)
+
+func intToInt32(i int) int32 {
```
bojand, line 17: maybe I am not getting it, but do we really need this function?

backend/pkg/api/connect/service/transform/service.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (s *Service) DeployTransform(ctx context.Context, c *connect.Request[v1alpha1.DeployTransformRequest]) (*connect.Response[v1alpha1.DeployTransformResponse], error) {
+	if !s.cfg.Redpanda.AdminAPI.Enabled {
+		return nil, apierrors.NewConnectError(
+			connect.CodeUnimplemented,
+			errors.New("the redpanda admin api must be configured to deploy a transform"),
+			apierrors.NewErrorInfo(v1alpha1.Reason_REASON_FEATURE_NOT_CONFIGURED.String()),
+			apierrors.NewHelp(apierrors.NewHelpLinkConsoleReferenceConfig()),
+		)
+	}
```
bojand, line 39: super minor nit, but maybe make this into it's own function so we can just do something like:
```go
if err := checkAdminAPIEnabled(errors.New("specific message here")); err != nil {
  return nil, err
}

File: backend/pkg/api/connect/service/transform/mapper.go (hunk 1 above)
Comment (gene-redpanda) [reply backend/pkg/api/connect/service/transform/mapper.go]: nope. I thought I needed a pointer initially, realized I didn't and then just didn't bother cleaning it up at the time. gone now!

File: backend/pkg/api/connect/service/transform/service.go (hunk 2 above)
Comment (gene-redpanda) [reply backend/pkg/api/connect/service/transform/service.go:39]: i mirrored the pattern in user. wouldn't mind having a function that does that check in a util package or something somewhere (sorta like the auth hooks) and then sharing it between the two/other locs. because i agree it definitely feels repetitious.

File: backend/pkg/api/routes.go [hunk 3]
```diff
@@ -235,6 +235,12 @@ func (api *API) routes() *chi.Mux {
 				r.Post("/users", api.handleCreateUser())
 				r.Delete("/users/{principalID}", api.handleDeleteUser())
 
+				// Wasm Transforms
+				r.Get("/transforms", api.handleListTransforms())
+				r.Get("/transforms/{transformID}", api.handleGetTransform())
+				r.Put("/transforms", api.handleDeployTransform())
+				r.Delete("/transforms/{transformID}", api.handleDeleteTransform())
+
```
bojand, line 243: I could be missing some context, but do we actually need this? This unfortunately might be a case of looking at the existing codebase and doing extra work that may not be necessary.
The connect gateway package already exposes HTTP REST API based on `google.api.http` annotations in the proto files. So unless we specifically want these paths and implement these handlers, I think this is unnecessary? @weeco to chime in.

gene-redpanda: I'm fine either way. I was honestly confused when I saw the connect stuff in the standard API because as you say, it should already exist. But I didn't want to not do it and need it 😅

proto/redpanda/api/dataplane/v1alpha1/transform.proto
```diff
@@ -0,0 +1,206 @@
...
+import "buf/validate/validate.proto";
+import "google/api/annotations.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+
+message TransformMetadata {
+  string name = 1;
+  string input_topic = 2;
+  repeated string output_topics = 3;
+  repeated PartitionTransformStatus status = 4;
+}
```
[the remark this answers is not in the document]
gene-redpanda: are you sure it isn't intput_topic_name rather than input_topic_name**s**?

proto/redpanda/api/dataplane/v1alpha1/transform.proto
```diff
@@ -0,0 +1,206 @@
...
+
+  // WASM binary for the transform. The binary must be a valid WASM binary
+  bytes wasm_binary = 2;
+
+  // 'input_topic' specifies the name of the input topic for the transform.
+  string input_topic = 3;
+
+  // 'output_topics' is a list of output topics for the transform. This field can contain
+  // multiple strings, each representing the name of an output topic.
+  repeated string output_topics = 4;
```
bojand, line 87: similar comment regarding topic name fields.

backend/pkg/api/connect/service/transform/mapper.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+	v1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/dataplane/v1alpha1"
+)
+
+func statusToPartitionTransformStatus_PartitionStatus(s string) (v1alpha1.PartitionTransformStatus_PartitionStatus, error) {
+	switch s {
+	case "unspecified":
+		return v1alpha1.PartitionTransformStatus_PARTITION_STATUS_UNSPECIFIED, nil
```
rockwotj, line 20: We (core) never return `"unspecified"` here FWIW

backend/pkg/api/handle_transforms.go
```diff
@@ -0,0 +1,290 @@
...
+			return
+		}
+
+		if err := api.RedpandaSvc.DeployWasmTransform(r.Context(), adminapi.TransformMetadata{
+			Name:         msg.Name,
+			InputTopic:   msg.InputTopic,
+			OutputTopics: msg.OutputTopics,
+			Status:       nil,
+			Environment:  msg.Environment,
+		}, msg.WasmBinary); err != nil {
```
rockwotj, line 203: FWIW one thing we sanity check WRT WasmBinary in Redpanda (and I think RPK) is making sure the magic bytes for a wasm binary are valid to double check someone didn't try uploading a native binary.

Magic bytes are `\0asm` (null byte followed by ascii `a` `s` `m`).

rockwotj (commented):
Looks pretty good to me. Couple of small bits

proto/redpanda/api/dataplane/v1alpha1/transform.proto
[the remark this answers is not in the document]
bojand: Yes sorry `input_topic_name` and `output_topic_names` 🤦

backend/pkg/api/connect/service/transform/service.go
gene-redpanda: or dump it in some middleware somewhere 🤷‍♀️

backend/pkg/api/handle_transforms.go
gene-redpanda: looks like there isn't a check in RPK: https://github.com/redpanda-data/redpanda/blob/821bd6d1fd31ef4efc9c699eade449d40d089568/src/go/rpk/pkg/adminapi/api_transform.go#L24

IMO it'd make more sense to check it in RPK since we're doing the actual send with the admin client from rpk anyway.

gene-redpanda: https://github.com/redpanda-data/redpanda/issues/16230 created an issue

gene-redpanda: and a PR here https://github.com/redpanda-data/redpanda/pull/16231

rockwotj: This happens in the deploy command not the validation at the moment.


As for not doing it here and only in the rpk, isn't part of the point of these APIs is so people don't have to use RPK?

backend/pkg/api/connect/service/transform/service.go
bojand: I wouldn't put it in middleware. I think it's fine to leave it as is too. It was just a thought. Not a big deal. Fine the way it is too.

backend/pkg/api/routes.go
birdayz: Yeah this is obsolete, it will happen automatically through grpc-gateway.

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		transformClient := v1alpha1connect.NewTransformServiceClient(http.DefaultClient, s.httpAddress())
+		_, err = transformClient.DeployTransform(ctx, connect.NewRequest(&v1alpha1.DeployTransformRequest{
+			Name:             "foo",
+			WasmBinary:       buf,
+			InputTopicName:   "wasm-tfm-create-test-i",
+			OutputTopicNames: []string{"wasm-tfm-create-test-o"},
+			Environment:      nil,
+		}))
+		require.NoError(err)
```
bojand, line 67: minor nit, but maybe this should be an assert, since _this is the test_?

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,95 @@
...
+			require.NoError(err)
+			require.NoError(despawnTopic(ctx, topicClient, "wasm-tfm-create-test-i"))
+			require.NoError(despawnTopic(ctx, topicClient, "wasm-tfm-create-test-o"))
+		}()
+	})
+}
+
+func (s *APISuite) TestGetTransform() {
+	t := s.T()
+	// require := rqr.New(t)
```
bojand, line 82:
```suggestion
```

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,95 @@
...
+	assert := assert.New(t)
+
+	t.Run("create transform with valid request (connect-go)", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+		defer cancel()
+
+		transformClient := v1alpha1connect.NewTransformServiceClient(http.DefaultClient, s.httpAddress())
+		_, err := transformClient.GetTransform(ctx, connect.NewRequest(&v1alpha1.GetTransformRequest{
+			Name: "test-transform",
+		}))
```
bojand, line 92: should we verify the metadata in the response to match the expectations?

backend/pkg/api/routes.go
bojand: I think we can probably delete these routes and `backend/pkg/api/handle_transforms.go`.

proto/redpanda/api/dataplane/v1alpha1/transform.proto
```diff
@@ -0,0 +1,206 @@
...
+
+import "buf/validate/validate.proto";
+import "google/api/annotations.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+
+message TransformMetadata {
+  string name = 1;
+  string input_topic_name = 2;
+  repeated string output_topic_names = 3;
+  repeated PartitionTransformStatus status = 4;
```
bojand, line 13: We should maybe put some validation annotations in this message? At least the topic names can be validated like:
```proto
string input_topic_name = 2 [
      (buf.validate.field).string.min_len = 1,
      (buf.validate.field).string.max_len = 249,
      (buf.validate.field).string.pattern = "^[a-zA-Z0-9._\\-]*$"
    ];
```

backend/pkg/api/connect/integration/transform_test.go
gene-redpanda: I think some wires got crossed as that test file was very much not ready for review yet (as you noticed)! I had pushed it up purely to ask some questions about handling the binary from a couple people. 

The complete testfile is now up and available for review.

proto/redpanda/api/dataplane/v1alpha1/transform.proto
gene-redpanda: I don't think it makes sense to validate TransformMetadata as it is purely an output element. 

I do have validation on fields on incoming elements though

gene-redpanda: Well, except for these two on DeployRequest that were pending me copying the validation rules from Topic lol 


// 'input_topic_name' specifies the name of the input topic for the transform.
  string input_topic_name = 3 [
    (buf.validate.field).string.min_len = 1,
    (buf.validate.field).string.max_len = 255,
    (buf.validate.field).string.pattern = "^[a-z0-9.]+$"
  ];

  // 'output_topic_names' is a list of output topics for the transform. This field can contain
  // multiple strings, each representing the name of an output topic.
  repeated string output_topic_names = 4 [
    (buf.validate.field).string.min_len = 1,
    (buf.validate.field).string.max_len = 255,
    (buf.validate.field).string.pattern = "^[a-z0-9.]+$"
  ];

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,259 @@
...
+		require.NoError(spawnTopic(ctx, 3, topicClient, "wasm-tfm-create-test-o"))
+
+		transformClient := v1alpha1connect.NewTransformServiceClient(http.DefaultClient, s.httpAddress())
+		stErr := spawnTransform(s.api.Cfg.REST, "test-transform",
+			"wasm-tfm-create-test-i",
+			[]string{"wasm-tfm-create-test-o"},
+			[]adminapi.EnvironmentVariable{{Key: "foo", Value: "bar"}},
+			technicallyATransform)
+
+		assert.NoError(stErr, fmt.Sprintf("error: %v", stErr))
```
bojand, line 123: in `spawnTransform` should we read and parse the response into JSON struct, and then in test we can verify the metadata returned to assert it is what we expect based on input?

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,259 @@
...
+	return nil
+}
+
+func (s *APISuite) TestDeployTransform() {
+	t := s.T()
+
+	require := rqr.New(t)
+	assert := asrt.New(t)
+
+	t.Run("create transform with valid request (connect-go)", func(t *testing.T) {
```
bojand, line 108: Maybe get rid of connect-go part since it's an HTTP multi part upload

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,259 @@
...
+		ctx, cancel := context.WithTimeout(context.Background(), transformTimeout)
+		defer cancel()
+
+		topicClient := v1alpha1connect.NewTopicServiceClient(http.DefaultClient, s.httpAddress())
+		require.NoError(spawnTopic(ctx, 2, topicClient, "wasm-tfm-create-test-i"))
+		require.NoError(spawnTopic(ctx, 3, topicClient, "wasm-tfm-create-test-o"))
```
bojand, line 114: can we maybe change the tests to do deferred despawnTopic of created topics after we have successfully created the topics. so it's clearer that the deferred topic cleanup is immediately after creation in the code.

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,259 @@
...
+		require.NoError(spawnTransform(s.api.Cfg.REST, "test-transform-get", "wasm-tfm-create-test-c", []string{"wasm-tfm-create-test-d"}, []adminapi.EnvironmentVariable{{Key: "foo", Value: "bar"}}, technicallyATransform))
+
+		msg, err := transformClient.GetTransform(ctx, connect.NewRequest(&v1alpha1.GetTransformRequest{
+			Name: "test-transform-get",
+		}))
+		assert.NoError(err)
+
+		assert.Equal(msg.Msg.Transform.Name, "test-transform-get")
+		assert.Equal(msg.Msg.Transform.InputTopicName, "wasm-tfm-create-test-c")
+		assert.Equal(msg.Msg.Transform.OutputTopicNames, []string{"wasm-tfm-create-test-d"})
```
bojand, line 159: expected values go in the first param in asserts, so I think these should be switched around.

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,259 @@
...
+				Name: "test-transform1",
+			},
+		}))
+		assert.NoError(err)
+
+		for _, transform := range transforms.Msg.Transforms {
+			assert.Condition(func() bool {
+				return transform.Name == "test-transform1" || transform.Name == "test-transform2"
+			}, "transform name should be test-transform1 or test-transform2")
+			assert.Equal(transform.InputTopicName, "wasm-tfm-list-test-a")
```
bojand, line 200: here too, switch expected and actual value params

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,259 @@
...
+	assert := asrt.New(t)
+
+	t.Run("list transforms with valid request (connect-go)", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), transformTimeout)
+		defer cancel()
+
+		transformClient := v1alpha1connect.NewTransformServiceClient(http.DefaultClient, s.httpAddress())
+		topicClient := v1alpha1connect.NewTopicServiceClient(http.DefaultClient, s.httpAddress())
+		require.NoError(spawnTopic(ctx, 2, topicClient, "wasm-tfm-list-test-a"))
+		require.NoError(spawnTopic(ctx, 2, topicClient, "wasm-tfm-list-test-b"))
```
bojand, line 184: here too lets maybe defer topic cleanup at creation site.

backend/pkg/api/connect/service/transform/util.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		pts = append(pts, &v1alpha1.PartitionTransformStatus{
+			NodeId:    int32(ts.NodeID),
+			Status:    st,
+			Lag:       int32(ts.Lag),
+			Partition: int32(ts.Partition),
+		})
+	}
+	return pts, nil
+}
```
bojand, line 79: 🤔 To me it feels like these 2 should go in `mapper.go` perhaps since they're converting / mapping messages, in this case admin api <-> proto. so following conventions in other serices, something in form of `fromAdminAPITransformToProto()` and similarly for status.

backend/pkg/api/connect/service/transform/util.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/redpanda-data/redpanda/src/go/rpk/pkg/adminapi"
+
+	apierrors "github.com/redpanda-data/console/backend/pkg/api/connect/errors"
+	commonv1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/common/v1alpha1"
+	v1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/dataplane/v1alpha1"
+)
+
+func findTransformByName(transforms []*v1alpha1.TransformMetadata, name string) (*v1alpha1.TransformMetadata, error) {
+	for _, transform := range transforms {
+		if strings.Contains(transform.Name, name) {
```
rockwotj, line 26: Do you **really** want `Contains` here?

That means that `rpk transform delete foo` could delete a transform called `foobar`. Same with `get` requests. Maybe we should do exact equality?

backend/pkg/api/connect/service/transform/service.go
```diff
@@ -0,0 +1,164 @@
...
+			apierrors.NewErrorInfo(v1alpha1.Reason_REASON_REDPANDA_ADMIN_API_ERROR.String()),
+		)
+	}
+
+	tfs, err := transformsConversion(transforms)
+	if err != nil {
+		return nil, err
+	}
+
+	transform, err := findTransformByName(tfs, c.Msg.Name)
```
rockwotj, line 138: You're going to want to use a different predicate for list vs get. get and delete should use the same predicate.

rockwotj (commented):
Did we decide to loosen validation on the wasm binaries? That seems fine to me. I recently improved the core side validation anyways: https://github.com/redpanda-data/redpanda/pull/16498

backend/pkg/api/connect/service/transform/util.go
gene-redpanda: I followed topic on this one https://github.com/redpanda-data/console/blob/e724c8d3461e80aff70808fa710283a3f9171fbf/backend/pkg/api/connect/service/topic/service.go#L59

backend/pkg/api/connect/service/transform/service.go
gene-redpanda: ok, added new func findExact which uses equality rather than contains

gene-redpanda:
> Did we decide to loosen validation on the wasm binaries? That seems fine to me. I recently improved the core side validation anyways: [redpanda-data/redpanda#16498](https://github.com/redpanda-data/redpanda/pull/16498)

I decided it didn't make sense to be super fixated on evaluating the binary when

a) its a quick shoot over to core
b) we aren't actually compiling it so can't really say if it is/isn't valid to begin with


I did put a 50mb limit, but I am open to making it bigger as it was fairly arbitrary.

backend/pkg/api/connect/service/transform/util.go
rockwotj: But it does not use that for get and delete. I think it's fine for list

bojand (approved):

rockwotj (approved):

backend/pkg/api/handle_transforms.go
```diff
@@ -0,0 +1,170 @@
...
+			rest.SendRESTError(w, r, api.Logger, &rest.Error{
+				Err:      err,
+				Status:   http.StatusInternalServerError,
+				Message:  "Could not list wasm transforms from Kafka cluster",
+				IsSilent: false,
+			})
+			return
+		}
+
+		t, err := findTransformByName(transforms, msg.Name)
```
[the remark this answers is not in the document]
gene-redpanda: huh, i guess it should. dunno why i felt differently earlier 😅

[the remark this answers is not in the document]
gene-redpanda: oh lol i see now. the issue here is that it is under api rather than connect. so it isn't using the util functions i created for connect. the find function here is actually using == over contains. 

i'll rename it for sanity purposes

rockwotj (approved):

backend/pkg/tools/openapi-generator/main.go
```diff
@@ -481,6 +482,82 @@ func updateUsers(doc3 *openapi3.T) {
 	}
 }
 
+func updateTransforms(doc3 *openapi3.T) {
```
weeco, line 485: I assume we need to add an entirely new node to document the deploy trnasforms endpoint?

backend/pkg/api/api_integration_test.go
```diff
@@ -582,6 +582,30 @@ func (a *assertHooks) IsProtectedKafkaUser(_ string) bool {
...
+	return rv.BoolValue, rv.Err
+}
+
+func (a *assertHooks) CanDeleteTransform(_ context.Context) (bool, *rest.Error) {
+	if !a.isCallAllowed("any") {
+		assertHookCall(a.t)
+	}
+	rv := a.getCallReturnValue("any")
+	return rv.BoolValue, rv.Err
+}
```
weeco, line 607: Remove

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,293 @@
...
+			Configs: []*v1alpha1.CreateTopicRequest_Topic_Config{
+				{
+					Name:  "cleanup.policy",
+					Value: kadm.StringPtr("compact"),
+				},
+			},
+		},
+	}))
+	return err
+}
```
weeco, line 56: Instead of using the connect client, could we use the kadm client like we did in other integration tests as well? If something breaks in the CreateTopic endpoint it would be cool that only the tests for this endpoint fail and not the entire integration testsuite because everyone uses this endpoint as well if that makes sense

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	var metaResp response
+	if err := json.NewDecoder(resp.Body).Decode(&metaResp); err != nil {
+		return nil, err
+	}
+
+	return metaResp.Transform, nil
+}
+
+func (s *APISuite) TestDeployTransform() {
```
weeco, line 118: Please use the lower dependencies (i.e. kafka client, redpanda admin client, etc) to create the pre-requisites to avoid cascading test failures

backend/pkg/api/handle_transforms.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		if !canList {
+			rest.SendRESTError(w, r, api.Logger, &rest.Error{
+				Err:      fmt.Errorf("requester has no permissions to list wasm transforms"),
+				Status:   http.StatusForbidden,
+				Message:  "You don't have permissions to list wasm transforms",
+				IsSilent: false,
+			})
+			return
+		}
```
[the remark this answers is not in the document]
gene-redpanda: removed hooks. open to suggestions re: moving file

backend/pkg/tools/openapi-generator/main.go
gene-redpanda: wasn't sure how to handle that. i had initially used the same semantics as the rest but because this endpoint is not grpc it can't be done that way. do you want me to just put a custom explanation there and call it good? or something else in mind?

backend/pkg/api/handle_transforms.go
```diff
@@ -0,0 +1,170 @@
...
+
+		if !api.Cfg.Redpanda.AdminAPI.Enabled {
+			rest.SendRESTError(w, r, api.Logger, &rest.Error{
+				Err:      fmt.Errorf("you must enable the admin api to manage wasm transforms"),
+				Status:   http.StatusServiceUnavailable,
+				Message:  "you must enable the admin api to manage wasm transforms",
+				IsSilent: false,
+			})
```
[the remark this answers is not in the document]
gene-redpanda: unsure what you're looking for with the desired end state here

weeco:
@gene-redpanda I moved the file that contained the HTTP handler into the connect/transform package, but haven't fixed most stuff that is broken due to that. One pending issue are the calls to return errors using `rest.SendRESTError()`. The error scheme would be very different to what we return in error cases on the programmatic API where a connect.Error is translated into JSON by the grpc gateway.

What we can do instead is to use the original protobuf errors and use the `connect.ErrorWriter` to translate them into JSON like so:

```go
		connectErr := apierrors.NewConnectError(
			connect.CodeUnauthenticated,
			fmt.Errorf(publicMessage),
			apierrors.NewErrorInfo(reason))
		if errorWriter.IsSupported(r) {
			_ = errorWriter.Write(w, r, connectErr)
		} else {
			w.WriteHeader(http.StatusUnauthorized)
			_, _ = w.Write([]byte(connectErr.Error()))
		}
```

backend/pkg/tools/openapi-generator/main.go
weeco: We need to add the node for the WASM endpoint there so that the result in the OpenAPI spec is similar to what is generated for the other endpoints. Would need to do a bit research / check the library a bit to figure out what to do exactly as well.

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,677 @@
...
+		body := &bytes.Buffer{}
+		writer := multipart.NewWriter(body)
+
+		inputTopicName := "wasm-tfm-create-test-i"
+		deployRequestMetadata := deployTransformRequest{
+			Name:             "deploy-transform-incomplete-metadata",
+			InputTopicName:   inputTopicName,
+			OutputTopicNames: []string{"some-random-topic"},
+			EnvironmentVariables: []KeyVal{{
+				Key:   "READPANDA_KEYS_ARE_PROTECTED", // This should be flagged
```
bojand, line 265: since this is a test for "no wasm binary" in requests, this env var, and comment, is a little confusing. The env var is "OK" because it's R**E**ADPANDA_KEYS_ARE_PROTECTED instead of `RADPANDA_KEYS_ARE_PROTECTED`, (no **E**), but the comment is still "// This should be flagged". I think generally this part can be cleaned up maybe, just so the env key is "FOO", and no comment?

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,677 @@
...
+		err := requests.
+			URL(s.httpAddress() + "/v1alpha1/transforms/" + tfName).
+			ToJSON(&httpRes).
+			AddValidator(requests.ValidatorHandler(
+				requests.CheckStatus(http.StatusOK), // Allows 2xx otherwise
+				requests.ToString(&errResponse),
+			)).
+			Fetch(ctx)
+		assert.Empty(errResponse)
+		require.NoError(err)
```
bojand, line 416: should we check the response, similar to the above one?

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,677 @@
...
+		ctx, cancel := context.WithTimeout(ctx, 6*time.Second)
+		t.Cleanup(cancel)
+
+		msg, err := transformClient.GetTransform(ctx, connect.NewRequest(&v1alpha1.GetTransformRequest{
+			Name: "does-not-exist",
+		}))
+		assert.Nil(msg)
+		assert.Error(err)
+		assert.Equal(connect.CodeNotFound.String(), connect.CodeOf(err).String())
+	})
```
bojand, line 429: minor thing, maybe add HTTP version of this test?

backend/pkg/api/connect/integration/transform_test.go
```diff
@@ -0,0 +1,677 @@
...
+		for _, transform := range httpRes.Transforms {
+			assert.Condition(func() bool {
+				return transform.Name == tfNameOne || transform.Name == tfNameTwo
+			}, fmt.Sprintf("transform name should be %s or %s", tfNameOne, tfNameTwo))
+			assert.Equal(inputTopicName, transform.InputTopicName)
+			require.Len(transform.OutputTopicNames, 1)
+			assert.Equal(outputTopicName, transform.OutputTopicNames[0])
+		}
+		_, err = transformClient.ListTransforms(ctx, connect.NewRequest(&v1alpha1.ListTransformsRequest{}))
+		assert.NoError(err)
```
bojand, line 544: why do we have this `ListTransforms()` call again?

backend/pkg/api/connect/integration/transform_test.go
weeco: It would be `/v1alpha1/transforms` instead of `/v1alpha1/transforms/{name}`which would match the list endpoint and not the resource get endpoint.

backend/pkg/api/connect/integration/transform_test.go
weeco: Haha two issues in one place, indeed it should not be there at all.

backend/pkg/api/connect/integration/transform_test.go
bojand: Maybe I am missing something or causing confusion. We have `"get non-existent transform (connect-go)"` test, but we do not have `"get non-existent transform (http)"` test... Like we have `"get transform with valid request (http)" and another one with special characters... So my suggestion was just to have an HTTP test for unknown transform, ex `/v1alpha1/transforms/some_unknown_wasm_tranform_1234`, and ensure a `404` similar to the connect-go test. Minor thing.

PR #1025, opened 2024-01-23
Regexp search for tables in console

jvorcak:
Screenshot 2024-01-23 at 16 32 23
Screenshot 2024-01-23 at 16 31 56


rikimaru0345 (commented):
Some of these use toLowerCase on the regex pattern. Is that intended?

rikimaru0345:
solves https://github.com/redpanda-data/console/issues/1026

rikimaru0345 (approved):

PR #1030, opened 2024-01-24
backend: increase SR client timeout and limit concurrency when gettin…

bojand:
…g all schemas and subjects

### DIFF
--- a/backend/pkg/schema/client.go
+++ b/backend/pkg/schema/client.go
@@ -18,9 +18,11 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/go-resty/resty/v2"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/redpanda-data/console/backend/pkg/config"
 )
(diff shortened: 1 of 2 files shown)


backend/pkg/schema/client.go
```diff
@@ -697,9 +697,18 @@ func (c *Client) GetSchemasIndividually(ctx context.Context, showSoftDeleted boo
...
+	guard := make(chan struct{}, 10)
+
 	// Describe all subjects concurrently one by one
 	for _, subject := range subjectsRes.Subjects {
+		guard <- struct{}{}
+
 		go func(s string) {
+			defer func() {
+				<-guard
+			}()
```
weeco, line 710: Can we refactor to use errgroup with SetLimit here?

backend/pkg/schema/client.go
```diff
@@ -52,7 +52,7 @@ func newClient(cfg config.Schema) (*Client, error) {
 		SetHeader("Accept", "application/vnd.schemaregistry.v1+json").
 		SetHeader("Content-Type", "application/vnd.schemaregistry.v1+json").
 		SetError(&RestError{}).
-		SetTimeout(5 * time.Second)
+		SetTimeout(30 * time.Second)
```
weeco, line 55: Why 30s? Seems like a very long time for a single request to be made. I think not even the frontend would wait for such a long time for the response.

backend/pkg/schema/client.go
bojand: Updated.

backend/pkg/schema/client.go
bojand: We were seeing request timeouts to SR when trying to fetch data from `.../versions` endpoint.
```
... context deadline exceeded (Client.Timeout exceeded while awaiting headers)
``` 
Although could be due to the amount of requests we were sending to SR as well.

weeco (approved):

PR #1033, opened 2024-01-25
pprof: bind it to the loopback interface by default

nicolaferraro:


backend/pkg/api/routes.go
```diff
@@ -207,7 +207,14 @@ func (api *API) routes() *chi.Mux {
...
 			r.Handle("/admin/health", api.handleLivenessProbe())
 			r.Handle("/admin/startup", api.handleStartupProbe())
+		})
 
+		// Debug routes
+		router.Group(func(r chi.Router) {
+			api.Hooks.Route.ConfigInternalRouter(r)
+			if api.Cfg.REST.Debug.ForceLoopback {
+				r.Use(forceLoopbackMiddleware(api.Logger))
+			}
```
weeco, line 217: Wondering if we should check whether debug endpoints are enabled at all first. The enabled check could even be outside line 213 where we create the router group no?

weeco:
Can we add the new config props here for docs purposes: https://github.com/redpanda-data/console/blob/master/docs/config/console.yaml with a comment describing what they do. I'll then create a ticket for our docs team so that lands on the redpanda documentation too.

nicolaferraro:
> Can we add the new config props here for docs purposes: https://github.com/redpanda-data/console/blob/master/docs/config/console.yaml with a comment describing what they do. I'll then create a ticket for our docs team so that lands on the redpanda documentation too.

Thanks, I should have addressed the comments

weeco (approved):

PR #1037, opened 2024-01-26
backend: do not base64 encode json payload in msgpack serde

### LINKED ISSUE #1034: msgpack filtering and display broken
After updating to version v2.4.0 / 2024-01-23 I noticed that display and filtering of msgpack encoded messages and headers is broken.

Using `git bisect` I found commit 89f4152b1442f4f62f5e7f942e6ff7606b9a99a4 that introduced this issue.

What it looked before commit 89f4152b1442f4f62f5e7f942e6ff7606b9a99a4:

image
image

And before I could use filter like `headers["X-Correlation-ID"] == "771b4325b16b4128b6cdba71ed590c13"` and it worked.

How it look now:

image
image

And filtering by header values doesn't work.

bojand:
Fixes #1034.

image


skarrok:
Thank you for a quick fix!

I checked out this branch and can confirm value shows up correctly and filtering by value fields just works.

However headers are still show up as text instead of being decoded as msgpack.

image

weeco (approved):

PR #1040, opened 2024-01-26
Proto schema registry refresh optimizations

bojand:
Iterative work for optimizing some of the proto schema registry refreshing.

# Changes

- In schema Service we not keep a map of proto schemas, and proto file descriptors indexed by schema ID.
- As we refresh the schemas, we determine which ones are new or have been updated, and merge those into the index. We only parse and compile these updated / new ones.
- We also ensure that the removed ones are removed from the mapping.
- Changes the returned types of `GetSchemasBySubject()`, `GetSchemas()`, `GetSchemasIndividually()` to return pointer types.
- `gci` seems to incorrectly sort the `"maps"` import. Updated the tooling hoping it would address the issue, but it didn't seem to help, so I commented that step out for now.

I will try and add some tests for this soon.


backend/pkg/schema/client.go
```diff
@@ -687,35 +688,45 @@ func (c *Client) CreateSchema(ctx context.Context, subjectName string, schema Sc
...
+		return nil, []error{fmt.Errorf("failed to get subjects to fetch schemas for: %w", err)}
 	}
 
 	g, ctx := errgroup.WithContext(ctx)
 	g.SetLimit(10) // limit max concurrency to 10 requests at a time
 
-	schemas := make([]SchemaVersionedResponse, 0, len(subjectsRes.Subjects))
+	schemas := make([]*SchemaVersionedResponse, 0, len(subjectsRes.Subjects))
+	errors := make([]error, len(subjectsRes.Subjects))
+	hasErrors := atomic.Bool{}
```
weeco, line 702: I think we should do `errors := make([]error, 0, len(subjectsRes.Subjects))` and then we can check the length of errors to see if we have errors. I think in most cases we do not expect every single subject to return an error.

It feels a bit weird to return a slice of schemas and a slice of errors. At this point we can probably just refactor the function to return schema + error via a channel and handle the error and returned schema on the receiving side.

bojand: The intent is to be able to get all the schemas we successfully fetched, even if we failed to retrieve some subset of all the schemas in SR. So we return all the successfully fetched schemas and a collection of errors.
This is the case for both `GetSchemasIndividually()` and `GetSchemas()`, and then `GetProtoDescriptors()` we are able to log all the failed schemas, but still be able to parse and compile the successfully retrieved ones so we could potentially use them.

backend/pkg/schema/service.go
```diff
@@ -67,35 +74,69 @@ func (s *Service) CheckConnectivity(ctx context.Context) error {
...
 	// duplicate requests against the schema registry
 	key := "get-proto-descriptors"
-	v, err, _ := s.requestGroup.Do(key, func() (any, error) {
-		schemasRes, err := s.registryClient.GetSchemas(ctx, false)
-		if err != nil {
-			// If schema registry returns an error we want to retry it next time, so let's forget the key
-			s.requestGroup.Forget(key)
-			return nil, fmt.Errorf("failed to get schema from registry: %w", err)
+	_, err, _ := s.requestGroup.Do(key, func() (any, error) {
+		schemasRes, errs := s.registryClient.GetSchemas(ctx, false)
```
weeco, line 84: schemaRes will be nil as far as I can tell if all retrieved schemas errored (e.g. schema registry not reachable). I think this func will still work if this is the case, but would be good to test the case where schemaRes is nil.

Rather than using singleflight we could use https://github.com/twmb/go-cache to simplify some things, but I assume you tried to keep as much of the original code as possible

weeco (approved):

PR #1056, opened 2024-01-31
frontend: don't re-request messages on save to get raw data, instead …

rikimaru0345:
…always include it in every search; otherwise on fast moving topics our reuqest is outdated by the time the user clicks the save messages button (also, we are not even guaranteed that we get the messages from the exact same partitions and offsets each time)

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages/index.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages/index.tsx
@@ -618,7 +618,7 @@ export class TopicMessageView extends Component<TopicMessageViewProps> {
                 Save Messages
             </Button>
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -784,14 +783,10 @@ class SaveMessagesDialog extends Component<{
                         </Checkbox>
                     </ModalBody>
                     <ModalFooter gap={2}>
-                        <Button variant="outline" colorScheme="red" onClick={onClose} isDisabled={this.isLoadingRawMessage}>
+                        <Button variant="outline" colorScheme="red" onClick={onClose}>
```
weeco, line 786: Disable button if no messages?

weeco (approved):

PR #1063, opened 2024-02-01
Rel 2.4/backend fix schema subject param unescape

rikimaru0345:


bojand (approved):

backend/pkg/api/handle_schema_registry.go
```diff
@@ -817,3 +819,33 @@ func (api *API) handleValidateSchema() http.HandlerFunc {
...
+		subjectPart = requestURI[:lastIndex]
+	}
+
+	subject, err := url.PathUnescape(subjectPart)
+	if err != nil {
+		subject = subjectPart
+	}
+
+	return subject
+}
```
weeco, line 851: @bojand I actually do not understand this, can you add a table unit test with expect inputs & outputs? That would help me understand this better in the future I believe

twmb (commented):
+1

PR #1067, opened 2024-02-02
Implement users filter

weeco:
This PR changes the filter for the ListUsersRequest and implements the missing filter functionality. Additionally it adds a few integration tests for the users API


backend/pkg/api/connect/integration/user_test.go
```diff
@@ -0,0 +1,213 @@
...
+			err := s.redpandaAdminClient.DeleteUser(childCtx, username)
+			assert.NoError(err)
+		}
+
+		username1 := "console-integration-test-list-users-1"
+		username2 := "console-integration-test-list-users-2"
+		createUser(username1)
+		createUser(username2)
+		defer deleteUser(username1)
+		defer deleteUser(username1)
```
krystianity, line 61: @weeco this seems to be a bug :)

weeco: Thanks @krystianity for the unexpected review :D, will fix

PR #1070, opened 2024-02-03
Connector reference doc edits

kbatuigas:
Language/style edits for kafka-connect


proto/redpanda/api/dataplane/v1alpha1/kafka_connect.proto
```diff
@@ -294,7 +294,7 @@ message UpsertConnectorRequest {
     (google.api.field_behavior) = REQUIRED,
     (buf.validate.field).required = true,
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
-      description: "The unique name of the connect cluster you are targeting",
+      description: "Unique name of target connect cluster. Use the cluster name defined in the console's configuration file.",
```
weeco, line 297: While this is technically true, this may be confusing for our Redpanda cloud users who don't have access or awareness of Redpanda Console or it's configuration file. They should always use "redpanda" as this is the cluster name that we configure behind the scenes.

Maybe worth to add a sentence that is specific to Redpanda Cloud here?

kbatuigas: @weeco @sago2k8 How about something like the following?

> For Redpanda self-hosted deployments, use the cluster name defined in the console's configuration file. For Redpanda Cloud, use `redpanda`.

weeco: SGTM

PR #1076, opened 2024-02-05
backend: list message concurrent stream send fix

### LINKED ISSUE #1073: Inconsitent filtering messages on a topic
Hi, version 2.4.1

filtering messages on a topic

sometimes it works correctly and returns the correct result, but more often it does not work properly

some example:
there is a message with the key Ao9JdNYRQPirTIGe2Nz64g
image

filtering out works
image

finding doesn't work
image

Often I see such error:
image

No errors in logs:
image

bojand:
Multiple goroutines may perform writes (stream `Send()`s) to the stream via stream reporter. In stream reporter ensure these are guarded using a mutex.

Minor frontend cleanup to only have a single Console service client.

Fixes #1073.

### DIFF
--- a/backend/pkg/api/connect/service/console/service.go
+++ b/backend/pkg/api/connect/service/console/service.go
@@ -52,16 +52,6 @@ func NewService(logger *zap.Logger,
 
 // ListMessages consumes a Kafka topic and streams the Kafka records back.
 func (api *Service) ListMessages(ctx context.Context, req *connect.Request[v1alpha.ListMessagesRequest], stream *connect.ServerStream[v1alpha.ListMessagesResponse]) error {
-	timeout := 35 * time.Second
-	if req.Msg.GetFilterInterpreterCode() != "" || req.Msg.GetStartOffset() == console.StartOffsetNewest {
-		// Push-down filters and StartOffset = Newest may be long-running streams.
-		// There's already a client-side provided timeout which we usually trust.
-		// But additionally we want to ensure it never takes much longer than that.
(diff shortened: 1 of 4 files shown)


backend/pkg/api/connect/service/console/service.go
```diff
@@ -122,6 +112,17 @@ func (api *Service) ListMessages(ctx context.Context, req *connect.Request[v1alp
...
 
+	timeout := 35 * time.Second
+	if req.Msg.GetFilterInterpreterCode() != "" || req.Msg.GetStartOffset() == console.StartOffsetNewest {
+		// Push-down filters and StartOffset = Newest may be long-running streams.
+		// There's already a client-side provided timeout which we usually trust.
+		// But additionally we want to ensure it never takes much longer than that.
+		timeout = 31 * time.Minute
+	}
+
+	ctx, cancel := context.WithTimeoutCause(ctx, timeout, errors.New("list fetch timeout"))
```
weeco, line 122: Very good to add the reason there 👍

frontend/src/state/backendApi.ts
```diff
@@ -415,6 +421,7 @@ const apiStore = {
                             this.messagesTotalConsumed = Number(res.controlMessage.value.messagesConsumed);
                             break;
                         case 'done':
+                            console.log('done phase')
```
weeco, line 424: Can we remove the console logs? I think this would always print a log message to the console on a successful list message response. We try to keep the console log relatively lean (only weird behaviour should make it to the browser console ideally)

PR #1081, opened 2024-02-06
Fixes invalid regexp scenario

jvorcak:


frontend/src/components/pages/schemas/Schema.List.tsx
```diff
@@ -107,11 +107,15 @@ class SchemaList extends PageComponent<{}> {
...
-            .filter(x => uiSettings.schemaList.showSoftDeleted || (!uiSettings.schemaList.showSoftDeleted && !x.isSoftDeleted))
-            .filter(x => x.name.toLowerCase().match(quickSearchRegExp));
+        let filteredSubjects = api.schemaSubjects
+
+        try {
+            const quickSearchRegExp = new RegExp(uiSettings.schemaList.quickSearch, 'i')
+            filteredSubjects = filteredSubjects.filter(x => uiSettings.schemaList.showSoftDeleted || (!uiSettings.schemaList.showSoftDeleted && !x.isSoftDeleted))
+                .filter(x => x.name.toLowerCase().match(quickSearchRegExp));
+        } catch (e) {
+            console.warn('Invalid expression')
```
malinskibeniamin, line 117: Ideally we should handle or rethrow the error or capture the exception. Maybe we should also check that the regexp should run, if there is no `uiSettings.schemaList.quickSearch` for example?

malinskibeniamin (commented):
https://github.com/redpanda-data/console/pull/1081/files#r1479625471 + linting

frontend/src/components/pages/schemas/Schema.List.tsx
jvorcak: @malinskibeniamin based on what I found there is no other way to check whether regexp is valid other than trying to compose it and catch with an exception. IF it's valid, we filter on it, if it's not valid, we keep the original array.

malinskibeniamin (approved):

frontend/src/components/pages/schemas/Schema.List.tsx
```diff
@@ -107,11 +107,15 @@ class SchemaList extends PageComponent<{}> {
         if (api.schemaOverviewIsConfigured == false) return renderNotConfigured();
         if (api.schemaSubjects === undefined) return DefaultSkeleton; // request in progress
 
-        const quickSearchRegExp = new RegExp(uiSettings.schemaList.quickSearch, 'i')
-
-        const filteredSubjects = api.schemaSubjects
-            .filter(x => uiSettings.schemaList.showSoftDeleted || (!uiSettings.schemaList.showSoftDeleted && !x.isSoftDeleted))
-            .filter(x => x.name.toLowerCase().match(quickSearchRegExp));
+        let filteredSubjects = api.schemaSubjects
```
malinskibeniamin, line 110: lint?

malinskibeniamin (approved):

PR #1090, opened 2024-02-07
backend: support AWS IAM role auth

vrmare:
Use AWS IAM role for auth. The console config I'm using:

```
kafka:
  brokers:
    - b-1.xyz.kafka.us-east-1.amazonaws.com:9098
    - b-2.xyz.kafka.us-east-1.amazonaws.com:9098
  clientId: redpanda-console
  schemaRegistry:
    enabled: false
  sasl:
    enabled: true
    mechanism: AWS_MSK_IAM
  tls:
    enabled: true
    insecureSkipTlsVerify: true
```

This is support my use case where I want to run console in a managed EKS with service accounts using a IAM role.


vrmare:
I am still figuring out a way to test this change in my EKS with IAM setup. How do I trigger a push to a registry mentioned here? https://github.com/redpanda-data/console/blob/master/docs/installation.md#docker-images

Thanks.

backend/pkg/kafka/config_helper.go
```diff
@@ -146,12 +147,34 @@ func NewKgoConfig(cfg *config.Kafka, logger *zap.Logger, hooks kgo.Hook) ([]kgo.
...
+				if err != nil {
+					return nil, fmt.Errorf("unable to get aws credentials: %v", err)
+				}
+				mechanism = aws.Auth{
+					AccessKey:    val.AccessKeyID,
+					SecretKey:    val.SecretAccessKey,
+					SessionToken: val.SessionToken,
+					UserAgent:    val.ProviderName,
+				}.AsManagedStreamingIAMMechanism()
+			} else {
```
weeco, line 169: I'm not familiar how the AWS IAM works, could you add some comments here to describe why we we create a new "session" here in case no accesskey/secretkey/session token is provided? It looks like we would use ephemeral credentials pulled upon startup in that case or what is happening there exactly?

weeco:
Hi @vrmare ,
thanks for your pull request. It's indeed not easy to get your own Docker image (because we don't have a Dockerfile in this repo anymore) - which is unfortunate for testing. Easiest way would probably to just compile the Go binary and start it without the frontend. Is that an option for you?

If you prefer you could also add your own Dockerfile to the repository and build a Docker image. The backend does not require anything special - any Dockerfile for Go applications should work fine. You can find **some** help in the https://github.com/redpanda-data/console/blob/master/CONTRIBUTING.md guide.

backend/pkg/kafka/config_helper.go
vrmare: Hello, 

I was able to finally build this locally and test on my EKS with IRSA setup, and it works!

```
❯ kubectl -n dev logs -f redpanda-console-5ffb68649-g8hhb
{"level":"info","ts":"2024-02-15T23:21:06.189Z","msg":"started Redpanda Console","version":"development","built_at":"<not set>"}
{"level":"info","ts":"2024-02-15T23:21:06.189Z","msg":"creating Kafka connect HTTP clients and testing connectivity to all clusters"}
{"level":"info","ts":"2024-02-15T23:21:06.190Z","msg":"tested Kafka connect cluster connectivity","successful_clusters":0,"failed_clusters":0}
{"level":"info","ts":"2024-02-15T23:21:06.190Z","msg":"successfully create Kafka connect service"}
{"level":"info","ts":"2024-02-15T23:21:06.318Z","msg":"connecting to Kafka seed brokers, trying to fetch cluster metadata"}
{"level":"info","ts":"2024-02-15T23:21:07.590Z","msg":"successfully connected to kafka cluster","advertised_broker_count":2,"topic_count":7,"controller_id":2,"kafka_version":"v3.6"}
{"level":"fatal","ts":"2024-02-15T23:21:07.596Z","msg":"failed to load index.html from embedded filesystem","error":"open index.html: file does not exist"}
```

With a config:

```
console:
  config:
    redpanda: {}
    kafka:
      brokers:
        - b-1.x1.kafka.us-east-1.amazonaws.com:9098
        - b-2.x1.us-east-1.amazonaws.com:9098
      clientId: redpanda-console
      schemaRegistry:
        enabled: false
      sasl:
        enabled: true
        mechanism: AWS_MSK_IAM
        awsMskIam: {}
      tls:
        enabled: true
        insecureSkipTlsVerify: true

# Additional environment variables for the Console Deployment
extraEnv:
  - name: AWS_REGION
    value: us-east-1
```

As you can see, under config `awsMskIam` I have not specified the accesskey/secretkey/session keys and let the IRSA do the sasl auth.

vrmare: This change actually was suggested here: https://github.com/twmb/franz-go/issues/39#issuecomment-1234999666

vrmare: If there are instructions on how to build and include UI, I'm happy to do that in the local docker file testing. Would love to run this for sometime ensuring the session credentials are renewed or if that needs handling as well.

backend/pkg/kafka/config_helper.go
```diff
@@ -146,12 +147,37 @@ func NewKgoConfig(cfg *config.Kafka, logger *zap.Logger, hooks kgo.Hook) ([]kgo.
...
+				// SessionToken is optional
+				mechanism = aws.Auth{
+					AccessKey:    cfg.SASL.AWSMskIam.AccessKey,
+					SecretKey:    cfg.SASL.AWSMskIam.SecretKey,
+					SessionToken: cfg.SASL.AWSMskIam.SessionToken,
+					UserAgent:    cfg.SASL.AWSMskIam.UserAgent,
+				}.AsManagedStreamingIAMMechanism()
+			} else {
+				cfgVal, err := awsConfig.LoadDefaultConfig(
+					context.TODO(),
```
weeco, line 162: Worth to use a context with a reasonable timeout here I believe?

backend/pkg/kafka/config_helper.go
```diff
@@ -146,12 +147,37 @@ func NewKgoConfig(cfg *config.Kafka, logger *zap.Logger, hooks kgo.Hook) ([]kgo.
...
+					awsConfig.WithRegion(cfg.SASL.AWSMskIam.Region),
+				)
+				if err != nil {
+					return nil, err
+				}
+
+				creds, err := cfgVal.Credentials.Retrieve(context.TODO())
```
weeco, line 169: Same here, let's use a ctx that times out in case we never get a response

backend/pkg/kafka/config_helper.go
```diff
@@ -146,12 +147,37 @@ func NewKgoConfig(cfg *config.Kafka, logger *zap.Logger, hooks kgo.Hook) ([]kgo.
...
 		// AWS MSK IAM
 		if cfg.SASL.Mechanism == config.SASLMechanismAWSManagedStreamingIAM {
-			mechanism := aws.Auth{
-				AccessKey:    cfg.SASL.AWSMskIam.AccessKey,
-				SecretKey:    cfg.SASL.AWSMskIam.SecretKey,
-				SessionToken: cfg.SASL.AWSMskIam.SessionToken,
-				UserAgent:    cfg.SASL.AWSMskIam.UserAgent,
-			}.AsManagedStreamingIAMMechanism()
+			var mechanism sasl.Mechanism
+			if cfg.SASL.AWSMskIam.AccessKey != "" &&
```
twmb, line 151: I'd say || rather than &&, or maybe even more clearly -- error if only one exists.

backend/pkg/kafka/config_helper.go
vrmare: updated to use a duration that is configurable for context. Unless someone is really working in a really crappy network, 10s is generous for getting credentials here.

backend/pkg/kafka/config_helper.go
vrmare: updated

backend/pkg/kafka/config_helper.go
vrmare: Good point. Used the existing validate func to capture this requirement on config

vrmare:
@weeco updated the config as well. Thanks for the reviews.

weeco (approved):

IlyaNakhaichuk:
@weeco Tell me, when do you plan to release this change?

Issue #1107, opened 2024-02-09
Invalid pushdown filter do not throw an error


weeco:
Invalid pushdown filters (e.g. just `ret`) are not handled well. The message search just continues running without any error prompt or results. This is particularly painful for more complex queries where users may not be certain that their filter works.

This is easy to reproduce by using an invalid filter. In my test I used `ret` and the frontend sends the following code to the backend:

base64:
```
ZnVuY3Rpb24gZmlsdGVyMSgpIHsKICAgICAgICAgICAgICAgICAgICByZXR1cm4gcmV0dTsNCgogICAgICAgICAgICAgICAgfQoKcmV0dXJuIGZpbHRlcjEoKQ
```

decoded to:

```js
function filter1() {
                    return retu;

                }

return filter1()
```

PR #1108, opened 2024-02-11
backend: add filter code checks to list message RPC

bojand:
Addresses #1107.

### DIFF
--- a/backend/pkg/api/connect/service/console/service.go
+++ b/backend/pkg/api/connect/service/console/service.go
@@ -98,7 +98,7 @@ func (api *Service) ListMessages(ctx context.Context, req *connect.Request[v1alp
 	if err != nil {
 		return apierrors.NewConnectError(
 			connect.CodeInvalidArgument,
-			err,
+			fmt.Errorf("failed decoding provided interpreter code: %w", err),
 			apierrors.NewErrorInfo(commonv1alpha1.Reason_REASON_INVALID_INPUT.String()),
 		)
 	}
--- a/backend/pkg/api/connect/service/console/service.go
+++ b/backend/pkg/api/connect/service/console/service.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"connectrpc.com/connect"
+	"github.com/dop251/goja"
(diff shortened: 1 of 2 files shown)


backend/pkg/api/connect/service/console/service.go
```diff
@@ -93,7 +94,25 @@ func (api *Service) ListMessages(ctx context.Context, req *connect.Request[v1alp
 		}
 	}
 
-	interpreterCode, _ := lmq.DecodeInterpreterCode() // Error has been checked in validation function
+	interpreterCode, err := lmq.DecodeInterpreterCode()
+	if err != nil {
+		return apierrors.NewConnectError(
+			connect.CodeInvalidArgument,
+			err,
```
weeco, line 101:
```suggestion
			fmt.Errorf("failed decoding provided interpreter code: %w", err),
```

NIT: Just to add some context where it is failing.

backend/pkg/api/connect/service/console/service.go
```diff
@@ -93,7 +94,25 @@ func (api *Service) ListMessages(ctx context.Context, req *connect.Request[v1alp
...
+		)
+	}
+
+	// test compile
+	code := fmt.Sprintf(`var isMessageOk = function() {%s}`, interpreterCode)
+	_, err = goja.Compile("", code, true)
+	if err != nil {
+		return apierrors.NewConnectError(
+			connect.CodeInvalidArgument,
+			err,
```
weeco, line 112:
```suggestion
			fmt.Errorf("failed to compile provided interpreter code: %w", err),
```

Same here

weeco (approved):
That's cool that we can check for syntax issue upfront - and only once. 

I imagine to solve this issue completely we would also need to consider issues that could happen for a specific message. For example the provided code tries to access a property inside an array and that fails because it hit a tombstone where this property doesn't exist. Do we already return an error in that case? 

I'm not even sure if we should - because we may be emitting thousands of errors potentially, but if the code fails for all messages in that topic it would also be weird to not inform about the issue.

PR #1109, opened 2024-02-11
backend: stream progress report when using newest message list messag…

bojand:
…e request

### DIFF
--- a/backend/pkg/api/connect/service/console/stream_progress_reporter.go
+++ b/backend/pkg/api/connect/service/console/stream_progress_reporter.go
@@ -37,17 +37,33 @@ type streamProgressReporter struct {
 }
 
 func (p *streamProgressReporter) Start() {
-	// If search is disabled do not report progress regularly as each consumed message will be sent through the socket
-	// anyways
-	if p.request.FilterInterpreterCode == "" {
+	tickerDuration := 0 * time.Second
+
+	// We should report progress in two scenarios.
+	// If filter is enabled it could take a while to find the matching record(s).
+	// If search is for newest records it could take a while to get new records.
+	// For the second scenario we also need to get around certain infrastructure and ingress idle connection limits.
+	// With "newest" request, we may wait a while until we get a record to send.
+	// While we wait for a record the connection is idle.
+	// Different ingress controllers, proxies, and load balancers have different default settings for idle connections.
+	// For example AWS LB has a default idle connection timeout of 1m.
+	// Essentially we need a ping / keep alive message in stream to work around these idle timeouts.
+
+	if p.request.FilterInterpreterCode != "" {
+		tickerDuration = 0 * time.Second
+	} else if p.request.StartOffset == -3 { // Newest
+		tickerDuration = 30 * time.Second
+	}
+
+	if tickerDuration == 0 {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


backend/pkg/api/connect/service/console/stream_progress_reporter.go
```diff
@@ -37,17 +37,33 @@ type streamProgressReporter struct {
...
 func (p *streamProgressReporter) Start() {
-	// If search is disabled do not report progress regularly as each consumed message will be sent through the socket
-	// anyways
-	if p.request.FilterInterpreterCode == "" {
+	tickerDuration := 0 * time.Second
+
+	// We should report progress in two scenarios.
+	// If filter is enabled it could take a while to find the matching record(s).
+	// If search is for newest records it could take a while to get new records.
+	// For the second scenario we also need to get around certain infrastructure and ingress idle connection limits.
```
weeco, line 45: > For the second scenario we also need to get around certain infrastructure

That's true for both cases as both can take a while until we actually send messages via the stream. The difference was that we already used to send a progress report (how many messages & bytes we have consumed) when we use pushdown filters.

backend/pkg/api/connect/service/console/stream_progress_reporter.go
```diff
@@ -37,17 +37,33 @@ type streamProgressReporter struct {
...
+	// With "newest" request, we may wait a while until we get a record to send.
+	// While we wait for a record the connection is idle.
+	// Different ingress controllers, proxies, and load balancers have different default settings for idle connections.
+	// For example AWS LB has a default idle connection timeout of 1m.
+	// Essentially we need a ping / keep alive message in stream to work around these idle timeouts.
+
+	if p.request.FilterInterpreterCode != "" {
+		tickerDuration = 0 * time.Second
```
weeco, line 53: In this case (pushdown filter enabled) we would return early, which is wrong I believe?

backend/pkg/api/connect/service/console/stream_progress_reporter.go
```diff
@@ -37,17 +37,33 @@ type streamProgressReporter struct {
...
+	} else if p.request.StartOffset == -3 { // Newest
+		tickerDuration = 30 * time.Second
+	}
+
+	if tickerDuration == 0 {
 		return
 	}
```
weeco, line 60: What do you think about something as simple as that?

```suggestion
     tickerDuration := 30 * time.Second
      if p.requestFilterInterpreterCode != "" {
        tickerDuration = time.Second
      }
```

Why? Because this issue applies to any streaming search that takes a very long time for whatever reason. It wouldn't hurt to send the progress update (as a "keep alive" signal) if it takes 30s for any reason. Only in the case of pushdown filters we want to report more frequently (because we are actually curious about the progress).

backend/pkg/api/connect/service/console/stream_progress_reporter.go
bojand: Oops that was suppose to be `1`.

backend/pkg/api/connect/service/console/stream_progress_reporter.go
bojand: Improved comments.

weeco (approved):

PR #1119, opened 2024-02-15
frontend: wip connector logs

rikimaru0345:
https://github.com/redpanda-data/console/assets/28133019/7559c752-9680-468f-8ca7-81c3e4e28071


frontend/src/components/pages/connect/Connector.Details.tsx
```diff
@@ -125,21 +132,28 @@ const KafkaConnectorMain = observer(
                                 <Tooltip placement="top" isDisabled={canEdit !== true} label={'You don\'t have \'canEditConnectCluster\' permissions for this connect cluster'} hasArrow={true}>
                                     <Button
                                         variant="outline"
-                                        style={{width: '200px'}}
+                                        style={{ width: '200px' }}
                                         disabled={(() => {
```
jvorcak, line 136: I think this property should be `isDisabled`. There some some TS bug in the UI libraryI think, because it doesn't warn us.

frontend/src/components/pages/connect/Connector.Details.tsx
```diff
@@ -475,3 +490,138 @@ const ConnectorDetails = observer((p: {
...
+            header: 'Timestamp',
+            accessorKey: 'timestamp',
+            cell: ({ row: { original: { timestamp } } }) => <TimestampDisplay unixEpochMillisecond={timestamp} format="default" />,
+            size: 30
+        },
+        {
+            header: 'Value',
+            accessorKey: 'value',
+            cell: ({ row: { original } }) => <MessagePreview msg={original} previewFields={() => []} isCompactTopic={topic ? topic.cleanupPolicy.includes('compact') : false} />,
+            size: 700
```
jvorcak, line 541: I think in order to expand this column, we need to set this to Infinity or Number.MAX_SAFE_INTEGER https://tanstack.com/table/v8/docs/guide/column-sizing

ivpanda:
Nitpick but there is so much space to the right of the `[+]` ? Can the timestamp and value be shifted left, so this is more consistent in appearance with the rest of our logs?

ivpanda:
Also, is it possible to have connector status be in title case (e.g. `Unhealthy`, `Paused`) rather than all caps?

rikimaru0345:
@ivpanda both things will be fixed later.
This whole pr has to be adjusted to other changes related to message search.
I'll continue with this PR once that is done.

frontend/src/components/pages/connect/Connector.Details.tsx
rikimaru0345: good catch

jvorcak (approved):

Issue #1104, opened 2024-02-09
Resetting pagination if the result returns fewer pages


ErykCh:
Hi,
this is not critical.

I'm on page 10:
![image](https://github.com/redpanda-data/console/assets/95491982/2bde7f26-bb52-4c8a-879c-2b5464e74129)

Now I change the filter and the results are on the first page, but they are not visible.
![image](https://github.com/redpanda-data/console/assets/95491982/ac27dddb-1370-49de-950a-37e6e713e5ce)

I need to turn to page one
![image](https://github.com/redpanda-data/console/assets/95491982/8240d767-4d83-4183-9945-b14946253119)

PR #1121, opened 2024-02-16
Limit last page to the last existing page in the table

jvorcak:
Fixes https://github.com/redpanda-data/console/issues/1104


jvorcak:
I'd appreciate if we try to test this properly before merging.

rikimaru0345 (approved):

Issue #1118, opened 2024-02-15
Even Kafka Connect is not enabled, still got "successfully create Kafka connect service"


jovezhong:
I think the log message is a bit misleading
```
console-1  | {"level":"info","ts":"2024-02-15T01:09:39.561Z","msg":"started Redpanda Console","version":"v2.4.3","built_at":"1707418068"}
console-1  | {"level":"info","ts":"2024-02-15T01:09:39.561Z","msg":"creating Kafka connect HTTP clients and testing connectivity to all clusters"}
console-1  | {"level":"info","ts":"2024-02-15T01:09:39.561Z","msg":"tested Kafka connect cluster connectivity","successful_clusters":0,"failed_clusters":0}
console-1  | {"level":"info","ts":"2024-02-15T01:09:39.561Z","msg":"successfully create Kafka connect service"}
console-1  | {"level":"debug","ts":"2024-02-15T01:09:39.561Z","msg":"creating new kafka client","config":{..}
console-1  | {"level":"debug","ts":"2024-02-15T01:09:39.561Z","msg":"configuring SCRAM-SHA-256 mechanism"}
console-1  | {"level":"info","ts":"2024-02-15T01:09:39.561Z","msg":"connecting to Kafka seed brokers, trying to fetch cluster metadata"}
console-1  | {"level":"debug","ts":"2024-02-15T01:09:39.823Z","logger":"kafka_client_hooks","msg":"kafka connection succeeded","host":"..","dial_duration":0.077249916}
```

bojand:
Fixed via #1122.

jovezhong:
Thanks for the quick turnaround.

PR #1122, opened 2024-02-16
backend: check for connect clusters in new connect service startup to…

bojand:
… avoid confusing logs.

Addresses #1118

### DIFF
--- a/backend/pkg/connect/service.go
+++ b/backend/pkg/connect/service.go
@@ -41,7 +41,6 @@ type ClientWithConfig struct {
 // NewService creates a new connect.Service. It tests the connectivity for each configured
 // Kafka connect cluster proactively.
 func NewService(cfg config.Connect, logger *zap.Logger) (*Service, error) {
-	// 1. Create a client for each configured Connect cluster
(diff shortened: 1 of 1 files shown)


backend/pkg/connect/service.go
```diff
@@ -41,45 +41,48 @@ type ClientWithConfig struct {
...
 
-		opts = append(opts, con.WithHost(clusterCfg.URL))
-		// TLS Config
-		tlsCfg, err := clusterCfg.TLS.TLSConfig()
-		if err != nil {
-			childLogger.Error("failed to create TLS config for Kafka connect HTTP client, fallback to default TLS config", zap.Error(err))
-			tlsCfg = &tls.Config{MinVersion: tls.VersionTLS12}
-		}
-		opts = append(opts, con.WithTLSConfig(tlsCfg))
+	if len(cfg.Clusters) > 0 {
```
weeco, line 47: I think it would be nicer to check at the beginning af the func and return early with the dummy struct `if len(cfg.Clusters) == 0`

weeco (approved):

PR #1124, opened 2024-02-19
Better handling of long names in the tables & breadcrumbs

jvorcak:
1. I think having it in the breadcrumb is a good thing when it comes to consistency. We currently have the detail items there for all the pages I know of.

2. I think it's however OK to cut it when needed (we have no other choice). I've worked on a prototype that takes all the possible space and adds `...` once needed.  (Please note that the whitespace at the end is needed for the refreshing text, if we could get rid of that it could go til the end).

Screenshot 2024-02-19 at 12 09 18 PM
Screenshot 2024-02-19 at 12 14 57 PM

However in this scenario we would need to add `copy to clipboard` icon so that users can copy it if it's too long. Sth like github does 
Screenshot 2024-02-19 at 12 17 39 PM

In the tables I prototyped the solution that would take `n` number of lines before it shows `...` 
Screenshot 2024-02-19 at 12 15 37 PM

WDYT?

### DIFF
--- a/frontend/src/components/layout/Header.tsx
+++ b/frontend/src/components/layout/Header.tsx
@@ -14,62 +14,67 @@ import { observer } from 'mobx-react';
 import { Link, useRouteMatch } from 'react-router-dom';
 import { isEmbedded } from '../../config';
 import { uiState } from '../../state/uiState';
-import { MotionDiv } from '../../utils/animationProps';
-import { ZeroSizeWrapper } from '../../utils/tsxUtils';
 import { UserPreferencesButton } from '../misc/UserPreferences';
 import DataRefreshButton from '../misc/buttons/data-refresh/Component';
 import { IsDev } from '../../utils/env';
-import { Breadcrumb, BreadcrumbItem, BreadcrumbLink, BreadcrumbLinkProps, ColorModeSwitch, Flex } from '@redpanda-data/ui';
+import { Box, Breadcrumb, BreadcrumbItem, BreadcrumbLink, BreadcrumbLinkProps, ColorModeSwitch, Flex } from '@redpanda-data/ui';
 
 const AppPageHeader = observer(() => {
     const showRefresh = useShouldShowRefresh();
 
-    return <MotionDiv identityKey={uiState.pageTitle} className="pageTitle" style={{ display: 'flex', paddingRight: '16px', alignItems: 'center', marginBottom: '10px' }}>
-        <Breadcrumb spacing="8px" separator={<ChevronRightIcon/>}>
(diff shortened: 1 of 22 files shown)


ivpanda:
I like the `...` solution for the breadcrumbs and in the tables. In the tables, I'd like to keep everything to one line rather than multiples. 

I would also agree that we need to add a copy ID button somewhere. To this end, I'm struggling a little with the difference here between console and the regular UI, where we have a title for topics and we could do something like the below: 

image

Instead, in this case, we'd have to put the copy button in the breadcrumb, which doesn't feel right to me. How much room do we have here to make/propose changes right now?

jvorcak:
> I like the `...` solution for the breadcrumbs and in the tables. In the tables, I'd like to keep everything to one line rather than multiples.
> 
> I would also agree that we need to add a copy ID button somewhere. To this end, I'm struggling a little with the difference here between console and the regular UI, where we have a title for topics and we could do something like the below:
> 
> image
> Instead, in this case, we'd have to put the copy button in the breadcrumb, which doesn't feel right to me. How much room do we have here to make/propose changes right now?

I think we have no limitations right now. What would help me is to showcase how it should be in the console (and cloud) for the long names and I can provide an implementation.

When it comes to the tables, so we will stick with one line only and `...` right?

ivpanda:
> When it comes to the tables, so we will stick with one line only and ... right?

Yes. The user can then click the table item to view the details page and copy the full name. 

> What would help me is to showcase how it should be in the console (and cloud) for the long names and I can provide an implementation.

My preference would be to have the two be consistent, and console echo Cloud's header, with breadcrumbs and titles, rather than only breadcrumb as is in console: 

image

I understand that the breadcrumb may be smaller in Console, but making them similar in style and bringing the title of the page in below them would help a great deal.

jvorcak:
@ivpanda I think that is related to https://github.com/redpanda-data/console-enterprise/issues/190

Maybe it would be the best if we could have a call today and sync about all of the details.

ivpanda:
After a call, we've decided to add a copy button to the other buttons in the breadcrumb. In Cloud, we'll add the copy button to the same set of buttons that appear in the title: 

image

Then, when [this issue](https://github.com/redpanda-data/console-enterprise/issues/190) is implemented, we'll keep the button in the title.

rikimaru0345 (approved):

jvorcak:
Screenshot 2024-02-29 at 10 26 08
Screenshot 2024-02-29 at 10 26 02
Screenshot 2024-02-29 at 10 25 02
Screenshot 2024-02-29 at 10 24 54

PR #1126, opened 2024-02-20
backend: remove websocket code

bojand:


weeco:
Great, we should be able to remove the gorilla/websocket dependency as well in the end no?

weeco (approved):

PR #1130, opened 2024-02-22
backend: fix inconsistent error scheme on http

weeco:
Prior to this commit the error scheme on the HTTP
interface were inconsistent when using the
connect.ErrorWriter. In this commit we introduce
a new function that allows to write connect.Errors to the HTTP interface so that it gets serialized
in a consistent format.


sago2k8 (approved):
LGTM

bojand (approved):

PR #1145, opened 2024-02-29
Parse redpanda errors

weeco:
Properly parse admin api errors and bubble up the error message returned by Redpanda to the response.

### DIFF
--- a/backend/pkg/api/connect/errors/redpanda.go
+++ b/backend/pkg/api/connect/errors/redpanda.go
@@ -35,7 +35,6 @@ func NewConnectErrorFromRedpandaAdminAPIError(err error, prefixErrMsg string) *c
 
 		adminAPIErr, err := httpErr.DecodeGenericErrorBody()
 		if err != nil {
-
 			return NewConnectError(
 				connectCode,
 				errors.New(prefixErrMsg+httpErr.Error()),
--- a/backend/pkg/api/connect/errors/redpanda.go
+++ b/backend/pkg/api/connect/errors/redpanda.go
@@ -33,7 +33,7 @@ func NewConnectErrorFromRedpandaAdminAPIError(err error, prefixErrMsg string) *c
 	if errors.As(err, &httpErr) {
 		connectCode := CodeFromHTTPStatus(httpErr.Response.StatusCode)
 
-		adminApiErr, err := httpErr.DecodeGenericErrorBody()
+		adminAPIErr, err := httpErr.DecodeGenericErrorBody()
 		if err != nil {
 			return NewConnectError(
 				connectCode,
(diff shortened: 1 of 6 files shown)


backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -0,0 +1,111 @@
...
+	case http.StatusOK:
+		return 0
+	case 499:
+		return connect.CodeCanceled
+	case http.StatusInternalServerError:
+		return connect.CodeInternal
+	case http.StatusBadRequest:
+		return connect.CodeInvalidArgument
+	case http.StatusGatewayTimeout:
+		return connect.CodeDeadlineExceeded
```
bojand, line 93: This should be `UNAVAILABLE` according to [gRPC docs](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md)

backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -0,0 +1,111 @@
...
+	case http.StatusNotFound:
+		return connect.CodeNotFound
```
bojand, line 95: I think this should be `UNIMPLEMENTED` ?

backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -0,0 +1,111 @@
...
+	case http.StatusConflict:
+		return connect.CodeAlreadyExists
+	case http.StatusForbidden:
+		return connect.CodePermissionDenied
+	case http.StatusUnauthorized:
+		return connect.CodeUnauthenticated
+	case http.StatusTooManyRequests:
+		return connect.CodeResourceExhausted
```
bojand, line 103: This too should be `UNAVAILABLE`.

backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -0,0 +1,111 @@
...
+		connect.CodeInternal,
+		fmt.Errorf("%v%w", prefixErrMsg, err),
+		NewErrorInfo(v1alpha1.Reason_REASON_REDPANDA_ADMIN_API_ERROR.String()),
+	)
+}
+
+// CodeFromHTTPStatus converts an HTTP response status code into the
+// corresponding connect.Code. If no status code matches, CodeUnknown will be
+// returned.
+func CodeFromHTTPStatus(status int) connect.Code {
```
bojand, line 82: Should this eventually go into common-go heh?

weeco: Could go into the common library, but I'm not yet happy with the mapping since it's so ambiguous (e.g. 404 could be unimplemented or not found). Not sure if we can make everyone happy with our own opinionated mapping at the moment and you figured that connectrpc already has such a mapping.

backend/pkg/api/connect/errors/redpanda.go
bojand: My concern is this will lead to inconsistency from client perspective, some requests that are "not found" return `UNIMPLEMENTED` as is documented in grpc docs and is implemented by Connect, and then some return `CodeNotFound` status code, as is in this case.

I get this is a tricky and unfortunate ambiguity, and really I agree HTTP 404 -> NotFound status code... I do wonder if the mapping is historical or legacy. 🤷‍♂️

bojand (approved):

PR #1147, opened 2024-02-29
chore: remove unused helm workflows

weeco:
This repository used to host a Helm chart, but we moved it to a dedicated repository. Thus these workflows can be deleted.


weeco:
I also just removed the CodeQL action which seems to be deprecated. CodeQL v3 has to be setup via repository settings.

PR #1149, opened 2024-02-29
frontend: add e2e test job to frontend GitHub Actions

bojand:
- Renames `frontend-lint-and-compile-check.yml -> .github/workflows/frontend-verify.yml`
- Adds `e2e-test` job to the GitHub Action, and it depends on the `"lint-and-compile"` job.

Locally:
```sh
$ cd frontend
$ docker compose --file tests/config/docker-compose.yaml up
```
and 
```
$ cd frontend
$ npm i
$ npm run e2e-test
```

The GitHub Action essentially does the same.


.github/workflows/frontend-verify.yml
```diff
@@ -33,3 +33,23 @@ jobs:
           REACT_APP_BUILD_TIMESTAMP=$(date +%s) 
           REACT_APP_DEV_HINT=true
           npm run build 
+  
+  e2e-test:
+    needs: "lint-and-compile"
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
weeco, line 42:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/frontend-verify.yml
```diff
@@ -33,3 +33,23 @@ jobs:
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18.10.0'
```
weeco, line 45:
```suggestion
      - uses: actions/setup-node@v4
        with:
          node-version: '18'
```

Node v18.x for less maintenance effort at the cost of potential differences. In theory our frontend should be using the latest v18.x as well for the build process

.github/workflows/frontend-verify.yml
```diff
@@ -33,3 +33,23 @@ jobs:
...
+      - name: Install dependencies 
+        run: |
+          cd frontend
+          npm ci
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
+      - name: Run Playwright tests
+        run: |
+          cd frontend
```
weeco, line 54: Maybe we can use working-directory instead of the cd command?

.github/workflows/frontend-verify.yml
```diff
@@ -33,3 +33,23 @@ jobs:
...
+          npm run e2e-test
```
weeco, line 55: Does it not require a certain backend/api to be started beforehand? @jvorcak

bojand: Yes I am working on this. Not sure what's the best way to run the right version of backend with the appropriate configs.

jvorcak: @weeco Playwright should be able to do this, but I'm having some issues with running it as part of PW. @bojand can we start the frontend for this in GH actions ?

.github/workflows/frontend-verify.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Install dependencies 
+        working-directory: frontend
+        run: npm ci
+      - name: Run lint
+        working-directory: frontend
+        run: npm run lint
+      - name: Run build frontend
+        working-directory: frontend
+        run: |
+          REACT_APP_BUSINESS=true
```
weeco, line 43: Not sure if we want to run thefrontend in enterprise mode. I think we are running against the community backend version for the tests

.github/workflows/frontend-verify.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Log connect
+        if:  runner.debug == '1'
+        continue-on-error: true
+        run: docker logs connect
+      - name: Run Playwright tests
+        working-directory: frontend
+        run: npm run e2e-test
+      - name: Upload Playwright report artifact
+        uses: actions/upload-artifact@v4
+        if: always()
```
weeco, line 118: What do we need this if for? I assume always is a static true?

frontend/tests/config/console.config.yaml
```diff
@@ -0,0 +1,20 @@
+serveFrontend: false
+kafka:
```
weeco, line 2: Should we add the protobuf configuration too, so that protobuf encoded messages will be decoded?

frontend/tests/topic.spec.ts
```diff
@@ -16,7 +20,7 @@ test.describe('Topic', () => {
         const fillText = 'example content ';
         const content = fillText.repeat((maxMessageSize / fillText.length) + 1);
 
-        const monacoEditor = page.locator('.monaco-editor').nth(1);
+        const monacoEditor = page.locator('.monaco-editor').nth(0);
```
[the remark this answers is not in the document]
jvorcak: @rikimaru0345 this change was there because of the other change you're describing. We can as well provide more explicit test-id and search for a `monaco-editor` within that Locator. I'll investigate once I get to this.

frontend/tests/connector.spec.ts
```diff
@@ -7,6 +7,8 @@ export const SECRET_ACCESS_KEY = 'secretaccesskey';
 export const S3_BUCKET_NAME = 's3bucketname';
 
 test.describe('Connector', async () => {
+    test.skip()
```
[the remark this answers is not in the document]
jvorcak: I think we should always pass a GitHub issue when skipping https://github.com/redpanda-data/console-enterprise/issues/296

.github/workflows/frontend-verify.yml
bojand: It uploads report as an artifact. I changed the condition to be on failure only, as that's really when the report would be useful.

.github/workflows/frontend-verify.yml
bojand: Removed this. @rikimaru0345 to confirm it's the right action.

frontend/tests/topic.spec.ts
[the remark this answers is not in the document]
jvorcak: @rikimaru0345 fixed in https://github.com/redpanda-data/console/pull/1149/commits/0ef0a811c35d4ed29e05ad0f101cd69e414852a9

jvorcak (approved):
Looks OK from what I've checked.

weeco (approved):

.github/workflows/frontend-verify.yml
rikimaru0345: Yes, normal, non-business mode is fine.
It only introduces extra steps that we don't need.

rikimaru0345 (approved):

Issue #1105, opened 2024-02-09
Add support for Karapace (aiven) schema registry


Boniker:
**Error:**
`
{"level":"error","ts":"", "msg":"Sending REST error","route":"/api/schema-registry/mode","method":"GET","status_code":502,"remote_address":"127.0.0.1:35678","public_error":"Failed to retrieve schema registry mode from the schema registry: get mode request failed: Status code 404","error":"get mode request failed: Status code 404"}
`

**Configurations:**
```
kafka:
  brokers: <broker>:9096
  sasl:
    enabled: true
    mechanism: <SSL/PLAINTEXT>
    password: <password>
    username: <username>
  schemaRegistry:
    enabled: true
    password: <password>
    urls:
    - http://<path_to_k8s_service>:8081/
    username: <username>
```

**Product:** console
**Console chart version:** v0.7.20 (with problem, last version without problem: v0.7.15)
**Kafka security.protocol:** SASL_SSL
**SchemaRegistry:** karapace (aiven)

Boniker:
![Uploading Screenshot 2024-02-09 at 15.55.50.png…]()

Boniker:
![redpanda](https://github.com/redpanda-data/console/assets/48773101/b9820b68-6915-4c11-a6e5-33b654c7ece6)

weeco:
There are some schema registry implementations that do not support the `/mode` endpoint. Given we use this endpoint to only show a key/val pair in the "Summary bar" at the top, this should be optional and not cause the entire site to fail.

dweber019:
We have a similar issue, where we use Apicurio and some endpoints are protected and therefore responding 403.
The endpoints can be ignored and the redpanda console could show schemas.

cthtrifork:
We have the same issue after updating console. We are also using apicurio registry.

weeco:
@Boniker You can always overwrite the used Docker image tag in the helm chart. It's not necessary to wait for the Helm chart to update. I've filed a PR for it here though: https://github.com/redpanda-data/helm-charts/pull/1075

Boniker:
Oh, I have already did that. Thanks, everything is working fine.

It was just a reminder

weeco:
@dweber019 As far as I remember there are also other issues such as incompatibilities with the `validate` endpoint which you'll need for creating schemas via Console.

PR #1154, opened 2024-03-03
frontend: still render schema pages if mode request errors

weeco:
SchemaRegistries such as apicurio or karapace do not support the /mode endpoint. When opening the schema registry pages they failed to load because this request failed. The response to this request should not be mandatory. This is changed with this PR.


rikimaru0345 (approved):

weeco:
Fixes https://github.com/redpanda-data/console/issues/1105

PR #1156, opened 2024-03-04
Fix/long names 2

jvorcak:
Fixes

- Spacing in the Broker table
- Spacing in a Message Configration
  - We've made the table a bit wider
  - We widen it to 100% in case the keys are long, if it's not enough, they'll break (in this case we don't use ...)

Screenshot 2024-03-04 at 16 18 11
Screenshot 2024-03-04 at 16 18 39

### DIFF
--- a/frontend/src/components/misc/ConfigList.tsx
+++ b/frontend/src/components/misc/ConfigList.tsx
@@ -18,7 +18,7 @@ import { equalsIgnoreCase } from '../../utils/utils';
 
 import styles from './ConfigList.module.scss';
 
-import { DataTable, Flex, Tooltip } from '@redpanda-data/ui';
+import { DataTable, Flex, Tooltip, Text } from '@redpanda-data/ui';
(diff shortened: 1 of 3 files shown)


weeco:
@jvorcak I believe the config value was meant to be on the same line as the config key (especially in the happy case where we have enough space). Did you change this intentionally?

Good to stretch the width only if necessary by the way 👍

weeco:
I think this was only meant for when we are running out of horizontal space (i.e. your second example). If possible we would still like to show it in just one line (up to the maximum available width in that panel, rather than the fixed max width we have today). But Ivana can confirm.

jvorcak:
Screenshot 2024-03-04 at 16 44 18
Screenshot 2024-03-04 at 16 44 04

In that case I'd forget about breaking the line and just widen the table as Ivana requested. I've comitted the changes.

rikimaru0345 (approved):

PR #1170, opened 2024-03-12
Adds breadcrumbs for more pages

jvorcak:
Fixes https://github.com/redpanda-data/console-enterprise/issues/190

I've checked and only found these 2 occurrences where we missed proper breadcrumbs. In case there are more, feel free to comment. I need to check this in a CloudUI before merging, but I don't have access now.


weeco:
@jvorcak LGTM, but I think to close the referenced issue there are a couple more things form Ivana's recommendation to address. Most notably changing the layout of breadcrumbs to match those in our cloud.

rikimaru0345 (approved):

jvorcak:
> @jvorcak LGTM, but I think to close the referenced issue there are a couple more things form Ivana's recommendation to address. Most notably changing the layout of breadcrumbs to match those in our cloud (two layers where the description is on the one line and the value below).

I'll address that in a follow up PR

PR #1171, opened 2024-03-13
backend: flatten the structure of secrets API create and update endpo…

bojand:
…ints

Also seems some proto generate were missed.


proto/redpanda/api/dataplane/v1alpha1/secret.proto
```diff
@@ -207,8 +214,21 @@ message CreateConnectSecretRequest {
...
+    (buf.validate.field).map = {
+      values: {
+        string: {pattern: "^([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]*)$"}
+      },
+    },
+    (google.api.field_behavior) = IMMUTABLE
+  ];
+
+  // The secret data.
+  bytes secret_data = 4 [
```
weeco, line 228: Should we add validation for the max length of the secret data (seems 64KiB is enforced on AWS and GCP) for update and create?

weeco (approved):

proto/redpanda/api/dataplane/v1alpha1/secret.proto
bojand: Added.

PR #1186, opened 2024-03-20
frontend: Fix typo in topic page

comsky:
Screenshot 2024-03-20 at 8 28 01 PM
Screenshot 2024-03-20 at 8 18 35 PM

It seems like 'Bit' may be a typo of 'Bytes'.
I noticed the word "Bytes" on the "Edit retention.bytes" pop-up on the topic configuration page.


rikimaru0345 (changes requested):
"bytes" is not a valid member of `RetentionSizeUnit`
This needs more work to ensure nothing else breaks

PR #1201, opened 2024-03-28
proto: initial definition of security api

bojand:


bojand:
I will reopen this once I update a few things.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+}
+
+// DeleteRoleResponse is the response for DeleteRole.
+message DeleteRoleResponse {}
+
+// List role members for a role. That is user principals assigned to that role.
+message ListRoleMembersRequest {
+  // Filter options.
+  message Filter {
+    // Filter results to only mamber named with the prefix.
```
weeco, line 170:
```suggestion
    // Filter results to only member named with the prefix.
```

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+    string name_prefix = 1 [(buf.validate.field).string.max_len = 128];
+  }
+
+  // The role name.
+  string role = 1 [
```
weeco, line 175:
```suggestion
  string role_name = 1 [
```

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+
+// DeleteRoleResponse is the response for DeleteRole.
+message DeleteRoleResponse {}
+
+// List role members for a role. That is user principals assigned to that role.
+message ListRoleMembersRequest {
+  // Filter options.
+  message Filter {
+    // Filter results to only mamber named with the prefix.
+    string name_prefix = 1 [(buf.validate.field).string.max_len = 128];
```
weeco, line 171: Same thought here regarding `name_prefix` vs `name_contains`. Maybe we want to have both.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+    }
+  ];
+
+  // Value of the next_page_token field returned by the previous response.
+  // If not provided, the system assumes the first page is requested.
+  string page_token = 4;
+}
+
+// ListRoleMembersResponse is the response for ListRoleMembers.
+message ListRoleMembersResponse {
```
weeco, line 205: The `next_page_token` is missing here

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+  // Members assigned to the role.
+  repeated RoleMembership members = 1;
+}
+
+// RoleMembership is the role membership.
+message RoleMembership {
+  // The name of the principal assigned to the role.
+  string name = 1;
```
weeco, line 213: Again subjective but I'd prefer `principal` or `principal_name` here to show the connection to the `principal_type`

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+
+  // The principal type.
+  Security.PrincipalType principal_type = 2;
+}
+
+// UpdateRoleMembershipRequest is the request to UpdateRoleMembership.
+message UpdateRoleMembershipRequest {
+  // The role name.
+  string role = 1 [
```
weeco, line 222:
```suggestion
  string role_name = 1 [
```

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+message RoleMembership {
+  // The name of the principal assigned to the role.
+  string name = 1;
+
+  // The principal type.
+  Security.PrincipalType principal_type = 2;
+}
+
+// UpdateRoleMembershipRequest is the request to UpdateRoleMembership.
+message UpdateRoleMembershipRequest {
```
weeco, line 220: Just a thought, but the semantics for updating resources here is very different to what we do on the Kafka API for topic/broker configs for example. I assume you sticked closely to how the Redpanda API works. Personally I'd prefer consistency across endpoints for such things, but I'm unsure how well it would work out across all the upstream APIs we have to support.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+  repeated RoleMembership add = 3;
+
+  // Members to remove from the role.
+  repeated RoleMembership remove = 4;
+}
+
+// UpdateRoleMembershipResponse is the response for UpdateRoleMembership.
+message UpdateRoleMembershipResponse {
+  // The role name.
+  string role = 1 [
```
weeco, line 244:
```suggestion
  string role_name = 1 [
```

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+// ListUserRolesRequest is the request for ListUserRoles.
+message ListUserRolesRequest {
+  // Filter options.
+  message Filter {
+    // Filter results to only roles named with the prefix.
+    string name_prefix = 1 [
+      (buf.validate.field).string.max_len = 128,
+      (buf.validate.field).string.pattern = "^([a-zA-Z0-9-_]*)$"
+    ];
+  }
```
weeco, line 268: I have not looked at the Core RFC for a while, so I'm actually not sure what this endpoint does. I was unsure whether it's:

1. Return all users that belong to role x
2. Return all roles that are bound to a user x

Also wondering whether the Filter is actually optional or required here. Is the term `user` as part of the request/response name apropriate here or is it a principal?

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+    }
+  ];
+
+  // Value of the next_page_token field returned by the previous response.
+  // If not provided, the system assumes the first page is requested.
+  string page_token = 3;
+}
+
+// ListUserRolesResponse is the response for ListUserRoles.
+message ListUserRolesResponse {
```
weeco, line 292: Missing `next_page_token`

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+  message Filter {
+    // Filter results only roles named with the prefix.
+    string name_prefix = 1 [
+      (buf.validate.field).string.max_len = 128,
+      (buf.validate.field).string.pattern = "^([a-zA-Z0-9-_]*)$"
+    ];
+  }
+
+  // Optional filter.
+  optional Filter filter = 1;
```
weeco, line 309: If no filter is provided we will list all roles with their members? I assume that each role may have thousands of members potentially = problematic for pagination? How do we offer pagination here? One item = one role with all members? Seems massive, is this done in the Redpanda Admin API the same way?

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+    }
+  ];
+
+  // Value of the next_page_token field returned by the previous response.
+  // If not provided, the system assumes the first page is requested.
+  string page_token = 3;
+}
+
+// RoleWithMembers is role data with members assigned to the role.
+message RoleWithMembers {
```
weeco, line 330: Missing `next_page_token`

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+    PRINCIPAL_TYPE_GROUP = 2;
+    PRINCIPAL_TYPE_REDPANDA_ROLE = 3;
+  }
+}
+
+// Role defienes a role in the system.
+message Role {
+  // The name of the role.
+  string name = 1;
+}
```
weeco, line 23: Wondering whether this type is actually worth it. As of today it's just one property and we have some responses (e.g. `GetRoleResponse`) where we just return a string instead of a `Role`.

proto/redpanda/api/console/v1alpha1/security.proto
bojand: The frontend expressed the need for a call that lists roles _and_ the role members.
I share the concern about not being able to control the size of individual list of members.
@rikimaru0345 can share more details and perhaps what's needed and how this would be used and how we can control response size.

proto/redpanda/api/console/v1alpha1/security.proto
bojand: There is a comment on the actual RPC definition.
```
// ListUserRoles lists all the authenticated user's roles based on optional filter.
```

Maybe the RPC can be renamed to be clearer?

The filter is for filtering the list roles in the response.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+
+// ListRolesRequest is the request for ListRoles.
+message ListRolesRequest {
+  // Filter options.
+  message Filter {
+    // Filter results only roles named with the prefix.
+    string name_prefix = 1 [
```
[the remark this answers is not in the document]
bojand: The `name_prefix` parameter is exposed in the Admin API, so I included it here.

I have also added `name_containes` filter parameter to all the filters here too.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+      (buf.validate.field).string.max_len = 128,
+      (buf.validate.field).string.pattern = "^([a-zA-Z0-9-_]*)$"
+    ];
+
+    // Return only roles assigned to this principal.
+    optional string principal = 2 [(buf.validate.field).string.max_len = 128];
+
+    // Return only roles of this principal type.
+    // Qualifies parameter principal has no effect on its own.
+    optional Security.PrincipalType principal_type = 3;
```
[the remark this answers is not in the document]
bojand: I think generally it's considered best practice that if a field is optional semantically or conceptually to set it as such in the type as well. It communicated intent that it is not required. Furthermore it can be more performant as optional fields _when not set_ are not serialized and sent over the wire, and not parsed / deserialized on the other end.

proto/redpanda/api/console/v1alpha1/security.proto
bojand: Hmm I am not sure if RFC got updated, or perhaps my definition here was incorrect from start, but `GetRoleResponse` should return the role and all members. I have updated the type to reflect that.
Although we don't adhere to it strictly perhaps, it's considered best practice for requests and responses not to have primitive types, except for ids and tokens, but to contain the "resource object" as a field. I have changed the API to reflect that.

proto/redpanda/api/console/v1alpha1/security.proto
weeco: The frontend should not necessarily be the driver for the API. Keep in mind that we want to expose these APIs in the future. Also if one request may multiply many requests towards the backend, we need to be **very careful** - I want to avoid this as much as possible. Otherwise we run into a bunch of problems in regards to performance and effective rate limiting.

proto/redpanda/api/console/v1alpha1/security.proto
weeco: > lists all the **authenticated** user's

Authenticated user is the Console superuser here? That endpoint wouldn't help us in Console as of today? Why would we expose an endpoint that returns the Console user's role?

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+// ListRolesWithMembersResponse is the response for ListRolesWithMembers.
+message ListRolesWithMembersResponse {
+  // The roles in the system.
+  repeated RoleWithMembers roles = 1;
+}
+
+service SecurityService {
+  // ListRoles lists all the roles based on optional filter.
+  rpc ListRoles(ListRolesRequest) returns (ListRolesResponse) {
+    option (google.api.http) = {get: "/v1alpha1/security/roles"};
```
weeco, line 374: I would remove all HTTP annotations since these endpoints are just for Console only as of today, right? I imagine the `/security` prefix does not align with the existing structure of the endpoints.

bojand: Removed all `google.http.api` and grpc annotations.

proto/redpanda/api/console/v1alpha1/security.proto
bojand: I remove `ListUserRoles` rpc.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+  // Filter options.
+  message Filter {
+    // Filter results only roles named with the prefix.
+    string name_prefix = 1 [
+      (buf.validate.field).string.max_len = 128,
+      (buf.validate.field).string.pattern = "^([a-zA-Z0-9-_]*)$"
+    ];
+
+    // Return only roles assigned to this principal.
+    optional string principal = 2 [(buf.validate.field).string.max_len = 128];
```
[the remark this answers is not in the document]
bojand: I have removed the principal type to align with ACL API and rpk.

proto/redpanda/api/console/v1alpha1/security.proto
bojand: I have removed `ListRolesWithMembers` rpc.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,239 @@
...
+    ];
+
+    // Filter results to only roles with names which contain the string.
+    string name_contains = 2 [
+      (buf.validate.field).string.max_len = 128,
+      (buf.validate.field).string.pattern = "^([a-zA-Z0-9-_]*)$"
+    ];
+
+    // Return only roles assigned to this principal.
+    optional string principal = 3 [(buf.validate.field).string.max_len = 128];
```
weeco, line 35: Reviving the discussion regarding the `optional` keyword here. I think similar to how no optional keyword was used for `name_prefix` and `name_contains` we shouldn't need to apply it here either. 

It's not possible to have role with an empty principal, so for this list request we can assume that `principal = ""` means we don't want to apply a filter on the principal name.

We follow Google's [AIP-149](https://google.aip.dev/149)

Quote:
> Services defined in protocol buffers should use the optional keyword for primitives if and only if it is necessary to distinguish setting the field to its default value (0, false, or empty string) from not setting it at all

bojand: I have removal the optional property from that field.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,751 @@
...
+  string role = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.min_len = 1,
+    (buf.validate.field).string.max_len = 128,
+    (buf.validate.field).string.pattern = "^[a-zA-Z0-9-_]+$"
+  ];
+
+  // Whether to delete the ACLs bound to the role.
+  optional bool delete_acls = 2;
```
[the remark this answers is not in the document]
weeco: cc @bojand

proto/redpanda/api/console/v1alpha1/security.proto
weeco: @bojand

proto/redpanda/api/console/v1alpha1/security.proto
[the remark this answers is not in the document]
oleiman: Ah. Under the hood a role is more like:
- Some name (in the case of your example, `x`)
- A set of members (i.e. user principals)

^^ that's what you're interacting with over the admin API

Then way over in the authorization subsystem there are ACLs. Some ACLs might be assigned to `RedpandaRole:x`. If a request comes in from `User:martin`, we check whether that principal is a member of role `x`. If so, the ACLs from role `x` are applied to the request (along with any ACLs assigned directly to `User:martin`.

In that sense a "Role" resource (from the Admin API view) is more analogous to like a SASL/SCRAM user, where the _identity_ is managed through Admin and the _permissions_ are managed through Kafka/acl.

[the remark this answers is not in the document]
bojand: We create a role `foo`.
Then we can create users, `Bob` and `Alice` and assign them to the role `foo`.
Then we create the ACL with "allow all" operations (for example), and we allow principal `"RedpandaRole:foo"`.
Then when we delete role `foo` we can set the `delete_acl` flag to `true` to also delete the ACL we just created.

proto/redpanda/api/console/v1alpha1/security.proto
weeco: Okay let's remain close to what we get from upstream and we'll revisit this (having consistent semantics across all endpoints) at some point in the future, but let's always keep it in mind when adding new API endpoints.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,236 @@
...
+message GetRoleResponse {
+  // The Role.
+  Role role = 1;
+
+  // Members assigned to the role.
+  repeated RoleMembership members = 2;
+}
+
+// UpdateRoleRequest is the request for UpdateRole.
+message UpdateRoleRequest {
```
weeco, line 94: You to update a role you provide the role_name and the role of type `role`. Type `role` however only contains a name. Is this supposed to be a role rename then only?

proto/redpanda/api/console/v1alpha1/security.proto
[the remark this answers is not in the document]
weeco: Got it, I thought we were using the Kafka ACL API to create roles, but apparently we use the Redpanda admin api which allows that then. 

I think we should remove the `optional` keyword here.

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,236 @@
...
+    lte: 1000
+  }];
+
+  // Value of the next_page_token field returned by the previous response.
+  // If not provided, the system assumes the first page is requested.
+  string page_token = 4;
+}
+
+// ListRoleMembersResponse is the response for ListRoleMembers.
+message ListRoleMembersResponse {
```
weeco, line 161: Wondering if it's worth to include the role name here again?

proto/redpanda/api/console/v1alpha1/security.proto
```diff
@@ -0,0 +1,236 @@
...
+  string role_name = 1 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.min_len = 1,
+    (buf.validate.field).string.max_len = 128,
+    (buf.validate.field).string.pattern = "^[a-zA-Z0-9-_]+$"
+  ];
+
+  // Create the role if it doesn't already exist.
+  // If the role is created in this way, the “add” list will be respected, but the “remove” list will be ignored.
+  optional bool create = 2;
```
weeco, line 187: Remove the optional?

proto/redpanda/api/console/v1alpha1/security.proto
bojand: Correct, currently a role only has a single property, name. So we need the existing role name, provided via `role_name` which could be used to map to a path param for HTTP API should we need to in the future. Then the content for the update is in the `Role role` field.

proto/redpanda/api/console/v1alpha1/security.proto
bojand: Added `string role_name = 1;` to the response message.

proto/redpanda/api/console/v1alpha1/security.proto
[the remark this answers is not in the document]
bojand: Removed the `optional` keyword.

weeco (approved):

rikimaru0345 (approved):

PR #1202, opened 2024-03-29
frontend: update connectors image in e2e tests

bojand:


frontend/tests/config/docker-compose.yaml
```diff
@@ -68,7 +68,7 @@ services:
       - redpanda
   
   connect:
-    image: docker.cloudsmith.io/redpanda/cloudv2-dev/connectors:v1.0.0-8cf37ab
+    image: docker.cloudsmith.io/redpanda/connectors-unsupported/connectors:v1.0.0-5014b61
```
weeco, line 71: I read that they changed to rolling images where they keep only the last three images. This is why I had to change to the latest tag in the different PR after the previously used image tag got deleted. I'd prefer hashes as well, but it doesn't seem an option anymore :(

bojand: Yup I had seen discussions about reducing number of images and retention time. Had debated whether to use latest or an actual tag. 

I have changed the PR code to use `latest` tag.

weeco (approved):

PR #1210, opened 2024-04-11
Backend/security api local adminapi

bojand:
Creating a draft PR to get Docker builds.


backend/pkg/api/connect/service/security/service.go
```diff
@@ -0,0 +1,332 @@
...
+			if restErr.Err != nil {
+				err = restErr.Err
+			}
+			return nil, apierrors.NewConnectError(
+				connect.CodePermissionDenied,
+				err,
+				apierrors.NewErrorInfo(commonv1alpha1.Reason_REASON_PERMISSION_DENIED.String()),
+			)
+		}
+	}
```
weeco, line 314: This is very hard to read and more nested than necessary. In fact it's not handling the case where restErr is nil & canCreateRoles = false. 

This would be much quicker to digest (even though a bit more verbose). FWIW the `restErr` will only be set if there's an issue such as the user session not being set in the request context. It's more like an internal server error, but we can still reject the request as permission denied in that case.

```go
	// Check if logged in user is allowed to create roles for the given request
	canCreateRoles, restErr := s.authHooks.CanCreateKafkaRoles(ctx)
	if restErr != nil {
	      return nil, apierrors.NewConnectError(
		      connect.CodePermissionDenied,
		      // The returned `restErr` is in our control, so whether or not Err is set
		      // should be in our control. It should always be set!
		      restErr.Err,
		      apierrors.NewErrorInfo(commonv1alpha1.Reason_REASON_PERMISSION_DENIED.String()),
	      )
	}

	if !canCreateRoles {
	    return nil, apierrors.NewConnectError(
		      connect.CodePermissionDenied,
		      errors.New("You are not allowed to create roles")
		      apierrors.NewErrorInfo(commonv1alpha1.Reason_REASON_PERMISSION_DENIED.String()),
	      )
	}
```

backend/pkg/api/connect/service/security/service.go
```diff
@@ -0,0 +1,332 @@
...
```
weeco, line 79: Let's refactor similarly like the other case.

backend/pkg/api/connect/service/security/service.go
```diff
@@ -0,0 +1,332 @@
...
```
weeco, line 139: Let's refactor similarly like the other case.

backend/pkg/api/connect/service/security/service.go
```diff
@@ -0,0 +1,332 @@
...
```
weeco, line 171: Let's refactor similarly like the other case.

backend/pkg/api/connect/service/security/service.go
```diff
@@ -0,0 +1,332 @@
...
```
weeco, line 204: Let's refactor similarly like the other case.

backend/pkg/api/connect/service/security/service.go
```diff
@@ -0,0 +1,332 @@
...
```
weeco, line 244: Let's refactor similarly like the other case.

PR #1215, opened 2024-04-15
backend: use commong-go rpadmin admin api module

bojand:


backend/go.mod
```diff
@@ -35,7 +35,8 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/prometheus/client_golang v1.19.0
 	github.com/redpanda-data/common-go/api v0.0.0-20240410132648-06b2f85d3d52
-	github.com/redpanda-data/redpanda/src/go/rpk v0.0.0-20240105044330-c094966ca0cf
+	github.com/redpanda-data/common-go/net v0.1.0
+	github.com/redpanda-data/common-go/rpadmin v0.1.0
```
bojand, line 39: For now I published this as `v0.1.0` until we use it and get a little bit of soak time.
We can decide how we want to publish it later. I think following redpanda versioning would be cumbersome and impractical, even if useful in some ways? There is also room for some improvements later, so having a < 1.0.0 version would probably be convenient for now.

weeco: The current rpk module was not versioned at all and hasn't caused us issues so far, so not sure if we even need versioned releases.

weeco (approved):

backend/go.mod
bojand: I'll keep in mind and test it... maybe I was doing something wrong..., or maybe it has to do with the monorepo multi-module setup, but I had trouble bringing in the new net package w/o tagging it first.

PR #1230, opened 2024-04-24
Feat/rbac users renamed to principals

jvorcak:


frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -163,19 +163,27 @@ const UsersTab = observer(() => {
                 <Button variant="outline"
                     isDisabled={!Features.createUser}
                     onClick={() => appGlobal.history.push('/security/users/create')}>
-                    Create user
+                    Create Principal
```
rikimaru0345, line 166: this should still be "create user" a prinipal is something abstract.
```suggestion
                    Create User
```

rikimaru0345 (changes requested):

frontend/src/components/pages/acls/Acl.List.tsx
jvorcak: Screenshot 2024-04-24 at 14 24 39

What confuses me how it's called in an empty action in a table.

rikimaru0345 (approved):

frontend/src/components/pages/acls/Acl.List.tsx
ivpanda: You're right— in the empty table it should be Create User. That's what is being created: a redpanda SASL/SCRAM user.

jvorcak: @ivpanda @rikimaru0345 so I've changed the buttons to be "Create user" https://github.com/redpanda-data/console/commit/579667c2d7b525ca75934cc88f6fd06d84ae99f1

PR #1244, opened 2024-05-02
Backend: incompatible redpanda version error

bojand:
This PR adds incompatible redpanda version error to common errors.

### DIFF
--- a/backend/pkg/api/connect/errors/common_errors.go
+++ b/backend/pkg/api/connect/errors/common_errors.go
@@ -12,9 +12,8 @@ package errors
 import (
 	"errors"
 
+	commonv1alpha1 "buf.build/gen/go/redpandadata/common/protocolbuffers/go/redpanda/api/common/v1alpha1"
 	"connectrpc.com/connect"
-
-	v1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/dataplane/v1alpha1"
 )
 
 // NewRedpandaAdminAPINotConfiguredError is a standard error to return if an endpoint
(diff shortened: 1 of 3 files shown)


proto/redpanda/api/dataplane/v1alpha1/error.proto
```diff
@@ -11,4 +11,5 @@ enum Reason {
   REASON_KAFKA_CONNECT_API_ERROR = 5;
   REASON_TYPE_MAPPING_ERROR = 6;
   REASON_SECRET_STORE_ERROR = 7;
+  REASON_FEATURE_NOT_SUPPORTED = 8;
```
weeco, line 14: I think this should go into the common proto package. We already have one similar enum for features: https://buf.build/redpandadata/common/file/main:redpanda/api/common/v1alpha1/common.proto#L41

backend/pkg/api/connect/errors/common_errors.go
```diff
@@ -27,3 +27,14 @@ func NewRedpandaAdminAPINotConfiguredError() *connect.Error {
...
+// NewRedpandaFeatureNotSupportedError is a standard error to return if an endpoint
+// requires the Redpanda feature that is not supported with current running version and instance.
+func NewRedpandaFeatureNotSupportedError(feature string) *connect.Error {
+	return NewConnectError(
+		connect.CodeUnimplemented,
+		errors.New("redpanda version does not support feature "+feature),
+		NewErrorInfo(v1alpha1.Reason_REASON_FEATURE_NOT_SUPPORTED.String()),
+		NewHelp(NewHelpLinkConsoleReferenceConfig()),
+	)
+}
```
weeco, line 40: Maybe we can use a more generic error here instead? We already have something similar for Kafka API endpoints if the target cluster version does not implement a certain Kafka API version.

backend/pkg/api/connect/errors/common_errors.go
```diff
@@ -27,3 +27,14 @@ func NewRedpandaAdminAPINotConfiguredError() *connect.Error {
 		NewHelp(NewHelpLinkConsoleReferenceConfig()),
 	)
 }
+
+// NewRedpandaFeatureNotSupportedError is a standard error to return if an endpoint
+// requires the Redpanda feature that is not supported with current running version and instance.
+func NewRedpandaFeatureNotSupportedError(feature string) *connect.Error {
+	return NewConnectError(
+		connect.CodeUnimplemented,
+		errors.New("redpanda version does not support feature "+feature),
```
weeco, line 36: We could put the feature into the metadata instead. I assume it's some sort of an enum?

backend/pkg/api/connect/errors/common_errors.go
bojand: I am not sure I understand exactly what you mean by a more generic error? Can you elaborate?

weeco (approved):

PR #1253, opened 2024-05-10
backend: refresh aws creds when using iam auth after token expiration

mislavmandaricaxilis:
Currently, when AWS IAM auth fetches token, it only does so once on startup and the same token is reused. This change introduces refresh mechanism by retrieving the token asynchronously when it's needed by the kafka auth system.

Token is internally cached and reused until it expires. Once it expires, a new token is fetched and used.


twmb (approved):

tachang:
For some reason this is still not working. I am not sure exactly why but the credentials are not being pulled again after the session timeout.

Issue #232, opened 2021-07-08
Feature Request: Republishing an existing message


AtulKhanduri:
As mentioned in https://github.com/cloudhut/kowl/issues/221#issuecomment-865106925, it is very helpful to have a feature to republish the message from a topic.

So, we could have 2 options in the message menu behind the three dots - "re-publish as-is" and "re-publish edit...".

cobolbaby:
Could this feature be included in the goal of the next phase?

PR #1285, opened 2024-06-12
Add copy message to publish for Republishing an existing message

ying234:
Feature #232 
Add a copy message to publish menu item to copy the key and value to publish page.


weeco:
Republishing a record as is requires copying other attributes such as the offset, partition, headers as well to be copied. I believe just copying key/value is only a slight advantage over what's already possible by copying the key & value individually into the new publish modal.

jvorcak (approved):

jvorcak:
I've checked when it comes to code, I'll leave to @weeco to approve whether it makes sense or not to merge in this state.

ying234:
> Republishing a record as is requires copying other attributes such as the offset, partition, headers as well to be copied. I believe just copying key/value is only a slight advantage over what's already possible by copying the key & value individually into the new publish modal.

For me, I was using such copy a lot and I need to open two browser tabs. One tab is for copy and another is for paste to the input. Every time with several extra clicks. So, this change can save me from these extra clicks.

ying234:
> Republishing a record as is requires copying other attributes such as the offset, partition, headers as well to be copied. I believe just copying key/value is only a slight advantage over what's already possible by copying the key & value individually into the new publish modal.

Add message key and value encoding, and headers to the copying.

ying234:
@weeco any more suggestion or comment?

Issue #1064, opened 2024-02-01
Add an indicator that message count is an estimate on the main topic page


patrickangeles:
We already show that message count is an estimate in the Partitions tab. Would be good to add this to the main page as well. A tooltip when hovering the number (something non-intrusive) could work fine.

![image](https://github.com/redpanda-data/console/assets/135980/bc318b83-a82e-4bce-85ed-ddad7689110b)

jvorcak:
![Screenshot 2024-06-11 at 17 53 38](https://github.com/redpanda-data/console/assets/1083817/7923d599-6497-4392-a7d5-2baf29c80391)

Something like this could work?

weeco:
@jvorcak We should just change the label description from "Messages" to "Estimated Messages" - regardless of the cleanup policy. Even for cleanup.policy = delete it may be an estimate only because the cleanup policy may have been compact or because of control records which are not actual messages but would still be counted as messages. 

I assume the tooltip could be used to explain why it's an estimate only though! Maybe something along those lines:

> The number of messages shown is an estimate. This is calculated by summing the differences between the highest and lowest offsets in each partition. The actual number of messages may vary due to factors such as message deletions, log compaction, and uncommitted or transactional messages.

PR #1286, opened 2024-06-12
Information about estimation of the message count

jvorcak:
Fixes https://github.com/redpanda-data/console/issues/1064

Screenshot 2024-06-12 at 9 28 29
Screenshot 2024-06-12 at 9 28 27

I know @ivpanda wanted to use sentence case elsewhere, so I've used it here as well. Please let me know if I should change some phrasing.

### DIFF
--- a/frontend/src/components/misc/Statistic.tsx
+++ b/frontend/src/components/misc/Statistic.tsx
@@ -13,17 +13,19 @@ export function Statistic(p: {
 
     return (
         <Stat key={key} className={className} flexBasis="auto" flexGrow={0} marginRight="2rem" {...rest}>
-            <Flex gap={1}>
-                <StatNumber>{value}</StatNumber>
-                {hint && (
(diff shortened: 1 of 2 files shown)


weeco:
I think the info tooltip behind the number looks weird. I'd either remove it and make it hoverable on the label (Estimated messages) or put the tooltip there instead. Personally I think info icons irritate too much and the explanation is not super important to learn about.

rikimaru0345:
The code itself looks fine from my side.

As for the text itself, or icon placement: you two guys will need to figure that out.

jvorcak:
@weeco I see your point, but on the other hand, it's good to have some clue that there is an additional info located somewhere. Otherwise it'll give me a feeling I need to hover over other stuff just to learn more. But I'm open to take whatever approach. WDYT @ivpanda?

jvorcak:
Screenshot 2024-06-19 at 19 44 17

PR #1290, opened 2024-06-18
proto: add pipeline management API

simon0191:
Add the pipeline management API proto following the conventions from the cloud public API.

### Implementation notes
1. `Pipeline.ConfigYAML` has `YAML` as a suffix so that (1) it's explicit that the configuration should be provided as YAML, and (2) allow us to replace it in the future with an strongly typed `config` field.
2. About the state machine
   - It models a pipeline as a long running process which can be stopped and restarted
   - There's no distinction between Degraded and Error state. IMO it might be hard to differentiate, so I propose we simply use `Error` and provide more details in other fields like the `Pipeline.Statis.Error` field. If more details are needed we can add more fields to the `status`.
3. The API follows the conventions of the Cloud Public API
   - Dedicated messages for the payloads of the create and update requests. This is to allow us to have different validation rules for each case. For example, on creation the spec is required, but for update it's not.
4. About resources
   - The request and limit fields were modelled closely following k8s resource requests
   - `Pipeline.Request` field is the minimum amount of resources that the pipeline needs to be running. The underlying scheduler should guarantee the availability of those resources.
   - `Pipeline.Limit` is the maximum amount of resources to be allocated to the pipeline

### Pipeline management API methods

- Create
- Get
- Update
- List
- Start
- Stop
- Delete

### Pipeline state machine

```mermaid
stateDiagram-v2
    [*] --> Starting
    Starting --> Running
    Running --> Stopping
    Stopping --> Stopped
    Running --> Error
    Error --> Stopped
    Error --> Running
    Stopped --> Starting
    Stopped --> [*]
```

## Refs
- Depends on #1293
- Depends on #1295


proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+  Resources limit = 6 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The maximum amount of resources that should be assigned to the pipeline."}];
+
+  Status status = 7 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  message Spec {}
+
+  message Resources {
+    string memory_shares = 1;
+    string cpu_shares = 2;
+  }
```
weeco, line 52: Would be good to have a comment on what the expected values look like. I assume something like `128Mi` or `1Gi` will work and `100m` or any integer for cpu_shares.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+  // The maximum amount of resources that should be assigned to the pipeline
+  Resources limit = 6 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The maximum amount of resources that should be assigned to the pipeline."}];
+
+  Status status = 7 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  message Spec {}
+
+  message Resources {
+    string memory_shares = 1;
+    string cpu_shares = 2;
```
weeco, line 51: Should each of these be required inputs or are we going to set some default?

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+
+package redpanda.api.dataplane.v1alpha1;
+
+import "buf/validate/validate.proto";
+import "google/api/field_behavior.proto";
+import "google/protobuf/field_mask.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+
+message Pipeline {
+  string id = 1 [
```
weeco, line 11: Can we restrict the character set here as well? Ideally to something that's natively supported by SpiceDB as well. Similar naming constraints as we have for Kafka topic names would make sense to me.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.len = 20
```
weeco, line 14: Where is the id coming from that it's required to be 20chars I wonder? I thought it could be an arbitrary id specified by the user. For example `my-snowflake-sink-connector`

tomasz-sadura (approved):
LGTM

bochenekmartin (approved):

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+    STATE_ERROR = 5;
+  }
+
+  message Status {
+    State state = 1;
+    string error = 2;
+  }
+}
+
+service PipelineService {
```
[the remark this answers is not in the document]
simon0191: Because my understanding is that we're following a [resource-oriented design](https://google.aip.dev/121) where each resource has its own service.

RedpandaConnectService would be too broad and doesn't refer to a single resource.

Instead if we want to manage more resources we should create extra services for each resource. For example `SecretService`.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
simon0191: The ID should be generated by the backend that implements the service, that's why the create method doesn't receive the ID.
I'll clarify in a comment.

weeco: Okay that makes sense to me. Maybe worth to add what kind of id it is supposed to be. I assume you had something specific in mind since you restricted the length to 20 chars.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+
+  string display_name = 2 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_ /]+$",
+    (buf.validate.field).string.min_len = 3,
+    (buf.validate.field).string.max_len = 128
+  ];
+
+  string description = 3 [(buf.validate.field).string.max_len = 1000];
```
weeco, line 25: Being curious: Where do we store this in the case of cloud? I imagine we have some limits there. Possibly worth to restricting it to less characters (256 or so)? We need to find something that supported across all our cloud providers.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Limit the paginated response to a number of items. Defaults to 100. Use -1 to disable pagination.",
+      minimum: -1,
+      maximum: 1000
+    }
+  ];
+  string page_token = 3;
+
+  message Filter {
+    optional string display_name = 1;
```
weeco, line 136:
```suggestion
    string display_name = 1;
```

NIT: I think we don't need to differentiate between *string and string here no? `display_name` is a required field and searching for `display_name=""` doesn't make sense, so we can assume this as being unset. No strong opinion on this.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,212 @@
...
+    STATE_UNSPECIFIED = 0;
+    STATE_STARTING = 1;
+    STATE_RUNNING = 2;
+    STATE_STOPPING = 3;
+    STATE_STOPPED = 4;
+    STATE_ERROR = 5;
+  }
+
+  message Status {
+    State state = 1;
```
birdayz, line 64: is there a specific reason to have status inside status, and not as a top-level field? we have it as top-level field pretty much anywhere else.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
birdayz: This comes from:

See https://google.aip.dev/149

```
Important: Services should not need to distinguish between the default value and unset most of the time; if an alternative design does not require such a distinction, it is usually preferred. In practice, this means optional should only ever be used for integers and floats.
```

It's also done like this here: https://github.com/redpanda-data/cloudv2/blob/a00cbf359bff86374c23aa36ab6d90aa28fccd89/proto/public/cloud/redpanda/api/controlplane/v1beta2/cluster.proto#L1296

empty string is not a legit value, so it's preferable to use it for the field presence check.

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
simon0191: Good point. This will be stored in k8s

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
simon0191: Done

bojand: Should this be `IMMUTABLE` and `OUTPUT_ONLY` ?

weeco (approved):

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,240 @@
...
+  // See https://docs.redpanda.com/redpanda-connect/configuration/about/
+  string config_yaml = 4 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The configuration of the Pipeline as YAML. See https://docs.redpanda.com/redpanda-connect/configuration/about for more details."}
+  ];
+
+  // The minimum amount of resources needed for the pipeline to run
+  Resources request = 5 [
+    (google.api.field_behavior) = REQUIRED,
```
bojand, line 37: Just curious... Is this required? Or can we have some sane default based on like tier or config (number of input, processors, etc..) or something?

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,240 @@
...
+
+  string display_name = 2 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_ /]+$",
+    (buf.validate.field).string.min_len = 3,
+    (buf.validate.field).string.max_len = 128
+  ];
+
+  string description = 3 [(buf.validate.field).string.max_len = 256];
```
bojand, line 25: max len in `PipelineUpdate` and `PipelineCreate` is `1000` and the response messages to those calls has a `Pipeline` message which has a max len of `256`?

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,240 @@
...
+    STATE_STOPPED = 4;
+    STATE_ERROR = 5;
+  }
+
+  message Status {
+    string error = 2;
+  }
+}
+
+service PipelineService {
```
bojand, line 97: Should we have HTTP / gRPC gateway option definitions for all of these?

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
```diff
@@ -0,0 +1,240 @@
...
+  ];
+  string description = 2 [(buf.validate.field).string.max_len = 1000];
+
+  string config_yaml = 3 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
+  ];
+  Pipeline.Resources request = 4 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
```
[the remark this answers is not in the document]
simon0191: yeah, good point. I'll make it not required and the backend can set a default

proto/redpanda/api/dataplane/v1alpha1/pipeline.proto
simon0191: Good catch

weeco (approved):

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -0,0 +1,386 @@
...
+  ];
+  string description = 2 [(buf.validate.field).string.max_len = 256];
+
+  string config_yaml = 3 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
+  ];
+  Pipeline.Resources request = 4 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
```
bojand, line 269: Do we want to remove this required too?

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -0,0 +1,386 @@
...
```
bojand, line 335: Do we want to remove this required too?

simon0191: Good catch. In updates everything should be optional.

bojand (approved):

Issue #1222, opened 2024-04-22
Exponential backoff retries logic never kicks in if redpanda is down


vncorpacc:
We upgraded from version 2.3.5 to a version 2.4.6 and noticed this regression.

On version `2.3.5` we relied on exponential backoff retries (configured to Infinite) for our environment restart operations in case `redpanda-console` starts up before redpanda. Those worked properly as coded in `kafka.NewService`.

Apparently this commit: https://github.com/redpanda-data/console/commit/d06eaa34d47e117a74905adf092ba347c6249fe0
breaks this behaviour by attempting a "admin client connectivity" check in `redpanda.NewService` without any retries. Failure to establish it is fatal.

bojand:
Should be addressed in latest [unstable builds](https://hub.docker.com/r/redpandadata/console-unstable/tags) and will be in the next release.

Similar to Kafka backoff options, now Redpanda Admin API config can include options for dictating the startup behaviour:

```yaml
redpanda:
  adminApi:
    enabled: true
    urls: ["http://localhost:19644"]
    startup:
      establishConnectionEagerly: true # false to skip the startup test
      maxRetries: 5
      retryInterval: 1s
      maxRetryInterval: 60s
      backoffMultiplier: 2
```

PR #1294, opened 2024-06-19
backend: add startup options to redpanda admin api config

bojand:
Addresses https://github.com/redpanda-data/console/issues/1222.

This adds similar startup options to Redpanda Admin API config, allowing to specify backoff.

Since we need cluster version there is no EstablishConnectionEagerly option to turn off connection test. That means if Admin API is enabled, we require the test to succeed at some point. Here's a filing scenario:

```sh
{"level":"info","ts":"2024-06-19T15:25:09.888-0300","msg":"started Redpanda Console","version":"development","built_at":"<not set>"}
{"level":"info","ts":"2024-06-19T15:25:09.888-0300","msg":"testing admin client connectivity","urls":["http://localhost:19634"]}
{"level":"warn","ts":"2024-06-19T15:25:12.992-0300","msg":"Failed to test Redpanda Admin connection, going to retry in 1s","remaining_retries":5}
{"level":"info","ts":"2024-06-19T15:25:13.993-0300","msg":"testing admin client connectivity","urls":["http://localhost:19634"]}
{"level":"warn","ts":"2024-06-19T15:25:17.136-0300","msg":"Failed to test Redpanda Admin connection, going to retry in 2s","remaining_retries":4}
{"level":"info","ts":"2024-06-19T15:25:19.136-0300","msg":"testing admin client connectivity","urls":["http://localhost:19634"]}
{"level":"warn","ts":"2024-06-19T15:25:22.226-0300","msg":"Failed to test Redpanda Admin connection, going to retry in 4s","remaining_retries":3}
{"level":"info","ts":"2024-06-19T15:25:26.227-0300","msg":"testing admin client connectivity","urls":["http://localhost:19634"]}
{"level":"warn","ts":"2024-06-19T15:25:29.353-0300","msg":"Failed to test Redpanda Admin connection, going to retry in 8s","remaining_retries":2}
{"level":"info","ts":"2024-06-19T15:25:37.354-0300","msg":"testing admin client connectivity","urls":["http://localhost:19634"]}
{"level":"warn","ts":"2024-06-19T15:25:40.500-0300","msg":"Failed to test Redpanda Admin connection, going to retry in 16s","remaining_retries":1}
{"level":"fatal","ts":"2024-06-19T15:25:56.502-0300","msg":"failed to create Redpanda service","error":"failed to test kafka connection: failed to test admin client connectivity: Get \"http://localhost:19634/v1/brokers\": dial tcp 127.0.0.1:19634: connect: connection refused"}
```

### DIFF
--- a/backend/pkg/redpanda/service.go
+++ b/backend/pkg/redpanda/service.go
@@ -20,6 +20,7 @@ import (
 	adminapi "github.com/redpanda-data/common-go/rpadmin"
 	"go.uber.org/zap"
 
+	"github.com/redpanda-data/console/backend/pkg/backoff"
 	"github.com/redpanda-data/console/backend/pkg/config"
 )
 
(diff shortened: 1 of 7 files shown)


backend/pkg/redpanda/service.go
```diff
@@ -66,26 +67,65 @@ func NewService(cfg config.Redpanda, logger *zap.Logger) (*Service, error) {
 		return nil, fmt.Errorf("failed to create admin client: %w", err)
 	}
 
+	// Ensure Kafka connection works, otherwise fail fast. Allow up to 5 retries with exponentially increasing backoff.
```
weeco, line 70:
```suggestion
	// Ensure Redpanda connection works, otherwise fail fast. Allow up to 5 retries with exponentially increasing backoff.
```

backend/pkg/redpanda/service.go
```diff
@@ -66,26 +67,65 @@ func NewService(cfg config.Redpanda, logger *zap.Logger) (*Service, error) {
 		return nil, fmt.Errorf("failed to create admin client: %w", err)
 	}
```
weeco, line 69: We are not checking for `EstablishConnectionEagerly` here. Maybe add a check and return early if it's disabled so that we won't test?

weeco (commented):
LGTM, eventually we need to do the same for the schema registry as well

backend/pkg/redpanda/service.go
bojand: I am not sure I understand..? There is a check if admin API is enabled [at the start of `NewService()` method](https://github.com/redpanda-data/console/blob/master/backend/pkg/redpanda/service.go#L36).

weeco: But you introduced a new property `EstablishConnectionEagerly` which we had for the Kafka service as well. If this is enabled (default) we'll check if we can reach the API on startup. If we can't we will fail the startup process. This property is not used anywhere yet as far as I can see.

bojand: `EstablishConnectionEagerly` is already present in `KafkaStartup` config, along with other backoff config properties. What I did was refactor these backoff properties into `ServiceStartupAttemptsOptions` config struct. Then `KafkaStartup` embeds the new backoff struct, along with `EstablishConnectionEagerly` property.
For `RedpandaAdminAPI` config struct I added just the new `ServiceStartupAttemptsOptions` as `startup` property. So the Admin API startup does not have `EstablishConnectionEagerly` because Admin API connection is kinda required on startup. We can't do it eagerly or lazily. Also I feel like `EstablishConnectionEagerly` doesn't belong in backoff options, but for scenarios where it is reasonable, we can just embed the backoff options like I did for Kafka service. Hope this makes sense.

weeco: Ah correct, it's only in the Kafka config. Why do you think the admin api is required in contrast to the Kafka API? Clusters can totally work without any redpanda admin api connection at all, so I think it's actually less required than the Kafka API. I think we should add this property too, so that Console can also start (if desired) if the admin api is currently unreachable for whatever reason.

bojand: We were saving the cluster version into the redpanda service struct... I had thought it was needed for something like endpoint compatibility checks... Upon further inspection we were not using the cluster version at all. I think I had added this previously when I was working on RBAC compatibility checks; but it was not needed. I have removed the cluster version from the struct, and added `EstablishConnectionEagerly` to the new generic `ServiceStartupAttemptsOptions`, which now is essentially the same as `KafkaStartup` was. `KafkaStartup` is removed and both Kafka and Redpanda configs now use `ServiceStartupAttemptsOptions`. In Redpanda service we perform the get cluster version connection check only if `EstablishConnectionEagerly` is `true` similar to the Kafka Service.

PR #1304, opened 2024-06-26
Better message detail

jvorcak:
Background will be part of https://github.com/redpanda-data/console-enterprise/issues/382


weeco:
@jvorcak Can we add the key/value size as part of tab menu titles as well as the number of headers for the headers?

weeco (approved):

rikimaru0345 (approved):

PR #1305, opened 2024-06-26
New statistics in a topic detail

jvorcak:
Fixes https://github.com/redpanda-data/console-enterprise/issues/381
Screenshot 2024-07-02 at 21 57 10
Screenshot 2024-07-02 at 21 56 53

### DIFF
--- a/frontend/src/components/pages/topics/QuickInfo.tsx
+++ b/frontend/src/components/pages/topics/QuickInfo.tsx
@@ -15,44 +15,32 @@ import { api } from '../../../state/backendApi';
 import '../../../utils/arrayExtensions';
 import { prettyBytesOrNA } from '../../../utils/utils';
 import { formatConfigValue } from '../../../utils/formatters/ConfigValueFormatter';
-import { Flex } from '@redpanda-data/ui';
-import { Statistic } from '../../misc/Statistic';
-
-const StatsSeparator = () => <div style={{ width: '1px', background: '#8883', margin: '0 1.5rem', marginLeft: 0 }} />
-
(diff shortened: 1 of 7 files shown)


frontend/src/components/pages/topics/QuickInfo.tsx
```diff
@@ -15,74 +15,104 @@ import { api } from '../../../state/backendApi';
...
+        <Flex gap={2}>
+          {cleanupPolicy==='compact' && <><Text as="dt" fontWeight="bold">Segment:</Text>
+              {segmentMs && segmentBytes && <Text as="dd">~
+                  <>{formatConfigValue(segmentMs.name, segmentMs.value, 'friendly')} or {formatConfigValue(segmentBytes.name, segmentBytes.value, 'friendly')}</>
+              </Text>}
+          </>}
+
+          {cleanupPolicy==='delete' && <><Text as="dt" fontWeight="bold">Retention:</Text>
+            {retentionMs && retentionBytes && <Text as="dd">~
+                {retentionMs.value === '-1' && retentionBytes.value === '-1' ? 'Unlimited':
```
jvorcak, line 109: @weeco @rikimaru0345 do you know how it's with the type here? TS says it's always string or undefined, but in the ticket it says it's -1 like a number.

weeco: `retention.ms` is a long with valid values between -1 and max long. The same should apply to retention.bytes

jvorcak: I think in that case we need to improve the typings, but it's probably out of scope of this PR

rikimaru0345 (approved):

Issue #1220, opened 2024-04-17
Change ACL modification mechanism in Console


jason-da-redpanda:
Currently in console when ACL's are modified,  we delete the existing ones..   then recreate with existing acls + additions.

We have seen this can cause problems on clients ...

Have been able to reproduce consistently with following steps  : 

- In Console configure acls for a user-A.  . give permisions  for  topic-1, + consumergroup1.   (then repeat with topic2, consumer group2 ... up to 5 sets of resources for topic + group) [Save/ok]

- Start java consumer. Confirm it's  reading from topic1 , Group 1 running..  + no issues with authorisation.

- In console Amend ACL's  for user-A..    Give additional permissions for  topic6 + consumergroup6.   [Save/ok]

- Java Consumer now fails with  auth errors
        Java log :
`[com.redpanda.ConsumerExample.main()] WARN org.apache.kafka.clients.consumer.internals.Fetcher - [Consumer clientId=consumer-firefox-1, groupId=firefox] Not authorized to read from partition topic-1-0.
org.apache.kafka.common.errors.TopicAuthorizationException: Not authorized`

       Redpanda Reports :
`2024-04-16 11:43:10.200REDPANDA redpanda INFO 2024-04-16 11:43:10,200 [shard 0:fetc] kafka - 31.54.228.122:62370 failed authorization - connection_context.cc:179 - proto: kafka rpc protocol, sasl state: complete, acl op: read, principal: type {user} name {redpanda-chat-account}, resource: {topic1} `

- Restart java consumer no issues with ACLS / authorisation

- The ACL's are as expected in consol/rpk acl list...e.g the additional topic/group was added we didnt lose any

**Requested Feature change :** 
Either :
1. change the UI for ACLs significantly
2. make console smarter so that it can do better diffs without removing

PR #1320, opened 2024-07-09
Avoid deleting ACLs when editing

jvorcak:
Fixes https://github.com/redpanda-data/console/issues/1220

Please try to test it yourself as well, I tested on my setup, but I'd appreciate if somebody could double-check this.


frontend/package.json
```diff
@@ -39,6 +39,7 @@
         "array-move": "^4",
         "framer-motion": "^7",
         "js-base64": "^3.7.5",
+        "lodash-es": "^4.17.21",
```
rikimaru0345, line 42: Not worth adding a dependency for just isEqual.

There are 3 options that are better:

You can either iterate over `Object.entries` and compare property by property, or use mobx `comparer.structural` (which essentially does the same thing, but also nesting), or simply write a custom isEqual function,

rikimaru0345 (changes requested):

rikimaru0345:
How does this prevent that ACLs that existed before (and don't need to be modified) are not created again? or is that not an issue? if not, why?

jvorcak:
If they don't need to be modified, we remote then from `allToCreate` array thus they're not created.

frontend/package.json
jvorcak: @rikimaru0345 done

frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
```diff
@@ -42,22 +51,29 @@ export const AclPrincipalGroupEditor = observer((p: {
...
-                else throw new Error('No targeted resources. You can delete this ACL group from the list view.');
+            if (allToCreate.length == 0) {
+                if (p.type == 'create') {
+                    throw new Error('Creating an ACL group requires at least one resource to be targeted. Topic/Group targets with an empty selector are not valid.');
+                } else {
+                    throw new Error('No targeted resources. You can delete this ACL group from the list view.');
+                }
+            }
 
             // Delete all ACLs in group
```
rikimaru0345, line 62:
```suggestion
            // Ignore creation of ACLs that already exist, and delete ACLs that are no longer needed
```

frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
```diff
@@ -42,22 +51,29 @@ export const AclPrincipalGroupEditor = observer((p: {
...
-                        host: group.host,
-                        operation: 'Any',
-                        permissionType: 'Any'
-                    });
+                if (group.sourceEntries.length > 0) {
+                    const requests = group.sourceEntries.map(acl => {
+                        // try to find this in allToCreate
+                        const foundIdx = allToCreate.findIndex(x => comparer.structural(acl, x))
+                        if(foundIdx !== -1) {
+                            allToCreate.splice(foundIdx, 1)
```
rikimaru0345, line 69:
```suggestion
                            // acl already exists, remove it from the list
                            allToCreate.splice(foundIdx, 1)
```

frontend/src/components/pages/acls/PrincipalGroupEditor.tsx
```diff
@@ -42,22 +51,29 @@ export const AclPrincipalGroupEditor = observer((p: {
...
+                            return Promise.resolve()
+                        } else {
+                            return api.deleteACLs(acl)
```
rikimaru0345, line 72:
```suggestion
                            // acl should no longer exist, delete it
                            return api.deleteACLs(acl)
```

rikimaru0345 (approved):

PR #1323, opened 2024-07-10
backend: add endpoint compatibility check for wasm transforms

bojand:
When no WASM transforms (ex. Redpanda 22.3.25):

```json
{
  "endpoint": "redpanda.api.console.v1alpha1.TransformService",
  "method": "POST",
  "isSupported": false
}
```

When we do have WASM transform capability (Redpanda 24.2.1-rc3):

```json
{
  "endpoint": "redpanda.api.console.v1alpha1.TransformService",
  "method": "POST",
  "isSupported": true
}
```

### DIFF
--- a/backend/pkg/console/endpoint_compatibility.go
+++ b/backend/pkg/console/endpoint_compatibility.go
@@ -125,6 +125,12 @@ func (s *Service) GetEndpointCompatibility(ctx context.Context) (EndpointCompati
 			HasRedpandaAPI:  true,
 			RedpandaFeature: "role_based_access_control",
 		},
+		{
+			URL:             consolev1alpha1connect.TransformServiceName,
+			Method:          "POST",
+			HasRedpandaAPI:  true,
+			RedpandaFeature: "wasm_transforms",
+		},
 	}
 
 	endpoints := make([]EndpointCompatibilityEndpoint, 0, len(endpointRequirements))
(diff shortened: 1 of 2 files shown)


backend/pkg/console/endpoint_compatibility.go
```diff
@@ -125,6 +125,12 @@ func (s *Service) GetEndpointCompatibility(ctx context.Context) (EndpointCompati
 			HasRedpandaAPI:  true,
 			RedpandaFeature: "role_based_access_control",
 		},
+		{
+			URL:             consolev1alpha1connect.TransformServiceName,
+			Method:          "POST",
+			HasRedpandaAPI:  true,
```
weeco, line 131: That alone (redpanda admin api available) doesn't indicate whether the transform service is supported though no?

weeco (approved):

bojand:
Turns out the [features API endpoint](https://docs.redpanda.com/api/admin-api/#get-/features) is not a reliable mechanism to determine cluster capability ([ref](https://redpandadata.slack.com/archives/C01819S3ARH/p1720621385130579)). More appropriate mechanism it to attempt to utilize the functionality and if it doesn't work, interpret the result as cluster lacking said functionality.

For example cluster has WASM data transforms only if we are able to list transforms. Cluster config does have `data_transforms_enabled` bool flag, but even that is not reliable, because it could be `true` but until the cluster is restarted transforms may not be usable.

Similarly for RBAC, attempt to list roles. If the call works, cluster has the capability, otherwise it does not.

weeco (approved):
Great that you spotted this by double checking/testing it

PR #1325, opened 2024-07-10
v1alpha2 refactor

bojand:
Moves ConnectRPC public API services to respective `v1alpha1` and `v1alpha2` folders.
`v1alpha2` is the true implementation.
`v1alpha1` services wrap the `v1alpha2` implementation, mapping requests and responses.


weeco (approved):
Wondering if we should separate integration tests between each version. One dir for v1alpha1 another for v1alpha2. No strong opinion, LGTM

bojand:
Yea I would prefer that too but I couldn't do it nicely due to the way the test suite and testify works. It would have to be separate tests suites and would require more work.

PR #1335, opened 2024-07-19
Quotas column modification

jvorcak:
Screenshot 2024-07-19 at 12 53 36
Screenshot 2024-07-19 at 12 53 33

### DIFF
--- a/frontend/src/components/pages/quotas/Quotas.List.tsx
+++ b/frontend/src/components/pages/quotas/Quotas.List.tsx
@@ -14,7 +14,7 @@ import { PageComponent, PageInitHelper } from '../Page';
 import { api } from '../../../state/backendApi';
 import { computed, makeObservable } from 'mobx';
 import { appGlobal } from '../../../state/appGlobal';
-import { DefaultSkeleton } from '../../../utils/tsxUtils';
+import { DefaultSkeleton, InfoText } from '../../../utils/tsxUtils';
 import { SkipIcon } from '@primer/octicons-react';
 import { toJson } from '../../../utils/jsonUtils';
 import { prettyBytes, prettyNumber } from '../../../utils/utils';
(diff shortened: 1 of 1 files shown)


mattschumpert:
@dotnwat  @michael-redpanda @twmb  @pgellert should we call this last column 'metadata operations' since 'controller' is more of an internal detail of Redpanda.   The user story here is 

"Limit rate of topic mutation requests, including create, add, and delete partition, in number of partitions per second"

mattschumpert (approved):
Approving but suggest we change the last column name per comment

rikimaru0345 (approved):

dotnwat:
> @dotnwat @michael-redpanda @twmb @pgellert should we call this last column 'metadata operations' since 'controller' is more of an internal detail of Redpanda. The user story here is

Seems debatable. "Controller" is an implementation detail, but after KIP-500, Kafka also has a "controller" so it's definitely an abstraction that isn't completely hidden.

Using "metadata operations" is also ambiguous in the sense that there are kafka client operations that are metadata which don't fall into producer or consumer. If the metric is really counting controller operations, then something like "controller operations" makes sense. "mutation rate" sounds a bit too low-level for a dashboard like this?

pgellert:
I think the 'Controller Mutation Rate' column name makes sense because this is consistent with:
* The naming used in [KIP-599](https://cwiki.apache.org/confluence/plugins/servlet/mobile?contentId=148648680#KIP599:ThrottleCreateTopic,CreatePartitionandDeleteTopicOperations-ControllerMutationQuotaManager) which introduced this quota.
* How Redpanda operators would create the quota with rpk: `rpk cluster quota alter --default client-id --add controller_mutation_rate=100`
* How Kafka operators would create the quota with kafka-configs.sh: `kafka-configs.sh --bootstrap-server localhost:9092 --alter --client-defaults --add-config controller_mutation_rate=100`

The quota is counting topic partition mutations (number of topic partitions added/deleted), so perhaps 'Partition Mutation Rate' would be more descriptive. But, in my opinion, it is better to be consistent here with the naming of this quota everywhere else and stay with 'Controller Mutation Rate'.

jvorcak:
@dotnwat @pgellert thanks for your comments. I'd need somebody to decide on a final answer. So it it `Controller Mutation Rate` if I understand correctly and the description would stay the same?

mattschumpert:
Agreed.' Controller Mutation Rate' is correct.  That said a tooltip explaining this cryptic Kafka name and what it means would be great

weeco (approved):

PR #1338, opened 2024-07-22
fix: Broken links in overview page

joostdecock:
The links in the security section of the overview page point to `/acl/` (relative to the basepath) but that returns an error.

Seems they should point to `/security/acl/` instead, so that's what I've changed.

Note that this is a little bit of an opportunistic PR as I am not familiar with console. But I felt that submitting a PR was more valuable than opening an issue.


weeco:
I assume this was fixed via https://github.com/redpanda-data/console/pull/1343 cc @jvorcak

PR #1339, opened 2024-07-22
Add generic TLS config options

bojand:


backend/pkg/config/tls.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// SetDefaults for the TLS config.
+func (c *TLS) SetDefaults() {
+	c.Enabled = false
+}
+
+// Validate the TLS config.
+func (*TLS) Validate() error {
+	return nil
```
weeco, line 26: Wondering wehther we should call TLSConfig() here to check early whether the tls settings may be valid. Classic issues may be:

- Wrong path to certs
- Console user has no read permissions on the filesystem
- Certs are not in expected PEM format
- Wrong CA for given key pair

I'm unsure whether defaults are set before or after our call to `Validate()`, this may be important. WDYT?

weeco: If we think calling TLSConfig() is problematic, we can still check if os.ReadFile() works if path is != "" for any of these.

weeco (approved):

backend/pkg/config/tls.go
bojand: Was thinking of doing that. `RedpandaAdminAPITLS` that this replaces did in fact to that.
Added so Validate performs `TLSConfig()`

weeco:
One more thing that we can address in a follow-up PR is to automatically set Enabled = true for HTTP clients. Whether or not we should use TLS should be clear based on the given scheme in the URL. Only for the Kafka API it's not clear upfront which is why I had the `Enabled` bool and blindly copy pasted it for other clients too.

weeco (approved):

PR #1348, opened 2024-07-30
Feature/rpconnect pipeline flow

weeco:
This PR is for the reactflow rendering of the graph pipeline. This PR contains code ported from Benthos studio.


backend/pkg/rpconnect/docs/studio_exclusive.go
```diff
@@ -0,0 +1,3 @@
+package docs
+
+var _, _ = marshalYAML(nil)
```
Jeffail, line 3: Let's delete this file, it's not doing anything for us. I'm not entirely sure why this existed in the first place.

backend/pkg/rpconnect/docs/sync.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cp ../../../../go/benthos/internal/docs/* .
+rm json_schema.go benchmark_test.go bloblang_markdown.go
+sed -i 's/github.com\/benthosdev\/benthos\/v4\/internal\/docs/dullploy\/blobstudio\/internal\/docs/g' ./*.go
```
Jeffail, line 5: Let's delete this file as well. The reason this exists is because we essentially have a fork of the `benthos/internal/docs` so that we can implement our tree builder, and this script lets us quickly bring in bug fixes. However, I think in reality we will go with one of these two options going forwards:

1. Reduce this package and maintain it separately
2. Remove this and we add the tree builder to benthos

Option 2 is definitely preferred, but it would also be a public API on Benthos which would need to adhere to backwards compatibility. Therefore, I need a bit of time to figure out what this functionality would look like, and so option 1 will need to be enough for now.

backend/pkg/rpconnect/graph/graph_test.go
```diff
@@ -0,0 +1,75 @@
...
+	stream, resources, err := graph.ConfigToTree(confNode, lints)
+	assert.NoError(t, err)
+
+	sj, _ := json.Marshal(stream)
+	fmt.Println("stream:")
+	fmt.Println(string(sj))
+
+	rj, _ := json.Marshal(resources)
+	fmt.Println("resources:")
+	fmt.Println(string(rj))
```
Jeffail, line 72: This needs tidying, ideally into a proper test assertion.

Jeffail (commented):
Left a couple of comments, mostly just a bit of tidying.

From what I understand thus far this is just going to be a readonly view for now. Also, if we add an ability to edit the config from the visual view in future we're probably going to implement it differently to how it works right now in Benthos Studio.

If my understanding is correct then I'd probably also suggest removing all of the "Action" functionality. It's adding complexity but doesn't actually do anything for us if we're displaying the tree in readonly mode.

PR #1367, opened 2024-08-09
frontend: show pipeline service in sidebar for serverless

ajyoung:
Attempt to show pipeline service in the sidebar for serverless clusters. ~~It doesn't quite work so I'll ask for some help from console team to fix it.~~ Works now but let me know if this is the right/best way.

related to: https://github.com/redpanda-data/console-enterprise/pull/435

### DIFF
--- a/frontend/src/components/routes.tsx
+++ b/frontend/src/components/routes.tsx
@@ -277,7 +277,9 @@ export const APP_ROUTES: IRouteEntry[] = [
         routeVisibility(true, [Feature.GetQuotas], ['canListQuotas'])
     ),
 
-    MakeRoute<{}>('/connect-clusters', KafkaConnectOverview, 'Connectors', LinkIcon, true),
+    MakeRoute<{}>('/connect-clusters', KafkaConnectOverview, 'Connectors', LinkIcon, true,
+        routeVisibility(true, [Feature.PipelineService])
+    ),
     MakeRoute<{ clusterName: string }>('/connect-clusters/:clusterName', KafkaClusterDetails, 'Connect Cluster'),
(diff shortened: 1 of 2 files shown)


frontend/src/components/routes.tsx
```diff
@@ -277,7 +277,9 @@ export const APP_ROUTES: IRouteEntry[] = [
         routeVisibility(true, [Feature.GetQuotas], ['canListQuotas'])
     ),
 
-    MakeRoute<{}>('/connect-clusters', KafkaConnectOverview, 'Connectors', LinkIcon, true),
+    MakeRoute<{}>('/connect-clusters', KafkaConnectOverview, 'Connectors', LinkIcon, true,
+        routeVisibility(true, [Feature.PipelineService])
```
rikimaru0345, line 281: If you do this `routeVisibility(true, [Feature.PipelineService])`, this will disable the connect overview page when the pipeline service is not available.

At first this sounds like it is intended. 
But think about the case where someone has the pipeline service disabled, but still wants to see what is going on with the old/basic kafka connect.

I think you don't want to add this routeVisibility filter.
The `KafkaConnectOverview` page is an exception in that it is aware that it needs to deal with 2 different apis (rp connect aka pipelines, and kafka connect) and has automatic handling for this.

rikimaru0345: Your removal of that line from `routesIgnoredInServerless` should be enough for the intended effect (unless I'm missing something)

rikimaru0345 (changes requested):

frontend/src/components/routes.tsx
ajyoung: Thanks Moritz! Fixed.

rikimaru0345 (approved):

PR #1372, opened 2024-08-12
Don't filter properties in getConfigObject

jvorcak:
This needs to be properly tested.


rikimaru0345:
> This needs to be properly tested.

Has it been tested?

rikimaru0345 (approved):

jvorcak:
> > This needs to be properly tested.
> 
> Has it been tested?

Yes from myside, but if someone else can test it or it can be included in some testing deployment to test for other, it would be highly appreciated.

PR #1373, opened 2024-08-13
Removed blocking route change from EmbeddedApp

jvorcak:
Fixes https://github.com/redpanda-data/console-enterprise/issues/422

Needs proper testing


sago2k8:
How should we test this,should we create an image and try it out in a cluster, it's ok for it to go main and let it go through the certification tests imho wdyt @weeco @jvorcak

sago2k8 (approved):

PR #1387, opened 2024-08-19
Cbor payload serde

omnilight:
Support for CBOR serde.


backend/pkg/cbor/service.go
```diff
@@ -0,0 +1,55 @@
...
+package cbor
+
+import (
+	"regexp"
+
+	"github.com/redpanda-data/console/backend/pkg/config"
+)
+
+// Service represents messagepack cfg, topic name regexes.
+type Service struct {
```
weeco, line 21: I think we don't need a service for this. Instead of creating a separate struct here I think we can handle this inside the serde directly no? The sole reason for this service seems to be for checking whether a topic is allowed or not.

backend/pkg/config/cbor.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Cbor represents the CBOR config.
+type Cbor struct {
+	Enabled bool `yaml:"enabled"`
+
+	// TopicNames is a list of topic names that shall be considered for cbor decoding.
+	// These names can be provided as regex string (e. g. "/.*/" or "/prefix-.*/") or as plain topic name
+	// such as "frontend-activities".
+	// This defaults to `/.*/`
+	TopicNames []string `yaml:"topicNames"`
```
weeco, line 22: I believe we can use the type `RegexpOrLiteral` here. This is also done in the proto topic mappings and parses these strings automatically into a regex. This should help removing the cbor/service.go

backend/pkg/serde/cbor.go
```diff
@@ -0,0 +1,155 @@
...
+	var obj any
+	err := cbor.Unmarshal(payload, &obj)
+	if err != nil {
+		return &RecordPayload{}, fmt.Errorf("decoding cbor payload: %w", err)
+	}
+
+	jsonObj := cborObjToJson(obj)
+	jsonBytes, err := json.Marshal(jsonObj)
+	if err != nil {
+		return &RecordPayload{}, fmt.Errorf("decoding cbor payload: %w", err)
```
weeco, line 64:
```suggestion
		return &RecordPayload{}, fmt.Errorf("failed converting cbor decoded object to JSON: %w", err)
```

backend/pkg/serde/cbor.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Name returns the name of the serde payload encoding.
+func (CborSerde) Name() PayloadEncoding {
+	return PayloadEncodingCbor
+}
+
+// DeserializePayload deserializes the kafka record to our internal record payload representation.
+func (d CborSerde) DeserializePayload(_ context.Context, record *kgo.Record, payloadType PayloadType) (*RecordPayload, error) {
+	if d.CborService == nil {
```
weeco, line 41: If the service is removed we should check the config instead whether CBOR is enabled and the topic is enabled to test for deserialization

weeco (commented):
Hey, thanks for your PR. Overall LGTM, but I propose to remove the CBOR service as the functionality is very minor and could be handled inside the stateless CBOR serde instead.

backend/pkg/cbor/service.go
omnilight: Yep, I can change that, but this is just the same way as it implemented for msgpack: https://github.com/redpanda-data/console/blob/master/backend/pkg/msgpack/service.go

Should msgpack be redesigned too?

backend/pkg/config/cbor.go
omnilight: Ok

backend/pkg/serde/cbor.go
omnilight: Ok

backend/pkg/cbor/service.go
weeco: You don't need to change msgpack. The regexp config type was introduced after the msgpack PR was provided by another community user.

backend/pkg/serde/cbor.go
omnilight: Replaced CborSvc with direct config usage

backend/pkg/serde/cbor.go
omnilight: Fixed

backend/pkg/config/cbor.go
omnilight: Replaced with RegexpOrLiteral

backend/pkg/cbor/service.go
omnilight: Fixed

omnilight:
@weeco Hi! I have fixed review comments, please take a look!

weeco (approved):
Very good work, thank you @omnilight for the contribution 👍

PR #1401, opened 2024-08-21
proto: add RPC for getting Pipeline Service config schema

bojand:
**Changes**
Adds comments.

Changes the List Pipelines filter to use `name_contains` as that is the standard that is used in all of dataplane APIs. 
The List RPC implementation already performs this behaviour, just the name of the param was not consistent with the rest of the APIs. Also makes the name_contains param match validation rules of the name property.

Adds RPC for getting the Redpanda Connect Schema.

Essentially there are 4 supported formats that match [connect list options](https://github.com/redpanda-data/connect/blob/v4/lib/service/list.go)

The default formatting will be a flat list text format equivalent to:

```sh
❯ rpk connect list
Inputs:
  - amqp_0_9
  - amqp_1
  - aws_kinesis
...
Bloblang Functions:
  - batch_index
  - batch_size
  - content
```

There is a JSON format which is a strict JSON serialization of the listing without any metadata and examples for different components. Equivalent to `rpk connect list formant --json`.

There is a JSON FULL format which seems to include some metadata, and examples. 

```json
 {
      "status": "stable",
      "name": "parse_duration",
      "description": "Attempts to parse a string as a duration and returns an integer of nanoseconds. A duration string is a possibly signed sequence of decimal numbers, each with an optional fraction and a unit suffix, such as \"300ms\", \"-1.5h\" or \"2h45m\". Valid time units are \"ns\", \"us\" (or \"µs\"), \"ms\", \"s\", \"m\", \"h\".",
      "params": {},
      "categories": [
        {
          "Category": "Timestamp Manipulation",
          "Description": "",
          "Examples": [
            {
              "mapping": "root.delay_for_ns = this.delay_for.parse_duration()",
              "summary": "",
              "results": [
                [
                  "{\"delay_for\":\"50us\"}",
                  "{\"delay_for_ns\":50000}"
                ]
              ],
              "skip_testing": false
            },
            {
              "mapping": "root.delay_for_s = this.delay_for.parse_duration() / 1000000000",
              "summary": "",
              "results": [
                [
                  "{\"delay_for\":\"2h\"}",
                  "{\"delay_for_s\":7200}"
                ]
              ],
              "skip_testing": false
            }
          ]
        }
      ],
      "impure": false
    },
```

There is "JSON FULL SCRUBBED" which includes some metadata, but no details and examples:

```json
...
{
      "status": "stable",
      "name": "parse_duration",
      "params": {},
      "impure": false
    },
...
```
This is what's included in [lint.go](https://github.com/redpanda-data/console/blob/master/backend/pkg/rpconnect/lint.go#L20-L24)

Finally there is JSON SCHEMA representation of the full schema. I am not sure how this is generated. But essentially this will be what is [rp-connect-schema.json](https://github.com/redpanda-data/console/blob/master/frontend/src/assets/rp-connect-schema.json). This is useful for frontent integration into Monaco editor for linting and auto completion.


proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -246,6 +246,24 @@ service PipelineService {
       }
     };
   }
+  rpc GetPipelineServiceConfigSchema(GetPipelineServiceConfigSchemaRequest) returns (GetPipelineServiceConfigSchemaResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha2/redpanda-connect/config_schema"
+      response_body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Get Redpanda Connect pipeline service config schema"
```
weeco, line 255: "pipeline service config schema" sounds weird to me. I'm unsure what the pipeline **service** refers to. How about this?

```suggestion
      summary: "Retrieve the schema for configuring Redpanda Connect pipelines."
```

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -378,3 +396,16 @@ message StartPipelineRequest {
 message StartPipelineResponse {
   Pipeline pipeline = 1;
 }
+
+message GetPipelineServiceConfigSchemaRequest {
+  enum Format {
+    FORMAT_UNSPECIFIED = 0;
+    FORMAT_JSONSCHEMA = 1;
```
weeco, line 403:
```suggestion
    FORMAT_JSON_SCHEMA = 1;
```

rockwotj, line 403: nit: generally `FORMAT_JSON_SCHEMA` for enum naming.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -378,3 +396,16 @@ message StartPipelineRequest {
 message StartPipelineResponse {
   Pipeline pipeline = 1;
 }
+
+message GetPipelineServiceConfigSchemaRequest {
+  enum Format {
+    FORMAT_UNSPECIFIED = 0;
```
rockwotj, line 402: If we allow unspecified, should we return the format in the response?

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -378,3 +396,16 @@ message StartPipelineRequest {
...
+    FORMAT_JSONSCHEMA = 1;
+  }
+
+  Format format = 1;
```
rockwotj, line 406: Do we anticipate multiple formats in the future? Should we omit until it's needed (In true [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) style?)

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -246,6 +246,24 @@ service PipelineService {
       }
     };
   }
+  rpc GetPipelineServiceConfigSchema(GetPipelineServiceConfigSchemaRequest) returns (GetPipelineServiceConfigSchemaResponse) {
+    option (google.api.http) = {
+      get: "/v1alpha2/redpanda-connect/config_schema"
```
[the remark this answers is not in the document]
bojand: Updated.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
bojand: Updated.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
bojand: Updated.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
bojand: I have added more format enums and have updated description with additional information regarding the formats.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -299,10 +354,18 @@ message ListPipelinesRequest {
...
   string page_token = 3;
 
   message Filter {
-    string display_name = 1;
+    // Substring match on pipeline name. Case-sensitive.
+    string name_contains = 1 [
+      (buf.validate.field).string.pattern = "^[A-Za-z0-9-_ /]+$",
+      (buf.validate.field).string.min_len = 3,
+      (buf.validate.field).string.max_len = 128
+    ];
```
bojand, line 368: This was changed to use `name_contains` as that is the standard that is used in all of dataplane APIs. 
The List RPC implementation already performs this behaviour, just the name of the param was not consistent with the rest of the APIs. Also the param now matches validation rules of the name property from create request.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
rockwotj: Is there a use case for each of these? It's going to me more work in the dataplane to add support for each of these formats.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
weeco: Good point, wondering if it should actually be min_len = 3 since we may want to use this for quicksearch as people type. I believe we don't do this on other filters.

bojand: Good point. I've removed min_len.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
bojand: We want at least the JSON Schema one so we can generate it for frontend integration. I think one of the JSON ones (probably full or full scrubbed) would also be useful in case we ever want to have backend linting or validation within Console. Basically I added all for completeness to match Redpanda Connect. But for right now we really only need JSON Schema. I am open to just basically having unspecified (this is kinda required according to best practices and buf linting) and JSON Schema, and by default it would just return JSON Schema. If we need more later, we can add formats later.

simon0191: I agree with keeping it simple until we really need a new format.
If right now we only need JSON Schema, then let's remove the enum.
If/When we need another format, the enum can be easily added.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -400,7 +452,15 @@ message StartPipelineResponse {
 message GetPipelineServiceConfigSchemaRequest {
   enum Format {
     FORMAT_UNSPECIFIED = 0;
-    FORMAT_JSONSCHEMA = 1;
+
+    // The default text format that lists all components.
+    FORMAT_TEXT = 1;
```
simon0191, line 456: This looks more like another endpoint. It could be `ListPipelineServiceConfigSupportedComponents`

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -459,8 +459,14 @@ message GetPipelineServiceConfigSchemaRequest {
     // The service configuration is returned in JSON format.
     FORMAT_JSON = 2;
 
+    // The service configuration is returned in basic JSON format with metadata and examples.
+    FORMAT_JSON_FULL = 3;
```
simon0191, line 462: Do we need JSON and JSON FULL? what's the difference?

bojand: This is described in the PR description.

bojand:
I have removed the Format property. By default we will return JSON Schema structured representation of the Redpanda Connect Configuration Schema.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -378,3 +447,9 @@ message StartPipelineRequest {
 message StartPipelineResponse {
   Pipeline pipeline = 1;
 }
+
+message GetPipelineServiceConfigSchemaRequest {}
+
+message GetPipelineServiceConfigSchemaResponse {
+  string config_schema = 1;
```
rockwotj, line 454: nit: comments on what this is would be helpful to future readers. Maybe something like:

```
// config_schema is the JSON schema of the configuration files that are allowed for pipelines.
```

rockwotj (approved):

weeco (approved):

PR #1421, opened 2024-08-29
connect: update config schema

bojand:


rockwotj (approved):

taskfiles/connect.yaml
```diff
@@ -0,0 +1,7 @@
+version: 3
+tasks:
+ ai-schema:generate:
+    desc: generates the JSON schema for the AI version of connect
+    cmds:
+      # we use jsonschema format since frontend uses JSON Schema structured reprenentation of the config schem
+      - docker run redpandadata/connect:4.34.0-ai -- list --format jsonschema > frontend/src/assets/rp-connect-schema.json
```
weeco, line 7:
```suggestion
      - docker run redpandadata/connect:4.34.0-ai -- list --format jsonschema > {{.FRONTEND_ROOT}}/src/assets/rp-connect-schema.json
```

I think we should use the variable here for more robustness. I haven't tested this

weeco (approved):

taskfiles/connect.yaml
bojand: Just tested it locally it works.

PR #1424, opened 2024-09-02
rpcn: remove limit from the pipeline resource

simon0191:
Remove limit from the pipeline resource in the dataplane API.

Goals:
1. simplify the API. We'll make request=limit in the backend
2. by making request=limit in the backend it will be more predictable for the user


frontend/src/protogen/redpanda/api/dataplane/v1alpha2/pipeline_pb.ts
```diff
@@ -653,19 +637,12 @@ export class PipelineUpdate extends Message<PipelineUpdate> {
   configYaml = "";
 
   /**
-   * Pipeline resource requests.
+   * The amount of resources that is guaranteed to be assigned to the pipeline.
```
nicolaferraro, line 640: I'm ok with the single number describing the resources assigned to a pipeline in the public facing API, although we were already [tweaking request/limits in serverless](https://github.com/redpanda-data/serverless/blob/b1e5e138cce0b70012912081d90dac88f31121ca/infra/terraform/provisioners/kubernetes-proxy/modules/kubernetes-serverless-proxy-aws/files/console.yaml.tftpl#L47) to tune the over-subscription levels.

I'd be more in favor to remove the k8s naming and semantics from this field (req/limit) and just use a generic `resources` field. The reason is that in serverless, users will always see a certain amount of resources in the UI e.g. `100 millicores`, but under the hood we may decide to set a "k8s request" e.g. of `10 millicores` if CPU is mostly unused on average on some nodes.

So, TLDR the value the user configures in serverless is basically the "limit", not the "request" (which is on us to tune). Hence, better calling it just "resources"..

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -37,11 +40,10 @@ message Pipeline {
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The configuration of the Pipeline as YAML. See https://docs.redpanda.com/redpanda-connect/configuration/about for more details."}
   ];
 
-  // The minimum amount of resources needed for the pipeline to run
-  Resources request = 5 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The minimum amount of resources needed for the pipeline to run."}];
-
-  // The maximum amount of resources that should be assigned to the pipeline
-  Resources limit = 6 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "The maximum amount of resources that should be assigned to the pipeline."}];
+  // The amount of resources that is requested for the pipeline. Depending on
+  // the backend, the resources might be guaranteed, or might be tweaked based
```
nicolaferraro, line 44: nit: it's always "guaranteed", just in a "different way".. we may have to shuffle some pods a bit to guarantee it, but the eventual target is the same. It's a bin packing problem we need to solve

nicolaferraro (approved):
LGTM. Btw this needs the console to be adapted to the new proto, I think we could just rename the field and keep the old field number (one of them) to have it backward compatible (if the console is not using JSON encoding, on which I'm not sure)

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
simon0191: Comment adjusted

rikimaru0345 (approved):

bojand (approved):

PR #1427, opened 2024-09-03
routes: attempt to fix v1alpha2 http routes

ajyoung:
wip: testing with @bojand


bojand:
This is not the issue. The issue is that the old user svc implementation is registered later on. 
https://github.com/redpanda-data/console/pull/1428

ajyoung:
Abandoned in favor of #1428

PR #1425, opened 2024-09-03
gha: on run buf on merge to master

andrewhsu:
quick fix to gha that uses secrets. ensure that it only runs on push to master branch.

feel free to close this PR if something else is desired.

PR #1434, opened 2024-09-05
build: run buf action only on triggers from within the repo

bojand:
Substitute for https://github.com/redpanda-data/console/pull/1425.

Currently the this action fails on PRs from forks due to `aws-actions/configure-aws-credentials` workflow which does not have access to secrets and vars in that context. 

This PR makes the GH Action conditional to only run on PRs from within branches, along with push to master.
This should cover most cases since normally the Console team should be updating the protos.
Exceptions are docs updates by docs team, or external contributions. But we can work around those cases or offer manual fixes.

### DIFF
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -3,18 +3,20 @@ name: Buf CI
 on:
   push:
     paths:
+      - .github/workflows/buf.yml
       - proto/**.proto
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled]
     paths:
       - proto/**.proto
   delete:
-permissions:
-  id-token: write
-  contents: read
-  pull-requests: write
 jobs:
   push-module:
+    if: github.repository == 'redpanda-data/console'
+    permissions:
+      id-token: write
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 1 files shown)


weeco (approved):

.github/workflows/buf.yml
```diff
@@ -3,18 +3,20 @@ name: Buf CI
...
     paths:
       - proto/**.proto
   delete:
-permissions:
-  id-token: write
-  contents: read
-  pull-requests: write
 jobs:
   push-module:
+    if: github.repository == 'redpanda-data/console'
```
andrewhsu, line 15:
```suggestion
    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'redpanda-data/console' }}
```
i suggest similar changes to https://github.com/redpanda-data/connect/pull/2825#discussion_r1745848682

andrewhsu (changes requested):
suggested `if:` guard to have the step skipped if PR from forked repo, but allow the step to run after merge to `master`

.github/workflows/buf.yml
```diff
@@ -3,18 +3,20 @@ name: Buf CI
 on:
   push:
     paths:
+      - .github/workflows/buf.yml
       - proto/**.proto
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled]
     paths:
       - proto/**.proto
```
andrewhsu, line 11:
```suggestion
      - .github/workflows/buf.yml
      - proto/**.proto
```
i suggest also having this workflow run on PRs that change the workflow yaml file so you can also ensure the `if:` stmt is doing what you want

andrewhsu (approved):
i made suggestion, but in general LGTM

PR #1440, opened 2024-09-09
proto: remove field mask from pipeline put api

bojand:
None of our existing public APIs use partial updates. FieldMasks are supported in gRPC via `PATCH`. This introduces a new paradigm.
All existing APIs all explicit total updates via `PUT`.
This PR changes the Pipeline API to follow established patterns.


bojand:
@rikimaru0345 I made attempt at frontend changes to address this. Needs review.

bochenekmartin (approved):

simon0191:
@birdayz says fieldmask is autogenerated. So we don't need to remove it from the proto

PR #1455, opened 2024-09-21
License nagging

weeco:
This pull request changes:

1. Add new `ListEnterpriseFeatures` RPC to console pkg (may return CodeUnimplemented if unreachable or unimplemented)
2. Add `violation` bool in ListLicenses response (defaults to false if admin api endpoint is unreachable or unimplemented)

Has a dependency on: https://github.com/redpanda-data/common-go/pull/24


rikimaru0345 (approved):

backend/pkg/api/connect/service/license/service.go
```diff
@@ -135,12 +135,18 @@ func (s Service) ListLicenses(ctx context.Context, _ *connect.Request[v1alpha1.L
 		}
 	}
 
+	isViolation := false
+	enterpriseFeatures, err := s.adminapiCl.GetEnterpriseFeatures(ctx)
+	if err == nil {
+		isViolation = enterpriseFeatures.Violation
+	}
```
bojand, line 142: if `GetEnterpriseFeatures` fails we default to "no violation"... is that intentional? 
Couldn't just someone turn off / ensure this endpoint fails and always have a valid license?

weeco: Yes this is intentional, since we will also be dealing with Redpanda clusters that do not yet implement this new endpoint. It is not a problem, because this is just some soft nagging / reminding the users that they are violating the license. If we are not certain they are violating the license we better be quiet was the idea here

bojand (approved):

frontend/src/state/backendApi.ts
```diff
@@ -1549,20 +1552,30 @@ const apiStore = {
...
+            .finally(() => {
+                this.licensesLoaded = true;
+            })
+
+        // Handle the first response
+        this.enterpriseFeaturesUsed = enterpriseFeaturesResponse.features;
+
+        // Handle the second response
+        this.licenses = licensesResponse.licenses;
+        this.licenseViolation = licensesResponse.violation;
```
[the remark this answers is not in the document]
jvorcak: @rikimaru0345 can you please check now?

ivpanda:
LGTM! Just two small comments on the errors, if we could show them using our error components, like this: 

### On the Admin page: 
image

### In the Details box: 
image

Thanks!

jvorcak:
> LGTM! Just two small comments on the errors, if we could show them using our error components, like this:
> 
> ### On the Admin page:
> image
> ### In the Details box:
> image
> Thanks!

I'd try to avoid showing "... or upload a new license" WDYT?
In my opinion it somehow might imply that there is a problem with a configuration and when I upload a new license it can fade away. But this error is shown in a corner case like network error or similar that are irrelevant to the actual configuration.

jvorcak:
Screenshot 2024-09-26 at 17 13 25
Screenshot 2024-09-26 at 17 08 06

mattschumpert:
@jvorcak 
Right, in that case just 'failed to load license info'.  If the license key itself was found to be invalid, I assume that would fail differently, in the upload page.

Also,

I am confused why we still see different license status in license details. Uploading an Enterprise license is supposed to also push the license to the Redpanda cluster (what now says 'Core Community').  This part should should say 'Repanda Community Edition' or 'Redpanda Enterprise Edition', and after upload of a license at a minimum both should be 'Enterprise'.  The Console one should be 'Console Community Edition', 'Console Enterprise Edition.

jvorcak:
@mattschumpert I've addressed the details we talked about in https://github.com/redpanda-data/console/pull/1464

PR #1458, opened 2024-09-24
Remove @ant-design/icons dependency

jvorcak:
Screenshot 2024-09-24 at 13 11 57
Screenshot 2024-09-24 at 13 07 55
Screenshot 2024-09-24 at 13 02 27
Screenshot 2024-09-24 at 12 58 49
Screenshot 2024-09-24 at 12 08 15
Screenshot 2024-09-24 at 11 54 00
Screenshot 2024-09-24 at 11 46 58
Screenshot 2024-09-24 at 11 42 23
Screenshot 2024-09-24 at 11 40 07
Screenshot 2024-09-24 at 11 34 59
Screenshot 2024-09-24 at 11 34 34
Screenshot 2024-09-24 at 15 07 46
Screenshot 2024-09-24 at 14 55 53
Screenshot 2024-09-26 at 11 33 48


frontend/src/components/pages/connect/CreateConnector.tsx
```diff
@@ -310,7 +308,6 @@ const ConnectorWizard = observer(({ connectClusters, activeCluster }: ConnectorW
         {
             title: 'Properties',
             description: 'Configure basic connection properties.',
-            icon: <ApiOutlined />,
```
jvorcak, line 313: After the migration, we don't support icons in the Wizards

frontend/src/components/pages/consumers/Modals.tsx
```diff
@@ -605,7 +589,7 @@ class ColAfter extends Component<{
                                     There is no offset for this partition at or after the given timestamp (<code>{new Date(this.props.selectedTime ?? 0).toLocaleString()}</code>). As a fallback, the last offset in that partition will be used.
                                 </div>
                             }
-                            icon={<WarningOutlined/>}
+                            icon={<MdOutlineWarningAmber size={14} />}
```
malinskibeniamin, line 592: Any reason why we have different icon sizes in various locations? I think 16 is the default?

frontend/src/components/pages/reassign-partitions/Step1.Partitions.tsx
```diff
@@ -323,9 +320,7 @@ function renderPartitionErrorsForTopic(_partitionsWithErrors: number) {
     >
         <span>
             <ZeroSizeWrapper justifyContent="center" alignItems="center" width="20px" height="18px">
-                <span style={{fontSize: '19px'}}>
-                    <WarningTwoTone twoToneColor="orange"/>
-                </span>
+                <MdOutlineWarningAmber color="orange" size={20} />
```
malinskibeniamin, line 323: Why is there different sizing for MdOutlineWarningAmber in various places?

malinskibeniamin (commented):
Left some comments, I will let @rikimaru0345 take a final look

frontend/src/components/pages/reassign-partitions/Step1.Partitions.tsx
jvorcak: I think I'll answer both of your question in one comment. At first, I tried to keep everything unified, but the current implementation also did have various sizes on various places + MD icons are visually positioned differently than ant-design icons if we keep the same dimensions.

So I tried to adjust this to reduce/minify visual regressions, even though it was a bit handy to do. I think we can make it continuously more consistent over time (and we should). But that would require much more time now, and I think removing this dependency is not so business critical now.

rikimaru0345 (approved):
LGTM

PR #1462, opened 2024-09-26
rpcn: add a "Completed" state for pipelines

nicolaferraro:
This is for pipelines that terminate successfully after doing a series of steps.


rockwotj (approved):
LGTM, might be worthwhile in the comments to clarify the difference between `STOPPED` and `COMPLETED`. Really `STOPPED` is more like a user requested pause as I understand it

bojand (approved):

PR #1478, opened 2024-10-17
frontend: fix load env behavior

malinskibeniamin:
rspack has different behavior compared to webpack when it comes to environment variables.

This has not been an issue in the case of cloud and admin UIs because we don't add more variables later on, everything is within `.env` file

```ts
// Good
console.log('1: ', process.env.REACT_APP_BUILD_TIMESTAMP); // 32503680000
// Bad
console.log('2: ', process.env['REACT_APP_BUILD_TIMESTAMP']); // 32503680000 (only possible thanks to spreading all process.env secrets inside rsbuild config which leaks secrets into frontend html, usually undefined)
// Bad
console.log('3: ', process.env?.REACT_APP_BUILD_TIMESTAMP); // undefined
```

How CRA does it:
https://github.com/facebook/create-react-app/blob/dd420a6d25d037decd7b81175626dfca817437ff/packages/react-scripts/config/env.js#L101-L107

New docs (latest Rspack release)
https://github.com/web-infra-dev/rsbuild/issues/3718 (issue)
https://github.com/web-infra-dev/rsbuild/pull/3734/files (new docs)

https://rsbuild.dev/guide/migration/cra#env-variables
https://rsbuild.dev/guide/advanced/env-vars#processenv-replacement

---

**BEFORE**
Screenshot 2024-10-17 at 12 09 36

---

**AFTER**
Screenshot 2024-10-17 at 12 06 18


sago2k8 (approved):
great job @malinskibeniamin

frontend/rsbuild.config.ts
```diff
@@ -46,7 +46,10 @@ export default defineConfig({
         },
     },
     source: {
-        define: publicVars,
+        define: {
+            ...publicVars,
+            'process.env': JSON.stringify(rawPublicVars),
```
sago2k8, line 51: this will be filtered by `REACT_APP_` variables? just making sure.

PR #1490, opened 2024-10-30
Fix/special chars fix

jvorcak:
https://www.loom.com/share/a314db899566463289368b03e995011c

@weeco @bojand does it make any sense to validate fieldname on GET requests etc? I'd personally prefer to only set it on requests that modify state, wdyt?


weeco:
> does it make any sense to validate fieldname on GET requests etc?

IMO it makes sense for security reasons.

weeco (approved):

bojand (approved):

Issue #1087, opened 2024-02-06
Topic configs - Info tooltip UI regression


weeco:
Today:

- Sizing of the fonts looks weird
- Styling of that tooltip feels misaligned

![ss-wl2wJJNN](https://github.com/redpanda-data/console/assets/23424570/6db2f615-9656-4356-b036-0c21c42c7b53)

ivpanda:
## Design proposal: 

We can actually remove all of this information that's in the popover to begin with that makes it crowded. The outcome is what's important to the user, as well as the source. We can make that less obscure by doing the following: 

### Infinite 

<img width="381" alt="image" src="https://github.com/redpanda-data/console/assets/129790723/df1dd78a-6f39-48a5-b26b-4b766eae16f5">

### Numeric value

<img width="377" alt="image" src="https://github.com/redpanda-data/console/assets/129790723/6dc4338e-74c1-4587-bda5-4dda31e273c0">

These would use the `heading-s` text formatting for the title, and `text-s` for the copy.

jvorcak:
@ivpanda I like the way you present it, but this would require rephrasing the texts and changing the API, because the texts come from there. I've made the quick fix that looks, I think, better than what we have now. https://github.com/redpanda-data/console/pull/1498

If you still want to implement this design, feel free to re-open.

PR #1498, opened 2024-11-06
Better markup in topic configuration help popup

jvorcak:
Screenshot 2024-11-06 at 10 42 19

### DIFF
--- a/frontend/src/components/pages/topics/TopicConfiguration.tsx
+++ b/frontend/src/components/pages/topics/TopicConfiguration.tsx
@@ -1,6 +1,6 @@
 import { PencilIcon } from '@heroicons/react/solid';
 import { AdjustmentsIcon } from '@heroicons/react/outline';
-import { Alert, AlertDescription, AlertIcon, Box, Button, Flex, Grid, GridItem, Icon, Input, Modal, ModalBody, ModalContent, ModalFooter, ModalHeader, ModalOverlay,
+import { Alert, AlertDescription, AlertIcon, Box, Button, Flex, Icon, Input, Modal, ModalBody, ModalContent, ModalFooter, ModalHeader, ModalOverlay,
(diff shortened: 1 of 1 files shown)


jvorcak:
Fixes https://github.com/redpanda-data/console/issues/1087

weeco:
@jvorcak I think we can remove the "Value" row because it's already shown in the table where we show the tooltip. And maybe we can incorporate the source as a sentence after the description (coming from the backend) to get closer to the desired mock.

So just append a string on the description like this (pseudo code). I think this allows us to get rid of the value and source in this tooltip right?

```
switch topicConfig.Source {
  case DEFAULT_CONFIG:
    return "Inherited from DEFAULT_CONFIG".
  case DYNAMIC_TOPIC_CONFIG:
    return "This is a custom setting for this topic"
  case DYNAMIC_BROKER_CONFIG, STATIC_BROKER_CONFIG:
    return "This is a custom setting set on the BROKER_CONFIG level.
  default:
    return ""
}
```

jvorcak:
Screenshot 2024-11-06 at 12 51 08

@weeco wdyt?

frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -382,23 +361,37 @@ export const ConfigEntryEditor = observer((p: {
...
         case 'STATIC_BROKER_CONFIG':
-            return 'Set on the broker by either a config file or environment variable';
-
+            return 'This is a custom setting set on the BROKER_CONFIG level.';
         default:
             return '';
     }
 }
+
+// function getConfigSourceExplanation(source: string) {
```
weeco, line 378: @jvorcak Do you want to remove this prior to merging?

weeco (approved):

PR #1505, opened 2024-11-08
proto: add debug_bundle_process_not_started property to GetDebugBundl…

bojand:
…eStatusResponse.

This is convenience property so that the frontend knows whether the debug bundle process was not started.


proto/redpanda/api/console/v1alpha1/debug_bundle.proto
```diff
@@ -166,6 +166,10 @@ message GetDebugBundleStatusResponse {
   }
 
   repeated DebugBundleBrokerStatus broker_statuses = 1;
+
+  // true IFF all broker statuses return
+  // "debug bundle process" not started error code
+  bool debug_bundle_process_not_started = 2;
```
weeco, line 172: NIT: I'm not a friend of negated bools. I'd prefer to call if `debug_bundle_process_started` instead. WDYT?

bojand: Changed to `has_debug_process`.

weeco (approved):

Issue #1405, opened 2024-08-23
Set default compression type


anhnmt:
![image](https://github.com/user-attachments/assets/01e12c56-7516-4663-a2ef-cd3f90efa178)

Is there any way I can leave the default Compress type as UNCOMPRESSED?
I'm testing old programs and having to select UNCOMPRESSED every time is inconvenient

weeco:
We could store the last selection in the frontend. That would be relatively easy to add, do you want to give it a shot?

anhnmt:
@weeco Of course, I really need this feature. Also, if you can do a message recall feature, it would be even better.

![image](https://github.com/user-attachments/assets/6966f7de-ff7c-4a9f-b1fc-169ec56c4038)

Sometimes I want to reuse old messages, this will help save more time

weeco:
What do you mean by a recall feature? Reproducing an existing message and possibly editing it along the way?

anhnmt:
@weeco Sorry I don't know what to call this function, I just want to copy the old message and push it to the end of the topic with all the information from before.

real-mday:
This issue is currently causing us to get spikes of >1 million messages over the course of hours on our development cluster when a test message is created and our software currently is not able to recover from trying to read a compressed message

PR #1511, opened 2024-11-12
Feature: Modify Default Producer Compression Type

real-mday:
Adding ability to configure default producer compression type (related to https://github.com/redpanda-data/console/issues/1405)

Example usage:
```apiVersion: v1
kind: ConfigMap
metadata:
  name: redpanda-console
  labels:
    app.kubernetes.io/name: console
    app.kubernetes.io/instance: redpanda-console
    app.kubernetes.io/version: v2.7.2
data:
  config.yaml: |
    # from .Values.console.config
    connect:
      enabled: false
    kafka:
      brokers:
      - broker1:9092
      - broker2:9092
      - broker3:9092
      producerCompressionType: 
        value: "uncompressed"
      schemaRegistry:
        enabled: false

### DIFF
--- a/backend/pkg/api/handle_producer_compression_type.go
+++ b/backend/pkg/api/handle_producer_compression_type.go
@@ -1,0 +1,33 @@
+// Copyright 2024 Redpanda Data, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+//
+// As of the Change Date specified in that file, in accordance with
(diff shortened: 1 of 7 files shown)


weeco:
The compression type config should only be stored in the frontend settings, so that the selection is persisted there. I think the endpoint & config addition is some extra complexity we should omit for a niche usability feature.

real-mday:
> #1405

Am I understanding correctly that although the issue is called "Set default compression type & reuse old messages" that you are only accepting features that implement "reuse old messages"? I can open a second PR which only addresses that but also maybe good to change the name of the issue

PR #1520, opened 2024-11-15
Adds support for debug bundle expiration

jvorcak:


frontend/src/components/pages/admin/Admin.DebugBundle.tsx
```diff
@@ -82,9 +82,14 @@ export class AdminDebugBundle extends Component<{}> {
         return (
             <Box>
                 <Header />
-                {api.canDownloadDebugBundle && <Text mt={4} fontWeight="bold">Latest debug bundle:</Text>}
-                <DebugBundleLink statuses={api.debugBundleStatuses} showDeleteButton />
-                {api.debugBundleStatuses.length === 0 && <Text>No debug bundle available for download.</Text>}
+                <Box mt={4}>
+                    {(api.canDownloadDebugBundle || api.isDebugBundleExpired) && <Text fontWeight="bold">Latest debug bundle:</Text>}
+                    {api.isDebugBundleExpired && <Text>Your previous bundle has expired and cannot be downloaded.</Text>}
+                    {api.isDebugBundleError && <Text>This debug bundle was cancelled by the user and is not available for download.</Text>}
```
bojand, line 88: This doesn't necessarily mean it was cancelled? Just because there is an error ... it _could have been cancelled_, but it _could_ have errored out for any number of other reasons.

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
```diff
@@ -82,9 +82,14 @@ export class AdminDebugBundle extends Component<{}> {
...
+                    {api.canDownloadDebugBundle && <DebugBundleLink statuses={api.debugBundleStatuses} showDeleteButton/>}
```
bojand, line 89: I am not sure if I understand React code completely ... Will they be able to delete an expired bundle with this logic? They should be able to.

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
jvorcak: I agree, in Figma we have this wording but I also wanted to make it a bit more universal. Currently there is just no way to identify whether it's cancelled or it just errored for all brokers. Right now we assume that if all broker have error state, it's cancelled, but I think we need more universal wording. cc @ivpanda

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
jvorcak: They can generate a new bundle in that case. If we want to keep an explicit delete action as well, I can add that. Just let me know cc @ivpanda.

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
ivpanda: Just to clarify, if there WAS a different error, would we say what that is? E.g. something like: 

"Your debug bundle isn't available for download. [REASON]"

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
ivpanda: @bojand I'm assuming that when they generate a new one, the old one is overwritten? In this case, I'd prefer to not add a delete button— if a user sees that they can't download something, it's contradictory UX to tell them that they should delete it, too. I'd rather delete it for them on expiry if they can't download it.

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
jvorcak: @ivpanda right now we don't have a clear error message provided by the backend, at least not in case I can reproduce (when I interrupt the process). I'd go with something universal for now and we can always improve it.

frontend/src/components/pages/admin/Admin.DebugBundle.tsx
jvorcak: I've already added it before I saw this comment, but it's super easy to remove.

PR #1526, opened 2024-11-19
Preserve path encoded subjects in middleware

### LINKED ISSUE #1525: Non-default basePath causes issues with URL path escaped subject names
## Problem

Loading the schema registry UI or attempting to click 'Produce Record' on a topic integrated with the schema registry fails. The console shows 500 errors, and when looking at network requests there is a 200 request for the schema in the topic but 404 errors attempting to fetch the topic's schema references (if the schema references require URL path escaping - i.e. `%2F` for `/`).

## Steps to reproduce

* Configure a schema registry for a Protobuf topic (i.e. `some-topic-value`) which has one or more references with `/` in their names.
* Configure the console with a valid `schemaRegistry` and a non-default `server` -> `basePath`.
* Attempt to access 'Schema Registry' on the navigation on the left or attempt to produce a record to a schema registry enabled topic.

## Suggested Fix

The problem appears to be related to the middleware which is responsible for stripping the `basePath` from incoming requests: https://github.com/redpanda-data/console/blob/8a6526487b6af22a5155a9a940cb70c3c8296681/backend/pkg/api/middlewares.go#L75-L82

This section of code needs to use the `r.URL.RawPath` (if present) to build the `rctx.RoutePath`, not the `r.URL.Path`. Here's an example middleware from Chi: https://github.com/go-chi/chi/blob/6ceb498b221efa11f559602beb2463b8b52c2161/middleware/clean_path.go#L18-L22. It should probably also remove the prefix from `RawPath` as well for consistency.

I'm happy to raise a PR with the suggested fix.

pkwarren:
Update basePathMiddleware to use `RawPath` (if present) over `Path` to ensure that URL path encoded subjects are preserved when routing requests.

Update the middleware to also remove the prefix from `RawPath` for consistency.

Fixes #1525.

### DIFF
--- a/backend/pkg/api/middlewares.go
+++ b/backend/pkg/api/middlewares.go
@@ -75,11 +75,12 @@ func (b *basePathMiddleware) Wrap(next http.Handler) http.Handler {
 		// Strip prefix from the request url
 		var path string
 		rctx := chi.RouteContext(r.Context())
-		if rctx.RoutePath != "" {
+		switch {
+		case rctx.RoutePath != "":
 			path = rctx.RoutePath
-		} else if r.URL.RawPath != "" {
+		case r.URL.RawPath != "":
 			path = r.URL.RawPath
-		} else {
+		default:
 			path = r.URL.Path
 		}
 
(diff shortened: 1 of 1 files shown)


weeco:
Thanks for raising the issue and taking the time to fix it with this PR @pkwarren !

PR #1527, opened 2024-11-22
Add redpanda connect secret manager to console

andresaristizabal:
This PR introduces a new tab for Redpanda Connect

Features:
- Create secrets
- Edit secrets
- Delete secrets
- Update secrets
- Secret autocomplete on pipeline RPCN editor
- Choose which connect tab is showing as default based on query param `defaultTab`

Pending:

- [x] Handle BYOC, Serverless and Hosted cluster
- [x] Avoid duplications auto complete secrets in the pipeline editor **-- disable autocomplete**
- [ ] Disable buttons in the "Update Secret" form while an update request is in progress.

## Dedicated integration cluster

https://github.com/user-attachments/assets/88c0ab66-3efe-4091-b821-3436e19109b8

## Serverless integration cluster

https://github.com/user-attachments/assets/ef115086-2ec2-44fd-b014-43b7c69008a5

### Before:
image

## Enterprice self-hosted

https://github.com/user-attachments/assets/f1987f9e-16e5-464c-99ec-872a17a797c1

### DIFF
--- a/frontend/src/components/pages/connect/Overview.tsx
+++ b/frontend/src/components/pages/connect/Overview.tsx
@@ -10,14 +10,14 @@
  */
 
 import { observer, useLocalObservable } from 'mobx-react';
-import { Component } from 'react';
+import {Component, FunctionComponent} from 'react';
 import { appGlobal } from '../../../state/appGlobal';
 import { api } from '../../../state/backendApi';
 import { ClusterConnectorInfo, ClusterConnectors, ClusterConnectorTaskInfo } from '../../../state/restInterfaces';
 import { uiSettings } from '../../../state/ui';
 import { Code, DefaultSkeleton } from '../../../utils/tsxUtils';
 import Tabs, { Tab } from '../../misc/tabs/Tabs';
-import { PageComponent, PageInitHelper } from '../Page';
+import {PageComponent, PageInitHelper} from '../Page';
 import { ConnectorClass, ConnectorsColumn, errIcon, mr05, NotConfigured, OverviewStatisticsCard, TasksColumn, TaskState } from './helper';
 import Section from '../../misc/Section';
 import PageContent from '../../misc/PageContent';
@@ -28,31 +28,45 @@ import { RedpandaConnectIntro } from '../rp-connect/RedpandaConnectIntro';
 import { Features } from '../../../state/supportedFeatures';
 import {isServerless} from '../../../config';
 import RpConnectSecretsList from '../rp-connect/secrets/Secrets.List';
+import { useLocation } from 'react-router-dom';
 
 enum ConnectView {
-    kafkaConnect = 'kafka-connect',
+    KafkaConnect = 'kafka-connect',
     RedpandaConnect = 'redpanda-connect',
-    RedpandaConncetSecret = 'redpanda-connect-secret',
+    RedpandaConnectSecret = 'redpanda-connect-secret',
 }
 
-const getDefaultView = (defaultView: string): ConnectView => {
+/**
+ * The Redpanda Connect Secret Manager introduces a new tab in Redpanda Connect.
+ * this logic determines which tab should be opened based on the `defaultTab`
+ * query parameter in the URL.
+ */
+const getDefaultView = (defaultView: string): { initialTab: ConnectView, redpandaConnectTab: ConnectView } => {
 
     const showPipelines = Features.pipelinesApi
-
+    const showKafkaTab = {initialTab: ConnectView.KafkaConnect, redpandaConnectTab: ConnectView.RedpandaConnect}
+    const showRedpandaConnectTab = {initialTab: ConnectView.RedpandaConnect, redpandaConnectTab: ConnectView.RedpandaConnect}
     if (!showPipelines) {
-        return ConnectView.kafkaConnect;
+        return showKafkaTab;
     }
 
     switch (defaultView) {
         case 'kafka-connect':
-            return ConnectView.kafkaConnect;
+            return showKafkaTab;
         case 'redpanda-connect':
-            return ConnectView.RedpandaConnect;
+            return showRedpandaConnectTab;
         case 'redpanda-connect-secret':
-            return ConnectView.RedpandaConncetSecret;
+            return {initialTab: ConnectView.RedpandaConnect, redpandaConnectTab: ConnectView.RedpandaConnectSecret};
         default:
-            return showPipelines ? ConnectView.RedpandaConnect : ConnectView.kafkaConnect;
+            return showRedpandaConnectTab;
     }
+}
+
+const WrapUseSearchParamsHook: FunctionComponent<{matchedPath: string}> = (props) => {
+    const {search}= useLocation();
+    const searchParams = new URLSearchParams(search);
+    const defaultTab = searchParams.get('defaultTab') || '';
+    return <KafkaConnectOverview defaultView={defaultTab} {...props}/>
 }
 
 @observer
@@ -77,8 +91,6 @@ class KafkaConnectOverview extends PageComponent<{ defaultView: string }> {
     }
 
     render() {
-        const showPipelines = Features.pipelinesApi
-
         const tabs = [
             {
                 key: ConnectView.RedpandaConnect,
@@ -88,11 +100,11 @@ class KafkaConnectOverview extends PageComponent<{ defaultView: string }> {
                         <Text mb={4}>
                             Redpanda Connect is an alternative to Kafka Connect. Choose from a growing ecosystem of readily available connectors. <Link href="https://docs.redpanda.com/redpanda-cloud/develop/connect/about/" target="_blank">Learn more.</Link>
                         </Text>
-                        <TabRedpandaConnect defaultView={getDefaultView(this.props.defaultView)}/>
+                        <TabRedpandaConnect defaultView={getDefaultView(this.props.defaultView).redpandaConnectTab}/>
                     </Box>,
             },
             {
-                key: ConnectView.kafkaConnect,
+                key: ConnectView.KafkaConnect,
                 title: <Box minWidth="180px">Kafka Connect</Box>,
                 content:
                 <Box>
@@ -104,7 +116,7 @@ class KafkaConnectOverview extends PageComponent<{ defaultView: string }> {
         ] as Tab[];
 
         if (isServerless())
-            tabs.removeAll(x => x.key == 'kafkaConnect');
+            tabs.removeAll(x => x.key == ConnectView.KafkaConnect);
 
         return (
             <PageContent>
@@ -116,14 +128,14 @@ class KafkaConnectOverview extends PageComponent<{ defaultView: string }> {
                             ? tabs[0].content()
                             : tabs[0].content
                     )
-                    : <Tabs tabs={tabs} defaultSelectedTabKey={showPipelines ? ConnectView.RedpandaConnect : ConnectView.kafkaConnect} />
+                    : <Tabs tabs={tabs} defaultSelectedTabKey={getDefaultView(this.props.defaultView).initialTab} />
                 }
             </PageContent>
         );
     }
 }
 
-export default KafkaConnectOverview;
+export default WrapUseSearchParamsHook;
 
 @observer
 class TabClusters extends Component {
@@ -371,7 +383,7 @@ const TabRedpandaConnect = observer((_p: {defaultView: ConnectView}) => {
         },
     ] as Tab[];
 
-    return <Tabs tabs={tabs} defaultSelectedTabKey={_p.defaultView === ConnectView.RedpandaConncetSecret ? 'secrets' : 'pipelines'}/>
+    return <Tabs tabs={tabs} defaultSelectedTabKey={_p.defaultView === ConnectView.RedpandaConnectSecret ? 'secrets' : 'pipelines'}/>
 })
 
 export type ConnectTabKeys = 'clusters' | 'connectors' | 'tasks';
--- a/frontend/src/components/pages/connect/Overview.tsx
+++ b/frontend/src/components/pages/connect/Overview.tsx
@@ -27,9 +27,36 @@ import RpConnectPipelinesList from '../rp-connect/Pipelines.List';
 import { RedpandaConnectIntro } from '../rp-connect/RedpandaConnectIntro';
 import { Features } from '../../../state/supportedFeatures';
 import {isServerless} from '../../../config';
+import RpConnectSecretsList from '../rp-connect/secrets/Secrets.List';
+
+enum ConnectView {
+    kafkaConnect = 'kafka-connect',
+    RedpandaConnect = 'redpanda-connect',
+    RedpandaConncetSecret = 'redpanda-connect-secret',
+}
+
+const getDefaultView = (defaultView: string): ConnectView => {
+
+    const showPipelines = Features.pipelinesApi
+
+    if (!showPipelines) {
+        return ConnectView.kafkaConnect;
+    }
+
+    switch (defaultView) {
+        case 'kafka-connect':
+            return ConnectView.kafkaConnect;
+        case 'redpanda-connect':
+            return ConnectView.RedpandaConnect;
+        case 'redpanda-connect-secret':
+            return ConnectView.RedpandaConncetSecret;
+        default:
+            return showPipelines ? ConnectView.RedpandaConnect : ConnectView.kafkaConnect;
+    }
+}
 
 @observer
-class KafkaConnectOverview extends PageComponent {
+class KafkaConnectOverview extends PageComponent<{ defaultView: string }> {
     initPage(p: PageInitHelper): void {
         p.title = 'Overview';
         p.addBreadcrumb('Connect', '/connect-clusters');
@@ -54,18 +81,18 @@ class KafkaConnectOverview extends PageComponent {
 
         const tabs = [
             {
-                key: 'redpandaConnect',
+                key: ConnectView.RedpandaConnect,
                 title: <Box minWidth="180px">Redpanda Connect <Badge ml={2}>Recommended</Badge></Box>,
                 content:
                     <Box>
                         <Text mb={4}>
                             Redpanda Connect is an alternative to Kafka Connect. Choose from a growing ecosystem of readily available connectors. <Link href="https://docs.redpanda.com/redpanda-cloud/develop/connect/about/" target="_blank">Learn more.</Link>
                         </Text>
-                        <TabRedpandaConnect />
+                        <TabRedpandaConnect defaultView={getDefaultView(this.props.defaultView)}/>
                     </Box>,
             },
             {
-                key: 'kafkaConnect',
+                key: ConnectView.kafkaConnect,
                 title: <Box minWidth="180px">Kafka Connect</Box>,
                 content:
                 <Box>
@@ -89,7 +116,7 @@ class KafkaConnectOverview extends PageComponent {
                             ? tabs[0].content()
                             : tabs[0].content
                     )
-                    : <Tabs tabs={tabs} defaultSelectedTabKey={showPipelines ? 'redpandaConnect' : 'kafkaConnect'} />
+                    : <Tabs tabs={tabs} defaultSelectedTabKey={showPipelines ? ConnectView.RedpandaConnect : ConnectView.kafkaConnect} />
                 }
             </PageContent>
         );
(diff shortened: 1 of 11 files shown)


frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -27,9 +27,50 @@ import RpConnectPipelinesList from '../rp-connect/Pipelines.List';
...
+/**
+ * The Redpanda Connect Secret Manager introduces a new tab in Redpanda Connect.
+ * this logic determines which tab should be opened based on the `defaultTab`
+ * query parameter in the URL.
+ */
+const getDefaultView = (defaultView: string): { initialTab: ConnectView, redpandaConnectTab: ConnectView } => {
+
+    const showPipelines = Features.pipelinesApi
+    const showKafkaTab = {initialTab: ConnectView.kafkaConnect, redpandaConnectTab: ConnectView.RedpandaConnect}
+    const showRedpandaConnectTab = {initialTab: ConnectView.RedpandaConnect, redpandaConnectTab: ConnectView.RedpandaConnect}
```
malinskibeniamin, line 48: Nothing to say about this change but overall we desperately need a linter 😓 I will try to put something together when I have spare 5 mins

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -27,9 +27,50 @@ import RpConnectPipelinesList from '../rp-connect/Pipelines.List';
...
+        return showKafkaTab;
+    }
+
+    switch (defaultView) {
+        case 'kafka-connect':
+            return showKafkaTab;
+        case 'redpanda-connect':
+            return showRedpandaConnectTab;
+        case 'redpanda-connect-secret':
+            return {initialTab: ConnectView.RedpandaConnect, redpandaConnectTab: ConnectView.RedpandaConncetSecret};
```
malinskibeniamin, line 59: `RedpandaConncetSecret` -> `RedpandaConnectSecret`

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -27,9 +27,50 @@ import RpConnectPipelinesList from '../rp-connect/Pipelines.List';
...
+        default:
+            return showRedpandaConnectTab;
+    }
+}
+
+const WrapUseSearchParamsHook: FunctionComponent<{matchedPath: string}> = (props) => {
+    const {search}= useLocation();
+    const searchParams = new URLSearchParams(search);
+    const defaultTab = searchParams.get('defaultTab') || '';
+    return <KafkaConnectOverview defaultView={defaultTab} {...props}/>
```
malinskibeniamin, line 69: What happens if `defaultTab` is `''`? I think we need to return default to `ConnectView.RedpandaConnect` tab by default?

frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
```diff
@@ -156,7 +164,60 @@ export default RpConnectPipelinesCreate;
...
+    const addCustomAutocomplete = (monaco: Monaco) => {
+        monaco.languages.registerCompletionItemProvider('yaml', {
+            // Display suggestions for lines beginning with ${ or . to assist with variable interpolation
+            triggerCharacters: ['${', '.'],
+            provideCompletionItems(model: editor.ITextModel, position: Position): languages.ProviderResult<languages.CompletionList> {
+                const wordInfo = model.getWordUntilPosition(position);
+                const previousInfo = model.getWordUntilPosition({lineNumber: position.lineNumber, column: wordInfo.startColumn - 1});
+                const range = {
+                    startLineNumber: position.lineNumber,
+                    endLineNumber: position.lineNumber,
```
malinskibeniamin, line 180: It's always going to start and end on the same line, is that intentional?

frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
```diff
@@ -156,7 +164,60 @@ export default RpConnectPipelinesCreate;
...
+                    startColumn: wordInfo.startColumn,
+                    endColumn: wordInfo.endColumn,
+                };
+                const last_chars = model.getValueInRange({startLineNumber: position.lineNumber, startColumn: 0, endLineNumber: position.lineNumber, endColumn: position.column});
+                const words = last_chars.replace('\t', '').replace('\{', '').split(' ');
+                const active_typing = words[words.length - 1];
+                const empty = {suggestions: []}
```
malinskibeniamin, line 187: Maybe we can continue using `camelCase`

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+        const alreadyExists = (rpcnSecretManagerApi.secrets || []).any(x => x.id == this.id);
+        const isIdEmpty = this.id.trim().length == 0;
+        const isSecretEmpty = this.secret.trim().length == 0;
+
+        return (
+            <PageContent>
+                <ToastContainer/>
+                <Flex flexDirection="column" gap={5}>
+                    <FormField label="Secret name" isInvalid={alreadyExists} errorText="Secret name is already in use">
```
malinskibeniamin, line 95: Is that always going to be the reason for the error? What if some invalid character is used? Wouldn't it show wrong error text?

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
```diff
@@ -0,0 +1,172 @@
...
+        if (!Features.pipelinesApi) return;
+
+        rpcnSecretManagerApi.refreshSecrets(force)
+            .catch((err) => {
+                if (String(err).includes('404')) {
+                    // Hacky special handling for OSS version, it is possible for the /endpoints request to not complete in time for this to render
+                    // so in this case there would be an error shown because we were too fast (with rendering, or the req was too slow)
+                    // We don't want to show an error in that case
+                    return;
+                }
```
malinskibeniamin, line 51: I think maybe it sounds worse than it is, but thanks for explaining 😬

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
```diff
@@ -0,0 +1,172 @@
...
+                    return false;
+                } catch {
+                    return false;
+                }
+            });
+
+        return (
+            <PageContent>
+                <Section>
+                    <ToastContainer/>
```
malinskibeniamin, line 89: Do we need ToastContainer copied everywhere? There's no global provider?

frontend/src/components/pages/rp-connect/secrets/Secrets.Update.tsx
```diff
@@ -0,0 +1,134 @@
...
+    @observable isUpdating = false;
+
+    constructor(p: any) {
+        super(p);
+        makeObservable(this, undefined, {autoBind: true});
+    }
+
+    initPage(p: PageInitHelper) {
+        p.title = 'Update Secret';
+        p.addBreadcrumb('Redpanda Connect Secret Manager', '/rp-connect/secret/update');
```
malinskibeniamin, line 28: Note: I think in the future we should have a global place with all the routes to ensure we can easily change name/forbid accessing it across the app if we needed to. Just in the router file there should be a place to get all the route paths.

frontend/src/components/pages/rp-connect/secrets/Secrets.Update.tsx
```diff
@@ -0,0 +1,134 @@
...
+        return (
+            <PageContent>
+                <ToastContainer/>
+                <Flex flexDirection="column" gap={5}>
+                    <FormField label="Secret name">
+                        <Flex alignItems="center" gap="2">
+                            <Input
+                                placeholder="Enter a secret name..."
+                                data-testid="secretId"
+                                pattern="[a-zA-Z0-9_\-]+"
```
malinskibeniamin, line 94: Maybe we can reuse this regex for secrets, actually I see this pattern is different
`^[A-Z][A-Z0-9_]*$` vs `[a-zA-Z0-9_\-]+`, which one should we use?

frontend/src/state/ui.ts
```diff
@@ -234,6 +234,10 @@ const defaultUiSettings = {
         quickSearch: ''
     },
 
+    rpncSecretList: {
```
malinskibeniamin, line 237: I think you meant
`rpcnSecretList` -> `rpncSecretList`

frontend/src/components/pages/connect/Overview.tsx
andresaristizabal: fix

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -27,9 +27,50 @@ import RpConnectPipelinesList from '../rp-connect/Pipelines.List';
 import { RedpandaConnectIntro } from '../rp-connect/RedpandaConnectIntro';
 import { Features } from '../../../state/supportedFeatures';
 import {isServerless} from '../../../config';
+import RpConnectSecretsList from '../rp-connect/secrets/Secrets.List';
+import { useLocation } from 'react-router-dom';
+
+enum ConnectView {
+    kafkaConnect = 'kafka-connect',
```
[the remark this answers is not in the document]
andresaristizabal: fix

frontend/src/components/pages/connect/Overview.tsx
andresaristizabal: it should default from `getDefaultView()`

frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
andresaristizabal: yes, it is intencional, but I disabled autocomplete, because it needs more testing and there isn't time now

andresaristizabal: remove this function for now, I need to improve the autocomplete

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
andresaristizabal: I copy and past this handle.

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
andresaristizabal: Again, I use pipeline as model, we can improve it later, when we have all console context

frontend/src/components/pages/rp-connect/secrets/Secrets.Update.tsx
andresaristizabal: I don't know what was the reason for this change, we can discuss it later

frontend/src/components/pages/rp-connect/secrets/Secrets.Update.tsx
andresaristizabal: I updated this based on proto

frontend/src/state/ui.ts
andresaristizabal: right

frontend/src/components/pages/rp-connect/secrets/Secrets.Update.tsx
weeco: Should it be secret or secrets?

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
```diff
@@ -0,0 +1,142 @@
...
+import {formatPipelineError} from '../errors';
+import {CreateSecretRequest, Scope} from '../../../../protogen/redpanda/api/dataplane/v1alpha2/secret_pb';
+
+const {ToastContainer, toast} = createStandaloneToast();
+
+const returnSecretTab = '/connect-clusters?defaultTab=redpanda-connect-secret'
+
+
+@observer
+class RpConnectSecretCreate extends PageComponent {
```
jvorcak, line 18: When we try to load these new pages and RP-connect is not enabled, we get infinite loading, this applies for both new URLs that have been added to the router. Can we please show something that prompts the user to configure it?

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
```diff
@@ -0,0 +1,142 @@
...
+    initPage(p: PageInitHelper) {
+        p.title = 'Create Secret';
+        p.addBreadcrumb('Redpanda Connect Secret Manager', '/rp-connect/secret/create');
+        p.addBreadcrumb('Create Secret', '');
+
+        this.refreshData(true);
+        appGlobal.onRefresh = () => this.refreshData(true);
+    }
+
+    refreshData(_force: boolean) {
```
jvorcak, line 38: very nit: maybe we could keep the convention and just use `force` to be consistent with other pages

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
```diff
@@ -0,0 +1,142 @@
...
+    @observable isCreating = false;
+
+    constructor(p: any) {
+        super(p);
+        makeObservable(this, undefined, {autoBind: true});
+    }
+
+    initPage(p: PageInitHelper) {
+        p.title = 'Create Secret';
+        p.addBreadcrumb('Redpanda Connect Secret Manager', '/rp-connect/secret/create');
```
jvorcak, line 31: are we sure about these shortcuts in the URLs? wouldn't `/redpanda-connect/secret/create` be nicer?

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
```diff
@@ -0,0 +1,142 @@
...
+        this.secret = '';
+        this.id = '';
+        appGlobal.history.push(returnSecretTab);
+    }
+
+    async createSecret() {
+        this.isCreating = true;
+
+        //create function given string return base64 encoded Uint8Array
+        function base64Encode(str: string): Uint8Array {
```
jvorcak, line 52: maybe we could move this function to utils?

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
```diff
@@ -0,0 +1,142 @@
...
+            .finally(() => {
+                this.isCreating = false;
+            });
+    }
+
+    render() {
+        if (!rpcnSecretManagerApi.secrets) return DefaultSkeleton;
+
+        const alreadyExists = (rpcnSecretManagerApi.secrets || []).any(x => x.id == this.id);
+        const isIdEmpty = this.id.trim().length == 0;
```
jvorcak, line 88: Let's use `===` just to make the usage of it through the console more common and standardised, there are corner cases where `==` behaves differently and I've found (and fixed) some already in the past. I know it's not the case here.

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
```diff
@@ -0,0 +1,191 @@
...
+import {uiSettings} from '../../../../state/ui';
+import {Link} from 'react-router-dom';
+import {PencilIcon, TrashIcon} from '@heroicons/react/outline';
+import EmptyConnectors from '../../../../assets/redpanda/EmptyConnectors.svg';
+import {DeleteSecretRequest, Secret} from '../../../../protogen/redpanda/api/dataplane/v1alpha2/secret_pb';
+
+const {ToastContainer, toast} = createStandaloneToast();
+
+const CreateSecretButton = () => {
+    return (<Box style={{display: 'flex', marginBottom: '.5em'}}>
```
jvorcak, line 19: When we migrated to Chakra, we tried to make most of the code consistent and utilising chakra's args instead of inline styles. It'd be nice if we could stick to that.

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <Link to={'/rp-connect/secret/create'}><Button>Create secret</Button></Link>
```
jvorcak, line 20: This is wrong when it comes to a11y, having a button instead of a link is semantically wrong. I think something like `<Button as={ReactRouterLink} to="...">` would be better

frontend/src/components/pages/connect/Overview.tsx
```diff
@@ -89,14 +128,14 @@ class KafkaConnectOverview extends PageComponent {
                             ? tabs[0].content()
                             : tabs[0].content
                     )
-                    : <Tabs tabs={tabs} defaultSelectedTabKey={showPipelines ? 'redpandaConnect' : 'kafkaConnect'} />
+                    : <Tabs tabs={tabs} defaultSelectedTabKey={getDefaultView(this.props.defaultView).initialTab} />
```
jvorcak, line 131: I'd personally try to have tabs support different URLs for us to be able to link to the pages. I've re-used some pattern in Settings from some other page. Maybe we could apply here as well?

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
andresaristizabal: complete agree, I copied this logic 😅, I also like `===` in order to avoid cohercion types

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
andresaristizabal: complete agree, again, I used Pipeline page as example for making this component, let me fix it

frontend/src/components/pages/connect/Overview.tsx
andresaristizabal: could you point me to that logic?

andresaristizabal: one of my principal goal is to share url with tab in mind

bojand:
Maybe it doesn't have to be part of this PR, but it's related, is there any way we can add the well known predefined contextual env vars to auto completion. The env vars added in https://github.com/redpanda-data/cloudv2/pull/18566.
Defined:
https://github.com/redpanda-data/cloudv2/blob/main/apps/redpanda-connect-api/config/config.go#L229
https://github.com/redpanda-data/cloudv2/blob/main/apps/redpanda-connect-api/config/config.go#L382-L391

andresaristizabal:
> Maybe it doesn't have to be part of this PR, but it's related, is there any way we can add the well known predefined contextual env vars to auto completion. The env vars added in [redpanda-data/cloudv2#18566](https://github.com/redpanda-data/cloudv2/pull/18566). Defined: https://github.com/redpanda-data/cloudv2/blob/main/apps/redpanda-connect-api/config/config.go#L229 https://github.com/redpanda-data/cloudv2/blob/main/apps/redpanda-connect-api/config/config.go#L382-L391

Good idea, we need to check how we can add it on autocomplete, I disabled autocomplete because I want to spend more time on it

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
andresaristizabal: could you explain me? 🤔 , I get this from pipeline page

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
andresaristizabal: Thank you, I noticed there was a function like this on util

frontend/src/components/pages/rp-connect/secrets/Secrets.List.tsx
andresaristizabal: fixed

frontend/src/components/pages/connect/Overview.tsx
jvorcak: solved in a DM

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
jvorcak: I mean if it's already there, we can just keep it, I just prefer nice readable URLs instead of a shortcuts.

frontend/src/components/pages/rp-connect/secrets/Secrets.Update.tsx
andresaristizabal: it should be secrets, I fixing it

frontend/src/components/pages/rp-connect/secrets/Secrets.Create.tsx
andresaristizabal: when it happens, rpcn-secret continue working:

https://redpandadata.slack.com/archives/C074X8HD68L/p1732568383289989?thread_ts=1732294471.397329&cid=C074X8HD68L

malinskibeniamin (approved):

jvorcak (approved):

PR #1535, opened 2024-11-28
Fix/update rpnc secret

andresaristizabal:
Fix update secret


frontend/package.json
```diff
@@ -19,6 +19,7 @@
         "test": "jest",
         "test-interpreter-code": "ts-node --project src/utils/interpreter/tsconfig.json src/utils/interpreter/findFunction.test.ts",
         "lint": "biome check src/* --write --unsafe",
+        "linter": "biome.ci.src/*",
```
malinskibeniamin, line 22: should be
```suggestion
        "linter": "biome ci src/*",
```

malinskibeniamin (approved):

PR #1537, opened 2024-11-29
Fixes a rendering error if schemaVersion is requested by not supported

jvorcak:


malinskibeniamin (approved):
Glad we made it in the end! Awesome work

PR #1543, opened 2024-12-02
License color change in certain cases

jvorcak:


frontend/src/components/license/licenseUtils.spec.ts
```diff
@@ -94,7 +94,7 @@ describe('licenseUtils', () => {
   });
 
   describe('prettyExpirationDate', () => {
-    test('should return a formatted expiration date for an expiring license', () => {
```
jvorcak, line 97: I'll be fixing and adding more unit tests in another PR

bojand (approved):

PR #1546, opened 2024-12-02
fix(backend): make sure license type is mapped correctly

sago2k8:
Make sure license type is mapped correctly in console


r-vasquez (approved):
LGTM

bojand (approved):

PR #1547, opened 2024-12-02
Adds NullFallbackBoundary around license banners

jvorcak:
Introduced `NullFallbackBoundary` for silently failing, we shouldn't affect the functionality in case there would be some problem with the code


malinskibeniamin (approved):
Do you think it's worth using some generic Redpanda error page instead of rendering null or are you ok with not showing specific sections of the page where the error might happen?

jvorcak:
> Do you think it's worth using some generic Redpanda error page instead of rendering null or are you ok with not showing specific sections of the page where the error might happen?

In most of the cases we show this Redpanda generic error. But when it comes to license enforcements, it's mostly hints to the users and I'd like not to block the UX for them in case something goes wrong.

PR #1557, opened 2024-12-10
Connect/add quick add

andresaristizabal:
Add quick add secret in pipeline editor

https://github.com/user-attachments/assets/1f22fe3f-4667-4e85-8e3c-4023c6c2119d

### DIFF
--- a/frontend/.editorconfig
+++ b/frontend/.editorconfig
@@ -12,7 +12,7 @@ max_line_length = off
 
 [*.{ts,tsx}]
 indent_style = space
-indent_size = 4
+indent_size = 2
 
 [*.{diff,md}]
 trim_trailing_whitespace = false
--- a/frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
+++ b/frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
@@ -14,16 +14,20 @@ import {
   AlertIcon,
   Box,
   Button,
+  Link as ChLink,
   Flex,
   FormField,
+  Heading,
   Input,
   NumberInput,
   Text,
   createStandaloneToast,
+  useDisclosure,
 } from '@redpanda-data/ui';
-import { Link as ChLink } from '@redpanda-data/ui';
 import { action, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
+import { editor } from 'monaco-editor';
+import React, { useState } from 'react';
 import { Link } from 'react-router-dom';
 import { PipelineCreate } from '../../../protogen/redpanda/api/dataplane/v1alpha2/pipeline_pb';
 import { appGlobal } from '../../../state/appGlobal';
@@ -34,7 +38,9 @@ import PipelinesYamlEditor from '../../misc/PipelinesYamlEditor';
 import Tabs from '../../misc/tabs/Tabs';
 import { PageComponent, type PageInitHelper } from '../Page';
 import { formatPipelineError } from './errors';
+import { SecretsQuickAdd } from './secrets/Secrets.QuickAdd';
 import { MAX_TASKS, MIN_TASKS, tasksToCPU } from './tasks';
+
 const { ToastContainer, toast } = createStandaloneToast();
 
 const exampleContent = `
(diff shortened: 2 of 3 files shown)


frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
```diff
@@ -194,12 +200,46 @@ class RpConnectPipelinesCreate extends PageComponent<{}> {
...
+interface QuickActions {
+  editorInstance: editor.IStandaloneCodeEditor | null;
+}
+
+const QuickActions = ({ editorInstance }: QuickActions) => {
+  const { isOpen: isAddSecretOpen, onOpen: openAddSecret, onClose: closeAddSecret } = useDisclosure();
+
+  if (editorInstance === null) {
+    return <Box minW={300} />;
+  }
```
malinskibeniamin, line 212: Why not some Empty state?

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
```diff
@@ -0,0 +1,190 @@
...
+      return 'The name you entered is invalid. It must start with an letter (A–Z) and can only contain alphanumeric and underscores (_).';
+    }
+    if (id.length > 255) {
+      return 'The secret name must be fewer than 255 characters.';
+    }
+    return '';
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClose={closeModal} isCentered={true} size={'md'}>
```
malinskibeniamin, line 117: Is there any base modal we can reuse from the component library?

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const addSecret = async (id: string) => {
+    id = id.toUpperCase();
+    if (isNewSecret) {
+      setIsCreating(true);
+      const result = await rpcnSecretManagerApi
+        .create(
+          new CreateSecretRequest({
+            id: id,
+            // @ts-ignore js-base64 does not play nice with TypeScript 5: Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBuffer>'.
+            secretData: base64ToUInt8Array(encodeBase64(secret)),
```
malinskibeniamin, line 58: I left this comment in another component, can you please check if it continues to fail or maybe we can get rid of the ignore?

malinskibeniamin (commented):
The recording did not finish uploading

frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
andresaristizabal: Because if we don't get the editor instance, any of these operations can be applied, and also this case should have no impact on the customer experience, this case is not a break.

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
andresaristizabal: yes it continue failing

frontend/.editorconfig
```diff
@@ -12,7 +12,7 @@ max_line_length = off
 
 [*.{ts,tsx}]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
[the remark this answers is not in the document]
andresaristizabal: No, this is our code configuration. In my case, when I add a new line, it uses this configuration, but the biome is fixed when I save the file.

andresaristizabal:
> The recording did not finish uploading

Fixed

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const isNameValid = (id: string) => {
+    if (id === '') {
+      return '';
+    }
+    if (!/^[A-Za-z][A-Za-z0-9_]*$/.test(id)) {
+      return 'The name you entered is invalid. It must start with an letter (A–Z) and can only contain alphanumeric and underscores (_).';
+    }
+    if (id.length > 255) {
+      return 'The secret name must be fewer than 255 characters.';
+    }
```
malinskibeniamin, line 112: You can do this with `.max(255)` from Yup but that works too

malinskibeniamin (approved):

weeco:
@ivpanda @andresaristizabal Can we add a list with all secret references that were created + a copy button? If you happen to lose the reference when editing the JSON, they are kind of lost I assume? I think that could be a quick/small quality of life improvement.

weeco (approved):

PR #1558, opened 2024-12-10
rpcn: fix display of validation errors

nicolaferraro:
This provides more useful info e.g. when the pipeline name contains invalid characters.


weeco:
Is this still valid @nicolaferraro ?

nicolaferraro:
> Is this still valid @nicolaferraro ?

I forgot about it. Yes, it is. That's why errors are still displayed as generic...

andresaristizabal (approved):

PR #1572, opened 2025-01-03
Topic config modal uses react-hook-form

jvorcak:
https://www.loom.com/share/23e06762b02c4eeb9ceb2721850991b1

### DIFF
--- a/frontend/src/components/pages/topics/TopicConfiguration.tsx
+++ b/frontend/src/components/pages/topics/TopicConfiguration.tsx
@@ -1,12 +1,11 @@
-import { AdjustmentsIcon } from '@heroicons/react/outline';
 import { PencilIcon } from '@heroicons/react/solid';
 import {
   Alert,
-  AlertDescription,
   AlertIcon,
   Box,
   Button,
   Flex,
+  FormField,
   Icon,
   Input,
   Modal,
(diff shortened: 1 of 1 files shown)


weeco:
@mattschumpert We should have a look at all topic configs and put them into categories. We did that once, but haven't updated/set the categories anymore after we added further configs in Redpanda.

cc @mo-redpanda for a review

andresaristizabal:
Can you add a video for this change?

mattschumpert:
@weeco you mean categorizing the ones in 'other'?.  Whats the easiest way to supply this categorization?

jvorcak:
> Can you add a video for this change?

@andresaristizabal there is video in the description, or do you mean something else?

frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -361,89 +360,73 @@ function isTopicConfigEdittable(
...
+      return (
+        <SingleSelect<T>
+          options={[
+            { value: 'false' as T, label: 'False' },
+            { value: 'true' as T, label: 'True' },
+          ]}
+          value={value}
+          onChange={onChange}
+        />
+      );
```
malinskibeniamin, line 381: Any reason why we are not just using a Checkbox/Toggle/Switch components here?

malinskibeniamin: I realise it's code from before but just curious if you may know why

malinskibeniamin (approved):

frontend/src/components/pages/topics/TopicConfiguration.tsx
jvorcak: I don't know, it's a bit out of scope of this PR, but I'm happy to refactor this if we want to have this cleaner.

PR #1578, opened 2025-01-07
Fix setting of custom BASE_URL in index.html

YannikRudolph0108:
Since a default base tag was added to the index.html, the mechanism for setting a custom BASE_URL is broken. When multiple base tags are present in a html file, only the first tag is used (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base#multiple_base_elements). Therefore, we need to adapt the existing base tag and not append a newly created one.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -34,13 +34,12 @@
                 return window.location;
             };
             let l = window.__getLocation();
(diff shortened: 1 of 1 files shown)


jvorcak (approved):
This is great - thank you!

PR #1580, opened 2025-01-08
Show default config for topic configurations

jvorcak:
- We needed to remove altering data structures in the backendAPI (moved to ConfigList where it was needed)
- We added an information about default value to the topic config form

Screenshot 2025-01-08 at 14 52 27

### DIFF
--- a/frontend/src/components/pages/topics/TopicConfiguration.tsx
+++ b/frontend/src/components/pages/topics/TopicConfiguration.tsx
@@ -127,6 +127,8 @@ const ConfigEditorForm: FC<{
 
   const valueType = watch('valueType');
 
+  const defaultConfigSynonym = editedEntry.synonyms?.find((x) => x.source === 'DEFAULT_CONFIG');
+
   return (
     <Modal isOpen onClose={onClose}>
       <form onSubmit={handleSubmit(onSubmit)}>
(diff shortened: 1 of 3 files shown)


frontend/src/components/pages/topics/TopicConfiguration.tsx
```diff
@@ -127,6 +127,8 @@ const ConfigEditorForm: FC<{
 
   const valueType = watch('valueType');
 
+  const defaultConfigSynonym = editedEntry.synonyms?.find((x) => x.source === 'DEFAULT_CONFIG');
```
weeco, line 130: That's not always correct. A topic config may be inherited from:

- Default Config (defined in Redpanda / Apache Kafka if nothing is set)
- A default for all topics is set at the broker/cluster level (STATIC_BROKER_CONFIG / DYNAMIC_DEFAULT_BROKER_CONFIG / DYNAMIC_BROKER_CONFIG), basically to override the default config

or it is set at the Topic level (DYNAMIC_TOPIC_CONFIG) which is the custom/override case.

jvorcak: @weeco in that case FE doesn't have this information. On this place we show this information only for default tab, so we don't care about a topic level intentionaly. However, if not set at a topic level, it can be set at multiple levels above, is there a hierarchy for this so that I can determine the default from synonyms array?

weeco: Yes there's a hierarchy, it should be this order (top to bottom):

- DYNAMIC_TOPIC_CONFIG
- DYNAMIC_BROKER_CONFIG
- DYNAMIC_DEFAULT_BROKER_CONFIG
- STATIC_BROKER_CONFIG
- DEFAULT_CONFIG

jvorcak: Screenshot 2025-01-09 at 15 47 37

weeco (approved):

PR #1602, opened 2025-01-29
Update tag for User Service

kbatuigas:
Give that we are about to [expose IAM endpoints in the Control Plane](https://github.com/redpanda-data/cloudv2/pull/20537/files#diff-8cc6ab814f958996301a7db8a9142c8e8adc38525c7ebf6cad475db9c3a06b44), which also has the concepts/ entities for users, roles, etc, we can update the OpenAPI spec to specify that the endpoints here are used to interact with users within the data plane. 

As of now there is an "unofficial" consensus that we refer to the users, etc in the control plane as "Control Plane Users" etc - that is, we add the qualifier "control plane" - see also  [Slack thread](https://redpandadata.slack.com/archives/C05R2EEA1T6/p1737641983093419?thread_ts=1737476584.262319&cid=C05R2EEA1T6) So I think the quickest solution for the time being, while we hash it out or perhaps get Product's guidance, is to just also use the label or qualifier "data plane" here.

cc @sago2k8


sago2k8 (approved):

weeco:
@kbatuigas @sago2k8 Eventually we should come up with something better. This API should/will become available for self-hosted users where the concept of Dataplane / Controlplane doesn't exist (or irrelevant to them). For this audience the terminology will be confusing. I'm still in favour of separating the docs more between controlplane and dataplane as suggested in Slack

sago2k8:
> more

I agree with this, should we keep it as User no prefix ? The risk is low (imo) since we have different documentation files

kbatuigas:
> @kbatuigas @sago2k8 Eventually we should come up with something better. This API should/will become available for self-hosted users where the concept of Dataplane / Controlplane doesn't exist (or irrelevant to them). For this audience the terminology will be confusing. I'm still in favour of separating the docs more between controlplane and dataplane as suggested in Slack

@weeco @sago2k8 Yup, the split is underway: https://github.com/redpanda-data/docs/pull/929 This should be merged this week. 

Previews of split reference:
- [Control Plane API](https://deploy-preview-929--redpanda-docs-preview.netlify.app/api/cloud-controlplane-api/#overview)
- [Data Plane API](https://deploy-preview-929--redpanda-docs-preview.netlify.app/api/cloud-dataplane-api/#overview)

I need to merge a couple of recent commits from cloudv2/console, but this is more or less what we'll have in the docs. You'll see though that the "Data Plane" descriptor is only in the page title and the overview section. I think it's better to also add it to this User service so that in the event a reader stumbles on this with no other context, they'll at least know which plane and which type of users this API interacts with. 

I do agree that we can do more to explain control plane vs data plane. We have a little bit in the Cloud docs [overview](https://docs.redpanda.com/redpanda-cloud/get-started/cloud-overview/#redpanda-cloud-architecture) and Cloud API overviews, but definitely with regards to "users" and "roles" we have to flesh that out more. It's a conversation we've (Michele and I from Docs) have started with Denis, Mo, Santi because of Cloud RBAC and the introduction of the IAM endpoints in the Control Plane. But that's still ongoing and we've yet to finalize the best "story" behind these concepts.

But in the meantime, adding the respective CP vs DP descriptors at least to the services helps to wave a flag.

weeco:
I believe we added this across all endpoints already in a newer PR. So closing this.

PR #1603, opened 2025-01-30
api: fix schema reference for create topic response type

bojand:


weeco (approved):

kbatuigas (approved):
lgtm!

proto/redpanda/api/dataplane/v1alpha2/topic.proto
```diff
@@ -246,7 +246,7 @@ service TopicService {
         value: {
           description: "Topic created"
           schema: {
-            json_schema: {ref: ".redpanda.api.dataplane.v1alpha2.Topic"}
```
sago2k8, line 249: Based on the line https://github.com/redpanda-data/console/pull/1603/files#diff-b77c277ef1c6ab384b61eab06ad8168e65586171bdcd714658d16faf756dafb5L239 I don't think the response is correct.

PR #1604, opened 2025-01-30
Add API_CONSOLE API type for permissions

bojand:
Also fixes some test and linting issues


backend/pkg/serde/protobuf_schema.go
```diff
@@ -102,9 +102,9 @@ func (d ProtobufSchemaSerde) DeserializePayload(ctx context.Context, record *kgo
 
 	// Marshal proto message into JSON
 	o := protojson.MarshalOptions{
-		UseProtoNames: true, // use snake_case
+		UseProtoNames: false, // use loweCamelCase
 		// Do not use EmitUnpopulated, so we don't emit nulls (they are ugly, and provide no benefit. they transport no information, even in "normal" json).
-		EmitUnpopulated: false,
+		EmitUnpopulated: true,
```
bojand, line 107: These changes keep the functionality and behaviour same as is in current master

backend/pkg/serde/service_integration_test.go
```diff
@@ -1258,8 +1258,8 @@ func (s *SerdeIntegrationTestSuite) TestDeserializeRecord() {
 		obj, ok := (dr.Value.DeserializedPayload).(map[string]any)
 		require.Truef(ok, "parsed payload is not of type map[string]any")
 		assert.Equal("345", obj["id"])
-		assert.Len(obj["dec_val"], 1)
-		assert.Len(obj["color"], 3)
+		assert.Len(obj["decVal"], 1)
+		assert.Len(obj["color"], 4)
```
bojand, line 1262: These are adjusted to reflect existing behaviour and functionality as is in master branch.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
```diff
@@ -381,7 +381,7 @@ service PipelineService {
     };
     option (redpanda.api.auth.v1.authorization) = {
       required_permission: PERMISSION_VIEW
-      api: API_UNSPECIFIED
+      api: API_CONSOLE
```
weeco, line 384: I believe all of these pipeline protos would be "REDPANDA_CONNECT" and not Console because that's the upstream API we may want to have user impersonation for at some point. Since we do not even have authentication for RP Connect at the moment it doesn't matter though

backend/pkg/serde/protobuf_schema.go
weeco: Since we cut a new major release we would be allowed of doing so. Imo that would be a slight improvement over what we currently do, since it matches the proto field names more closely. WDYT?

weeco (approved):

backend/pkg/serde/protobuf_schema.go
bojand: > since it matches the proto field names more closely. WDYT?

Regarding JSON field names and casing, the [default JSON representation](https://protobuf.dev/programming-guides/json/) _is_ `lowerCamelCase`. 

Similarly regarding emitting unpopulated fields. Default behaviour is to _omit_ those fields.

I do not feel strongly on this but part of me thinks it's going against more standard convention and defaults. 
I can be convinced otherwise especially if it's a new major release, it relaxes some of the compatibility issue I brought up.

proto/redpanda/api/dataplane/v1alpha2/pipeline.proto
bojand: Yup was curious about this... Changed the Pipeline API type to `API_REDPANDA_CONNECT`.

backend/pkg/serde/protobuf_schema.go
weeco: Okay fine for me to change as you suggested 👍

PR #1605, opened 2025-01-31
fix: do not return repeated directly

sago2k8:
Use the complete response when the response is a list.

for some method the example does not match the actual response and we probably want to ensure the response is not nested. maybe the question is around breaking the API, so happy to do it for V1


proto/redpanda/api/dataplane/v1alpha2/topic.proto
```diff
@@ -292,10 +292,7 @@ service TopicService {
     };
   }
   rpc GetTopicConfigurations(GetTopicConfigurationsRequest) returns (GetTopicConfigurationsResponse) {
-    option (google.api.http) = {
-      get: "/v1alpha2/topics/{topic_name}/configurations"
-      response_body: "configurations"
-    };
+    option (google.api.http) = {get: "/v1alpha2/topics/{topic_name}/configurations"};
```
weeco, line 295: Should we do it, despite knowing the number of entries is very limited and will never exceed an unreasonable number that couldn't be retrieved in one go?

proto/redpanda/api/dataplane/v1alpha2/topic.proto
```diff
@@ -292,10 +292,7 @@ service TopicService {
     };
   }
   rpc GetTopicConfigurations(GetTopicConfigurationsRequest) returns (GetTopicConfigurationsResponse) {
```
weeco, line 294: Wondering if we also need to rename this from Get to List accordingly @sago2k8 @bojand . One could argue for both cases. It could be argued as one resource (the configuration of a topic which happens to have several properties) or a list of resources (configurations for topics)

bojand: Good observation. Maybe `ListTopicConfigurations()` is a better choice. Agree it kinda depends on how you look at it. But I do not feel strongly about it.

weeco:
I guess we can close this, since V1 is already released?

Issue #1536, opened 2024-11-29
Copy context menu items on message do not work


paul-green:
The **Copy ..** menu items on each message do not work;

- Copy Message
- Copy Key
- Copy Value
- Copy Epoch Timestamp

I tried each of these and nothing is put in the clipboard

jvorcak:
As far as I can check it should work now, feel free to re-open in case you can still experience this issue.

PR #1606, opened 2025-02-03
Fixes a bug when downloading messages

jvorcak:


jvorcak:
Fixes https://github.com/redpanda-data/console/issues/1536#event-16125815205

I verified that download does't work, but as for copy to clipboard, everything works for me on all the platforms, browsers, etc.

PR #1614, opened 2025-02-10
proto: add dataplane v1 api

bojand:
The only change is `v1alpha2` -> `v1`


weeco:
There are some compile errors in the backend after this change and I assume the frontend also needs some updates of references (pipeline types changed to v1). If we merge it as is, master is not buildable

bojand:
I reverted one commit that caused compile errors... But I am confused by these new lint warnings. I do not get them locally. Also actual (non-generated) Go files were changed...

weeco (approved):

PR #1625, opened 2025-02-25
proto: add SecretService for Console package

bojand:
I also changed the frontend as far as I know... People should review thoroughly and double check.


frontend/src/state/backendApi.ts
```diff
@@ -2112,35 +2112,41 @@ export const rpcnSecretManagerApi = observable({
...
+    const res = await client.listSecretScopes({ request: listSecretScopesRequest });
+
+    if (!res.response) {
+      this.isEnable = false;
+      return;
+    }
+
+    const isEnable = res.response?.scopes.some((scope) => scope === Scope.REDPANDA_CONNECT);
     this.isEnable = isEnable;
     return isEnable;
```
malinskibeniamin, line 2142: I feel like we can improve the `isEnable` flag is set, but it's ok to leave as-is for this PR, there are more refactors we may do later on

PR #1628, opened 2025-02-28
proto: change Connect secrets APIs to specify KafkaConnect

bojand:
I think this is a little bit of an improvement to make the RPC methods and message types explicit in their name for their purpose for Kafka Connect.


weeco:
I'm wondering if we happen to do this - shouldn't we move this straight into the Kafka Connect service? Should we even still expose a generic secret store as an API? I feel like we did a mistake there - the generic secret store should only exist under the hood, but via API we only have a secret store for RP Connect and one for Kafka Connect. What are your thoughts?

bojand:
The generic secrets API were re-added when we implemented secret management for Redpanda Connect. 

I think there is also a plan to use it for user managed cluster configuration within the dataplane.

So those endpoints are required and are included by design.

I am not sure about moving Kafka Connect secrets to a different service. Technically it's just management of secrets so kinda belongs here. This was meant to make naming more explicit. But we can move it too. I don't have strong opinions.

weeco:
> I think there is also a plan to use it for user managed cluster configuration within the dataplane.
> So those endpoints are required and are included by design.

I remember, but I'm not sure if that was the right decision to have one generic secret store since we want to namespace these secrets. For example we probably don't want secrets from the cluster configs show up for selection for Redpanda connect pipelines?

I also remember for kafka connect we had to do the distinction because we put all the kafka connect secrets in one specific location, so that we were able to setup strict permissions (e.g. kafka connect pods were only able to read secrets from within that directory). 

Possibly it's a larger change that we don't have to do as part of v1 though.

weeco (approved):

PR #1633, opened 2025-03-04
Console V3

weeco:
- Support for new user impersonation & authorization system
- Lazy loading schemas from schema registry
- Large refactoring including the use of client factories for better integration with the enterprise repository
- Breaking config changes


backend/pkg/api/connect/integration/api_suite_test.go
```diff
@@ -90,7 +90,7 @@ func (s *APISuite) SetupSuite() {
 
 	s.testSeedBroker = seedBroker
 
-	// 3. Start Kafka Connect Docker container
+	// 3. Start Kafka KafkaConnect Docker container
```
bojand, line 93: Super minor... Is this intentional "Kafka KafkaConnect"?

backend/pkg/api/connect/integration/api_suite_test.go
```diff
@@ -162,7 +162,7 @@ func (s *APISuite) TearDownSuite() {
 	ctx, cancel := context.WithTimeout(context.Background(), 45*time.Second)
 	defer cancel()
 
-	// 1. Terminate Kafka Connect container
+	// 1. Terminate Kafka KafkaConnect container
```
bojand, line 165: "Kafka KafkaConnect" stuttering in comment

backend/pkg/api/connect/interceptor/endpoint_check.go
```diff
@@ -47,7 +47,7 @@ func NewEndpointCheckInterceptor(cfg *config.ConsoleAPI, logger *zap.Logger) *En
 	}
 }
 
-// WrapUnary creates an interceptor to validate Connect requests.
+// WrapUnary creates an interceptor to validate KafkaConnect requests.
```
bojand, line 50: I think this is not intentional ... "Connect" here refers to ConnectRPC

backend/pkg/api/connect/service/clusterstatus/redpanda.go
```diff
@@ -0,0 +1,87 @@
...
+				version = "Redpanda " + brokerVersion[0]
+				break
+			}
+		}
+	}
+	return version
+}
+
+// ToPartitionBalancerViolationsProto maps a PartitionBalancerViolations Go type to its proto representation.
+func (*redpandaStatusChecker) partitionBalancerViolationsToProto(in rpadmin.PartitionBalancerViolations) *consolev1alpha1.GetRedpandaPartitionBalancerStatusResponse_Violations {
```
bojand, line 42: Minor nit, comment does not match function name. Seems like originally at some point in the implementation this was public and got changed to private.

backend/pkg/api/connect/service/clusterstatus/redpanda.go
```diff
@@ -0,0 +1,87 @@
...
+	if in.PartitionsPendingForceRecovery != nil {
+		v := int32(*in.PartitionsPendingForceRecovery)
+		out.PartitionsPendingForceRecoveryCount = &v
+	}
+
+	return out
+}
+
+// mapBalancerStatus converts the status string from the Go type to the proto enum.
+func (*redpandaStatusChecker) balancerStatusStringToProtoEnum(status string) consolev1alpha1.GetRedpandaPartitionBalancerStatusResponse_Status {
```
bojand, line 72: comment should have `balancerStatusStringToProtoEnum` as function name

backend/pkg/api/connect/service/clusterstatus/service.go
```diff
@@ -0,0 +1,332 @@
...
+	v1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/console/v1alpha1"
+	"github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/console/v1alpha1/consolev1alpha1connect"
+	v1alpha2 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/dataplane/v1alpha2"
+	"github.com/redpanda-data/console/backend/pkg/version"
+)
+
+var _ consolev1alpha1connect.ClusterStatusServiceHandler = (*Service)(nil)
+
+// Service that implements the UserServiceHandler interface. This includes all
+// RPCs to retrieve cluster information about all connected APIs.
```
bojand, line 41: "UserServiceHandler" is not correct here heh

backend/pkg/api/connect/service/clusterstatus/service.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+			clusterVersion = versions.VersionGuess()
+		})
+
+		return nil
+	})
+
+	if err := grp.Wait(); err != nil {
+		return nil, apierrors.NewConnectError(
+			connect.CodeInternal,
```
bojand, line 123: Probably not a big deal...
Only the metadata call returns an error from the 2 goroutines... 
Would we be able to map that error to a more appropriate error perhaps?

Almost feels like maybe we need something like `NewConnectErrorFromKafkaErrorCode` but it accepts an error, and checks the rest (is unauthorized first, and then returns internal).

backend/pkg/api/connect/service/clusterstatus/service.go
```diff
@@ -0,0 +1,332 @@
...
+		users, err := redpandaCl.ListUsers(ctx)
+		if err != nil {
+			s.logger.Warn("failed to list users via redpanda admin api", zap.Error(err))
+		} else {
+			userCount = int32(len(users))
+		}
+		return nil
+	})
+
+	// No error is returned, errors are already logged inside each go routine
```
bojand, line 227: is there a reason why we are dropping errors?

backend/pkg/api/connect/service/clusterstatus/util.go
```diff
@@ -0,0 +1,28 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0
+
+package clusterstatus
+
+import consolev1alpha1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/console/v1alpha1"
+
+func setStatus(s *consolev1alpha1.ComponentStatus, status consolev1alpha1.StatusType, reason string) {
+	if status > s.Status {
```
bojand, line 15: I am not sure I understand this `>` check?

backend/pkg/api/connect/service/kafkaconnect/v1/service.go
```diff
@@ -7,7 +7,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0
 
-// Package kafkaconnect implements the KafkaConnect interface for the Connect API.
+// Package kafkaconnect implements the KafkaConnect interface for the KafkaConnect API.
```
bojand, line 10: I think the original "Connect" in the comment was for "ConnectRPC"

backend/pkg/api/connect/service/kafkaconnect/v1alpha2/service.go
```diff
@@ -7,7 +7,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0
 
-// Package kafkaconnect implements the KafkaConnect interface for the Connect API.
+// Package kafkaconnect implements the KafkaConnect interface for the KafkaConnect API.
```
bojand, line 10: "ConnectRPC"

backend/pkg/api/handle_kafka_connect_integration_test.go
```diff
@@ -57,7 +57,7 @@ func (s *APIIntegrationTestSuite) TestHandleCreateConnector() {
 	seedBroker, err := redpandaContainer.KafkaSeedBroker(ctx)
 	require.NoError(err)
 
-	// Kafka Connect container
+	// Kafka KafkaConnect container
```
bojand, line 60: "Kafka KafkaConnect"

backend/pkg/api/hooks.go
```diff
@@ -401,3 +181,107 @@ func (*defaultHooks) CanCreateRedpandaRoles(_ context.Context) (bool, *rest.Erro
 func (*defaultHooks) CanDeleteRedpandaRoles(_ context.Context) (bool, *rest.Error) {
 	return true, nil
 }
+
```
bojand, line 184: I wonder if it's worth putting `AuthenticationDefaultHandler` into its own file?

backend/pkg/config/config.go
```diff
@@ -34,12 +33,14 @@ type Config struct {
...
 
-	Console  Console        `yaml:"console"`
-	Redpanda Redpanda       `yaml:"redpanda"`
-	Connect  Connect        `yaml:"connect"`
-	REST     Server         `yaml:"server"`
-	Kafka    Kafka          `yaml:"kafka"`
-	Logger   logging.Config `yaml:"logger"`
+	Console        Console      `yaml:"console"`
+	Redpanda       Redpanda     `yaml:"redpanda"`
+	Connect        KafkaConnect `yaml:"kafkaConnect"`
```
bojand, line 38: Minor... naming is hard... Should we call the actual variable something else too?

backend/pkg/config/connect.go
```diff
@@ -15,18 +15,18 @@ import (
 	"time"
 )
 
-// Connect defines all configuration options for connecting to one or more
-// Kafka Connect clusters.
-type Connect struct {
+// KafkaConnect defines all configuration options for connecting to one or more
```
bojand, line 18: Should we rename the file to `kafka_connect.go`?

backend/pkg/config/connect_cluster.go
```diff
@@ -47,7 +47,7 @@ func (c *ConnectCluster) Validate() error {
 	}
 
 	if c.URL == "" {
-		return fmt.Errorf("url to access the Connect cluster API must be set")
+		return fmt.Errorf("url to access the KafkaConnect cluster API must be set")
```
bojand, line 50: Minor and not necessary perhaps, should we make `ConnectCluster` explicit about "Kafka Connect" and rename it to `KafkaConnectCluster` and maybe rename the file similarly?

backend/pkg/console/api_versions.go
```diff
@@ -29,23 +56,31 @@ type APIVersion struct {
...
+		return nil, err
+	}
+	req := kmsg.NewApiVersionsRequest()
+	req.ClientSoftwareVersion = version.Version
+	req.ClientSoftwareName = "RPConsole"
+
+	versionsRes, err := req.RequestWith(ctx, cl)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get kafka api version: %w", err)
+		return nil, fmt.Errorf("failed to get kafka api vers: %w", err)
```
bojand, line 69: "vers"?

backend/pkg/console/api_versions.go
```diff
@@ -29,23 +56,31 @@ type APIVersion struct {
...
 	}
 
 	err = kerr.ErrorForCode(versionsRes.ErrorCode)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get kafka api version. Inner error: %w", err)
+		return nil, fmt.Errorf("failed to get kafka api vers. Inner error: %w", err)
```
bojand, line 74: "vers"?

backend/pkg/console/api_versions.go
```diff
@@ -29,23 +56,31 @@ type APIVersion struct {
...
 	}
 
 	versions := make([]APIVersion, len(versionsRes.ApiKeys))
-	for i, version := range versionsRes.ApiKeys {
+	for i, vers := range versionsRes.ApiKeys {
```
bojand, line 78: I am not sure the benefit of this naming change?

backend/pkg/api/connect/service/rpconnect/service.go
```diff
@@ -45,22 +43,13 @@ func NewService(logger *zap.Logger, authHooks hooks.AuthorizationHooks) (*Servic
...
-	canView, restErr := s.hooks.CanViewConnectCluster(ctx, "redpanda")
-	err := apierrors.NewPermissionDeniedConnectError(canView, restErr,
-		"you are not allowed to call this endpoint",
-	)
-	if err != nil {
-		return nil, err
-	}
-
+func (s *Service) LintConfig(_ context.Context, req *connect.Request[v1alpha1.LintConfigRequest]) (*connect.Response[v1alpha1.LintConfigResponse], error) {
+	// TODO: Ensure is authenticated here!
```
bojand, line 52: What is this TODO exactly? Should we take care of it?

backend/pkg/api/handle_topics_integration_test.go
```diff
@@ -376,16 +297,18 @@ func (s *APIIntegrationTestSuite) TestHandleGetTopics() {
...
+			require.Len(allLogs, 1)
+			assert.Equal("config resource response has an error", allLogs[0].Message)
+			assert.Equal(zap.WarnLevel, allLogs[0].Level)
+			contextMap := allLogs[0].ContextMap()
+			logObj, ok := contextMap["error"]
+			assert.True(ok)
+			fieldValue, ok := logObj.(string)
+			assert.True(ok)
+			assert.Equal("INVALID_TOPIC_EXCEPTION: The request attempted to perform an operation on an invalid topic.", fieldValue)
+		*/
```
bojand, line 312: If there are new log assertions to verify behaviour lets add them?
If the behaviour is different now that there are no meaningful log assertions lets remove this commented out code.

backend/pkg/api/connect/interceptor/endpoint_check.go
weeco: Too much of connect in my head :D

backend/pkg/api/connect/service/clusterstatus/service.go
weeco: That is intentional. We always want these cluster status rpcs to respond, even if the cluster is broken. If we would return a connectrpc error we cannot really inform about the perceived cluster status. If the metadata request fails, we cannot even figure out the health of the cluster and it has to be assumed lost connection, so that's the only case in my mind that makes sense to return an error.

backend/pkg/api/connect/service/clusterstatus/util.go
weeco: The status is an enum / iota where a greater number means => worse condition.

1 = Healthy
2 = Degraded
3 = Unhealthy.

Because we only want to return one status + error messsage, we use this comparison to see if the status we are setting is worse than the status we have so far.

backend/pkg/api/hooks.go
weeco: I wouldn't mind. Currently I opted for the way to put all default hooks in a single file but we can split the auth ones out I guessß

backend/pkg/schema/client.go
```diff
@@ -1,793 +1,390 @@
...
 
-	res, err := req.Get(url)
-	if err != nil {
-		return nil, fmt.Errorf("get schema by id request failed: %w", err)
+// NewCachedClient initializes and returns a new CachedClient instance with the
+// provided schema client factory and cache namespace function. It sets up
+// caching with specific settings.
+func NewCachedClient(schemaClientFactory schema.ClientFactory, cacheNamespaceFn func(context.Context) (string, error)) (*CachedClient, error) {
+	cacheSettings := []cache.Opt{
+		cache.MaxAge(30 * time.Second),
```
bojand, line 77: If I am understanding the code correctly... Now we do not have a long lived cache of all protobufs that we refresh at some interval from SR? But our registry is more dynamic based on this cache... Is this TTL _too short_?

backend/pkg/schema/client.go
```diff
@@ -1,793 +1,390 @@
...
-		SetContext(ctx).
-		SetResult([]string{})
-
-	if showSoftDeleted {
-		req.SetQueryParam("deleted", "true")
+		return protocompile.SearchResult{Source: bytes.NewReader(data)}, nil
+	})
+	compiler := protocompile.Compiler{
+		Resolver:       protocompile.WithStandardImports(resolveFromEmbeddedProtosFn),
+		SourceInfoMode: protocompile.SourceInfoStandard,
```
bojand, line 136: Is there benefit to using `SourceInfoStandard`. It results in more memory consumption. 

```
// If unspecified or set to SourceInfoNone, source code information will not
	// be included in the resulting descriptors. Source code information is
	// metadata in the file descriptor that provides position information (i.e.
	// the line and column where file elements were defined) as well as comments.
	//
	// If set to SourceInfoStandard, normal source code information will be
	// included in the resulting descriptors. This matches the output of protoc
	// (the reference compiler for Protocol Buffers). If set to
	// SourceInfoMoreComments, the resulting descriptor will attempt to preserve
	// as many comments as possible, for all elements in the file, not just for
	// complete declarations.
	//
	// If Resolver returns descriptors or descriptor protos for a file, then
	// those descriptors will not be modified. If they do not already include
	// source code info, they will be left that way when the compile operation
	// concludes. Similarly, if they already have source code info but this flag
	// is false, existing info will be left in place.
```

I am not sure I see benefit of having source code info for our purpoces? We can leave it for now but we should do some memory benchmarking as we do have customers with a lot of protos, that do have comments. In my testing turning off source code info could result in 50% memory savings.

backend/pkg/schema/client.go
```diff
@@ -1,793 +1,390 @@
...
+	sourceResolver := &protocompile.SourceResolver{
+		Accessor: protocompile.SourceAccessorFromMap(accessorMap),
 	}
-
-	parsed, ok := res.Result().(*ConfigResponse)
-	if !ok {
-		return nil, fmt.Errorf("failed to parse config for subject response")
+	compiler := protocompile.Compiler{
+		Resolver:       protocompile.WithStandardImports(c.standardImportsResolver(sourceResolver)),
+		SourceInfoMode: protocompile.SourceInfoStandard,
```
bojand, line 264: Same concern here

backend/pkg/serde/json_schema.go
```diff
@@ -95,17 +102,26 @@ func (d JSONSchemaSerde) SerializeObject(ctx context.Context, obj any, _ Payload
 		return nil, fmt.Errorf("first byte indicates this it not valid JSON, expected brackets")
 	}
 
-	schema, err := d.SchemaSvc.GetJSONSchemaByID(ctx, so.schemaID)
+	schema, err := d.schemaClient.SchemaByID(ctx, int(so.schemaID))
```
bojand, line 105: Technically wouldn't this also potentially find Proto or Avro Schemas?

We could have a helper function `JSONSchemaByID` that has the type test below? NBD but it might be useful.

backend/pkg/schema/client.go
```diff
@@ -1,793 +1,390 @@
...
-		SetPathParam("subject", subjectName).
-		SetQueryParam("normalize", "true").
-		SetBody(&schema).
-		Post("/subjects/{subject}/versions")
-	if err != nil {
-		return nil, fmt.Errorf("create schema failed: %w", err)
+// buildJSONSchemaWithReferences adds a schema and its references to the provided JSON schema compiler.
+// It iterates through all references in the schema, fetching and recursively adding them to the compiler.
+// Returns an error if any of the schema resources or references fail to be added.
+func (c *CachedClient) buildJSONSchemaWithReferences(ctx context.Context, compiler *jsonschema.Compiler, name string, sch sr.Schema) error {
```
bojand, line 327: This could potentially have a Type check

backend/pkg/schema/client.go
```diff
@@ -1,793 +1,390 @@
...
-		SetPathParam("version", version).
-		SetQueryParam("permanent", strconv.FormatBool(deletePermanently)).
-		Delete("/subjects/{subject}/versions/{version}")
-	if err != nil {
-		return nil, fmt.Errorf("delete subject version failed: %w", err)
+// ParseAvroSchemaWithReferences parses an avro schema that potentially has references
+// to other schemas. References will be resolved by requesting and parsing them
+// recursively. If any of the referenced schemas can't be fetched or parsed an
+// error will be returned.
+func (c *CachedClient) ParseAvroSchemaWithReferences(ctx context.Context, schema sr.Schema, schemaCache *avro.SchemaCache) (avro.Schema, error) {
```
bojand, line 279: This could potentially have a Type check.

backend/pkg/schema/client.go
```diff
@@ -1,793 +1,390 @@
...
+// CompileProtoSchemaWithReferences compiles the given schema and its referenced schemas into protobuf files.
+// It recursively loads all schema references into an accessor map, which is then used by the protobuf compiler.
+func (c *CachedClient) CompileProtoSchemaWithReferences(
+	ctx context.Context,
+	schema sr.Schema,
+	accessorMap map[string]string,
+) (linker.Files, error) {
+	// Helper function to recursively fetch and parse all schema references.
+	var loadReferencesFn func(s sr.Schema) error
+	loadReferencesFn = func(s sr.Schema) error {
```
bojand, line 235: This could potentially have a Type check.

backend/pkg/schema/client.go
weeco: That's correct. I think fetching a schema is relatively cheap. I feel like 30s is a good balance because

1. For deserializing messages we won't pull hundreds of times per second
2. If someone (hard) deletes a schema we'll be up-to-date within a reasonable time. If it's needed again after 30s, we can pull again no problem

Any specific cases you have in mind where it may be problematic?

backend/pkg/schema/client.go
weeco: Nope, nothing in mind. I remember you worked on improving the memory performance, what should I use?

backend/pkg/connect/service.go
```diff
@@ -24,14 +24,14 @@ import (
 
 // Service provides the API for interacting with all configured Kafka connect clusters.
 type Service struct {
-	Cfg    config.Connect
+	Cfg    config.KafkaConnect
```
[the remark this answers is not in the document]
weeco: I wouldn't mind renaming the package, but I agree it should be a separate PR.

backend/pkg/console/api_versions.go
weeco: just to avoid the naming collision with the imported package named `version`

backend/pkg/api/connect/service/clusterstatus/util.go
bojand: Looking at the code and context how it's used I get it now... A comment might be helpful.

backend/pkg/schema/client.go
bojand: I don't know maybe 30s is fine.

backend/pkg/schema/client.go
bojand: I am leaning towards disabling it for now.

bojand (approved):

backend/pkg/schema/client.go
weeco: Changed to `protocompile.SourceInfoNone,`

backend/pkg/serde/json_schema.go
weeco: Yes it will find other schemas as well, we check the `schema.Type` right afterwards though before trying to parse the schema. The schema client is a thing caching / abstraction layer for retrieving schemas so didn't feel appropriate to put the JSOnSchemaByID there. I also like the idea of having the JSON schema specific parsing inside the respective serde

jvorcak (approved):

PR #1635, opened 2025-03-06
pipeline api: add external address field

birdayz:
it can be used to provide the external address of a pipeline


proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -47,6 +47,11 @@ message Pipeline {
 
   Status status = 8 [(google.api.field_behavior) = OUTPUT_ONLY];
 
+  // External address. It is specific to the redpanda connect config and
+  // plugin, what address is returned, it could be
+  // http:// or wss:// or something else.
+  string external_address = 10;
```
bojand, line 53: Naming is hard and I don't have strong opinion... but I wonder if "url" is better than "address"? Ex `"external_url"`? For most PaaS, it's some variation of that. Ex, `"live_url"`, `"web_url"`, or just `"url"`.

bojand (approved):

birdayz:
@bojand i updated it to `url,` can you please re-check?

bojand (approved):

bojand:
👍

PR #1636, opened 2025-03-06
Serde extracts schema details from header and request via Api proto file

jboucourt:
Serde extracts schema details from header and request via Api proto file. FileDescriptor it's used to deserialize payload and return as JSON.


weeco:
I think this PR was not meant for this repository.

PR #1639, opened 2025-03-10
Console 35 consumer groups rework

jvorcak:


frontend/src/components/pages/consumers/Modals.tsx
```diff
@@ -741,173 +777,144 @@ export class DeleteOffsetsModal extends Component<{
...
-      } else {
-        this.props.onClose();
-      }
-    }
+              {mode === 'partition' && (
+                <Box>
+                  <Text>Group offsets will be deleted for partition:</Text>
+                  <UnorderedList fontWeight="600" my={2}>
+                    <ListItem>
+                      Topic: <span className="codeBox">{offsetsByTopic[0].topicName}</span>
```
malinskibeniamin, line 901: Why do we need to continue using classnames? Any way we can use `<Code>` component or similar?

malinskibeniamin (approved):
Can you record a video or take some screenshots before & after?

PR #1642, opened 2025-03-10
frontend: remove fetch-monaco scripts

malinskibeniamin:
We already use `monaco-editor-webpack-plugin` introduced by @andresaristizabal so we don't need the scripts


weeco:
Should we backport this into master as well once done?

malinskibeniamin:
> Should we backport this into master as well once done?

@weeco I think it's a quality of life change, not a big difference, I want to set up bun/npm workspaces in the next iteration, then we can have a new microfrontend project prepared for tomorrow

andresaristizabal (approved):
LGTM, those worker are handle now by rebuild + webpack

PR #1645, opened 2025-03-11
proto: add v1alpha2 to descriptions

bojand:


kbatuigas (approved):
Just one thing to check, but no blockers on the Docs side

PR #1649, opened 2025-03-12
frontend: generate connect query proto layer

malinskibeniamin:
We want to be able to use MobX less and less, and we need data fetching solution to be uniform across all UIs.

Instead of using promise client, we will have a transport and query client hoisted at the top of the frontend app directory, and that way we can avoid the boilerplate today + not have to go through MobX for the new work we are doing.

This will enable us to start calling `useQuery(listPipelines)` in functional components instead of having to do class components with some `PipelineService.listPipelines` singleton pattern.

It should also enable us to do testing in an easier way.

@bojand @weeco please shout if this is something that may cause problems for the release?


weeco:
Suggested to ben to use a separate branch for the POC, before we merge anything into the release-2.8 branch. That ensures the release-2.8 branch is always in a good state to apply critical fixes if needed

bojand:
I agree lets keep this branch w/o many architectural changes... Fixes and simple improvements only. Lets keep it boring and stable for cloud deployments while v3 in master matures.

malinskibeniamin:
Thanks both, closing PR

malinskibeniamin:
@jvorcak notified me over our UI sync on Friday 14th of March 2025 that he might be blocked if we don't generate proto layer, so this would be letting him continue using the new approach for data fetching using react query

proto/buf.gen.yaml
```diff
@@ -42,3 +42,9 @@ plugins:
   - plugin: connect-gateway
     out: backend/pkg/protogen
     opt: paths=source_relative
+  - plugin: buf.build/connectrpc/query-es:v1.4.0
+    opt:
+      - target=ts
+      - import_extension=
+      - js_import_style=legacy_commonjs
+    out: frontend/src/protogen
```
malinskibeniamin, line 50: @jvorcak this is relevant for connect query

PR #1650, opened 2025-03-13
backend: add tags to pipeline api

weeco:
Add support for tags for the pipeline API


proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -445,6 +462,15 @@ message PipelineUpdate {
 
   // The number of resources that are guaranteed to be assigned to the pipeline.
   Pipeline.Resources resources = 6;
+
+  // Optional: A map of tags to add or update.
+  // If a tag's value is empty, the server interprets that as a deletion.
```
weeco, line 467: Since we have no update mask, maybe worth to have them only delete if a key is set with an empty value? WDYT?
Also: I named it tags to align with some APIs in Control Plane. In our secrets API we went with "labels" though.

cc @bojand

bojand: I think either option is fine. 

I think what you propose is fine. Seems sensible heh not to complicate things. Especially since none of the APIs have field masks.

I _think_ we could add update field mask property to this request. Technically it's not a breaking change. Existing clients would not send it and existing servers would not know about it.
We might have to update all clients and servers somewhat in unison depending on the update implementation. I am not sure... Perhaps we could use something like 
```go
req.ProtoReflect().Has(req.ProtoReflect().Descriptor().Fields().ByName("update_mast"))
```

to check for presence of the update mask... if not present (from old clients?) we would treat it in backwards compatible way. But there is nuance to this... But I think f new clients don't set it you may end up with the same behaviour then this wouldn't really work... So just seems complicated to support in a compatible way. Obviously easier to support if we update all the clients and assume / require that update. Which isn't strictly backwards compatible.

bojand (approved):

bojand (approved):

PR #1651, opened 2025-03-13
Better ACL error message in the overview tab

jvorcak:
Screenshot 2025-03-13 at 10 18 09
Screenshot 2025-03-13 at 10 24 13

### DIFF
--- a/frontend/src/components/pages/overview/Overview.tsx
+++ b/frontend/src/components/pages/overview/Overview.tsx
@@ -321,7 +321,9 @@ function ClusterDetails() {
 
   const serviceAccounts = overview.redpanda?.userCount ?? 'Admin API not configured';
 
-  const aclCount = overview.kafkaAuthorizerInfo?.aclCount ?? 'Authorizer not configured';
+  const aclCount = overview.kafkaAuthorizerInfo
+    ? overview.kafkaAuthorizerInfo.aclCount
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/overview/Overview.tsx
```diff
@@ -321,7 +321,9 @@ function ClusterDetails() {
 
   const serviceAccounts = overview.redpanda?.userCount ?? 'Admin API not configured';
 
-  const aclCount = overview.kafkaAuthorizerInfo?.aclCount ?? 'Authorizer not configured';
+  const aclCount = overview.kafkaAuthorizerInfo
+    ? overview.kafkaAuthorizerInfo.aclCount
+    : (overview.kafkaAuthorizerError?.message ?? 'Authorizer not configured');
```
weeco, line 326: We should work with the error codes we return in connectrpc responses. Possibly we should tackle this as an initiative but these error codes have been introduced so that we have more control over how we can present errors in the UI.

jvorcak: I'm working on a new way of data fetching right now, so I'll take this into consideration.

malinskibeniamin: @jvorcak you can check how we render toasts and how queries/mutations in custom hooks handle errors

weeco (approved):

PR #1659, opened 2025-03-17
frontend: Add secrets store page

malinskibeniamin:
https://www.loom.com/share/580f2fe891554ab0958c4263f4514482?sid=96037fc0-9bfe-4e57-9f47-d5b9e7479324
Screenshot 2025-03-20 at 19 04 31
Screenshot 2025-03-20 at 19 04 42
Screenshot 2025-03-20 at 19 04 51
Screenshot 2025-03-20 at 19 04 58
Screenshot 2025-03-20 at 19 06 16
Screenshot 2025-03-20 at 19 06 40
Screenshot 2025-03-20 at 19 06 49


frontend/src/components/form/error-info-field.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const ErrorInfoField = ({ field, index }: ErrorInfoFieldProps) => {
+  if (field.state.meta.errors?.length === 0) {
+    return null;
+  }
+
+  return (
+    <FormErrorMessage>
+      <UnorderedList>
+        {field.state.meta.errors.map((error) => (
+          <ListItem key={index ?? error.message}>
```
jvorcak, line 24: If index is undefined, using error.message as a key might not be reliable (if messages are repeated). I'd try to use some unique id. Also I don't like that it's coming as a prop, and I use it as a key for `errors.map`

frontend/src/components/form/password-field.tsx
```diff
@@ -0,0 +1,33 @@
+import { ViewIcon, ViewOffIcon } from '@chakra-ui/icons';
```
jvorcak, line 1: Some time ago, we agreed to use MD icons from `react-icons` for consistency. There is not a rule for that, because it's not fully migrated yet. I'm actually surprised we still have this package.

frontend/src/components/form/password-field.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface PasswordFieldProps {
+  label: string;
+}
+
+export const PasswordField = ({ label }: PasswordFieldProps) => {
+  const field = useFieldContext<string>();
+  const [showValue, setShowValue] = useBoolean(false);
+
+  return (
+    <FormControl isInvalid={!!field.state.meta.errors?.length}>
```
jvorcak, line 15: nit: I though we had a PasswordInput component in the existing UI lib. Since we're using components from that library withing the file, could we use it as well?

frontend/src/components/pages/secrets/secrets-store-page.tsx
```diff
@@ -0,0 +1,242 @@
...
+import { AiOutlineDelete, AiOutlineEdit } from 'react-icons/ai';
+import { useListSecretsQuery } from 'react-query/api/secret';
+import { uiState } from 'state/uiState';
+import { CreateSecretModal } from './create-secret-modal';
+import { DeleteSecretModal } from './delete-secret-modal';
+import { UpdateSecretModal } from './update-secret-modal';
+
+// Hack for MobX to ensure we don't need to use observables
+export const updatePageTitle = () => {
+  runInAction(() => {
```
jvorcak, line 31: nit: I'd still prefer to stay consistent with the upper-level mobx classes to handle the logic and remove it with MobX later on. I think we would benefit from consistency.

jvorcak (approved):
I'll check the rest later, for now I've put some preliminary review.

frontend/src/components/form/password-field.tsx
malinskibeniamin: Fixed.

frontend/src/components/pages/secrets/secrets-store-page.tsx
malinskibeniamin: Agree, I am trying to move quite fast, so I had to resort to this to ensure the title is set properly, otherwise it was overriding it.

The problem I have is that with MobX decorators and class-based components or observer pattern I will never get out of the legacy/tech debt we have today, so I decided this is the best we can do for now.

frontend/src/components/form/error-info-field.tsx
malinskibeniamin: Fixed

PR #1661, opened 2025-03-17
frontend: use v1 data types for requests in frontend

bojand:
Bringing in this commit into master.
https://github.com/redpanda-data/console/commit/e9ce83d85bd0b73b45260b3d62b4f3f1925c8f5e

We should generally review and change dataplane `v1alpha2` usage throughout frontend


malinskibeniamin:
Looks like there is a typescript error, I will check

PR #1690, opened 2025-03-25
Misc/chat improvements

jvorcak:


frontend/src/components/pages/agents/details/agent-details-page.tsx
```diff
@@ -53,6 +53,7 @@ export const AgentDetailsPage = () => {
       name: 'Chat',
       component: (
         <AgentChatTab
+          id={agentData?.agent?.id ?? ''}
```
malinskibeniamin, line 56: Let's use agent that we provide in another prop, this id is not necessary

frontend/src/components/pages/agents/details/chat/agent-chat-tab.tsx
```diff
@@ -33,14 +35,14 @@ export const AgentChatTab = ({ agent }: AgentChatTabProps) => {
   // Use live query to listen for message changes in the database
   const messages =
     useLiveQuery(async () => {
-      const storedMessages = await chatDb.getAllMessages();
+      const storedMessages = await chatDb.getAllMessages(id);
```
malinskibeniamin, line 38:
```suggestion
      const storedMessages = await chatDb.getAllMessages(agent?.id);
```

frontend/src/components/pages/agents/details/chat/agent-chat-tab.tsx
```diff
@@ -52,9 +54,13 @@ export const AgentChatTab = ({ agent }: AgentChatTabProps) => {
 
   return (
```
malinskibeniamin, line 55: Let's add some form of agent missing indicator
```suggestion
if (!agent) {
return (
<div>Agent is not available right now</div>
)
}
```

frontend/src/components/pages/agents/details/chat/chat-input.tsx
```diff
@@ -79,9 +83,11 @@ export const ChatInput = ({ setIsTyping, agentUrl }: ChatInputProps) => {
       // Create error message
       const errorMessage: ChatMessage = {
         id: (Date.now() + 1).toString(),
+        agentId,
         content: 'Sorry, there was an error sending your message. Please try again later.',
         sender: 'system',
         timestamp: new Date(),
+        failure: true,
```
malinskibeniamin, line 90: 👍

frontend/src/database/chat-db.ts
```diff
@@ -14,31 +14,33 @@ import type { Table } from 'dexie';
...
+  failure: boolean;
 }
 
 class ChatDatabase extends Dexie {
   messages!: Table<ChatMessage, string>;
 
   constructor() {
     super('ChatDatabase');
-    this.version(1).stores({
-      messages: 'id, sender, timestamp',
```
malinskibeniamin, line 28: Does the previous version need to be kept as well or is it ok to just run the last migration?

jvorcak: I'dI think both are ok, I don't think we will modify db often though. We can set it to 1 again if you want

malinskibeniamin: All good, just not sure about the convention

malinskibeniamin (approved):

PR #1707, opened 2025-04-03
rpcn: warn when compute units are bumped

nicolaferraro:
Sometimes the API adapts automatically the number of compute units, for instance when using an Ollama connector or values in disallowed ranges e.g. 16-31 compute units (since they would use a little part of a dedicated big machine). This adds a warning when the backend does it.

To make it work, I had to do some hacks to have the useToast function available otherwise the local toast would disappear on page change (the "Pipeline created"/"Pipeline updated" toasts did already disappear actually). Pls advise if there's a better way.

### DIFF
--- a/frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
+++ b/frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
@@ -25,6 +25,8 @@ import {
   NumberInput,
   Text,
   useDisclosure,
+  useToast,
+  CreateToastFnReturn,
 } from '@redpanda-data/ui';
 import { action, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
@@ -41,9 +43,7 @@ import Tabs from '../../misc/tabs/Tabs';
 import { PageComponent, type PageInitHelper } from '../Page';
 import { formatPipelineError } from './errors';
 import { SecretsQuickAdd } from './secrets/Secrets.QuickAdd';
-import { MAX_TASKS, MIN_TASKS, tasksToCPU } from './tasks';
-
-const { ToastContainer, toast } = createStandaloneToast();
+import { MAX_TASKS, MIN_TASKS, tasksToCPU, cpuToTasks } from './tasks';
 
 const exampleContent = `
 `;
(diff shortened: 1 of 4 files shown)


frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
```diff
@@ -84,10 +83,24 @@ class RpConnectPipelinesCreate extends PageComponent<{}> {
...
+    const CreateButton = () => {
+      const toast = useToast();
+
+      return (
+        <Button
+          variant="solid"
+          isDisabled={alreadyExists || isNameEmpty || this.isCreating}
+          loadingText="Creating..."
+          isLoading={this.isCreating}
+          onClick={action(() => this.createPipeline(toast))}
```
jvorcak, line 95: nit: I don't really like that toast is being passed as a parameter to create / update fns. Wouldn't it be cleaner to provide onSuccess / onFailure callbacks and trigger the toasts in here instead of `createPipeline`?

jvorcak (approved):

frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
```diff
@@ -84,10 +83,24 @@ class RpConnectPipelinesCreate extends PageComponent<{}> {
     const alreadyExists = pipelinesApi.pipelines.any((x) => x.id === this.fileName);
     const isNameEmpty = this.fileName.trim().length === 0;
 
+    const CreateButton = () => {
+      const toast = useToast();
```
jvorcak, line 87: Do we need this component? Can't we stick to `createStandaloneToast` approach? I don't like that we create these components in the render function, that doesn't look right. I think using `onSuccess`/`onError` callbacks and trying to avoid abstraction above `<Button` would be more standard.

malinskibeniamin: I suggested to use `showToast` in a Slack message, as we don't want to always define a new toast provider, @jvorcak are there any common errors we may face if we don't provide a standalone toast?

frontend/src/components/pages/rp-connect/tasks.tsx
```diff
@@ -1,5 +1,5 @@
 export const MIN_TASKS = 1,
-  MAX_TASKS = 90;
+  MAX_TASKS = 72;
```
malinskibeniamin, line 2: Where does this value come from?

nicolaferraro: Initially I set a high value since we did not know what would be the max. Now we settled at 72 as max in cloud.

frontend/src/components/pages/rp-connect/Pipelines.Create.tsx
nicolaferraro: Yeah, the issue was with page change hiding the toast

jvorcak: @malinskibeniamin it should be fine. The only reason I wrote this is that I'm not a huge fan of this interface definition (like I wrote above). But it's definitely not a blocker.

jvorcak (approved):

malinskibeniamin (approved):

PR #1708, opened 2025-04-03
add new scope to the console

alenkacz:
We're continuing making the scope mandatory to follow the principle of least privilege and not allowing redpanda to have access to customer's connect secrets but also connect not having access to all customer's secrets.

The use case of needing to share the secret between both services/workloads is probably pretty rare anyway.


c4milo (approved):

c4milo:
@alenkacz, wow, thanks for the quick turnaround. This is great!

bojand (approved):

PR #1715, opened 2025-04-10
Simplifies the way how we handle URL params in topic detail

jvorcak:
https://www.loom.com/share/9da6736857de4415b491a9b869c34ef5

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages/index.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages/index.tsx
@@ -28,6 +28,7 @@ import type { TopicMessage } from '../../../../state/restInterfaces';
 import { Feature, isSupported } from '../../../../state/supportedFeatures';
 import {
   type ColumnList,
+  DEFAULT_SEARCH_PARAMS,
   type DataColumnKey,
   FilterEntry,
   PartitionOffsetOrigin,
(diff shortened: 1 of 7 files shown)


frontend/src/state/ui.ts
```diff
@@ -103,6 +103,7 @@ export const DEFAULT_SEARCH_PARAMS = {
   page: 0,
   pageSize: 10,
   sorting: [] as SortingState,
+  quickSearch: '',
```
malinskibeniamin, line 106: I believe this is the only non-linting/formatting change applied to this PR @jvorcak - would you please elaborate on how the `quickSearch` param works to prevent changing URL in topic details page?

jvorcak: @malinskibeniamin it's not the only change. Perhaps you missed the changes in `Tab.Messages/index`? Github hides them in a diff, because it's too big. 

In this PR, we change the way how we read parameters completely.

frontend/src/components/pages/topics/Tab.Messages/index.tsx
malinskibeniamin, line 1: Would it make sense to use `nuqs` library here at all? I think it would do quite a lot of heavy lifting for you out of the box.

I understand we use class components so it won't be allowed to use hooks directly, but we can maybe still manage to take advantage of other functions for parsing params.
https://github.com/47ng/nuqs

malinskibeniamin (approved):

PR #1724, opened 2025-04-15
Remove unwanted AdminAPI Check for Create Role/User

jvorcak:


frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -486,11 +485,10 @@ const RolesTab = observer(() => {
           variant="outline"
           onClick={() => appGlobal.history.push('/security/roles/create')}
           {...getCreateUserButtonProps()}
-          isDisabled={api.userData?.canCreateRoles === false || !api.isAdminApiConfigured}
+          isDisabled={api.userData?.canCreateRoles === false}
```
weeco, line 488: We should check for the endpoint compatibility hook here. It should be there in the form of `Features.RBAC`.

Endpoint equals to `/redpanda.api.console.v1alpha1.SecurityService`, Method: `POST` in case it doesn't exist yet.

jvorcak: Updated

bojand (approved):

malinskibeniamin (approved):

PR #1737, opened 2025-04-24
Add tokenFilePath support for providing SASL OAuth token via file

imrenagi:
Hi, currently I have use case where I use kubernetes JWT token stored in `/var/run/secrets/kafka/serviceaccount/token` in k8s pod as token to authenticate to kafka. This token is a mounted file and it is gets renewed periodically.  However, the current available options on the console is by using OIDC token endpoint or by providing token value configured via `KAFKA_SASL_OAUTH_TOKEN` env var. 

So, I am thinking to add additional optional configuration to the token file path and read the file to get the content of the content.

Please let me know what you think.

Thanks.

### DIFF
--- a/backend/pkg/config/kafka_sasl_oauth.go
+++ b/backend/pkg/config/kafka_sasl_oauth.go
@@ -29,7 +29,7 @@ type KafkaSASLOAuthBearer struct {
 	TokenEndpoint string                    `yaml:"tokenEndpoint"`
 	Scope         string                    `yaml:"scope"`
 	Extensions    []KafkaSASLOAuthExtension `yaml:"extensions"`
-	TokenFilePath string                    `yaml:"tokenFilePath"`
+	TokenFilepath string                    `yaml:"tokenFilepath"`
 }
 
 // KafkaSASLOAuthExtension is the struct for the SASL OAuth extension support(custom key-value pairs)
@@ -53,7 +53,7 @@ func (c *KafkaSASLOAuthBearer) Validate() error {
 		if c.ClientID == "" || c.ClientSecret == "" {
 			return fmt.Errorf("OAuth Bearer client credentials must be set")
 		}
-	} else if c.Token == "" && c.TokenFilePath == "" {
+	} else if c.Token == "" && c.TokenFilepath == "" {
 		return fmt.Errorf("OAuth Bearer token or token file path must be set")
 	}
 
--- a/backend/pkg/config/kafka_sasl_oauth.go
+++ b/backend/pkg/config/kafka_sasl_oauth.go
@@ -29,6 +29,7 @@ type KafkaSASLOAuthBearer struct {
 	TokenEndpoint string                    `yaml:"tokenEndpoint"`
 	Scope         string                    `yaml:"scope"`
 	Extensions    []KafkaSASLOAuthExtension `yaml:"extensions"`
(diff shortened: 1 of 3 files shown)


backend/pkg/factory/kafka/kafka.go
```diff
@@ -175,13 +177,22 @@ func NewKgoConfig(cfg config.Kafka, logger *zap.Logger, metricsNamespace string)
...
 				})
-			} else {
+			} else if cfg.SASL.OAUth.Token != "" {
 				mechanism = oauth.Auth{
 					Token:      cfg.SASL.OAUth.Token,
 					Extensions: kafkaSASLOAuthExtensionsToStrMap(cfg.SASL.OAUth.Extensions),
 				}.AsMechanism()
+			} else if cfg.SASL.OAUth.TokenFilePath != "" {
+				mechanism = oauth.Oauth(func(ctx context.Context) (oauth.Auth, error) {
+					token, err := os.ReadFile(cfg.SASL.OAUth.TokenFilePath)
```
walbertus, line 187: this should utilize filesystem service
https://github.com/redpanda-data/console/blob/master/backend/pkg/filesystem/service.go#L189

walbertus (changes requested):

backend/pkg/factory/kafka/kafka.go
walbertus: GetFileByFilename

imrenagi: I dont think we can use it because that is used to read the file that is cached. In this case, we cant really read from cache so that we get the most updated token from kubernetes. not the cached. @walbertus

backend/pkg/config/kafka_sasl_oauth.go
```diff
@@ -29,6 +29,7 @@ type KafkaSASLOAuthBearer struct {
 	TokenEndpoint string                    `yaml:"tokenEndpoint"`
 	Scope         string                    `yaml:"scope"`
 	Extensions    []KafkaSASLOAuthExtension `yaml:"extensions"`
+	TokenFilePath string                    `yaml:"tokenFilePath"`
```
weeco, line 32:
```suggestion
	TokenFilepath string                    `yaml:"tokenFilepath"`
```

For consistency

docs/config/console.yaml
```diff
@@ -53,6 +53,7 @@ kafka:
       # clientSecret: "example-client-secret"
       # tokenEndpoint: "https://accounts.google.com/token"
       # scope: "openid"
+      # tokenFilePath: "/path/to/token/file"
```
weeco, line 56: To be updated when config changes

weeco:
Thanks @imrenagi for the PR. LGTM , just one NIT to fix and then we can merge it

backend/pkg/config/kafka_sasl_oauth.go
imrenagi: updated. thanks

docs/config/console.yaml
imrenagi: updated. thanks

imrenagi:
Thanks @weeco . I have updated the PR.

imrenagi:
@weeco I have fixed the linter issue by using switch case as requested by the linter. thanks

weeco (approved):

PR #1740, opened 2025-04-24
Adds cursor:pointer for buttons

jvorcak:
I think this regression was caused when because we now do reset CSS with tailwind instead of Chakra, which we've disabled.

### DIFF
--- a/frontend/src/globals.scss
+++ b/frontend/src/globals.scss
@@ -32,3 +32,7 @@ h2 {
 b {
     font-weight: 600;
 }
+
+button {
+    cursor: pointer;
+}


weeco (approved):

weeco:
@jvorcak I guess it wasn't supposed to be a draft PR, so changed to normal PR and approved

malinskibeniamin (approved):

jvorcak:
@weeco it was, I intentionally changed it, because I have a suspicion that we have bad TW preflight configuration and I want to fix it properly.

jvorcak:
I've merged this untill I properly investigate TW setup, this shouldn't hurt.

PR #1742, opened 2025-04-25
frontend: Update to protobuf v2

malinskibeniamin:
This move required a lot of work, but it is a great improvement overall and unlocks future benefits such as protobuf reflection/validation etc.

**A few things that changed:**
New code generation APIs with TypeScript typing

**BEFORE**
`new Message(payload);`

**AFTER**
`create(MessageSchema, payload);`

---
**BEFORE**
`proto3.getEnumType(PayloadEncoding).findNumber(val.encoding)?.localName`

**AFTER**
`PayloadEncodingSchema.values.find((value) => value.number === key.encoding)?.localName`

---

This will also allow better streaming support with https://github.com/TanStack/query/discussions/9065 as well as https://github.com/connectrpc/connect-query-es/issues/524

**Changelog**
https://connectrpc.com/docs/node/migrating-to-v2
https://buf.build/blog/protobuf-es-v2
https://github.com/connectrpc/connect-es/blob/main/MIGRATING.md


jvorcak:
Looks good, I've briefly skimmed through the code and tested.

weeco (approved):

PR #1749, opened 2025-04-30
Properly fix cursor for buttons

jvorcak:
We've changed Chakra's resetCSS and included TS Preflight.

The whole issue was caused by this 

https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor

So this fixes it properly


malinskibeniamin (approved):

weeco:
Good to merge?

PR #1752, opened 2025-05-02
frontend: Use more gRPC services

malinskibeniamin:
I provided hooks for all connect query bindings, for example debug bundle service.

Security service missing queries/mutations have also been added.

Topic and user service are now using gRPC


frontend/src/react-query/api/transform.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const getTransformRequest = create(GetTransformRequestSchema, {
+    request: getTransformRequestDataPlane,
+  }) as MessageInit<GetTransformRequest> & Required<Pick<MessageInit<GetTransformRequest>, 'request'>>;
+
+  return useQuery(getTransform, getTransformRequest, {
+    enabled: options?.enabled !== false && input?.name !== '',
+  });
+};
+
+export const useDeleteTransformMutationWithToast = () => {
```
sago2k8, line 82: Was the previous code also simply showing a toast? I want to make sure we simply don't copy what we do in cloud-ui. I rather don't use toasts unless we really have a reason to

frontend/src/react-query/api/transform.tsx
```diff
@@ -0,0 +1,115 @@
...
+      await queryClient.invalidateQueries({
+        queryKey: createConnectQueryKey({
+          schema: TransformService.method.listTransforms,
+          cardinality: 'infinite',
+        }),
+        exact: false,
+      });
+
+      showToast({
+        id: TOASTS.TRANSFORM.DELETE.SUCCESS,
```
sago2k8, line 96: This ID is a big flaw in the cloud-ui I won't use it in here, it's better to generate a unique ID for instance.

frontend/src/utils/toast.utils.tsx
```diff
@@ -79,6 +79,58 @@ export const TOASTS = {
       ERROR: 'REDPANDA_CONNECT_LINT_CONFIG_ERROR',
     },
   },
+  TRANSFORM: {
```
sago2k8, line 82: I am not sure this is the best way to separate this, don't see the value in the cloud-ui. I have seen some degration like closing one toast and seeing all being closed.

frontend/src/components/pages/topics/create-topic-modal.tsx
```diff
@@ -19,7 +19,7 @@ import {
   CreateTopicRequest_TopicSchema,
   CreateTopicRequest_Topic_ConfigSchema,
 } from 'protogen/redpanda/api/dataplane/v1/topic_pb';
-import { useLegacyCreateTopicMutationWithToast, useLegacyListTopicsQuery } from 'react-query/api/topic';
+import { useCreateTopicMutationWithToast, useListTopicsQuery } from 'react-query/api/topic';
```
sago2k8, line 22: In here we have an error handling in the modal, why do we want a toast message also, shouldn't we pick one ?

sago2k8 (commented):
Left couple of comments, I also have 2 concerns:
1. I loved that we are using dataplane grcp api, want to double check authorization is property handled there, the (admin, writer, reader magic).
2. Do we want to use the same strategy we use in cloud, I am not sure showing toasts everywhere is desired.

malinskibeniamin:
> Left couple of comments, I also have 2 concerns:
> 
> 1. I loved that we are using dataplane grcp api, want to double check authorization is property handled there, the (admin, writer, reader magic).
> 2. Do we want to use the same strategy we use in cloud, I am not sure showing toasts everywhere is desired.

@sago2k8 I agree with toasts, removed them now, so we will let the `onSuccess` callback determine what to do inside the component layer.

Re: 1. We haven't been using the newly added gRPC wrappers anywhere else yet. This is just some preparation for the next steps so that the gRPC endpoints are easier to consume and we can slowly integrate them going forward and checking edge cases once they start being applied.

Right now the only endpoint I actually updated to use gRPC is to create a topic, and we just added a new e2e UI test for that, so if there's any issue, we should be able to see that.

frontend/src/react-query/api/transform.tsx
malinskibeniamin: Fixed.

frontend/src/react-query/api/transform.tsx
malinskibeniamin: Fixed.

frontend/src/utils/toast.utils.tsx
malinskibeniamin: Fixed.

frontend/src/components/pages/topics/create-topic-modal.tsx
malinskibeniamin: I have removed the toasts from mutations across the whole console codebase.

jvorcak:
> @sago2k8 I agree with toasts, removed them now, so we will let the `onSuccess` callback determine what to do inside the component layer.

So that means in a component I'll still be able to use `onSuccess` and `onError` because the promise is returned, right? I just want to avoid the situation like it's now in some of the calls when the error is passed as a "flag" to a success callback.

jvorcak (approved):

sago2k8:
>So that means in a component I'll still be able to use onSuccess and onError because the promise is returned, right? I just want to avoid the situation like it's now in some of the calls when the error is passed as a "flag" to a success callback.

Exactly not hidding errors, You should have the full API. Even using the promise directly should be allowed

sago2k8 (approved):

malinskibeniamin:
> > @sago2k8 I agree with toasts, removed them now, so we will let the `onSuccess` callback determine what to do inside the component layer.
> 
> So that means in a component I'll still be able to use `onSuccess` and `onError` because the promise is returned, right? I just want to avoid the situation like it's now in some of the calls when the error is passed as a "flag" to a success callback.

Yes still possible to use `onSuccess`/`onError` etc.

PR #1753, opened 2025-05-05
feat: hide throttle button for Redpanda clusters

drahoslove:
## Problem

Throttling cannot be used for Redpanda because Redpanda auto-throttles as needed. If someone tries to configure throttling the requests will fail. We need to adapt the page and remove the settings if we are running on a Redpanda cluster.

## Solution

Hide the `Throttle` setting button when the cluster is RedPanda


malinskibeniamin (approved):

malinskibeniamin:
@drahoslove maybe next time you can provide a small BEFORE and AFTER screenshot, just to make sure there is no layout shift or other issue if we hide a button in this case

PR #1761, opened 2025-05-09
Hide debug bundle button in embedded mode

drahoslove:
### Problem

The `debug bundle` button was visible when embedded in the Cloud

### Solution

Add a condition `!isEmbedded()` to only render the button in standalone mode


malinskibeniamin (approved):

drahoslove:
@sago2k8 I'm not entirely sure whether the `isEmbedded()` is what we want here, or if I should use `isServerless()` instead. Can you please take a look?

PR #1769, opened 2025-05-15
frontend: Fix AI agent pipeline filtering

malinskibeniamin:
We were passing the filter with empty string value - previously it did nothing but it was actually meant to be `undefined`.

We cannot filter by name partially because the tags can only be filtered by exact match and the agent name is stored as a tag, not a real pipeline name due to the abstraction living on the frontend. Once we have a dedicated agent service this problem will go away.

This resulted in people seeing this when they created an AI agent.
Screenshot 2025-05-15 at 19 37 47

Recording with fix
https://github.com/user-attachments/assets/623cba21-5834-4eae-aef4-5e9a96ff8d8b

### DIFF
--- a/frontend/src/react-query/api/agent.tsx
+++ b/frontend/src/react-query/api/agent.tsx
@@ -169,16 +169,26 @@ export const useListAgentsQuery = (
     refetchInterval?: number | false;
   },
 ) => {
+  // Ensure name contains is only used if a string is provided. It has to match the agent name exactly.
+  // We cannot use nameContains for filtering pipeline name because the name of the agent is stored as a tag, not pipeline name.
+  // TODO: Once dedicated "agent service" exists, we can use agentName for that endpoint to filter by agent name.
+  const agentName = input?.filter?.nameContains;
+
+  const listPipelinesRequestFilter = create(ListPipelinesRequest_FilterSchema, {
+    tags: agentName
+      ? {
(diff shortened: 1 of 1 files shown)


frontend/src/react-query/api/agent.tsx
```diff
@@ -169,16 +169,24 @@ export const useListAgentsQuery = (
...
   },
 ) => {
+  const listPipelinesRequestFilter = create(ListPipelinesRequest_FilterSchema, {
+    tags: input?.filter?.nameContains
+      ? {
+          __redpanda_cloud_pipeline_type: 'agent',
+          // Ensure name contains is only used if a string is provided. It has to match the agent name exactly.
+          // We cannot use nameContains because the name of the agent is stored as a tag, not pipeline name.
+          // TODO: Once dedicated "agent service" exists, we can use nameContains for that endpoint to filter by agent name.
+          __redpanda_cloud_agent_name: input?.filter?.nameContains,
```
weeco, line 179: This is not a nameContains filter, the tags only filter on exact matches.

weeco (approved):

frontend/src/react-query/api/agent.tsx
malinskibeniamin: Yes, I moved it to a comment because I want to keep using the filter schema from proto

PR #1776, opened 2025-05-21
frontend: edit user password and role

Mateoc:
Eng 374

Adds the ability to change user password also modifies the flow to edit role

https://www.loom.com/share/687423ae49db4028b57ff5fc1c344257?sid=9ebf711c-f642-4ab4-8af1-c743ad409d0e


frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -61,11 +61,12 @@ import {
   principalGroupsView,
 } from './Models';
 import { AclPrincipalGroupEditor } from './PrincipalGroupEditor';
-
+import { BsThreeDots } from 'react-icons/bs';
 import ErrorResult from '../../../components/misc/ErrorResult';
 import { FeatureLicenseNotification } from '../../license/FeatureLicenseNotification';
 import { NullFallbackBoundary } from '../../misc/NullFallbackBoundary';
 import { UserRoleTags } from './UserPermissionAssignments';
+import {ChangePasswordModal, ChangeRolesModal} from './UserEditModals';
```
malinskibeniamin, line 69: I think you need `biome format` here

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -441,6 +403,72 @@ const UsersTab = observer(() => {
...
+              }}
+            >
+              Change roles
+            </MenuItem>
+          )}
+          <DeleteUserConfirmModal
+            onConfirm={onConfirmDelete}
+            buttonEl={<MenuItem type="button">Delete</MenuItem>}
+            userName={user.name}
+          />
```
malinskibeniamin, line 464: Should we move the modal away from the menu list? Just below the menu, as we are using a react fragment

frontend/src/components/pages/acls/UserCreate.tsx
```diff
@@ -415,6 +416,59 @@ export const RoleSelector = observer((p: { state: string[] }) => {
...
+          noOptionsMessage={() => 'No roles found'}
+          // TODO: Selecting an entry triggers onChange properly.
+          //       But there is no way to prevent the component from showing no value as intended
+          //       Seems to be a bug with the component.
+          //       On 'undefined' it should handle selection on its own (this works properly)
+          //       On 'null' the component should NOT show any selection after a selection has been made (does not work!)
+          //       The override doesn't work either (isOptionSelected={()=>false})
+          value={undefined}
+          onChange={(val, meta) => {
+            console.log('onChange', { metaAction: meta.action, val });
```
malinskibeniamin, line 445: Let's remove the console log

frontend/src/components/pages/acls/UserCreate.tsx
```diff
@@ -415,6 +416,59 @@ export const RoleSelector = observer((p: { state: string[] }) => {
...
+          inputValue={searchValue}
+          onInputChange={setSearchValue}
+          placeholder="Select roles..."
+          noOptionsMessage={() => 'No roles found'}
+          // TODO: Selecting an entry triggers onChange properly.
+          //       But there is no way to prevent the component from showing no value as intended
+          //       Seems to be a bug with the component.
+          //       On 'undefined' it should handle selection on its own (this works properly)
+          //       On 'null' the component should NOT show any selection after a selection has been made (does not work!)
+          //       The override doesn't work either (isOptionSelected={()=>false})
```
malinskibeniamin, line 442: Thanks for the writeup. I wouldn't change legacy component library at this point, just focus on getting us to a more modern setup.

frontend/src/components/pages/acls/UserDetails.tsx
```diff
@@ -40,6 +41,11 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
 
   @observable selectedRoles: string[] = [];
 
+  @observable isChangePasswordModalOpen = false;
+  @observable isChangeRolesModalOpen = false
```
malinskibeniamin, line 45: Missing semicolon `;`

frontend/src/react-query/api/user.tsx
```diff
@@ -88,7 +88,7 @@ export const useListUsersQuery = (
   const listUsersResult = useInfiniteQueryWithAllPages(listUsers, listUsersRequest, {
     pageParamKey: 'pageToken',
     enabled: options?.enabled,
-    getNextPageParam: (lastPage) => lastPage.nextPageToken,
+    getNextPageParam: (lastPage) => lastPage.nextPageToken || undefined,
```
malinskibeniamin, line 91: What does this change help with?

frontend/src/react-query/api/security.tsx
```diff
@@ -69,32 +54,22 @@ export const useListRolesQuery = (
...
     enabled: options?.enabled,
-    getNextPageParam: (lastPage) =>
-      lastPage?.response?.nextPageToken
-        ? {
-            ...listRoleMembersRequestDataPlane,
-            pageToken: lastPage.response?.nextPageToken,
-          }
-        : undefined,
+    getNextPageParam: (lastPage) => lastPage?.nextPageToken || undefined,
+    pageParamKey: 'pageToken',
```
malinskibeniamin, line 69: Ae you sure this works well? I would be careful, make sure the infinite query works, try it out with page size of 1 and check that the next page gets retrieved after that.

frontend/src/components/pages/acls/UserEditModals.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
+    setSelectedRoles([...originalRoles]);
+  }, [JSON.stringify(originalRoles)]);
+
+  const onSaveRoles = async () => {
+    if (!Features.rolesApi) return;
+    const addedRoles = selectedRoles.except(originalRoles);
+    const removedRoles = originalRoles.except(selectedRoles);
```
malinskibeniamin, line 173: 👏

frontend/src/components/pages/acls/UserEditModals.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const toast = useToast()
+  const [selectedRoles, setSelectedRoles] = useState<string[]>([]);
+  const { mutateAsync: updateRoleMembership, isPending: isUpdateMembershipPending } = useUpdateRoleMembershipMutation();
+  const { data } = useListRolesQuery({ filter: { principal: userName } });
+  const originalRoles = data.roles.map((r) => r.name);
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
+    setSelectedRoles([...originalRoles]);
+  }, [JSON.stringify(originalRoles)]);
```
malinskibeniamin, line 168: This is quite ugly, maybe at least provide an explanation why we need to do this, if it's documented then people can be more aware of this

frontend/src/components/pages/acls/UserEditModals.tsx
```diff
@@ -0,0 +1,256 @@
...
+      setIsOpen(false);
+    } catch (err) {
+      toast({
+        status: 'error',
+        duration: null,
+        isClosable: true,
+        title: `Failed to update user ${userName}`,
+        description: String(err),
+      });
+    }
```
malinskibeniamin, line 54: This is a connect error, isn't it? Can we use `ConnectError.from(error)` and filter the toast using `formatMessageGrpc` util function?

frontend/src/components/pages/acls/UserCreate.tsx
Mateoc: To be honest most of this componenet is a copy paste from RoleSelector I just changed to useState rather than mobx

frontend/src/react-query/api/user.tsx
Mateoc: this fixes the isue where the query would keep fetching the endpoint [this](https://redpandadata.slack.com/archives/C081G37T71B/p1747333480572309?thread_ts=1747333453.921179&cid=C081G37T71B)

frontend/src/react-query/api/security.tsx
Mateoc: yes I did multiple checks

frontend/src/react-query/api/user.tsx
Mateoc: btw this endpoint if you set the size to 0 it would load [all users](https://github.com/redpanda-data/console/blob/master/backend/pkg/api/connect/service/user/v1/service.go#L109) though we could run into max payload size issues that is way I left the infinite query

frontend/src/components/pages/acls/UserEditModals.tsx
malinskibeniamin: You need to use `formatToastErrorMessageGRPC` to make sure toasts get rendered with consistent text.

malinskibeniamin: You can use `showToast(formatToastErrorMessageGRPC(ConnectError.from(error)))` or similar.

micheleRP:
Some little text suggestions please!

frontend/src/components/pages/acls/Acl.List.tsx
malinskibeniamin: Oh I remember, then let's leave it as-is, thanks for the reminder.

frontend/src/react-query/api/security.tsx
```diff
@@ -21,44 +22,28 @@ import {
...
     enabled: options?.enabled,
-    getNextPageParam: (lastPage) =>
-      lastPage?.response?.nextPageToken
-        ? {
-            ...listRolesRequestDataPlane,
-            pageToken: lastPage.response?.nextPageToken,
-          }
-        : undefined,
+    getNextPageParam: (lastPage) => lastPage?.nextPageToken || undefined,
+    pageParamKey: 'pageToken',
```
malinskibeniamin, line 43: @Mateoc Let's make sure this continues to work when there are multiple pages required to fetch, can try it with page size of 1 and check 2 requests are coming through

malinskibeniamin (approved):
Left 1 comment https://github.com/redpanda-data/console/pull/1776/files#r2107881529

frontend/src/react-query/api/security.tsx
Mateoc: yes it does

PR #1777, opened 2025-05-21
Fix infinite render in SearchBar component

drahoslove:
### Issue
The route /connected-clusters/redpanda contained infinite render loop

To reproduce:
 - run cloudv2 on port 3000
 - Go to http://localhost:3000/clusters/d0e8lbgpojcjpdtep250/connect-clusters/redpanda (or other cluster with connector)
 
### Cause
Caused by issue in the `SearchBar` component - it was reacting to any observable in the reactive context - some other observable being updated probably caused the issue (unsure which one exactly)
 
### Solution
Replaced the mobx's `autorun` in the `SearchBar` component by `reaction` - it does the same but is only triggered when the specified observable changes.

### DIFF
--- a/frontend/src/components/misc/SearchBar.tsx
+++ b/frontend/src/components/misc/SearchBar.tsx
@@ -10,7 +10,7 @@
  */
 
 import { SearchField } from '@redpanda-data/ui';
-import { type IReactionDisposer, autorun } from 'mobx';
+import { type IReactionDisposer, reaction } from 'mobx';
 import { observer } from 'mobx-react';
 import React, { Component } from 'react';
 import { AnimatePresence, MotionSpan, animProps_span_searchResult } from '../../utils/animationProps';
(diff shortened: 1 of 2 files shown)


malinskibeniamin (approved):
Great find!

Can you also record the video that shows it's resolved or have you checked it locally? I haven't pulled it yet

sago2k8 (approved):

jvorcak (approved):

drahoslove:
> Great find!
> 
> Can you also record the video that shows it's resolved or have you checked it locally? I haven't pulled it yet

@malinskibeniamin  I have tested it locally. There's a proof that it loads, it does not re-render infinitely (not spamming logs) a the search still works (the data are mocked but that should not matter): https://www.loom.com/share/e27f85e98c484df59da3b717ac4cd456

PR #1779, opened 2025-05-26
Ux 225/implement tls reloader

sago2k8:
Implement hot reload for TLS certificates, this is meant to be used by clients.

### DIFF
--- a/backend/pkg/tls/ca.go
+++ b/backend/pkg/tls/ca.go
@@ -1,0 +1,133 @@
+// Copyright 2025 Redpanda Data, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.md
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0
+
+package tls
+
+import (
+	"context"
+	"crypto/sha256"
+	libtls "crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"os"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+// MaybeWithDynamicClientCA returns a function that configures a CA on the tls
+// Config reading it from a file and updating whenever its content changes.
+// It is intended to be used on a client configuration to set the server CA it
+// expects to talk to.
+//
+// Asynchronous reloading is triggered when a new request is served, not before
+// 5 minutes from last reload.
+func MaybeWithDynamicClientCA(
+	quitCtx context.Context,
+	caPath string,
+	hostname string,
+	refreshInterval time.Duration,
+	logger *zap.Logger,
+) func(*libtls.Config) error {
+	return func(cfg *libtls.Config) error {
+		if caPath == "" {
+			return nil
+		}
+
+		r, err := newCAReloader(quitCtx, cfg, caPath, refreshInterval, logger)
+		if err != nil {
+			return err
+		}
+		// There's no direct client-side equivalent to GetConfigForClient;
+		// instead, we skip the standard server certificate validation and
+		// override VerifyConnection instead.
+		cfg.InsecureSkipVerify = true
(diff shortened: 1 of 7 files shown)


backend/pkg/tls/ca.go
```diff
@@ -0,0 +1,134 @@
...
+	digest := sha256.Sum256(ca)
+	if digest == r.digest {
+		r.logger.Debug("CA cert did not change", zap.String("path", r.caPath))
+		return nil
+	}
+	cfg := r.tc.Load().Clone()
+	if cfg.ClientCAs == nil {
+		cfg.ClientCAs = x509.NewCertPool()
+	}
+	if !cfg.ClientCAs.AppendCertsFromPEM(ca) {
```
andresaristizabal, line 120: I understand this fn add the new ca to the existing ones, do we remove the old ones at some point?

backend/pkg/tls/ca.go
```diff
@@ -0,0 +1,133 @@
...
+	ticker := time.NewTicker(refreshInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-quitCtx.Done():
+			return
+		case <-ticker.C:
+			if err := r.load(); err != nil {
+				r.logger.Error("error while reloading the CA", zap.Any("err", err))
```
andresaristizabal, line 100: Do you think we should include we are going to try again in x minutes? in order to user try to fix and it will try again automatically

backend/pkg/tls/keypair.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/twmb/tlscfg"
+	"go.uber.org/zap"
+)
+
+// MaybeWithDynamicDiskKeyPair returns a function that configures a certificate
+// on the tls Config that is read from a pair of files on disk and updated
+// whenever the content of those files changes.
+//
+// Files are polled every 5 minutes for modifications.
```
andresaristizabal, line 30: [nit]: I think we should use the config name in TLS instead of say 5 minutes

backend/pkg/tls/keypair_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/twmb/tlscfg"
+)
+
+func TestMaybeWithDynamicDiskKeyPair(t *testing.T) {
+	// Setup test environment
+	testSetup := setupTLSTest(t)
+	defer os.RemoveAll(testSetup.TmpDir)
+
+	t.Run("empty paths return nil", func(t *testing.T) {
```
andresaristizabal, line 30: I think we can add a test for ca path and key empty path, no?

backend/pkg/tls/ca.go
sago2k8: We don't, this is how Serverless does it and it makes sense imho. Certificates will be rotated by kubernetes because they will be expired soon, if the reason for rotation is a certificate breach I would expect the pod to be restarted.

backend/pkg/tls/tls_test_utils.go
```diff
@@ -0,0 +1,158 @@
...
+	err := os.WriteFile(path, certPEM, 0o600)
+	require.NoError(t, err, "Failed to write certificate to file")
+}
+
+// writeCertToPEMFile writes a certificate to a file in PEM format
+func writeCertToPEMFile(t *testing.T, path string, cert *x509.Certificate) {
+	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
+	err := os.WriteFile(path, certPEM, 0o600)
+	require.NoError(t, err, "Failed to write certificate to PEM file")
+}
```
malinskibeniamin, line 141: Why do we have 2 of these functions? They are exactly the same.

sago2k8: Good call. thanks

weeco (approved):

PR #1782, opened 2025-05-27
proto: add new rpc for linting pipeline configs

weeco:


proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -390,6 +391,32 @@ service PipelineService {
       api: API_REDPANDA_CONNECT
     };
   }
+
+  // Lints a Redpanda Connect pipeline configuration and returns zero or more
+  // issues (“hints”). An empty list means the config passed all lint checks.
+  rpc LintPipelineConfig(LintPipelineConfigRequest) returns (LintPipelineConfigResponse) {
+    option (google.api.http) = {
+      post: "/v1/redpanda-connect/pipelines:lint"
```
birdayz, line 399: i would call it lint-config. the RPC is lint config, therefore the custom method is lint-config. and that seems to be the right direction for me anyway, because this RPC actually lints a config, and not a pipeline itself.

birdayz (approved):
LGTM, thank you

PR #1792, opened 2025-06-03
proto: mcp resource

rockwotj:


proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    // The contents of the configuration file in YAML format.
+    string contents = 2 [
+      (google.api.field_behavior) = REQUIRED,
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.min_len = 1
+    ];
+  }
+
+  // All the configuration files for the MCP server.
+  repeated Configuration config_files = 4 [
```
birdayz: i wonder if we want to selectively update files.

if yes (i think so?), using a map<string, Configuration> will be easier with fieldmask. Updating specific indices (or even based on a field name) via fieldmask is not possible, however with maps it's quite easy, because we can use the map key in the fieldmask to update a specific map entry.
(i do not like this limitation of fieldmask, and would want to see a better fieldmask that allows selecting list entries based on index or field+value..)

so either use map (for this reason, we use map a lot in cloud), or clients need to always update the entire list.

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+  State state = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  Status status = 7 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  // URL to connect to the MCP server
+  string url = 8 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  // Tags are key-value pairs that can be assigned to a MCP server resource.
+  // They help organize MCP servers and enable filtering when listing them.
+  map<string, string> tags = 9 [(buf.validate.field).map = {
```
birdayz: [nit]
call me esoteric, but: since this is a new resource, i would move this field "up" around id/display_name. it was only at such a high field index because it was added to pipeline: later on 

having metadata in the first fields is kinda nice if tools print by field number. (debugging etc)

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    max_pairs: 16
+    values: {
+      string: {pattern: "^([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]*)$"}
+    }
+  }];
+
+  message Resources {
```
birdayz: can we share this type with pipeline?

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    // The MCP server is running.
+    STATE_RUNNING = 2;
+    // The MCP server is in the process of stopping.
+    STATE_STOPPING = 3;
+    // The MCP server is stopped and in paused state.
+    STATE_STOPPED = 4;
+    // The MCP server encountered an error.
+    STATE_ERROR = 5;
+    // The MCP server has completed the job successfully.
+    STATE_COMPLETED = 6;
```
birdayz: i think completed makes no sense for MCP. it's only for oneshot pipelines, but mcp is never oneshot ("server")

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+  enum State {
+    STATE_UNSPECIFIED = 0;
+    // The MCP server is starting.
+    STATE_STARTING = 1;
+    // The MCP server is running.
+    STATE_RUNNING = 2;
+    // The MCP server is in the process of stopping.
+    STATE_STOPPING = 3;
+    // The MCP server is stopped and in paused state.
+    STATE_STOPPED = 4;
```
birdayz, line 90: do we even want to allow stopping, stopped? since it's a server, it will always run?

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+service MCPServerService {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_tag) = {
+    name: "Redpanda Connect MCP servers"
+    description: "Create and manage Redpanda Connect MCP servers and their configurations."
+  };
+
+  // CreateMCPServer creates a Redpanda Connect MCP Server in the Redpanda cluster.
+  rpc CreateMCPServer(CreateMCPServerRequest) returns (CreateMCPServerResponse) {
+    option (google.api.http) = {
+      post: "/v1/redpanda-connect/mcp-servers"
```
birdayz, line 112: [question]: shall we keep putting this stuff under redpanda-connect? what is the benefit? should it be a completely separate resource? eg. /v1/mcp-servers

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_REDPANDA_MCP_SERVER
+    };
+  }
+
+  // GetMCPServersForSecret implements the get MCP servers for secret method which lists the MCP servers
+  // in the Redpanda cluster for the given secret.
+  rpc GetMCPServersForSecret(GetMCPServersForSecretRequest) returns (GetMCPServersForSecretResponse) {
```
birdayz: do we need this one?

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_REDPANDA_MCP_SERVER
+    };
+  }
+
+  // GetMCPServersBySecrets implements the get MCP servers by secrets method which lists the MCP servers
+  // in the Redpanda cluster for all secrets.
+  rpc GetMCPServersBySecrets(GetMCPServersBySecretsRequest) returns (GetMCPServersBySecretsResponse) {
```
birdayz: do we need this one? also, maybe we can rather integrate these into List, instead of having multiple GetBy*, using the filter field.

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+}
+
+message ListMCPServersRequest {
+  // List filter.
+  Filter filter = 1;
+
+  // Limit the paginated response to a number of items.
+  int32 page_size = 2 [
+    (buf.validate.field).int32 = {
+      gte: -1
```
birdayz, line 398: interesting, we support -1 ? why -1 and not 0? ❓

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+      lte: 1000
```
birdayz: probably 1k is a much too large number of pipelines, i'd put a much more conservative value, eg. 50 or 100 (or even less).

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    }
+  ];
+
+  // Value of the next_page_token field returned by the previous response.
+  // If not provided, the system assumes the first page is requested.
+  string page_token = 3;
+
+  message Filter {
+    // Substring match on MCP server name. Case-sensitive.
+    string name_contains = 1 [
```
birdayz: shouldn't it be `display_name_contains` ? since the field is display_name

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+message MCPServerUpdate {
+  option (google.api.resource) = {
+    type: "redpanda.api.dataplane.v1/MCPServerUpdate"
+    singular: "mcp_server"
+    plural: "mcp_servers"
+  };
+
+  // User-friendly MCP servers name.
+  string display_name = 1 [
+    (google.api.field_behavior) = REQUIRED,
```
birdayz: display_name is not required for an update

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_ /]+$",
+    (buf.validate.field).string.min_len = 3,
+    (buf.validate.field).string.max_len = 128
+  ];
+
+  // MCP server description.
+  string description = 2 [
+    (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED,
```
birdayz: IGNORE_IF_UNPOPULATED is not needed, we ignore validation errors if the update_mask does not include the field

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    (buf.validate.field).string.max_len = 256
+  ];
+
+  // All the configuration files for the MCP server.
+  repeated MCPServer.Configuration config_files = 3 [
+    (google.api.field_behavior) = REQUIRED,
```
birdayz: not required, eg. if the update only changes a tag, this field is not present.

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,664 @@
...
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_EDIT
+      api: API_REDPANDA_MCP_SERVER
+    };
+  }
+
+  // The configuration schema includes available components and processors in this Redpanda Connect MCP Server instance.
+  rpc GetMCPServerServiceConfigSchema(GetMCPServerServiceConfigSchemaRequest) returns (GetMCPServerServiceConfigSchemaResponse) {
+    option (google.api.http) = {get: "/v1/redpanda-connect/mcp-config-schema"};
```
birdayz: this would be a custom method. https://google.aip.dev/136

eg. /v1/redpanda-connect/mcp-servers:getConfigSchema

but..isn't it for a specific "mcp server instance", eg. we need an ID?

/v1/redpanda-connect/mcp-servers/{id}:getConfigSchema

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: I have no idea... This is copied from RPCN

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: I've been burned from using maps in the past (ie you want to change your keys, etc), but you're right it's cleaner. Will switch

proto/redpanda/api/dataplane/v1/mcp.proto
weeco: Good question, no strong opinion on either, but I have a slight preference to group it under redpanda-connect I believe. Purely because I think it's a more specific resource type of redpanda connect and doesn't seem important/distinct enough for me to entertain it's own resources.

proto/redpanda/api/auth/v1/authorization.proto
```diff
@@ -22,6 +22,7 @@ enum API {
   API_REDPANDA_CONNECT = 4;
   API_KAFKA_CONNECT = 5;
   API_CONSOLE = 6;
+  API_REDPANDA_MCP_SERVER = 7;
```
weeco: Is `API_REDPANDA_MCP_SERVER` the best name for this? I'd omit `REDPANDA` or use `API_REDPANDA_CONNECT_MCP_SERVER`.

Also worth to consider re-using `API_REDPANDA_CONNECT` here since it's literally the same API that we are talking too. Not sure it makes sense to separate these out since we could create MCP servers using the pipelines API as well I believe? No strong opinion here either

proto/redpanda/api/dataplane/v1/mcp.proto
weeco: I think we should keep it, because in practice this state can exist for whatever reason or someone actually wants to pause their server (saving costs or some other reason)

proto/redpanda/api/dataplane/v1/mcp.proto
weeco: Hmm we were using this endpoint for pipelines to lookup whether a secret is still in use and warn the user when deleting. Not sure if we need it or whether we would still be able to retrieve all pipelines including MCP servers when calling that endpoint on the pipelines endpoint.

proto/redpanda/api/dataplane/v1/mcp.proto
birdayz: @sago2k8 maybe you know

proto/redpanda/api/auth/v1/authorization.proto
birdayz: hm, this enum is used for console only right? i'd defer to martin's suggestion, what fits console's needs best. redpanda_connect seems generic enough to cover both pipeline api but also others.

if we go w/ mcp_server:
agree w. the REDPANDA_ prefix being obsolete (we don't need to prefix every enum with the company name, basically)

proto/redpanda/api/dataplane/v1/mcp.proto
weeco: Seems to be done consistently across the dataplane API, not sure why. 0 Should opt out of page size as well where allowed (some places make sense to have the ability to opt out)

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: I moved it up above the output parameters, but I think that it should be after the "main" required configuration, since this is optional

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: Done.

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: done

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: I think yes we should group under connect it is a subcommand of connect and a different way of exposing the same stuff from the connectors repo

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: They will be separate, so you'll need to call both MCP and pipelines

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: Moved into List

proto/redpanda/api/auth/v1/authorization.proto
rockwotj: Since this is used for authz, I think it makes sense to split it out and have fine grained resources

proto/redpanda/api/dataplane/v1/mcp.proto
rockwotj: This isn't for a specific instance but for the editor in the console to give a schema. Switched to the custom method

proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -0,0 +1,555 @@
...
+    STATE_STOPPING = 3;
+    // The MCP server is stopped and in paused state.
+    STATE_STOPPED = 4;
+    // The MCP server encountered an error.
+    STATE_ERROR = 5;
+  }
+
+  // MCP server status may contain an error message.
+  message Status {
+    string error = 2;
```
weeco: NIT: I guess wasn't on purpose

```suggestion
    string error = 1;
```

rockwotj: Nope, thanks!

PR #1795, opened 2025-06-04
Add Knowledgebase

birdayz:
- Add KB proto
- Add KB UI

### DIFF
--- a/frontend/src/components/pages/knowledgebase/KnowledgeBase.EditTabs.tsx
+++ b/frontend/src/components/pages/knowledgebase/KnowledgeBase.EditTabs.tsx
@@ -999,12 +999,19 @@ export class KnowledgeBaseEditTabs extends React.Component<KnowledgeBaseEditTabs
             helperText="All credentials are securely stored in your Secrets Store"
           />
         ) : (
-          <ProtoDisplayField
-            messageSchema={KnowledgeBase_VectorDatabase_PostgresSchema}
-            fieldName="dsn"
-            label="PostgreSQL DSN"
-            value={postgres?.dsn || 'Not configured'}
-          />
+          <Box>
+            <Text fontSize="sm" fontWeight="medium" color="gray.700" mb={1}>
+              PostgreSQL DSN
+            </Text>
+            <Text fontSize="sm" color="gray.500" mb={2}>
+              All credentials are securely stored in your Secrets Store
+            </Text>
+            <Box p={3} bg="gray.50" borderRadius="md" border="1px solid" borderColor="gray.200">
+              <Text fontSize="sm" color="gray.900">
+                {postgres?.dsn || 'Not configured'}
+              </Text>
+            </Box>
+          </Box>
         )}
 
         {postgres && (
@@ -1604,10 +1611,9 @@ export class KnowledgeBaseEditTabs extends React.Component<KnowledgeBaseEditTabs
         ) : (
           <>
             {/* Display fields for view mode */}
-            <ProtoDisplayField
-              label="Provider"
-              value={generation?.provider?.provider.case === 'openai' ? 'OpenAI' : 'OpenAI'}
-            />
+            <Heading size="sm">
+              {generation?.provider?.provider.case === 'openai' ? 'OpenAI' : 'OpenAI'} Configuration
+            </Heading>
 
             <ProtoDisplayField label="Model" value={generation?.model || ''} />
 


proto/redpanda/api/dataplane/v1/knowledge_base.proto
```diff
@@ -0,0 +1,75 @@
...
+import "google/api/field_behavior.proto";
+
+// KnowledgeBase defines the configuration and metadata for a retrieval-augmented generation (RAG) Knowledge Base.
+message KnowledgeBase {
+  // KnowledgeBase ID.
+  string id = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_/]+$"
+  ];
```
weeco: We should autogenerate the id and not have it user-specified I think. 

Also this id will likely go into a Kafka topic name (e.g. `__redpanda_cloud__rag_<kb-id>_source_confluence-with-id`) so we need to limit the character set. Slashes wouldn't be allowed.

proto/redpanda/api/dataplane/v1/knowledge_base.proto
```diff
@@ -0,0 +1,75 @@
...
+  // Optional pipeline description.
+  string description = 3 [(buf.validate.field).string.max_len = 256];
+
+  message VectorDatabase {
+    message Postgres {
+      string dsn = 1;
+    }
+
+    oneof vector_database {
+      Postgres postgres = 1;
```
malinskibeniamin: Should we differentiate as db type, since there are many flavours of postgres based DBs?

proto/redpanda/api/dataplane/v1/knowledge_base.proto
birdayz: oh yeah, absolutely. my mistake. should be xid

proto/redpanda/api/dataplane/v1/knowledge_base.proto
birdayz: what different fields would we support? in the end, it's a conn via PG protocol. and, i guess, it needs pgvector, we could also call it PGVector

rockwotj: +1 call it PGVector - we shouldn't be specific and will want to support different extensions I'm sure

proto/redpanda/api/dataplane/v1/knowledge_base.proto
```diff
@@ -0,0 +1,75 @@
...
+
+  // Optional pipeline description.
+  string description = 3 [(buf.validate.field).string.max_len = 256];
+
+  message VectorDatabase {
+    message Postgres {
+      string dsn = 1;
+    }
+
+    oneof vector_database {
```
rockwotj: Does this need to be required with buf.validate?

proto/redpanda/api/dataplane/v1/knowledge_base.proto
```diff
@@ -0,0 +1,75 @@
...
+    (buf.validate.field).string.min_len = 3,
+    (buf.validate.field).string.max_len = 128
+  ];
+
+  // Optional pipeline description.
+  string description = 3 [(buf.validate.field).string.max_len = 256];
+
+  message VectorDatabase {
+    message Postgres {
+      string dsn = 1;
```
rockwotj:
```suggestion
      string dsn = 1 [(buf.validate.field).string.uri = true];
```

proto/redpanda/api/dataplane/v1/knowledge_base.proto
```diff
@@ -0,0 +1,75 @@
...
+        string model = 2 [
+          (google.api.field_behavior) = REQUIRED,
+          (buf.validate.field).required = true
+        ];
+      }
+
+      oneof provider {
+        option (buf.validate.oneof).required = true;
+        OpenAI openai = 1;
+      }
```
rockwotj:
```suggestion
      message OpenAI {
        string api_key = 1 [
          (google.api.field_behavior) = REQUIRED,
          (buf.validate.field).required = true
        ];
        string model = 2 [
          (google.api.field_behavior) = REQUIRED,
          (buf.validate.field).required = true
        ];
      }
      message Cohere {
        string base_url = 1 [
          (google.api.field_behavior) = REQUIRED,
          (buf.validate.field).required = true
        ];
        string api_key = 2 [
          (google.api.field_behavior) = REQUIRED,
          (buf.validate.field).required = true
        ];
        string model = 3 [
          (google.api.field_behavior) = REQUIRED,
          (buf.validate.field).required = true
        ];
        int32 dimensions = 4 [
          (buf.validate.field).ignore = IGNORE_IF_UNPOPULATED,
          (buf.validate.field).int32 = { in: [256, 512, 1024, 1536] }
        ];
      }

      oneof provider {
        option (buf.validate.oneof).required = true;
        OpenAI openai = 1;
        Cohere cohere = 2;
      }
```

proto/redpanda/api/dataplane/v1/knowledge_base.proto
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  // Configuration of embeddings generator.
+  EmbeddingGenerator embedding_generator = 6 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
+  ];
+
+  // API url for retrieval API, that allows lookup of documents.
+  string retrieval_api_url = 4 [(google.api.field_behavior) = OUTPUT_ONLY];
```
rockwotj: funky field number ordering?

proto/redpanda/api/dataplane/v1/knowledge_base.proto
birdayz: this is the resource. so it's required, because it's never an input

proto/redpanda/api/dataplane/v1/knowledge_base.proto
birdayz: yeah i will get it ordered before merging, still iterating a bit

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
```diff
@@ -118,18 +118,22 @@ const SecretsQuickAdd = ({ isOpen, onAdd, onCloseAddSecret }: SecretsQuickAddPro
     <Modal isOpen={isOpen} onClose={closeModal} isCentered={true} size={'md'}>
       <ModalOverlay />
       <ModalContent>
-        <ModalHeader>Add secret</ModalHeader>
+        <ModalHeader>Select secret</ModalHeader>
```
birdayz: i changed this, because, even in existing places where it's used, the text did not make sense. it says add secret, but actually you select a secret, but may add one and then select.

rockwotj: +1 I have tripped up trying to add a secret while making a pipeline before.

I do think it might be nice to be able to add a secret from the create pipeline view too

frontend/src/components/routes.tsx
```diff
@@ -383,6 +395,27 @@ export const APP_ROUTES: IRouteEntry[] = [
...
+  MakeRoute<{}>(
+    '/knowledgebases',
+    KnowledgeBaseList,
+    'Knowledge Bases',
+    BookOpenIcon,
+    true,
+    routeVisibility(
+      // Do not display knowledge bases if feature flag is disabled
+      () => isFeatureFlagEnabled('enableKnowledgeBaseInConsoleUi'), // Needed to pass flags to current routing solution
+      [Feature.PipelineService],
```
birdayz, line 405: i am a bit confused by this. shall i add a Feature. entry for this?
why do we have featureFlagEnabled function and this one (both) ?

frontend/src/state/backendApi.ts
```diff
@@ -2242,6 +2247,84 @@ export const pipelinesApi = observable({
   },
 });
 
+export const knowledgebaseApi = observable({
+  knowledgeBases: undefined as undefined | KnowledgeBase[],
+  knowledgeBasesError: null as ConnectError | null,
+
+  async refreshKnowledgeBases(_force: boolean): Promise<void> {
```
birdayz, line 2254: is this the correct pattern? all calls must be added in this file?

frontend/src/state/backendApi.ts
```diff
@@ -2242,6 +2247,84 @@ export const pipelinesApi = observable({
...
+    await client.deleteKnowledgeBase({ id });
+  },
+  async createKnowledgeBase(knowledgeBase: KnowledgeBaseCreate) {
+    const client = appConfig.knowledgebaseClient;
+    if (!client) throw new Error('knowledgebase client is not initialized');
+
+    console.log('=== Console gRPC Client Debug ===');
+    console.log('Full knowledgeBase object:', knowledgeBase);
+    console.log('knowledgeBase.indexer:', knowledgeBase.indexer);
+    console.log('Serialized request:', JSON.stringify({ knowledgeBase }, null, 2));
```
weeco: Should probably remove all the console.logs here.

proto/redpanda/api/dataplane/v1alpha1/knowledge_base.proto
```diff
@@ -0,0 +1,607 @@
...
+import "google/api/annotations.proto";
+import "google/api/field_behavior.proto";
+import "google/api/resource.proto";
+import "google/protobuf/field_mask.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "redpanda/api/auth/v1/authorization.proto";
+
+message KnowledgeBase {
+  option (google.api.resource) = {
+    type: "redpanda.api.dataplane.v1/KnowledgeBase"
```
weeco: This is going into v1alpha1, should the type be .v1 ? The v1alpha1 api is deprecated as a package. I think we want to get rid of it entirely.

cc @sago2k8

frontend/src/components/routes.tsx
malinskibeniamin: I believe the idea is that before March 2025 we never had feature flags in Console. Previously we were hiding features based on which services were available, in this case a pipeline service needed to be configured to show the route/allow navigation.

frontend/package-lock.json
malinskibeniamin, line 1: We should use `bun` exclusively on the frontend, which means we should only have `bun.lock` and `yarn.lock` files. `package-lock.json` should not be part of the codebase anymore.

frontend/src/state/backendApi.ts
malinskibeniamin: I will write this down in a way that hopefully can be used as a prompt for Claude Code.

This is quite legacy and I see Claude Code has decided to follow the existing pattern here.

Ideally instead of knowledgebaseApi we just use gRPC connect query, calling specific rpc like listKnowledgeBases, createKnowledgeBase etc. and we don't rely on an observable mobx pattern. We don't need additional side effects from mobx, we can just reuse a `useQuery` hook from connect query.

frontend/src/utils/protobuf-reflection.ts
malinskibeniamin, line 1: Very interesting piece of code, could be a beginning of a way to integrate it with auto form component to use grpc for validating the form directly without any intermediate measures.

frontend/src/state/backendApi.ts
malinskibeniamin: The way to do it right now is to follow a pattern established for another service, for example frontend/src/react-query/api/secret.tsx for secrets, that way we can have reusable hooks that isolate all data fetching logic

frontend/src/config.ts
```diff
@@ -188,6 +194,8 @@ const setConfig = ({ fetch, urlOverride, jwt, isServerless, featureFlags, ...arg
     transformsClient: transformClient,
     rpcnSecretsClient: secretGrpcClient,
     clusterStatusClient: clusterStatusGrpcClient,
+    knowledgebaseClient: knowledgebaseGrpcClient,
+    userClient: userGrpcClient,
```
malinskibeniamin, line 198: What do we use `userClient` for?

frontend/src/components/pages/knowledgebase/KnowledgeBase.List.tsx
```diff
@@ -0,0 +1,275 @@
...
+    <Flex alignItems="center" justifyContent="center" flexDirection="column" gap="4" mb="4">
+      <Image src={EmptyConnectors} />
+      <Box>You have no knowledge bases.</Box>
+      <CreateKnowledgeBaseButton />
+    </Flex>
+  );
+};
+
+@observer
+class KnowledgeBaseList extends PageComponent<{}> {
```
malinskibeniamin: This is an outdated pattern as well.

We should use functional React components (not class-based), and try to do something similar to frontend/src/components/pages/agents/create/create-agent-page.tsx where you do not rely on any MobX state at all.

frontend/src/state/backendApi.ts
birdayz: oh lord. sorry. removing!

frontend/src/config.ts
birdayz: i added it for the "auto create users", but had to stop implementing it, it was a bit too tricky.
i can remove the client - i did keep it because it's reasonable to have this client available

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
weeco: > secret, but may add one and then select.

How about "Select or add secret" then? :D

frontend/src/components/pages/knowledgebase/KnowledgeBase.List.tsx
birdayz: done, could you check the result is what you expected?

proto/redpanda/api/dataplane/v1/knowledge_base.proto
birdayz: tbh, i did not go for PGVector. In the end, this is just postgres: dsn, table, ...
not suuper important for me though

frontend/src/components/pages/rp-connect/secrets/Secrets.QuickAdd.tsx
birdayz: changed modal header to "Select or add secret"

proto/redpanda/api/dataplane/v1alpha1/knowledge_base.proto
birdayz: yes should go into v1alpha1, good find. thanks. updated

frontend/src/state/backendApi.ts
birdayz: updated, what do you think?

birdayz:
migrated to v1alpha3

buf.yaml
```diff
@@ -27,3 +27,4 @@ breaking:
   ignore:
     - redpanda/api/console/v1alpha1
     - redpanda/api/dataplane/v1alpha1
+    - proto/redpanda/api/dataplane/v1alpha1
```
weeco, line 30:
```suggestion
    - proto/redpanda/api/dataplane/v1alpha3
```

weeco: I guess this can be removed now?

birdayz: removed

malinskibeniamin (approved):

PR #1800, opened 2025-06-06
feat: implement TLS certificate reloading for clients

sago2k8:
Add automatic TLS certificate reloading capability for Kafka and Schema Registry clients. This makes clients reload certificates when they change on disk.

- Update single client providers to create clients on demand
- Add TLSConfigWithReloader usage for dynamic certificate loading
- Extract hostnames from connection URLs for proper certificate validation
- Pass logger to relevant components for certificate reload logging

### DIFF
--- a/backend/pkg/factory/kafka/kafka.go
+++ b/backend/pkg/factory/kafka/kafka.go
@@ -26,6 +26,7 @@ import (
 	"github.com/jcmturner/gokrb5/v8/client"
 	krbconfig "github.com/jcmturner/gokrb5/v8/config"
 	"github.com/jcmturner/gokrb5/v8/keytab"
+	netpkg "github.com/redpanda-data/common-go/net"
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kgo"
 	"github.com/twmb/franz-go/pkg/sasl"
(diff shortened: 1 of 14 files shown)


backend/pkg/factory/kafka/kafka.go
```diff
@@ -273,7 +274,14 @@ func NewKgoConfig(cfg config.Kafka, logger *zap.Logger, metricsNamespace string)
...
 	if cfg.TLS.Enabled {
-		tlsConfig, err := cfg.TLS.TLSConfig()
+		// Extract hostname from the first broker URL
+		_, host, err := netpkg.ParseHostMaybeScheme(cfg.Brokers[0])
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse broker url scheme: %w", err)
+		}
+		hostname, _ := netpkg.SplitHostPortDefault(host, 9092)
+
+		tlsConfig, err := cfg.TLS.TLSConfigWithReloader(ctx, logger, hostname)
```
weeco: I think this wouldn't work, because the `ctx` will quit the Go routine that watches the hot-reloading process. I think the context that is passed into this function is the request context though, and therefore rather short-lived.

sago2k8: good catch I will review this.

backend/pkg/factory/schema/schema.go
```diff
@@ -34,23 +36,41 @@ var _ ClientFactory = (*SingleClientProvider)(nil)
...
 	}
 
+	return &SingleClientProvider{
+		cfg:    cfg,
+		logger: l,
+	}, nil
+}
+
+// GetSchemaRegistryClient returns a schema registry client for the given context.
+func (p *SingleClientProvider) GetSchemaRegistryClient(ctx context.Context) (*sr.Client, error) {
```
sago2k8: context is short living in this case (note to self)

PR #1817, opened 2025-06-23
Provides console with news from Redpanda

jvorcak:
Screenshot 2025-06-23 at 19 53 33
Screenshot 2025-06-23 at 19 53 24

### DIFF
--- a/frontend/src/components/pages/overview/Overview.tsx
+++ b/frontend/src/components/pages/overview/Overview.tsx
@@ -50,6 +50,7 @@ import {
 import { NullFallbackBoundary } from '../../misc/NullFallbackBoundary';
 import { Statistic } from '../../misc/Statistic';
 import ClusterHealthOverview from './ClusterHealthOverview';
+import { ResourcesAndUpdates } from './ResourcesAndUpdates';
 
 @observer
 class Overview extends PageComponent {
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/useNewsFeed.ts
```diff
@@ -0,0 +1,48 @@
...
+    'link-text': string;
+    message: string;
+    name: string;
+    'link-url': string;
+    slug: string;
+    segment: string;
+  };
+}
+
+export const useNewsFeed = (segment?: 'redpanda' | 'kafka') => {
```
malinskibeniamin: Should we use react query? That way we don't need custom wrapping around fetch()

frontend/src/components/pages/overview/ResourcesAndUpdates.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <Text>Failed to load news: {error}</Text>
+      ) : (
+        <Flex direction="column" gap={4}>
+          {news.map((item) => (
+            <Link
+              key={item.id}
+              href={item.fieldData['link-url']}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="!no-underline"
```
malinskibeniamin, line 37: Why should we hide underline for links?

jvorcak: I wanted to follow https://redpanda-marketing-glance.lovable.app/ as close as possible + this will be all replaced by the usage of UI registry anyway. But I can put it back if you want

malinskibeniamin: Oh sorry had no idea about that URL, let's leave it as-is, we can use proper registry later.

PR #1830, opened 2025-07-02
Schema registry ACL proposal 2

andresaristizabal:
This pull request introduces new functionality to the `SecurityService` in the Redpanda Data Plane API. Specifically, it adds support for managing Schema Registry ACLs, including listing, creating, and deleting them. 

#### Updates to `SecurityService` Interface and Methods:
* Added new RPC methods for managing Schema Registry ACLs (`ListSchemaRegistryACLs`, `CreateSchemaRegistryACL`, and `DeleteSchemaRegistryACLs`) to the `SecurityServiceClient` interface.
* Updated the `SecurityServiceHandler` interface to include implementations for the new Schema Registry ACL methods.

This proposal updates the current ACL definition, and exposes new methods on security service for the schema registry  ACL, This proposal makes allow use current acls and its validations, but some validations need to change with this approach, for example [Host property needs to have](https://github.com/redpanda-data/console/pull/1830/commits/4fef739579e5b3e0645e04b1120561f88b224baa#diff-2d1263ee7c909821b46541f38e20c47c519328d0688056c8bb23a76e518a48ffL152) `User:` prefix, but for SR we need to support Role too.


proto/redpanda/api/dataplane/v1/acl.proto
```diff
@@ -50,6 +52,7 @@ message ACL {
     OPERATION_IDEMPOTENT_WRITE = 12;
     OPERATION_CREATE_TOKENS = 13;
     OPERATION_DESCRIBE_TOKENS = 14;
+    OPERATION_REMOVE = 15;
```
Mateoc: is this different from OPERATION_DELETE

andresaristizabal: Yes, one delete is for Kafka acls and this one is for schema registry acls, I agree with it could confuse

Mateoc: Can't you reuse it ? The operation, as far as I can tell, is not tied to the resource if not I would go with the other option

proto/redpanda/api/dataplane/v1/security.proto
```diff
@@ -196,6 +199,217 @@ message UpdateRoleMembershipResponse {
...
+}
+
+message CreateSchemaRegistryACLRequest {
+  option (buf.validate.message).cel = {
+    id: "resource_name_must_be_set_except_for_cluster_resource_type"
+    expression:
+      "this.resource_type == 4 && size(this.resource_name) == 0 ? ''"
+      ": this.resource_type == 4 && this.resource_name != 'kafka-cluster' ? 'Field resource_name must be set to \"kafka-cluster\" or empty when using resource_type=CLUSTER'"
+      ": this.resource_type != 4 && size(this.resource_name) == 0 ? 'Field resource_name must be set'"
+      ": ''"
```
weeco: We can/should work on a good CEL to validate the requests properly. Several combinations aren't possible, especially if we have a separate "CreateSchemaRegistryACLRequest" rather than a generic one that spans both Kafka & SR.

andresaristizabal: Fixed, now we will handle both ACL in a single endpoint.

PR #1831, opened 2025-07-03
frontend: upgrade Biome to v2

malinskibeniamin:
This PR upgrades Biome to v2, resuting in performance upgrades, being able to format more file extensions, better formatting/linting defaults, and more.

I attempted to upgrade to ultracite but there are issues https://github.com/haydenbleasel/ultracite/issues/196. Down the line the hope is to use Ultracite MCP with very good recommended defaults, which will help us get automatic formatting/linting at the LLM level, see [here](https://github.com/haydenbleasel/ultracite/blob/ae88a06966f0ba633c0975cf6e5a8255df7fc624/docs/content/mcp.mdx#L2) for more information if you are curious.

Descriptions for Biome rules have been taken from https://github.com/haydenbleasel/ultracite/blob/main/biome.jsonc


biome.json
malinskibeniamin, line 1: We are using jsonc for biome now

frontend/.github/copilot-instructions.md
malinskibeniamin, line 1: This should help us stay consistent when using LLMs through IDEs like Cursor/Windsurf or with plugins like GitHub Copilot.

Mateoc (approved):

PR #1843, opened 2025-07-21
backend: add slog-based logger package to replace zap

weeco:
## Summary
- Add new logger package based on Go's native slog to replace uber/zap
- Provides Prometheus metrics integration for log message counts by level
- Supports configurable handlers with middleware-style extensibility
- Includes JSON/text output formats and custom timestamp formatting
- Preparation for migrating existing zap-based logging to slog

### DIFF
--- a/backend/.golangci.yaml
+++ b/backend/.golangci.yaml
@@ -80,7 +80,7 @@ linters:
       attr-only: true
       no-global: all
       context: all
-      msgstyle: lowercased
+      msg-style: lowercased
       key-naming-case: snake
       forbidden-keys:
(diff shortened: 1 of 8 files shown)


backend/pkg/schema/client_test.go
sago2k8, line 1: I this part of this PR, it looks like it does not belong here.

sago2k8 (commented):
Just a comment for now.

backend/pkg/schema/client_test.go
weeco: Fixed

PR #1851, opened 2025-07-25
Handle errors for create/delete secrets

jvorcak:
Screenshot 2025-07-25 at 7 33 35
Screenshot 2025-07-25 at 7 32 09

- Fixes error handling for create secret modal
- Fixes error handling for delete secret modal
- Resets both forms on clicking cancel button, for now it only worked when you closed modal the other way (esc, click outside)

### DIFF
--- a/frontend/src/components/pages/secrets/create-secret-modal.tsx
+++ b/frontend/src/components/pages/secrets/create-secret-modal.tsx
@@ -27,7 +27,6 @@ import { useCreateSecretMutation, useListSecretsQuery } from 'react-query/api/se
 import { base64ToUInt8Array, encodeBase64 } from 'utils/utils';
 import type { z } from 'zod';
 import { secretSchema } from './form/secret-schema';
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/secrets/create-secret-modal.tsx
```diff
@@ -82,8 +89,17 @@ export const CreateSecretModal = ({ isOpen, onClose, customSecretSchema, helperT
         labels: labelsMap,
       });
 
-      await createSecret({ request });
-      handleClose();
+      try {
+        await createSecret({ request });
+        handleClose();
+      } catch (err: any) {
```
malinskibeniamin: Why not subscribe to the onError callback for the use mutation hook? That way we will know what error type it will be

jvorcak: Done

PR #1855, opened 2025-07-30
Backend: ListSRACL support in ListACL

r-vasquez:
This initial PR introduces:

- Migration to our common-go package `rpsr`, which wraps franz-go/sr and extends the client for Redpanda's SR ACL endpoints.
- Add ListSRACL support to our current ListACL response in our dataplane grpc endpoint, this is only called if:
  - Schema Registry is enabled
  - The filter contains any SR-relevant filter (i.e resource_type must be one of `ANY`, `SUBJECT`, or `REGISTRY`.
  - I would love to improve this to don't even call the endpoint if the user don't have a valid license.


backend/pkg/api/connect/service/acl/v1/service.go
```diff
@@ -105,7 +108,34 @@ func (s *Service) ListACLs(ctx context.Context, req *connect.Request[v1.ListACLs
 		}
 		resources[i] = aclResProto
 	}
-
+	// TODO: how to tell if authz is enabled for Schema Registry reliably? - Listing is not guarded against license
+	//  and a user can create an ACL before authz is enabled. Is there a way to check if license is valid here?
```
weeco: We could add a variable to this proto service (maybe with expiry since they may upgrade/enable the SR during the Console runtime) that checks if this method is available by submitting a list request and see if that route exists in the configured schema registry?

This is possibly the only good way to figure it out? We can't do it eagerly (e.g. at create time of the service) due to impersonation.

backend/pkg/api/connect/service/acl/v1/service.go
```diff
@@ -105,7 +108,34 @@ func (s *Service) ListACLs(ctx context.Context, req *connect.Request[v1.ListACLs
...
+	if s.cfg.SchemaRegistry.Enabled {
```
weeco: We should refactor this to a return early:

```
if !s.cfg.SchemaRegistry.Enabled {
  return ....
}

// Rest of the code follows
```

backend/pkg/api/connect/service/acl/v1/service.go
```diff
@@ -105,7 +108,34 @@ func (s *Service) ListACLs(ctx context.Context, req *connect.Request[v1.ListACLs
...
+		srFilter := s.srClientMapper.listACLFilterToDescribeACLSR(req.Msg.Filter)
+		// Filter can be nil if the filter contains no fields that are relevant
+		// for Schema Registry ACLs.
+		if srFilter != nil {
```
weeco: Too avoid the deep nesting probably another early return here. Probably you can build the filter beforehand and then have a single condition:

```go
srFilter := s.srClientMapper.listACLFilterToDescribeACLSR(req.Msg.Filter)
checkForSchemaRegistryACLs := s.cfg.SchemaRegistry.Enabled && srFilter != nil
if !checkForSchemaRegistryACLs {
  // return early
}
```

backend/pkg/api/connect/service/acl/v1/mapper.go
```diff
@@ -232,3 +233,245 @@ func (k *kafkaClientMapper) deleteACLMatchingResultToProto(acl kmsg.DeleteACLsRe
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to map resource type: %w", err)
+		}
+		resourcePatternType, err := srm.srACLPatternTypeToProto(aclRes.PatternType)
+		if err != nil {
+			return nil, fmt.Errorf("failed to map pattern type: %w", err)
+		}
+
+		// Create unique key for grouping as mentioned above.
+		key := fmt.Sprintf("%s:%s:%s", resourceType.String(), aclRes.Resource, resourcePatternType.String())
```
c-julin, line 261: minor: key could be a struct of these 3 things instead of trying to concatenate into a string

backend/pkg/api/connect/service/acl/v1/mapper.go
```diff
@@ -232,3 +233,245 @@ func (k *kafkaClientMapper) deleteACLMatchingResultToProto(acl kmsg.DeleteACLsRe
...
+		var principal, resourceName, host string
+		if filter.Principal != nil {
+			principal = *filter.Principal
+		}
+		if filter.ResourceName != nil {
+			resourceName = *filter.ResourceName
+		}
+		if filter.Host != nil {
+			host = *filter.Host
+		}
```
c-julin, line 311: `func deRefPtr[T any](ptr *T) T {
	if ptr == nil {
		var zero T
		return zero
	}
	return *ptr
}`

replacing this with a de referencing function  and you can call it directly in your struct, just something small to tidy this up

c-julin (commented):
Looks good, i have left the most minor comments, stylistic if anything. Will let Martin give the 👍

backend/pkg/api/connect/service/acl/v1/service.go
```diff
@@ -106,6 +110,41 @@ func (s *Service) ListACLs(ctx context.Context, req *connect.Request[v1.ListACLs
...
+		// without returning an error and just return the Kafka ACLs.
+		if se := (*sr.ResponseError)(nil); errors.As(err, &se) && se.StatusCode == http.StatusNotFound {
+			return connect.NewResponse(&v1.ListACLsResponse{Resources: resources}), nil
+		}
+
+		return nil, apierrors.NewConnectError(
+			connect.CodeInternal,
+			fmt.Errorf("error listing Schema Registry ACLs: %w", err),
+			apierrors.NewErrorInfo(v1.Reason_REASON_REDPANDA_SCHEMA_REGISTRY_ERROR.String()),
+		)
```
weeco, line 134: We should probably create our own helper for constructing rich schema registry errors for ConnectRPC as well. We have one for Kafka API and Redpanda Admin API errors already that convert the returned errors in the best possible ConnectRPC error.

See pkg/api/connect/errors/kafka.go and pkg/api/connect/errors/redpanda.go for example

proto/redpanda/api/dataplane/v1/error.proto
```diff
@@ -31,4 +31,7 @@ enum Reason {
 
   // The Redpanda enterprise license has expired and is no longer valid.
   REASON_ENTERPRISE_LICENSE_EXPIRED = 9;
+
+  // Redpanda Schema Registry returned and error.
+  REASON_REDPANDA_SCHEMA_REGISTRY_ERROR = 10;
```
weeco, line 36: Good!

NIT: Possibly rename to `REASON_SCHEMA_REGISTRY_ERROR` , since we are usally not dealing with custom schema registry methods. Seems a bit more appropriate, but really a NIT

weeco (approved):

backend/pkg/api/connect/service/acl/v1/mapper.go
r-vasquez: Good point, I changed all these direct access to the automatically-generated proto getter functions, which essentially does that and is safer to use as it prevents panic on nil structs when used well. Going to add that in the upcoming PR

PR #1856, opened 2025-08-01
backend: kafka quota APIs added to dataplane

c-julin:
### Summary
- Created proto and generated files
- Based on the docs found [here](https://docs.redpanda.com/current/manage/cluster-maintenance/manage-throughput/#client-throughput-limits)
- List Create and Delete specs and services defined
- Only supporting client-id and client-id-prefix entities. User and IP left out for now.
- Only supporting producer bytes max and consumer bytes max quotas. [Request rate quota](https://kafka.apache.org/documentation/#design_quotascpu) not supported in Redpanda
- created console service
- created the api to list, create and delete quotas 
- created integration tests for quotas api


taskfiles/backend.yaml
```diff
@@ -165,9 +165,11 @@ tasks:
   test-integration:
     dir: "{{.BACKEND_ROOT}}"
     desc: Run integration tests
+    deps:
+      - install-tparse
     cmds:
       - |
-        {{ .PATH_PREFIX }} go test -v -race --tags=integration -p=1 {{.CLI_ARGS | default "./..." }}
+        {{ .PATH_PREFIX }} go test -race --tags=integration -json -p=1 {{.CLI_ARGS | default "./..." }} | {{ .BUILD_ROOT }}/bin/go/tparse -notests
```
c-julin: happy to remove this but i thought it was nice to have the output tided up and presented in a table and surfaces failures to the top. I was struggling to find the failures when it wasn't parsed.

backend/pkg/api/connect/service/quota/v1/mapper.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (kafkaClientMapper) listQuotasRequestToKafka(req *v1.ListQuotasRequest) (*kmsg.DescribeClientQuotasRequest, error) {
+	var (
+		entityType string
+		matchType  kmsg.QuotasMatchType
+		match      *string
+	)
+
+	// Match entity type and filter type from the request
+	switch req.GetEntityType() {
```
weeco: We also have ENTITY_TYPE_ANY + 2 others that are not yet supported in Redpanda.

I guess especially ANY is a valid type that we want to support for listing? I think we can add support for the others too, since we also support Apache Kafka envs.

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+import "buf/validate/validate.proto";
+import "google/api/annotations.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "redpanda/api/auth/v1/authorization.proto";
+
+message Quota {
+  // The type of entity that the quota applies to.
+  enum EntityType {
+    ENTITY_TYPE_UNSPECIFIED = 0;
+    ENTITY_TYPE_ANY = 1;
```
weeco: What is this entity type for? I think it doesn't exist?

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+  option (buf.validate.message).cel = {
+    id: "name_required_for_filter_type_name"
+    message: "name is required when filter_type is FILTER_TYPE_NAME"
+    expression: "this.filter_type != 2 || this.name != ''"
+  };
+  // FilterType defines the type of filter to apply when listing quotas.
+  enum FilterType {
+    FILTER_TYPE_UNSPECIFIED = 0;
+    // The filter type any will return quotas for all entities.
+    FILTER_TYPE_ANY = 1;
```
weeco:
```suggestion
    // The filter type any will return quotas for all entities for the requested entity type.
    FILTER_TYPE_ANY = 1;
```

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+    // The filter type name when specified with name
+    // will return quotas for the specified client id.
+    FILTER_TYPE_NAME = 2;
+    // The filter type default returns the default client id
+    FILTER_TYPE_DEFAULT = 3;
+  }
+  Quota.EntityType entity_type = 1;
+  FilterType filter_type = 2;
+  string name = 3;
```
weeco: Add validation annotations. Enums are probably defined only and not unspecified or so, name probably just needs upper limits and maybe character set?

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+
+message DeleteQuotaResponse {}
+
+service QuotaService {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_tag) = {
+    name: "Quotas"
+    description: "Manage client throughput limits. See [Redpanda Client Throughput Limits](https://docs.redpanda.com/current/manage/cluster-maintenance/manage-throughput/#client-throughput-limits) for more information."
+  };
+
+  rpc ListQuotas(ListQuotasRequest) returns (ListQuotasResponse) {
```
weeco, line 197: We probably want to add pagination here, see other requests. The Kafka API doesn't support pagination but we offer it for resource-constrained frontend clients, including the possibility to run filters on the backend in case there or hundreds or thousands of resources.

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+    // The filter type name when specified with name
+    // will return quotas for the specified client id.
+    FILTER_TYPE_NAME = 2;
+    // The filter type default returns the default client id
+    FILTER_TYPE_DEFAULT = 3;
+  }
+  Quota.EntityType entity_type = 1;
+  FilterType filter_type = 2;
+  string name = 3;
+}
```
weeco: I think we can get rid of the filter type and setting the `QuotasMatchType` (in the client) dependent on what is given by the user here to simplify the API. WDYT?

Basically: If a field is present filter by it. Additional validation annotation still necessary.

```proto
message ListQuotasRequest {
  message Filter {
    // User or ClientID?
    Quota.EntityType entity_type = 1;
    
    // Optional: The exact name of the entity to filter for (e.g., a specific client ID).
    string entity_name = 2;

    // Optional: A prefix to match against entity names.
    string entity_name_prefix = 3;
  }

  Filter filter = 1;
  
  int32 page_size = 2 [
    (buf.validate.field).int32 = {
      gte: -1
      lte: 1000
    },
      description: "Limit the paginated response to a number of items. Defaults to 100. Use -1 to disable pagination."
      minimum: -1
      maximum: 1000
    }
  ];
  
  // Value of the next_page_token field returned by the previous response. If not provided, the system assumes the first page is requested.
  string page_token = 3;
}
```

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_KAFKA
+    };
+  }
+
+  rpc CreateQuota(CreateQuotaRequest) returns (CreateQuotaResponse) {
```
weeco: NIT: I would rename to SetQuotas since it's idempotent and a bit more of an upsert than create under the hood?

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+message ListQuotasResponse {
+  // QuotaEntry represents the quota for a specific entity type.
+  message QuotaEntry {
+    Quota.Entity entity = 1;
+    repeated Quota.Value values = 2;
+  }
+  repeated QuotaEntry quotas = 1;
+}
+
+message RequestEntity {
```
weeco: I think this name is too generic since it would be valid for the entire dataplane package.

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+    ENTITY_TYPE_IP = 5; // Not supported in Redpanda
+  }
+
+  // The type of value that the quota represents.
+  enum ValueType {
+    VALUE_TYPE_UNSPECIFIED = 0;
+    VALUE_TYPE_ANY = 1;
+    VALUE_TYPE_PRODUCER_BYTE_RATE = 2;
+    VALUE_TYPE_CONSUMER_BYTE_RATE = 3;
+  }
```
weeco:
```suggestion
  enum ValueType {
    VALUE_TYPE_UNSPECIFIED = 0;
    // A special type that matches any quota value. Only used for deleting.
    VALUE_TYPE_ANY = 1;
    // The rate at which producers can send data.
    VALUE_TYPE_PRODUCER_BYTE_RATE = 2;
    // The rate at which consumers can fetch data.
    VALUE_TYPE_CONSUMER_BYTE_RATE = 3;
  }
```

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,222 @@
...
+
+  RequestEntity entity = 1 [(buf.validate.field).required = true];
+
+  // List of values that should be deleted for the specified entities
+  Quota.ValueType value_type = 2 [(buf.validate.field).required = true];
+}
+
+message DeleteQuotaResponse {}
+
+service QuotaService {
```
weeco, line 191: Possibly good to have batch **and** single quotas consistently. Currently we have a batch api for create but a singular rpc for delete.

```proto
  rpc ListQuotas(ListQuotasRequest) returns (ListQuotasResponse) {
    option (google.api.http) = {get: "/v1/quotas"};
    // ... (OpenAPI annotations)
  }

  rpc SetQuota(SetQuotaRequest) returns (SetQuotaResponse) {
    option (google.api.http) = {
      post: "/v1/quotas"
      body: "*"
    };
    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      summary: "Set a quota"
      description: "Create or update client throughput quotas for a single entity. This operation is an upsert."
    };
    // ... (Auth annotations)
  }

  rpc BatchSetQuotas(BatchSetQuotasRequest) returns (BatchSetQuotasResponse) {
    option (google.api.http) = {
      post: "/v1/quotas:batchSet"
      body: "*"
    };
    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      summary: "Set multiple quotas"
      description: "Create or update client throughput quotas for multiple entities in a single request."
    };
    // ... (Auth annotations)
  }

  rpc DeleteQuota(DeleteQuotaRequest) returns (DeleteQuotaResponse) {
    option (google.api.http) = {
      delete: "/v1/quotas/{entityType}/{entityName}" // Not sure if that would work?
      body: "*"
    };
    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      summary: "Delete a quota"
      description: "Delete client quotas for a single entity."
    };
    // ... (Auth annotations)
  }

  rpc BatchDeleteQuotas(BatchDeleteQuotasRequest) returns (BatchDeleteQuotasResponse) {
    option (google.api.http) = {
      delete: "/v1/quotas:batchDelete"
      body: "*"
    };
    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
      summary: "Delete multiple quotas"
      description: "Delete client quotas for multiple entities in a single request."
    };
```

backend/pkg/console/servicer.go
```diff
@@ -30,7 +30,8 @@ type Servicer interface {
 	DeleteConsumerGroupOffsets(ctx context.Context, groupID string, topics []kmsg.OffsetDeleteRequestTopic) ([]DeleteConsumerGroupOffsetsResponseTopic, error)
 	DeleteTopic(ctx context.Context, topicName string) *rest.Error
 	DeleteTopicRecords(ctx context.Context, deleteReq kmsg.DeleteRecordsRequestTopic) (DeleteTopicRecordsResponse, *rest.Error)
-	DescribeQuotas(ctx context.Context) QuotaResponse
+	DescribeQuotas(ctx context.Context, request kmsg.DescribeClientQuotasRequest) QuotaResponse
+	AlterQuotas(ctx context.Context, request kmsg.AlterClientQuotasRequest) error
```
weeco: I think we should convert those to the "Plain kafka requests" (see the same file further down). Then it's basically just a proxy that takes the original franz-go request and also returns the original franz-go response. We are trying to move towards a flatter architecture which allows us to do better error responses in the API package.

e.g.
```go
DescribeQuotas(ctx context.Context, request kmsg.DescribeClientQuotasRequest) (kmsg.DescribeClientQuotasResponse, error)
```

proto/redpanda/api/dataplane/v1/quota.proto
c-julin: You're right i was looking at some other enums in the project and thought it was a standard similar to how proto enums require an unspecified. Will remove.

proto/redpanda/api/dataplane/v1/quota.proto
c-julin: My thinking was we can differentiate between default and any without the filter type. I do agree its complicated as is. So open to it remove and user has to specify "<default>" as name if they want to list default quotas

proto/redpanda/api/dataplane/v1/quota.proto
c-julin: i tried to look this up, and i cant find any documentation to say there are any restrictions on name and don't want to arbitrarily assign a limit 😕

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,278 @@
...
+  ];
+  int64 value = 2;
+}
+
+message BatchSetQuotaRequest {
+  // To specify multiple entities to apply the same quota
+  repeated RequestQuotaEntity entities = 1 [(buf.validate.field).required = true];
+  // The values that should be set for the specified entities.
+  repeated RequestQuotaValue values = 2 [(buf.validate.field).required = true];
+}
```
weeco: I think we don't want this sort of batch request, but rather something like below. I think this parallel array is a bit confusing otherwise (your comments explain the purpose), but it wouldn't allow you to provide different quotas for a set of ACLs.

I think the following is more explicit and allows you to set different quotas for different entities in a single atomic request, just like the Kafka API allows. 

```proto
message BatchSetQuotaRequest {
  message QuotaSetting {
    RequestQuotaEntity entity = 1 [(buf.validate.field).required = true];
    repeated RequestQuotaValue values = 2 [(buf.validate.field).min_items = 1];
  }

  // A list of quota settings to apply. Each setting pairs an entity
  // with the specific values to be set for it.
  repeated QuotaSetting settings = 1 [(buf.validate.field).min_items = 1];
}
```

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,278 @@
...
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_ADMIN
+      api: API_KAFKA
+    };
+  }
+
+  // Delete quota for a specific entity and value type.
+  // If value type is unspecified, all values for the entity will be deleted.
+  rpc DeleteQuota(DeleteQuotaRequest) returns (DeleteQuotaResponse) {
+    option (google.api.http) = {delete: "/v1/quotas"};
```
weeco:
```suggestion
    option (google.api.http) = {post: "/v1/quotas:delete"};
    body: "*"
```

I'm leaning towards this because in REST world DELETE /v1/quotas would usually mean to delete the entire collection. We don't have an identifier or good way to use a set of parameters that we could include in the URL to make this more explicit (e.g. something like DELETE `/v1/quotas/produce-rate/client-id/franz` would be cool, but we have to deal with default quotas too).

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  message Entity {
+    Quota.EntityType entity_type = 1;
+    string entity_name = 2;
+  }
+
+  message Value {
+    Quota.ValueType value_type = 1;
+    int64 value = 2;
```
weeco:
```suggestion
    double value = 2;
```

The go type uses float64, I guess fractional units are allowed in some cases?

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,278 @@
...
+
+  // Value that should be deleted for the specified entity
+  repeated Quota.ValueType value_types = 2 [
+    (buf.validate.field).repeated.items.enum = {
+      defined_only: true
+      not_in: [0] /* Disallow VALUE_TYPE_UNSPECIFIED */
+    },
+    (buf.validate.field).required = true
+  ];
+}
```
weeco: If we adapt the BatchSet we also need to adapt this one then.

```proto
message BatchDeleteQuotaRequest {
  // A deletion pairs a single entity with the value types to remove.
  message QuotaDeletion {
    QuotaTarget entity = 1 [(buf.validate.field).required = true];
    // List of value types to delete. To delete all quotas for the entity, leave this empty.
    repeated Quota.ValueType value_types = 2;
  }
  repeated QuotaDeletion deletions = 1 [(buf.validate.field).min_items = 1];
}
```

backend/pkg/api/connect/service/quota/v1/service.go
```diff
@@ -0,0 +1,169 @@
...
+
+	resp, err := s.consoleSvc.DescribeClientQuotas(ctx, kafkaReq)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to describe quotas: %w", err))
+	}
+
+	// Handle errors from the console service
+	if err = kerr.ErrorForCode(resp.ErrorCode); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to describe quotas: %s", err))
+	}
```
weeco: We have a helper for that already, look for: `NewConnectErrorFromKafkaErrorCode`

backend/pkg/api/connect/service/quota/v1/service.go
```diff
@@ -0,0 +1,169 @@
...
+			return quotaEntries[i].Entity.EntityName < quotaEntries[j].Entity.EntityName
+		})
+		page, token, err := pagination.SliceToPaginatedWithToken(quotaEntries, int(req.Msg.PageSize), req.Msg.GetPageToken(), "quotaEntry", func(x *v1.ListQuotasResponse_QuotaEntry) string {
+			return fmt.Sprintf("%d_%s", x.Entity.EntityType, x.Entity.EntityName)
+		})
+		if err != nil {
+			return nil, connect.NewError(
+				connect.CodeInternal,
+				fmt.Errorf("failed to apply pagination: %w", err),
+			)
```
weeco: NIT: we try to add a reason on all errors (which can be used by the frontend potentially to differentiate)

```suggestion
			return nil, connect.NewError(
				connect.CodeInternal,
				fmt.Errorf("failed to apply pagination: %w", err),
				apierrors.NewErrorInfo(v1.Reason_REASON_CONSOLE_ERROR.String()),
			)
```

backend/pkg/api/connect/service/quota/v1/service.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// BatchSetQuota creates client throughput quotas for specified entities.
+func (s *Service) BatchSetQuota(ctx context.Context, req *connect.Request[v1.BatchSetQuotaRequest]) (*connect.Response[v1.BatchSetQuotaResponse], error) {
+	kafkaReq, err := s.kafkaClientMapper.batchSetQuotaRequestToKafka(req.Msg)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	_, err = s.consoleSvc.AlterClientQuotas(ctx, kafkaReq)
```
weeco: In the single SetQuota you checked for an error (`if err = kerr.ErrorForCode(resp.ErrorCode); err != nil {`), we don't want to error the response here if one of the quotas failed, we must add it as part of our response.

So in the proto `message BatchSetQuotaResponse {}` we have to mimic somewhat what the kafka response does I guess. Otherwise the user has no chance of knowing what quotas were set successfully and which weren't

backend/pkg/api/connect/service/quota/v1/service.go
```diff
@@ -0,0 +1,169 @@
...
+func (s *Service) DeleteQuota(ctx context.Context, req *connect.Request[v1.DeleteQuotaRequest]) (*connect.Response[v1.DeleteQuotaResponse], error) {
+	kafkaReq, err := s.kafkaClientMapper.deleteQuotaRequestToKafka(req.Msg)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	_, err = s.consoleSvc.AlterClientQuotas(ctx, kafkaReq)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to delete quota: %w", err))
+	}
```
weeco: We need to inspect the inner kafka error here as well. You did it with the list quota rpc

See:

```go
type AlterClientQuotasResponseEntry struct {
	// ErrorCode is the error code for an alter on a matched entity.
	ErrorCode int16

	// ErrorMessage is an informative message if the alter on this entity failed.
	ErrorMessage *string
```

backend/pkg/api/connect/service/quota/v1/service.go
```diff
@@ -0,0 +1,169 @@
...
+func (s *Service) SetQuota(ctx context.Context, req *connect.Request[v1.SetQuotaRequest]) (*connect.Response[v1.SetQuotaResponse], error) {
+	kafkaReq, err := s.kafkaClientMapper.setQuotaRequestToKafka(req.Msg)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	_, err = s.consoleSvc.AlterClientQuotas(ctx, kafkaReq)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to set quota: %w", err))
+	}
```
weeco: Check for inner kafka error

backend/pkg/api/connect/service/quota/v1/service.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
+	_, err = s.consoleSvc.AlterClientQuotas(ctx, kafkaReq)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to batch delete quota: %w", err))
+	}
+
+	response := &v1.BatchDeleteQuotaResponse{}
```
weeco: We need to adapt the response here (same problem as with batch set) - inner kafka errors may exist (e.g. authz issues)

proto/redpanda/api/dataplane/v1/quota.proto
```diff
@@ -0,0 +1,278 @@
...
+
+message ListQuotasRequest {
+  message Filter {
+    // The type of entity to filter by. If  unspecified, all entities will be returned.
+    Quota.EntityType entity_type = 1;
+
+    // Optional: The exact name of the entity to filter by. If not specified, all entities of the specified type will be returned.
+    // Defaults to returning all entities of the specified type. To return default entity (e.g., default client-id), use "<default>".
+    string entity_name = 2;
```
weeco: Ah i see why it's needed, but I don't like using this special `entity_name` <default> due to discoverability. Would it be a better idea to add a prop specifically for this case? To list the default entities then just leave entity_name blank? Not ideal either, but more discoverable maybe from an API perspective.

```proto
bool match_default = 3
```

proto/redpanda/api/dataplane/v1/quota.proto
c-julin: I agree, this makes it feel more like a batch api. allowing multiple entities would just be confusing. have updated.

proto/redpanda/api/dataplane/v1/quota.proto
c-julin: this makes for the neatest solution and yeah expecting the user to type <default> would have probably gone badly

proto/redpanda/api/dataplane/v1/quota.proto
weeco: We can assume the same constraints as we have in place for topic names. We can relax it as needed in the future:

```proto
[
      (buf.validate.field).string.max_len = 249,
      (buf.validate.field).string.pattern = "^[a-zA-Z0-9._\\-]*$"
    ];
```

PR #1861, opened 2025-08-04
Backend: Kafka Quota API

c-julin:


taskfiles/backend.yaml
```diff
@@ -165,9 +165,11 @@ tasks:
   test-integration:
     dir: "{{.BACKEND_ROOT}}"
     desc: Run integration tests
+    deps:
+      - install-tparse
```
c-julin, line 169: Introducing tparse to tidy up the output of the test, as it was hard to find the failures

PR #1862, opened 2025-08-05
fix: implement fault-tolerant protobuf schema compilation

weeco:
## Summary

Fixes an issue where users experienced compilation errors of their protobuf schemas that resulted in all schemas not being compiled. The problem occurred when some schemas had broken references (e.g., to soft-deleted dependencies in the schema registry).

## Changes

- **Fault-tolerant compilation**: Replace all-or-nothing approach with best-effort schema compilation
- **Graceful error handling**: Skip problematic schemas instead of aborting entire refresh
- **Enhanced logging**: Warn about failed schemas and provide success/failure metrics
- **Integration test**: Add test case that reproduces and validates the fix

## Technical Details

The issue manifested when Console encountered protobuf schemas referencing non-existent or soft-deleted schemas. Previously, this would cause the entire schema compilation process to abort, leaving users unable to deserialize any protobuf messages.

With this fix:
- Valid schemas are compiled successfully
- Broken schemas are logged as warnings and skipped
- Only fails if zero schemas can be compiled
- Provides clear metrics on compilation success/failure rates

## Impact

Users with mixed schema environments (some valid, some with broken references) will now be able to use Console's protobuf deserialization for working schemas instead of losing all functionality.

### DIFF
--- a/backend/pkg/schema/service.go
+++ b/backend/pkg/schema/service.go
@@ -11,6 +11,7 @@ package schema
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"maps"
 	"strconv"
@@ -113,7 +114,7 @@ func (s *Service) GetProtoDescriptors(ctx context.Context) (map[int]*desc.FileDe
 		// 3. Compile all fetched schemas into a new, temporary map.
 		// Skip schemas that fail compilation instead of aborting the entire refresh.
 		newProtoFDByID := make(map[int]*desc.FileDescriptor)
-		var compilationErrors []error
+		failedCount := 0
 		compileStart := time.Now()
 		for _, schema := range newProtoSchemasByID {
 			fd, err := s.compileProtoSchemas(schema, schemasBySubjectAndVersion)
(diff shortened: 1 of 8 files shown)


backend/pkg/schema/service.go
```diff
@@ -111,29 +111,38 @@ func (s *Service) GetProtoDescriptors(ctx context.Context) (map[int]*desc.FileDe
...
-				return nil, fmt.Errorf("failed to compile schema for subject %q (id %d): %w", schema.Subject, schema.SchemaID, err)
+				compilationErrors = append(compilationErrors, err)
+				continue // Skip this schema, continue with others
 			}
 			newProtoFDByID[schema.SchemaID] = fd
 		}
 
+		// Only fail if NO schemas could be compiled and we had schemas to compile
+		if len(newProtoFDByID) == 0 && len(newProtoSchemasByID) > 0 {
+			return nil, fmt.Errorf("failed to compile any protobuf schemas. First error: %w", compilationErrors[0])
```
c-julin: would it be worth doing errors.join() and  keeping the fmt'd error from line 124 so you don't lose any of the information?

weeco: Actually we already log them above in the loop as warn messages. That made me realized we shouldn't even collect them. I changed the code a bit to just count the number of failures and return that.

c-julin (approved):

PR #1864, opened 2025-08-07
Backend: Add Schema Registry ACL support to CreateACL and DeleteACL

r-vasquez:
This PR adds support for Schema Registry ACLs (create, list, and delete operations) to the existing methods in the ACL service.

The new methods detect whether Schema Registry ACLs are supported in the cluster, meaning SR is configured, enabled, and the cluster has a license, to avoid failures on clusters that only use Kafka ACLs, ensuring backward compatibility.

**Additional details:**

- Dispatcher pattern: Implements a unified approach for handling both Kafka and Schema Registry ACLs.
- Caching: Adds a 15-minute cache for detecting Schema Registry ACL capabilities.
- Error handling: Introduces dedicated error mapping and validation for Schema Registry operations.


backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -83,6 +85,72 @@ func NewConnectErrorFromRedpandaAdminAPIError(err error, prefixErrMsg string) *c
...
+// more insightful connect.Error messages for Schema Registry errors.
+func NewConnectErrorFromSchemaRegistryError(err error, prefixErrMsg string) *connect.Error {
+	var srErr *sr.ResponseError
+	if errors.As(err, &srErr) {
+		return NewConnectError(
+			codeFromSRError(srErr),
+			errors.New(prefixErrMsg+srErr.Error()),
+			NewErrorInfo(
+				v1.Reason_REASON_REDPANDA_SCHEMA_REGISTRY_ERROR.String(),
+				KeyVal{"sr_error_code", strconv.Itoa(srErr.ErrorCode)},
```
weeco: We could also add the actual error name here, see: https://github.com/twmb/franz-go/blob/master/pkg/sr/errors.go

Otherwise we have to map the status code to the name as well. I'd keep the sr_error_code and add sr_error_name maybe.

backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -83,6 +85,72 @@ func NewConnectErrorFromRedpandaAdminAPIError(err error, prefixErrMsg string) *c
...
+				KeyVal{"sr_status_code", strconv.Itoa(srErr.StatusCode)},
```
weeco:
```suggestion
				KeyVal{"sr_http_status_code", strconv.Itoa(srErr.StatusCode)},
```

Error_code and status_code are so similar

backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -83,6 +85,72 @@ func NewConnectErrorFromRedpandaAdminAPIError(err error, prefixErrMsg string) *c
 	)
 }
 
+// NewConnectErrorFromSchemaRegistryError enhances error handling by providing
+// more insightful connect.Error messages for Schema Registry errors.
+func NewConnectErrorFromSchemaRegistryError(err error, prefixErrMsg string) *connect.Error {
```
weeco: I would put this into a new `schema_registry.go` file. We already have kafka.go for Kafka API errors, redpanda.go for Redpanda admin api errors, I think that would split it nicely. WDYT?

backend/pkg/api/connect/errors/redpanda.go
```diff
@@ -83,6 +85,72 @@ func NewConnectErrorFromRedpandaAdminAPIError(err error, prefixErrMsg string) *c
...
+
+	// Fallback.
+	return NewConnectError(
+		connect.CodeInternal,
+		fmt.Errorf("%v%w", prefixErrMsg, err),
+		NewErrorInfo(v1.Reason_REASON_REDPANDA_SCHEMA_REGISTRY_ERROR.String()),
+	)
+}
+
+func codeFromSRError(srErr *sr.ResponseError) connect.Code {
```
weeco: Nice, potentially we could use https://github.com/twmb/franz-go/blob/29dc44e616cd48d1c37bfae651ee9ac7120c32f7/pkg/sr/errors.go#L181-L196 and/or extend it. But this works of course just fine too

backend/pkg/api/connect/service/acl/v1/dispatcher.go
```diff
@@ -0,0 +1,46 @@
...
+	apierrors "github.com/redpanda-data/console/backend/pkg/api/connect/errors"
+	v1 "github.com/redpanda-data/console/backend/pkg/protogen/redpanda/api/dataplane/v1"
+)
+
+// targetACL represents which ACL subsystem should be involved in an operation.
+type targetACL struct {
+	IncludeKafka bool // Query/modify Kafka ACLs.
+	IncludeSR    bool // Query/modify Schema Registry ACLs.
+	IsSROnly     bool // This is exclusively a Schema Registry operation.
+}
```
weeco: I think we can just make that an enum with some functions. The three bools that are directly related to each other made me think about it. Something like:

```go
type Target uint8

const (
	TargetNone  Target = 0
	TargetKafka Target = 1 << iota
	TargetSR
	TargetBoth = TargetKafka | TargetSR
)

func (t Target) IncludesKafka() bool { return t&TargetKafka != 0 }
func (t Target) IncludesSR() bool    { return t&TargetSR != 0 }
func (t Target) IsSROnly() bool      { return t == TargetSR }
```

backend/pkg/api/connect/service/acl/v1/mapper.go
```diff
@@ -348,130 +337,91 @@ func (srm *schemaRegistryMapper) srACLToProto(acl rpsr.ACL) (*v1.ListACLsRespons
...
-	case v1.ACL_PERMISSION_TYPE_ALLOW:
-		return rpsr.PermissionAllow
-	case v1.ACL_PERMISSION_TYPE_DENY:
-		return rpsr.PermissionDeny
-	case v1.ACL_PERMISSION_TYPE_ANY:
-		return "" // TODO: temporary solution, RP doesn't support ANY yet.
-	default:
-		return ""
+func (srm *schemaRegistryMapper) deleteACLMatchingResultsToProtos(acls []rpsr.ACL) ([]*v1.DeleteACLsResponse_MatchingACL, error) {
+	matchingACLs := make([]*v1.DeleteACLsResponse_MatchingACL, 0, len(acls))
```
weeco:
```suggestion
	matchingACLs := make([]*v1.DeleteACLsResponse_MatchingACL, len(acls))
```

Otherwise out of bounds for direct assignments in the loop

backend/pkg/api/connect/service/acl/v1/mapper_enums.go
```diff
@@ -180,3 +181,131 @@ func (*kafkaClientMapper) aclResourceTypeToProto(resourceType kmsg.ACLResourceTy
...
+}
+
+func (*schemaRegistryMapper) protoResourceTypeToSR(resourceType v1.ACL_ResourceType) rpsr.ResourceType {
+	switch resourceType {
+	case v1.ACL_RESOURCE_TYPE_REGISTRY:
+		return rpsr.ResourceTypeRegistry
+	case v1.ACL_RESOURCE_TYPE_SUBJECT:
+		return rpsr.ResourceTypeSubject
+	case v1.ACL_RESOURCE_TYPE_ANY:
+		return "" // TODO: temporary solution, RP doesn't support ANY yet.
```
weeco, line 256: Is this ANY then rejected in validation or what is supposed to happen in this case?

backend/pkg/api/connect/service/acl/v1/mapper_enums.go
```diff
@@ -180,3 +181,131 @@ func (*kafkaClientMapper) aclResourceTypeToProto(resourceType kmsg.ACLResourceTy
...
+}
+
+func (*schemaRegistryMapper) protoPatternTypeToSR(patternType v1.ACL_ResourcePatternType) rpsr.PatternType {
+	switch patternType {
+	case v1.ACL_RESOURCE_PATTERN_TYPE_LITERAL:
+		return rpsr.PatternTypeLiteral
+	case v1.ACL_RESOURCE_PATTERN_TYPE_PREFIXED:
+		return rpsr.PatternTypePrefix
+	case v1.ACL_RESOURCE_PATTERN_TYPE_ANY:
+		return "" // TODO: temporary solution, RP doesn't support ANY yet.
```
weeco, line 269: Same Q here as above - wondering whether this is case is fine to hit. As a user I'd be potentially trying the any type for listing

backend/pkg/api/connect/service/acl/v1/service.go
```diff
@@ -40,28 +41,117 @@ type Service struct {
...
+
+	srACLSupportCache *cache.Cache[string, bool]
 }
 
 // NewService creates a new ACL service handler.
 func NewService(cfg *config.Config,
 	logger *slog.Logger,
 	consoleSvc console.Servicer,
 ) *Service {
+	cacheSettings := []cache.Opt{
```
weeco, line 54: We need to clarify for the serverless / multi-tenant envs whether this is an all vclusters or no vclusters thing. If this can be enabled / disabled per VCluster we need to bring in the "cacheNamespaceFn" (see schema/client.go), where this will then be cached only for each individual user.

backend/pkg/api/connect/service/acl/v1/service.go
```diff
@@ -215,6 +296,7 @@ func (s *Service) DeleteACLs(ctx context.Context, req *connect.Request[v1.Delete
 		)
 	}
 
+	// TODO: rogger, verify what if there are no matching ACLs, for example: try to Delete with all ANY if only a SR ACL is present.
```
weeco: Do you want to resolve before merging or with the next PR?

backend/pkg/api/connect/service/acl/v1/mapper.go
r-vasquez: Oops, correct.

backend/pkg/api/connect/service/acl/v1/mapper_enums.go
r-vasquez: If we send ANY right now, it's rejected in validation. But Core will add support for it in the upcoming versions

backend/pkg/api/connect/service/acl/v1/mapper_enums.go
r-vasquez: An empty string is the same as ANY for the Schema Registry ACL, that's why I'm parsing it like this.

So, if the user is trying ANY for listing, then it will be ok to use it like this 👍

backend/pkg/api/connect/service/acl/v1/service.go
r-vasquez: I'm resolving it. This was a note to myself for testing, sorry about that.

Answer is: Kafka client returns no Matching ACLs and there are no issues 👍

backend/pkg/api/connect/service/acl/v1/service.go
r-vasquez: Just checked, it's an 'All VCluster` situation. So, IIUC we don't need to handle the cache for each user.

backend/pkg/api/connect/errors/schema_registry_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+			assert.Equal(t, tc.expectedCode, connectErr.Code(), "Expected connect code to match")
+			assert.Equal(t, tc.expectedErrorMsg, connectErr.Message(), "Expected error message to match")
+
+			// Check that the ErrorInfo contains the correct reason
+			details := connectErr.Details()
+			assert.NotEmpty(t, details, "Expected error details to be present")
+
+			// Verify that the error has the expected reason
+			found := false
```
c-julin: errorInfo := &errdetails.ErrorInfo{}
			for _, detail := range details {
				value, err := detail.Value()
				require.NoError(t, err)
				if errDetails, ok := value.(*errdetails.ErrorInfo); ok && errDetails.Reason == tc.expectedReason {
					errorInfo = errDetails
					break
				}
			}

			require.NotNil(t, errorInfo, "Expected to find ErrorInfo with correct reason")

			var srErr *sr.ResponseError
			if errors.As(tc.err, &srErr) {
				assert.Contains(t, errorInfo.Metadata, "sr_error_code", "Expected sr_error_code in metadata")
				assert.Contains(t, errorInfo.Metadata, "sr_http_status_code", "Expected sr_http_status_code in metadata")
				assert.Contains(t, errorInfo.Metadata, "sr_error_name", "Expected sr_error_name in metadata")
			}

could tidy it by doing this

backend/pkg/api/connect/service/acl/v1/mapper.go
```diff
@@ -298,29 +298,18 @@ func (srm *schemaRegistryMapper) listACLFilterToDescribeACLSR(filter *v1.ListACL
...
-			principal = *filter.Principal
-		}
-		if filter.ResourceName != nil {
-			resourceName = *filter.ResourceName
-		}
-		if filter.Host != nil {
-			host = *filter.Host
-		}
-
+	if rt := filter.GetResourceType(); rt == v1.ACL_RESOURCE_TYPE_REGISTRY || rt == v1.ACL_RESOURCE_TYPE_SUBJECT || rt == v1.ACL_RESOURCE_TYPE_ANY {
```
c-julin:
```go
schemaRegistryACLs := map[v1.ACL_ResourceType]struct{}{
		v1.ACL_RESOURCE_TYPE_REGISTRY: {},
		v1.ACL_RESOURCE_TYPE_SUBJECT:  {},
		v1.ACL_RESOURCE_TYPE_ANY:      {},
	}
	
	if _, ok := schemaRegistryACLs[filter.GetResourceType()]; !ok {
		return nil
	}
```

can be cleaned up with a map and return early if not found

proto/redpanda/api/dataplane/v1/acl.proto
```diff
@@ -112,9 +112,9 @@ message CreateACLRequest {
   option (buf.validate.message).cel = {
     id: "resource_name_must_be_set_except_for_cluster_resource_type"
     expression:
-      "this.resource_type == 4 && size(this.resource_name) == 0 ? ''"
+      "(this.resource_type == 4 || this.resource_type == 8) && size(this.resource_name) == 0 ? ''"
```
c-julin, line 120: very minor but i think its worth commenting the written version of the rule,  dont think we do it enough on our cel expressions (including me :'( )

andresaristizabal (approved):

proto/redpanda/api/dataplane/v1/acl.proto
r-vasquez: Agree, added a new commit with this.

backend/pkg/api/connect/service/acl/v1/mapper.go
r-vasquez: Abstracted this to a method of the mapper as is used in both list and delete Thanks!

PR #1866, opened 2025-08-12
refactor MCPServer resource

birdayz:
do not talk about config files, but about MCP tools. this makes the resource more natural to use:

Declare tools and their YAMLs.
Also add field "component_type"; we need to know what type it is to mount it into the correct folder.

this takes away the burden to understand the folder structure. in addition, it makes the folder structure an implementation detail, not part of the API.


proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -38,9 +38,10 @@ message MCPServer {
   string description = 3 [(buf.validate.field).string.max_len = 256];
 
   // A configuration file for the MCP server.
-  message Configuration {
+  message Tool {
+    string component_type = 1;
```
weeco: What's this component type? What are different component types we may want to use here for defining a set of MCP tools?

birdayz: it's either cache or processor. it is necessary to mount the config into the correct path

weeco: Maybe enum instead?

birdayz: updated to enum

birdayz:
moved to v1alpha3

proto/redpanda/api/dataplane/v1alpha3/mcp.proto
```diff
@@ -327,7 +335,7 @@ service MCPServerService {
 // MCPServerCreate contains the details for the MCP Server creation request.
 message MCPServerCreate {
   option (google.api.resource) = {
-    type: "redpanda.api.dataplane.v1/MCPServerCreate"
+    type: "redpanda.api.dataplane.v1alpha3alpha3/MCPServerCreate"
```
weeco: v1alpha3alpha3 Duplicate `alpha3`

birdayz: fixed

weeco (approved):

buf.yaml
```diff
@@ -25,5 +25,7 @@ breaking:
     - EXTENSION_NO_DELETE
     - FIELD_SAME_DEFAULT
   ignore:
-    - redpanda/api/console/v1alpha1
-    - redpanda/api/dataplane/v1alpha1
+    - proto/redpanda/api/console/v1alpha1
+    - proto/redpanda/api/dataplane/v1alpha1
+    - proto/redpanda/api/dataplane/v1alpha3
```
weeco, line 30: Wondering why that one is needed

rockwotj (approved):

buf.yaml
birdayz: it did not work, at all, without proto/ prefix. tested it out multiple times

PR #1876, opened 2025-08-19
Acl sr UI

andresaristizabal:
Add ACL and Role Management Features

  This PR improves Access Control List (ACL) and Role-Based Access Control (RBAC) management functionality in Console frontend.

  🎯 Overview

  This implementation provides a complete UI for managing security policies in Redpanda clusters, including:
  - ACL Management: Create, read, update, and delete access control rules
  - Role Management: Manage roles and their member assignments
  - Comprehensive Testing: End-to-end test coverage for all security features

  ✨ Features Added

  ACL Management

  - ACL Creation: Interactive form with resource type selection (Topic, Consumer Group, Cluster, etc.)
  - Rule Configuration: Support for custom operations, allow-all, and deny-all modes
  - Resource Selectors: Literal, prefix, and wildcard pattern matching
  - ACL Details: View and edit existing ACL configurations
  - Principal & Host Configuration: Shared configuration for all rules in an ACL

  Role Management

  - Role CRUD Operations: Complete create, read, update, delete functionality
  - Member Management: Add/remove users and principals from roles
  - Role Details: View role configuration and member assignments
  - Integration: Seamless integration with ACL system for role-based permissions
  
  Video: https://redpandadata.slack.com/archives/C081G37T71B/p1755619300758459?thread_ts=1755618971.432419&cid=C081G37T71B

### DIFF
--- a/frontend/src/components/pages/acls/Acl.List.tsx
+++ b/frontend/src/components/pages/acls/Acl.List.tsx
@@ -41,14 +41,13 @@ import { makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
 import { type FC, useEffect, useRef, useState } from 'react';
 import { BsThreeDots } from 'react-icons/bs';
-import { Link as ReactRouterLink } from 'react-router-dom';
+import { Link as ReactRouterLink, useNavigate } from 'react-router-dom';
 import ErrorResult from '../../../components/misc/ErrorResult';
 import { appGlobal } from '../../../state/appGlobal';
 import { api, rolesApi } from '../../../state/backendApi';
 import { AclRequestDefault } from '../../../state/restInterfaces';
 import { Features } from '../../../state/supportedFeatures';
 import { uiSettings } from '../../../state/ui';
-import { clone } from '../../../utils/jsonUtils';
 import { Code as CodeEl, DefaultSkeleton } from '../../../utils/tsxUtils';
 import { FeatureLicenseNotification } from '../../license/FeatureLicenseNotification';
 import { NullFallbackBoundary } from '../../misc/NullFallbackBoundary';
@@ -602,6 +601,8 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
   const [editorType, setEditorType] = useState<'create' | 'edit'>('create');
   const [edittingPrincipalGroup, setEdittingPrincipalGroup] = useState<AclPrincipalGroup | null>(null);
 
+  const nav = useNavigate();
+
   let groups = p.principalGroups.filter((g) => g.principalType === 'User');
   try {
     const quickSearchRegExp = new RegExp(uiSettings.aclList.configTable.quickSearch, 'i');
@@ -657,6 +658,7 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
           data-testid="create-acls"
           variant="outline"
           onClick={() => {
+            nav('create');
             setEditorType('create');
             setEdittingPrincipalGroup(
               observable({
@@ -694,8 +696,7 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
                       type="button"
                       className="hoverLink"
                       onClick={() => {
-                        setEditorType('edit');
-                        setEdittingPrincipalGroup(observable(clone(record)));
+                        nav(`/security/acls/${record.principalName}/details`);
                       }}
                     >
                       <Flex>
--- a/frontend/src/components/pages/acls/Acl.List.tsx
+++ b/frontend/src/components/pages/acls/Acl.List.tsx
@@ -9,6 +9,7 @@
  * by the Apache License, Version 2.0
  */
 
+import { create } from '@bufbuild/protobuf';
 import { PencilIcon, TrashIcon } from '@heroicons/react/outline';
 import {
   Alert,
@@ -39,10 +40,19 @@ import type { TabsItemProps } from '@redpanda-data/ui/dist/components/Tabs/Tabs'
 import { isServerless } from 'config';
 import { makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
-import { type FC, useEffect, useRef, useState } from 'react';
+import {
+  ACL_Operation,
+  ACL_PermissionType,
+  ACL_ResourcePatternType,
+  ACL_ResourceType,
+  type DeleteACLsRequest,
+  DeleteACLsRequestSchema,
+} from 'protogen/redpanda/api/dataplane/v1/acl_pb';
+import { type FC, useRef, useState } from 'react';
 import { BsThreeDots } from 'react-icons/bs';
 import { Link as ReactRouterLink, useNavigate } from 'react-router-dom';
 import ErrorResult from '../../../components/misc/ErrorResult';
+import { useDeleteAclMutation, useListACLAsPrincipalGroups } from '../../../react-query/api/acl';
 import { appGlobal } from '../../../state/appGlobal';
 import { api, rolesApi } from '../../../state/backendApi';
 import { AclRequestDefault } from '../../../state/restInterfaces';
@@ -537,6 +547,7 @@ const RolesTab = observer(() => {
                       as={ReactRouterLink}
                       to={`/security/roles/${encodeURIComponent(entry.name)}/details`}
                       textDecoration="none"
+                      data-testid={`role-list-item-${entry.name}`}
                     >
                       {entry.name}
                     </ChakraLink>
@@ -592,26 +603,49 @@ const RolesTab = observer(() => {
   );
 });
 
-const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
-  useEffect(() => {
-    void api.refreshAcls(AclRequestDefault, true);
-  }, []);
+const AclsTab = observer((_: { principalGroups: AclPrincipalGroup[] }) => {
+  const { data: principalGroups, isLoading } = useListACLAsPrincipalGroups();
+  const { mutateAsync: deleteACLMutation } = useDeleteAclMutation();
 
   const [aclFailed, setAclFailed] = useState<{ err: unknown } | null>(null);
   const [editorType, setEditorType] = useState<'create' | 'edit'>('create');
   const [edittingPrincipalGroup, setEdittingPrincipalGroup] = useState<AclPrincipalGroup | null>(null);
 
-  const nav = useNavigate();
+  const navigate = useNavigate();
 
-  let groups = p.principalGroups.filter((g) => g.principalType === 'User');
+  const deleteACLsForPrincipal = async (principal: string, host: string) => {
+    const deleteRequest: DeleteACLsRequest = create(DeleteACLsRequestSchema, {
+      filter: {
+        principal: principal,
+        resourceType: ACL_ResourceType.ANY,
+        resourceName: undefined,
+        host: host,
+        operation: ACL_Operation.ANY,
+        permissionType: ACL_PermissionType.ANY,
+        resourcePatternType: ACL_ResourcePatternType.ANY,
+      },
+    });
+    await deleteACLMutation(deleteRequest);
+    toast({
+      status: 'success',
+      description: (
+        <Text as="span">
+          Deleted ACLs for <CodeEl>{principal}</CodeEl>
+        </Text>
+      ),
+    });
+  };
+
+  let groups = principalGroups?.filter((g) => g.principalType === 'User') || [];
+
   try {
     const quickSearchRegExp = new RegExp(uiSettings.aclList.configTable.quickSearch, 'i');
-    groups = groups.filter((aclGroup) => aclGroup.principalName.match(quickSearchRegExp));
+    groups = groups?.filter((aclGroup) => aclGroup.principalName.match(quickSearchRegExp));
   } catch (_e) {
     console.warn('Invalid expression');
   }
 
-  if (api.ACLs?.aclResources === undefined) return DefaultSkeleton;
+  if (isLoading || !principalGroups) return DefaultSkeleton;
 
   return (
     <Flex flexDirection="column" gap="4">
@@ -622,7 +656,6 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
         without considering any permissions that may be derived from assigned roles. For a complete view of all
         effective permissions, including those granted through roles, refer to the Permissions List tab.
       </Box>
-
       {Features.rolesApi && (
         <Alert status="info">
           <AlertIcon />
@@ -630,14 +663,12 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
           hierarchies or large numbers of users.
         </Alert>
       )}
-
       <SearchField
         width="300px"
         searchText={uiSettings.aclList.configTable.quickSearch}
         setSearchText={(x) => (uiSettings.aclList.configTable.quickSearch = x)}
         placeholderText="Filter by name"
       />
-
       <Section>
         {edittingPrincipalGroup && (
           <AclPrincipalGroupEditor
@@ -658,7 +689,7 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
           data-testid="create-acls"
           variant="outline"
           onClick={() => {
-            nav('create');
+            navigate('create');
             setEditorType('create');
             setEdittingPrincipalGroup(
               observable({
@@ -678,8 +709,8 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
         </Button>
 
         <Box py={4}>
-          <DataTable<AclPrincipalGroup>
-            data={groups}
+          <DataTable<{ principal: string; host: string; principalType: string; principalName: string }>
+            data={groups || []}
             pagination
             sorting
             columns={[
@@ -696,12 +727,17 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
                       type="button"
                       className="hoverLink"
                       onClick={() => {
-                        nav(`/security/acls/${record.principalName}/details`);
+                        navigate(`/security/acls/${record.principalName}/details`);
                       }}
                     >
                       <Flex>
                         {/* <Badge variant="subtle" mr="2">{principalType}</Badge> */}
-                        <Text as="span" wordBreak="break-word" whiteSpace="break-spaces">
+                        <Text
+                          as="span"
+                          wordBreak="break-word"
+                          whiteSpace="break-spaces"
+                          data-testid={`acl-list-item-${record.principalName}-${record.host}`}
+                        >
                           {record.principalName}
                         </Text>
                       </Flex>
@@ -724,28 +760,11 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
                 header: '',
                 cell: ({ row: { original: record } }) => {
                   const userExists = api.serviceAccounts?.users.includes(record.principalName);
-                  const hasAcls = record.sourceEntries.length > 0;
 
                   const onDelete = async (user: boolean, acls: boolean) => {
                     if (acls) {
                       try {
-                        await api.deleteACLs({
-                          resourceType: 'Any',
-                          resourceName: undefined,
-                          resourcePatternType: 'Any',
-                          principal: `${record.principalType}:${record.principalName}`,
-                          host: record.host,
-                          operation: 'Any',
-                          permissionType: 'Any',
-                        });
-                        toast({
-                          status: 'success',
-                          description: (
-                            <Text as="span">
-                              Deleted ACLs for <CodeEl>{record.principalName}</CodeEl>
-                            </Text>
-                          ),
-                        });
+                        await deleteACLsForPrincipal(record.principal, record.host);
                       } catch (err: unknown) {
                         console.error('failed to delete acls', { error: err });
                         setAclFailed({ err });
@@ -779,7 +798,7 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
                       </MenuButton>
                       <MenuList>
                         <MenuItem
-                          isDisabled={!userExists || !Features.deleteUser || !hasAcls}
+                          isDisabled={!userExists || !Features.deleteUser}
                           onClick={(e) => {
                             void onDelete(true, true);
                             e.stopPropagation();
@@ -797,7 +816,6 @@ const AclsTab = observer((p: { principalGroups: AclPrincipalGroup[] }) => {
                           Delete (User only)
                         </MenuItem>
                         <MenuItem
-                          isDisabled={!hasAcls}
                           onClick={(e) => {
                             void onDelete(false, true);
                             e.stopPropagation();
--- a/frontend/src/components/pages/acls/UserDetails.tsx
+++ b/frontend/src/components/pages/acls/UserDetails.tsx
@@ -9,10 +9,15 @@
  * by the Apache License, Version 2.0
  */
 
-import { Box, Button, Link as ChakraLink, DataTable, Flex, Heading, Text } from '@redpanda-data/ui';
+import { useQuery } from '@connectrpc/connect-query';
+import { Box, DataTable, Text } from '@redpanda-data/ui';
+import { EmbeddedAclDetail } from 'components/pages/acls/new-acl/ACLDetails';
+import { Button } from 'components/redpanda-ui/components/button';
+import { Card, CardContent, CardHeader, CardTitle } from 'components/redpanda-ui/components/card';
 import { makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
-import { Link as ReactRouterLink } from 'react-router-dom';
+import type { ListACLsRequest } from '../../../protogen/redpanda/api/dataplane/v1/acl_pb';
+import { listACLs } from '../../../protogen/redpanda/api/dataplane/v1/acl-ACLService_connectquery';
 import { appGlobal } from '../../../state/appGlobal';
 import { api, rolesApi } from '../../../state/backendApi';
 import { AclRequestDefault } from '../../../state/restInterfaces';
@@ -21,7 +26,7 @@ import { DefaultSkeleton } from '../../../utils/tsxUtils';
 import PageContent from '../../misc/PageContent';
 import { PageComponent, type PageInitHelper } from '../Page';
 import { DeleteUserConfirmModal } from './DeleteUserConfirmModal';
-import { type AclPrincipalGroup, principalGroupsView } from './Models';
+import type { AclPrincipalGroup } from './Models';
 import { ChangePasswordModal, ChangeRolesModal } from './UserEditModals';
 import { UserRoleTags } from './UserPermissionAssignments';
 
@@ -82,6 +87,71 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
 
     return (
       <PageContent>
+        <div className="flex justify-between">
+          <div>
+            <h3>Permissions</h3>
+            <p className="text-sm text-gray-600">The following permissions are assigned to this principal.</p>
+          </div>
+          <div className="flex gap-3">
+            {Features.rolesApi && (
+              <Button variant="outline" onClick={() => (this.isChangeRolesModalOpen = true)} disabled={!canEdit}>
+                Assign roles
+              </Button>
+            )}
+            {api.isAdminApiConfigured && (
+              <Button variant="outline" onClick={() => (this.isChangePasswordModalOpen = true)} disabled={!canEdit}>
+                Change password
+              </Button>
+            )}
+            {/* todo: refactor delete user dialog into a "fire and forget" dialog and use it in the overview list (and here) */}
+            {isServiceAccount && (
+              <DeleteUserConfirmModal
+                onConfirm={async () => {
+                  await api.deleteServiceAccount(userName);
+
+                  // Remove user from all its roles
+                  const promises = [];
+                  for (const [roleName, members] of rolesApi.roleMembers) {
+                    if (members.any((m) => m.name === userName)) {
+                      // is this user part of this role?
+                      // then remove it
+                      promises.push(rolesApi.updateRoleMembership(roleName, [], [userName]));
+                    }
+                  }
+                  await Promise.allSettled(promises);
+                  await api.refreshServiceAccounts();
+                  await rolesApi.refreshRoleMembers();
+                  appGlobal.historyPush('/security/users/');
+                }}
+                buttonEl={
+                  <Button variant="destructive" disabled={!isServiceAccount}>
+                    Delete
+                  </Button>
+                }
+                userName={userName}
+              />
+            )}
+          </div>
+        </div>
+
+        <div className="grid grid-cols-5 gap-3 start">
+          <div className="sm:col-span-5 md:col-span-3">
+            <UserPermissionDetailsContent userName={userName} />
+          </div>
+
+          <div className="sm:col-span-5 md:col-span-2">
+            <Card size="full">
+              <CardHeader>
+                <CardTitle className="text-lg font-medium text-gray-900">Assignments</CardTitle>
+              </CardHeader>
+              <CardContent>
+                <UserRoleTags userName={userName} verticalView={false} />
+              </CardContent>
+            </Card>
+          </div>
+        </div>
+
+        {/*Modals*/}
         {api.isAdminApiConfigured && (
           <ChangePasswordModal
             userName={userName}
@@ -89,6 +159,7 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
             setIsOpen={(value: boolean) => (this.isChangePasswordModalOpen = value)}
           />
         )}
+
         {Features.rolesApi && (
           <ChangeRolesModal
             userName={userName}
@@ -96,57 +167,6 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
             setIsOpen={(value: boolean) => (this.isChangeRolesModalOpen = value)}
           />
         )}
-        <Flex gap="4">
-          {Features.rolesApi && (
-            <Button variant="outline" onClick={() => (this.isChangeRolesModalOpen = true)} isDisabled={!canEdit}>
-              Change roles
-            </Button>
-          )}
-          {api.isAdminApiConfigured && (
-            <Button variant="outline" onClick={() => (this.isChangePasswordModalOpen = true)} isDisabled={!canEdit}>
-              Change password
-            </Button>
-          )}
-          {/* todo: refactor delete user dialog into a "fire and forget" dialog and use it in the overview list (and here) */}
-          {isServiceAccount && (
-            <DeleteUserConfirmModal
-              onConfirm={async () => {
-                await api.deleteServiceAccount(userName);
-
-                // Remove user from all its roles
-                const promises = [];
-                for (const [roleName, members] of rolesApi.roleMembers) {
-                  if (members.any((m) => m.name === userName)) {
-                    // is this user part of this role?
-                    // then remove it
-                    promises.push(rolesApi.updateRoleMembership(roleName, [], [userName]));
-                  }
-                }
-                await Promise.allSettled(promises);
-                await api.refreshServiceAccounts();
-                await rolesApi.refreshRoleMembers();
-                appGlobal.historyPush('/security/users/');
-              }}
-              buttonEl={
-                <Button variant="outline-delete" isDisabled={!isServiceAccount}>
-                  Delete
-                </Button>
-              }
-              userName={userName}
-            />
-          )}
-        </Flex>
-
-        <Heading as="h3" mt="4">
-          Permissions
-        </Heading>
-        <Box>The following permissions are assigned to this principal.</Box>
-
-        <Heading as="h3" mt="4">
-          Assignments
-        </Heading>
-        <UserRoleTags userName={userName} />
-        <UserPermissionDetails userName={userName} />
       </PageContent>
     );
   }
@@ -154,51 +174,67 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
 
 export default UserDetailsPage;
 
-const UserPermissionDetails = observer((p: { userName: string }) => {
+const UserPermissionDetailsContent = observer((p: { userName: string }) => {
   // Get all roles and ACLs matching this user
-  // For each "AclPrincipalGroup" show its name, then a table that shows the details
-  const roles: string[] = [];
+  const roles: {
+    principalType: string;
+    principalName: string;
+  }[] = [];
+
   if (Features.rolesApi) {
     for (const [roleName, members] of rolesApi.roleMembers) {
       if (!members.any((m) => m.name === p.userName)) continue; // this role doesn't contain our user
-      roles.push(roleName);
+      roles.push({
+        principalType: 'RedpandaRole',
+        principalName: roleName,
+      });
     }
   }
 
-  // Get all AclPrincipal groups, find the ones that apply
-  const groups = principalGroupsView.principalGroups.filter((g) => {
-    if (g.principalType === 'User' && (g.principalName === p.userName || g.principalName === '*')) return true;
-    if (g.principalType === 'RedpandaRole' && roles.includes(g.principalName)) return true;
-    return false;
-  });
+  const { data: hasAcls } = useQuery(
+    listACLs,
+    {
+      filter: {
+        principal: `User:${p.userName}`,
+      },
+    } as ListACLsRequest,
+    {
+      enabled: !!p.userName,
+      select: (response) => {
+        return response.resources.length > 0;
+      },
+    },
+  );
 
-  return <PrincipalGroupTables groups={groups} />;
-});
+  if (hasAcls) {
+    roles.push({
+      principalType: 'User',
+      principalName: p.userName,
+    });
+  }
 
-// Renders an array of AclPrincipalGroup as multiple tabls below each other
-const PrincipalGroupTables = observer((p: { groups: AclPrincipalGroup[] }) => {
   return (
-    <>
-      {p.groups.map((r) => (
-        <>
-          {r.principalType === 'RedpandaRole' ? (
-            <ChakraLink key={r.principalName} as={ReactRouterLink} to={`/security/roles/${r.principalName}/details`}>
-              <Heading as="h3" mt="4">
-                {r.principalName}
-              </Heading>
-            </ChakraLink>
-          ) : (
-            <Heading key={r.principalName} as="h3" mt="4">
-              User: {r.principalName}
-            </Heading>
-          )}
-          <AclPrincipalGroupPermissionsTable key={r.principalName} group={r} />
-        </>
-      ))}
-    </>
+    <div className="gap-3 flex flex-col">
+      {roles.map((g) => {
+        return (
+          <EmbeddedAclDetail
+            principal={`${g.principalType}:${g.principalName}`}
+            key={`key-${g.principalType}:${g.principalName}`}
+          />
+        );
+      })}
+      {roles.length === 0 && (
+        <Card size="full">
+          <CardContent>
+            <p>No permissions assigned to this user.</p>
+          </CardContent>
+        </Card>
+      )}
+    </div>
   );
 });
 
+// TODO: remove this component when we update RoleDetails
 export const AclPrincipalGroupPermissionsTable = observer((p: { group: AclPrincipalGroup }) => {
   const entries: {
     type: string;
--- a/frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
+++ b/frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
@@ -1,0 +1,273 @@
+/**
+ * Copyright 2022 Redpanda Data, Inc.
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { Button } from 'components/redpanda-ui/components/button';
+import { Card, CardContent, CardHeader, CardTitle } from 'components/redpanda-ui/components/card';
+import { Edit } from 'lucide-react';
+import { useNavigate } from 'react-router-dom';
+import { useGetAclsByPrincipal } from '../../../../react-query/api/acl';
+import { MatchingUsersCard } from '../../roles/MatchingUsersCard';
+import {
+  formatLabel,
+  getIdFromRule,
+  getOperationsForResourceType,
+  getRuleDataTestId,
+  parsePrincipal,
+  type Rule,
+  type SharedConfig,
+} from './ACL.model';
+
+// Helper function to get resource name
+const getResourceName = (resourceType: string): string => {
+  const resourceNames: Record<string, string> = {
+    cluster: 'cluster',
+    topic: 'topic',
+    consumerGroup: 'consumer group',
+    transactionalId: 'transactional ID',
+    subject: 'subject',
+    schemaRegistry: 'schema registry',
+  };
+  return resourceNames[resourceType] || resourceType;
+};
+
+// Helper function to get plural resource name
+const getPluralResourceName = (resourceType: string): string => {
+  const pluralNames: Record<string, string> = {
+    cluster: 'clusters',
+    topic: 'topics',
+    consumerGroup: 'consumer groups',
+    transactionalId: 'transactional IDs',
+    subject: 'subjects',
+    schemaRegistry: 'schema registries',
+  };
+  return pluralNames[resourceType] || resourceType;
+};
+
+interface ACLDetailsProps {
+  sharedConfig: {
+    principal: string;
+    host: string;
+  };
+  rules: Rule[];
+  onUpdateACL: () => void;
+  showMatchingUsers?: boolean;
+  isSimpleView?: boolean; // this prop show SharedConfig, this is used for emmbedded this component on legacy user page
+}
+
+export function ACLDetails({
+  sharedConfig,
+  rules,
+  onUpdateACL,
+  showMatchingUsers = false,
+  isSimpleView = false,
+}: ACLDetailsProps) {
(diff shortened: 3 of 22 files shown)


frontend/src/components/redpanda-ui/accordion.tsx
andresaristizabal, line 1: omit files on `frontend/src/components/redpanda-ui/*`

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                </div>
+              </div>
+              <Button
+                className="bg-gray-900 hover:bg-gray-800 text-white"
+                onClick={onUpdateACL}
+                data-testid="update-acl-button"
+              >
+                <Edit className="w-4 h-4 mr-2" />
+                Edit
+              </Button>
```
eblairmckee: Button has an `icon` prop, pass Edit (without any classNames to it)

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+          <div className=" mx-auto">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center gap-4">
+                <div>
+                  <h1 className="text-xl font-semibold text-gray-900">ACL Configuration Details</h1>
+                  <p className="text-sm text-gray-600">View the created ACL configuration</p>
+                </div>
+              </div>
+              <Button
+                className="bg-gray-900 hover:bg-gray-800 text-white"
```
eblairmckee: there should be a variant for it? maybe `outline`

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+  return (
+    <div className=" ">
+      {/* Header */}
+      <div>
+        <div className=" py-4">
+          <div className=" mx-auto">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center gap-4">
+                <div>
+                  <h1 className="text-xl font-semibold text-gray-900">ACL Configuration Details</h1>
```
eblairmckee: https://redpanda-ui-registry.netlify.app/docs/typography#h1

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                  <p className="text-sm text-gray-600">View the created ACL configuration</p>
```
eblairmckee: https://redpanda-ui-registry.netlify.app/docs/typography#p

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+    hostType: sharedConfig.host === '*' ? 'Allow all hosts' : 'Specific host',
+  };
+
+  // Track open state for each matching section per rule
+  const [openMatchingSections, setOpenMatchingSections] = useState<Record<number, boolean>>({});
+
+  return (
+    <div className=" ">
+      {/* Header */}
+      <div>
```
eblairmckee: both divs are useless. maybe just a fragment?

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+        <div className=" py-4">
+          <div className=" mx-auto">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center gap-4">
+                <div>
```
eblairmckee: another empty div

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+  // Track open state for each matching section per rule
+  const [openMatchingSections, setOpenMatchingSections] = useState<Record<number, boolean>>({});
+
+  return (
+    <div className=" ">
+      {/* Header */}
+      <div>
+        <div className=" py-4">
+          <div className=" mx-auto">
+            <div className="flex items-center justify-between">
```
eblairmckee: these do the same thing. get rid of line 82

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+      </div>
+
+      {/* Main Content */}
+      <main className="pt-6  pb-6">
+        <div className=" mx-auto">
+          <div className={`grid grid-cols-1 ${showMatchingUsers ? 'lg:grid-cols-3' : ''} gap-8`}>
+            {/* Left Column - Configuration Details */}
+            <div className={`${showMatchingUsers ? 'lg:col-span-2' : ''} space-y-6`}>
+              {/* Shared Configuration */}
+              <Card className="border-gray-200">
```
eblairmckee: is this border just to match mo's UI? if it looks reasonable without it, don't add the classname

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                <CardHeader className="pb-4">
+                  <CardTitle data-testid={'share-config-title'} className="text-lg font-medium text-gray-900">
```
eblairmckee: same comment as above wrt classNames being passed in

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
eblairmckee: for context, Mo is using vanilla ShadCN, whereas ours have our specific styling and variants. whenever you see `classNames` being passed one of our components, try removing and see if it looks reasonable. doesn't have to match Mo's design exactly given that's not "our" style

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                    Shared Configuration
+                  </CardTitle>
+                </CardHeader>
+                <CardContent className="space-y-4">
+                  <div className="grid grid-cols-2 gap-6">
+                    <div>
+                      <div className="text-sm font-medium text-gray-700 mb-1">Principal</div>
```
eblairmckee: any time you have a div wrapping text, that should be a typography component.

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                      <div className="text-sm text-gray-900">
+                        <span className="text-gray-600">{data.principalType}:</span>{' '}
+                        <span>{data.sharedConfig.principal.split(':')[1]}</span>
+                      </div>
```
eblairmckee: should probably be a Label

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                    </div>
+                  </div>
+                </CardContent>
+              </Card>
+
+              {/* Rules */}
+              <Card className="border-gray-200">
+                <CardHeader className="pb-4">
+                  <CardTitle data-testid={'acl-rules-length'} className="text-lg font-medium text-gray-900">
```
eblairmckee: test ids should always be strings if it's not dynamic

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                      const allDeny =
+                        enabledOperations.length > 0 &&
+                        availableRules === enabledOperations.length &&
+                        enabledOperations.every((op) => op.value === 'deny');
+                      const showSummary = allAllow || allDeny;
+
+                      return (
+                        <div
+                          key={rule.id}
+                          className="border border-gray-200 rounded-lg p-4 space-y-3"
```
eblairmckee: should this be a Card?

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                                  padding: '8px 12px',
+                                  border: '1px none rgb(229, 231, 235)',
+                                }}
+                                onClick={() =>
+                                  setOpenMatchingSections((prev) => ({
+                                    ...prev,
+                                    [rule.id]: !prev[rule.id],
+                                  }))
+                                }
+                              >
```
eblairmckee, line 198: should this be a button?

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                                    stroke="currentColor"
+                                    viewBox="0 0 24 24"
+                                  >
+                                    <path
+                                      strokeLinecap="round"
+                                      strokeLinejoin="round"
+                                      strokeWidth={2}
+                                      d="M19 9l-7 7-7-7"
+                                    />
+                                  </svg>
```
eblairmckee: actually I think this whole thing should be an Accordion.... https://redpanda-ui-registry.netlify.app/docs/accordion

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                                </div>
+                                {openMatchingSections[rule.id] && (
+                                  <div className="mt-3 border-t-0">
```
eblairmckee: `border-t-0` does nothing..

eblairmckee: this whole div needs to go

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                                    <div className="mt-2 flex flex-wrap gap-2">
+                                      <span className="inline-flex items-center px-2 py-1 text-xs rounded-full font-medium bg-blue-100 text-blue-800">
+                                        {rule.resourceType === 'topic' ? 'orders' : 'analytics-group'}
+                                      </span>
```
eblairmckee: https://redpanda-ui-registry.netlify.app/docs/badge

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                            {enabledOperations.length > 0 ? (
+                              <div className="flex flex-wrap gap-2">
+                                {showSummary ? (
+                                  <span
+                                    className={`inline-flex items-center px-2 py-1 text-xs rounded-full font-medium ${
+                                      allAllow ? 'bg-green-100 text-green-800' : 'bg-red-100 text-red-800'
+                                    }`}
+                                  >
+                                    {allAllow ? 'Allow all' : 'Deny all'}
+                                  </span>
```
eblairmckee: https://redpanda-ui-registry.netlify.app/docs/badge

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+                                  enabledOperations.map((op) => (
+                                    <span
+                                      key={op.name}
+                                      className={`inline-flex items-center px-2 py-1 text-xs rounded-full font-medium ${
+                                        op.value === 'allow' ? 'bg-green-100 text-green-800' : 'bg-red-100 text-red-800'
+                                      }`}
+                                      data-testid={`detail-item-op-${getIdFromRule(rule, op.originName, op.value)}`}
+                                    >
+                                      {op.name}: {op.value}
+                                    </span>
```
eblairmckee: https://redpanda-ui-registry.netlify.app/docs/badge

frontend/src/components/pages/acls/new-acl/AutocompleteInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+  placeholder,
+  className,
+  suggestions = [],
+  'data-testid': dataTestId,
+}: AutocompleteInputProps) {
+  const [inputValue, setInputValue] = React.useState(value);
+  const [filteredSuggestions, setFilteredSuggestions] = React.useState<string[]>([]);
+  const [showSuggestions, setShowSuggestions] = React.useState(false);
+
+  React.useEffect(() => {
```
eblairmckee: this exists somewhere but not exposed in the registry. have an open ticket for it https://redpandadata.atlassian.net/browse/UX-385

frontend/src/components/pages/acls/new-acl/CreateACL.tsx
```diff
@@ -0,0 +1,1220 @@
...
+                  </div>
+                </div>
+              </div>
+
+              {/* Operations */}
+              <div className="grid grid-cols-3 gap-3">
+                {Object.entries(rule.operations).map(([operation, operationValue]) => (
+                  <div key={operation} className="flex items-center gap-1">
+                    <Select
+                      value={operationValue}
```
eblairmckee, line 464: The value of the select should be not set, allow, or deny... and the label (use Label) above the select should be the operation value?
Screenshot 2025-08-19 at 1 26 16 PM

eblairmckee: @mo-redpanda feels weird that we're using essentially the label as the display value.

frontend/src/components/pages/roles/MatchingUsersCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+                  return (
+                    <div
+                      key={member.principal}
+                      className="group flex items-center justify-between bg-white/40 border border-gray-200 rounded-lg text-sm font-normal text-gray-900 gap-2 hover:bg-white/60 transition-colors"
+                      style={{ padding: '8px 10px' }}
+                    >
+                      <span className="flex-1 text-left">{username}</span>
+                      <Button
+                        variant="ghost"
```
eblairmckee, line 149: use https://redpanda-ui-registry.netlify.app/docs/button#icon `<Button variant="outline" size="icon">`

frontend/src/components/pages/roles/MatchingUsersCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+                  const isDeleting = deletingPrincipal === member.principal;
+
+                  return (
+                    <div
+                      key={member.principal}
+                      className="group flex items-center justify-between bg-white/40 border border-gray-200 rounded-lg text-sm font-normal text-gray-900 gap-2 hover:bg-white/60 transition-colors"
+                      style={{ padding: '8px 10px' }}
+                    >
+                      <span className="flex-1 text-left">{username}</span>
+                      <Button
```
eblairmckee, line 148: move this into the card header (top right corner)

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
andresaristizabal: I can't find the `icon` prop in `<Button/>` component, only size='icon', could you point me to that?

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +605,88 @@ const RolesTab = observer(() => {
...
-  }, []);
+interface SimpleAcl {
+  host: string;
+  principal: string;
+  principalType: string;
+  principalName: string;
+  hasAcl: boolean;
+}
+
+const AclsTab = observer((_: { principalGroups: AclPrincipalGroup[] }) => {
```
malinskibeniamin, line 606: I would strongly advise to reconsider if we want to continue adding tech debt with the `observer()` MobX pattern. I think this is not going to be easy to maintain

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +605,88 @@ const RolesTab = observer(() => {
...
+  const { data: principalGroups, isLoading } = useQuery(listACLs, {} as ListACLsRequest, {
```
malinskibeniamin: I would rather have a wrapper like `useListACLsQuery` because then we can handle all the transformations there with `select`, also handle invalidating cache for mutations etc. We should keep that separate from the component layer.

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +605,88 @@ const RolesTab = observer(() => {
...
+    select: (response) => {
+      const groupsAcl = response.resources.reduce((acc, r) => {
+        r.acls.forEach((a) => {
+          if (!acc.has(`${a.principal}:${a.host}`)) {
+            acc.set(`${a.principal}:${a.host}`, {
+              host: a.host,
+              principal: a.principal,
+              principalType: a.principal.split(':')[0] || '',
+              principalName: a.principal.split(':')[1] || '',
```
malinskibeniamin: [nit] Would prefer 
```tsx
const [principalType, principalName] = (a.principal ?? '').split(':')
```

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +605,88 @@ const RolesTab = observer(() => {
...
+            });
+          }
+        });
+        return acc;
+      }, new Map<string, SimpleAcl>());
+      return groupsAcl.values().toArray();
+    },
+  });
+
+  const { mutateAsync: deleteACLMutation } = useMutation(deleteACLs);
```
malinskibeniamin: Could be a `useDeleteACLsMutation` custom hook with `onSuccess`/`onError` to invalidate `queryClient` queries.

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +605,88 @@ const RolesTab = observer(() => {
...
   const [aclFailed, setAclFailed] = useState<{ err: unknown } | null>(null);
   const [editorType, setEditorType] = useState<'create' | 'edit'>('create');
   const [edittingPrincipalGroup, setEdittingPrincipalGroup] = useState<AclPrincipalGroup | null>(null);
 
-  let groups = p.principalGroups.filter((g) => g.principalType === 'User');
+  const nav = useNavigate();
+
+  const deleteACLsForPrincipal = async (principal: string, host: string) => {
+    const deleteRequest: DeleteACLsRequest = {
+      filter: {
```
malinskibeniamin, line 618: Let's use the `create()` function from `@bufbuild/protobuf`.

frontend/tests/console/pages/RolePage.ts
```diff
@@ -0,0 +1,132 @@
...
+      // Click the confirm button to add the user
+      const confirmButton = this.page.getByTestId('confirm-add-user-button');
+      await confirmButton.click();
+
+      // Wait for the add operation to complete and UI to update
+      await this.page.waitForTimeout(500);
+
+      // Verify the user was added by checking if it appears in the list
+      await this.validateMemberExists(username);
+    }
```
malinskibeniamin, line 80: I think all the function names and the comments should become `test.step()` so that in CI when the test fails, you can see exactly what the last recorded log was from Playwright.

malinskibeniamin: https://playwright.dev/docs/api/class-test#test-step

frontend/src/state/supportedFeatures.ts
```diff
@@ -134,6 +138,9 @@ class SupportedFeatures {
   @computed get rpcnSecretsApi(): boolean {
     return isSupported(Feature.SecretService);
   }
+  @computed get schemaRegistryACLService(): boolean {
```
malinskibeniamin: I think the convention established so far is to do
```suggestion
  @computed get schemaRegistryACLApi(): boolean {
```

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+        description: 'Role ACLs updated successfully',
+      });
+
+      // Invalidate the listACLs query cache to force fresh data on next request
+      await queryClient.invalidateQueries({
+        queryKey: createConnectQueryKey({
+          schema: ACLService,
+          cardinality: 'finite',
+        }),
+      });
```
malinskibeniamin: This should be part of a custom hook.

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+        sharedConfig.host,
+      ).map((r) => ({
+        ...r,
+        id: getIdFromCreateACLRequest(r),
+      }));
+
+      const { toCreate, toDelete } = calculateACLDifference(currentRules, newRules);
+
+      console.log('Rules to create:', toCreate);
+      console.log('Rules to delete:', toDelete);
```
malinskibeniamin: Remove `console.log()`

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+      const createResults = toCreate.map((r) => createACLMutation(r));
+      const deleteResults = toDelete.map((r) => {
+        const d: DeleteACLsRequest = {
```
malinskibeniamin: You can create a new delete ACLs request schema

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+          filter: {
+            principal: r.principal,
+            resourceType: r.resourceType,
+            resourceName: r.resourceName,
+            host: r.host,
+            operation: r.operation,
+            permissionType: r.permissionType,
+            resourcePatternType: r.resourcePatternType,
+            $typeName: 'redpanda.api.dataplane.v1.DeleteACLsRequest.Filter',
```
malinskibeniamin: You shouldn't need to do `$typeName` manually, just use `create()`

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+        id: getIdFromCreateACLRequest(r),
+      }));
+      const newRules: ACLWithId[] = convertRulesToCreateACLRequests(
+        rules,
+        sharedConfig.principal,
+        sharedConfig.host,
+      ).map((r) => ({
+        ...r,
+        id: getIdFromCreateACLRequest(r),
+      }));
```
malinskibeniamin: Move to custom hook located in `frontend/src/react-query/api/acl.tsx`

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const { roleName = '' } = useParams<{ roleName: string }>();
+
+  useEffect(() => {
+    uiState.pageBreadcrumbs = [
+      { title: 'Security', linkTo: '/security' },
+      { title: 'Roles', linkTo: '/security/roles' },
+      { title: roleName, linkTo: `/security/roles/${roleName}/details` },
+      { title: 'Update Role', linkTo: '', heading: '' },
+    ];
+  }, [roleName]);
```
malinskibeniamin, line 44: This should be moved to `runInAction()` MobX util method, which should be ran in a `useEffect()` you have now, but you could do something like
```
// Hack for MobX to ensure we don't need to use observables
export const updatePageTitle = () => {
  runInAction(() => {
    uiState.pageTitle = 'Remote MCP';
    uiState.pageBreadcrumbs.pop(); // Remove last breadcrumb to ensure the title is used without previous page breadcrumb being shown
    uiState.pageBreadcrumbs.push({ title: 'Remote MCP', linkTo: '/remote-mcp', heading: 'Remote MCP' });
  });
};
```

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const toCreate = newRules.filter((rule) => !currentIds.has(rule.id));
+
+  // Rules to delete: in currentRules but not in newRules
+  const toDelete = currentRules.filter((rule) => !newIds.has(rule.id));
+
+  return {
+    toCreate,
+    toDelete,
+  };
+}
```
malinskibeniamin: I would move it to some other file if possible, sounds like it deserves its own tests too.

frontend/src/components/pages/roles/RoleUpdatePage.tsx
```diff
@@ -0,0 +1,214 @@
...
+  type DeleteACLsRequest,
+  type ListACLsRequest,
+} from 'protogen/redpanda/api/dataplane/v1/acl_pb';
+import { createACL, deleteACLs, listACLs } from 'protogen/redpanda/api/dataplane/v1/acl-ACLService_connectquery';
+import { uiState } from '../../../state/uiState';
+import PageContent from '../../misc/PageContent';
+
+interface ACLWithId extends CreateACLRequest {
+  id: string;
+}
```
malinskibeniamin: Why doesn't ACL have an ID by default? Sounds like a gap in the proto layer.

frontend/src/components/pages/roles/RoleCreatePage.tsx
```diff
@@ -0,0 +1,124 @@
...
+      const result = convertRulesToCreateACLRequests(rules, principal, host);
+
+      const results = await Promise.all(
+        result.map((r) => {
+          return createACLMutation(r);
+        }),
+      );
+
+      // TODO: handle partial failures
+      console.log(results);
```
malinskibeniamin: Remove `console.log()`

frontend/src/components/pages/roles/RoleCreatePage.tsx
```diff
@@ -0,0 +1,124 @@
...
+import CreateACL from 'components/pages/acls/new-acl/CreateACL';
+import { ACLService } from 'protogen/redpanda/api/dataplane/v1/acl_pb';
+import { createACL } from 'protogen/redpanda/api/dataplane/v1/acl-ACLService_connectquery';
+import { type CreateRoleRequest, SecurityService } from 'protogen/redpanda/api/dataplane/v1/security_pb';
+import { createRole } from 'protogen/redpanda/api/dataplane/v1/security-SecurityService_connectquery';
+import { uiState } from '../../../state/uiState';
+import PageContent from '../../misc/PageContent';
+
+const RoleCreatePage = () => {
+  const toast = useToast();
```
malinskibeniamin, line 30: We should use a new Sonner component with `<Toaster />` under providers. It would be more aligned with our new design system. I will make that change for the Remote MCP work and you can rebase, I don't think there should be any conflicts.

frontend/src/components/pages/roles/MatchingUsersCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+    setNewUserName('');
+    setIsAddingUser(false);
+  };
+
+  // Get members count
+  const membersCount = principalType === 'RedpandaRole' ? membersData?.members?.length || 0 : 3;
+
+  return (
+    <div className="lg:col-span-1">
+      <Card size="full" className="bg-slate-100">
```
malinskibeniamin, line 121: For now if you do `className="max-w-full"` then the issue with the Card component not spanning across the entire component will be resolved.

frontend/src/components/pages/roles/MatchingUsersCard.tsx
```diff
@@ -0,0 +1,267 @@
...
+    onSuccess: async (_, variables) => {
+      // Invalidate and refetch the role members query
+      await queryClient.invalidateQueries({
+        queryKey: createConnectQueryKey({
+          schema: SecurityService.method.listRoleMembers,
+          cardinality: 'finite',
+          input: { roleName },
+        }),
+      });
+      await refetch();
```
malinskibeniamin: Refetching is not good - we should invalidate the query and then it can be refetched in due course when the time is right. Having the refetch alongside the invalidate kind of defeats the purpose of having cache stale time. Ideally you just mark the query as invalidated, causing it to load automatically at another time.

frontend/src/components/pages/acls/new-acl/AutocompleteInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <div className="relative flex-1">
+      <Input
+        testId={dataTestId}
+        type="text"
+        placeholder={placeholder}
+        value={inputValue}
+        onChange={handleInputChange}
+        onBlur={() => setTimeout(() => setShowSuggestions(false), 200)}
```
malinskibeniamin, line 58: I think `setTimeout` is redundant here. Are we trying to debounce the input? We can use a `useDebounce` hook.

```tsx

/**
 */

    (...args: Parameters<T>) => {
      if (timeoutRef.current) {
      }

        callback(...args);
      }, delay);
    },
    [callback, delay],
  );

};

```

frontend/src/components/pages/acls/new-acl/AutocompleteInput.tsx
malinskibeniamin, line 1: Sounds like something that should exist in the UI registry. Created a ticket [here](https://redpandadata.atlassian.net/browse/UX-425)

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
andresaristizabal: After updating the library, I don't see this new property. Let's move to forward PR.

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: I fully agree, but we need to make this process more progressive.

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: make sense, I'll move it to a file as hooks

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: good one!

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: of course!

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: Nice, thank you!

frontend/tests/console/pages/RolePage.ts
andresaristizabal: Good idea! let claude do this 😄

frontend/src/state/supportedFeatures.ts
andresaristizabal: done

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: I'll move all update operation in that case

frontend/src/components/pages/acls/new-acl/AclCreatePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+const AclCreatePage = () => {
+  const toast = useToast();
+  const nav = useNavigate();
+  const queryClient = useQueryClient();
+
+  useEffect(() => {
+    uiState.pageBreadcrumbs = [
+      { title: 'Security', linkTo: '/security' },
+      { title: 'ACLs', linkTo: '/security/acls' },
+      { title: 'Create new ACL', linkTo: '' },
```
micheleRP:
```suggestion
      { title: 'Create ACL', linkTo: '' },
```

frontend/src/components/pages/roles/RoleCreatePage.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const nav = useNavigate();
+  const queryClient = useQueryClient();
+
+  useEffect(() => {
+    uiState.pageBreadcrumbs = [
+      { title: 'Security', linkTo: '/security' },
+      { title: 'Roles', linkTo: '/security/roles' },
+      { title: 'Create new Role', linkTo: '' },
```
micheleRP:
```suggestion
      { title: 'Create role', linkTo: '' },
```

frontend/src/components/pages/acls/UserDetails.tsx
```diff
@@ -82,273 +85,149 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
...
       <PageContent>
+        <div className="flex justify-between">
+          <div>
+            <h3>Permissions</h3>
+            <p className="text-sm text-gray-600">The following permissions are assigned to this principal.</p>
+          </div>
+          <div className="flex gap-3">
+            {Features.rolesApi && (
+              <Button variant="outline" onClick={() => (this.isChangeRolesModalOpen = true)} disabled={!canEdit}>
+                Change roles
```
micheleRP: Should this be Assign roles, instead of Change roles?

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
malinskibeniamin: We don't need to worry about styling any of our text-related HTML tags if we include `prose.css` - let's use the new `theme.css` which should include it. Then we will be able to do `<h1>` and get most of the styling done out of the box.

frontend/src/components/pages/acls/Acl.List.tsx
malinskibeniamin: Ok, maybe we can only make that specific part wrapped with an `observer()` MobX pattern

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: done

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: done

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: done

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: Yes sr, I'll move it to own hook

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: let's avoid to use `Mobx` we need to remove it, therfore not use on new components

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: Yes, I moved it to acl mode file, where we have the logic for acl

frontend/src/components/pages/roles/RoleUpdatePage.tsx
andresaristizabal: ACL doesn't have ids 😄, this our own id which is a set of all attributes

frontend/src/components/pages/roles/MatchingUsersCard.tsx
andresaristizabal: `full` works very well for me, is there a reason for change it?

frontend/src/components/pages/roles/MatchingUsersCard.tsx
andresaristizabal: make sense, I updated this logic to use the existing hooks

frontend/src/components/pages/acls/new-acl/AutocompleteInput.tsx
andresaristizabal: what do you think use this? https://redpanda-ui-registry.netlify.app/docs/tags

frontend/src/components/pages/acls/new-acl/AclCreatePage.tsx
andresaristizabal: done

frontend/src/components/pages/acls/UserDetails.tsx
andresaristizabal: Is there another verb? You can add or remove roles, i'll use that suggestion for now

frontend/src/components/pages/roles/RoleCreatePage.tsx
andresaristizabal: done

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    Shared configuration
+                  </CardTitle>
+                </CardHeader>
+                <CardContent className="space-y-4">
+                  <div className="grid grid-cols-2 gap-6">
+                    <div>
+                      <div className="text-sm font-medium text-gray-700 mb-1">Principal</div>
+                      <div className="text-sm text-gray-900">
+                        <span className="text-gray-600">{data.principalType}:</span>{' '}
+                        <span>{data.sharedConfig.principal.split(':')[1]}</span>
```
jvorcak: These `split (':')[1]` occur on quite a lot of places. Would it be an idea to create a pure function, add some good naming and documentation so that the code is more readable?

Sth like this (role, name values are arbitrary).

`const {role, name} = parsePrincipal(data.sharedConfig.principal)`

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +603,49 @@ const RolesTab = observer(() => {
...
+const AclsTab = observer((_: { principalGroups: AclPrincipalGroup[] }) => {
+  const { data: principalGroups, isLoading } = useListACLAsPrincipalGroups();
+  const { mutateAsync: deleteACLMutation } = useDeleteAclMutation();
 
   const [aclFailed, setAclFailed] = useState<{ err: unknown } | null>(null);
   const [editorType, setEditorType] = useState<'create' | 'edit'>('create');
   const [edittingPrincipalGroup, setEdittingPrincipalGroup] = useState<AclPrincipalGroup | null>(null);
 
-  let groups = p.principalGroups.filter((g) => g.principalType === 'User');
+  const nav = useNavigate();
```
jvorcak: nit; should we try to be consistent and use `const navigate = useNavigate()` like elsewhere and also in react-router docs?

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +603,49 @@ const RolesTab = observer(() => {
...
   try {
     const quickSearchRegExp = new RegExp(uiSettings.aclList.configTable.quickSearch, 'i');
-    groups = groups.filter((aclGroup) => aclGroup.principalName.match(quickSearchRegExp));
+    groups = groups?.filter((aclGroup) => aclGroup.principalName.match(quickSearchRegExp));
   } catch (_e) {
     console.warn('Invalid expression');
   }
 
-  if (api.ACLs?.aclResources === undefined) return DefaultSkeleton;
+  if (isLoading || !principalGroups) return DefaultSkeleton;
```
jvorcak, line 648: if there are no principalGroups , woudn't we show the loader forever? Wouldn't taking an error state from the hook solve the problem?

frontend/src/components/pages/acls/new-acl/AutocompleteInput.tsx
malinskibeniamin: Yes, but I think autocomplete input sounds like you are just inserting free text and getting suggestion based on that. I am not sure if tags are what you are looking for? They are essentially multi select

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
andresaristizabal: good idea

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: make sense, done

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
andresaristizabal: done

frontend/src/components/pages/acls/Acl.List.tsx
andresaristizabal: in this case, it shows the empty list

frontend/src/components/pages/roles/MatchingUsersCard.tsx
malinskibeniamin: Oh ok - all good then

frontend/src/components/pages/acls/new-acl/AutocompleteInput.tsx
andresaristizabal: make sense

frontend/src/components/pages/acls/new-acl/AclDetailPage.tsx
eblairmckee: we'll still need to add the `prose` className to this component, though. but yeah, good callout with the prose styles.

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +603,49 @@ const RolesTab = observer(() => {
...
+        resourceType: ACL_ResourceType.ANY,
+        resourceName: undefined,
+        host: host,
+        operation: ACL_Operation.ANY,
+        permissionType: ACL_PermissionType.ANY,
+        resourcePatternType: ACL_ResourcePatternType.ANY,
+      },
+    });
+    await deleteACLMutation(deleteRequest);
+    toast({
```
malinskibeniamin, line 629: We can use a showToast method which ensures toast behavior is consistent across all places within Console.

frontend/src/components/pages/acls/UserDetails.tsx
```diff
@@ -9,10 +9,15 @@
  * by the Apache License, Version 2.0
  */
 
-import { Box, Button, Link as ChakraLink, DataTable, Flex, Heading, Text } from '@redpanda-data/ui';
+import { useQuery } from '@connectrpc/connect-query';
+import { Box, DataTable, Text } from '@redpanda-data/ui';
```
malinskibeniamin, line 13: I think in subsequent Pars we can remove dependency on the library in favor of the new design system. Ideally we shouldn't mix them too much

frontend/src/components/pages/acls/UserDetails.tsx
```diff
@@ -82,123 +87,154 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
...
-            </Heading>
-          )}
-          <AclPrincipalGroupPermissionsTable key={r.principalName} group={r} />
-        </>
-      ))}
-    </>
+    <div className="gap-3 flex flex-col">
+      {roles.map((g) => {
+        return (
+          <EmbeddedAclDetail
```
malinskibeniamin, line 220: Does embedded mean nested in this context?

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
```diff
@@ -0,0 +1,273 @@
+/**
+ * Copyright 2022 Redpanda Data, Inc.
```
malinskibeniamin, line 2: 2025

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
```diff
@@ -0,0 +1,273 @@
...
+                          <div className="mb-3">
+                            <div className="font-medium text-gray-900">
+                              {(() => {
+                                const text =
+                                  rule.resourceType === 'cluster' || rule.resourceType === 'schemaRegistry'
+                                    ? getResourceName(rule.resourceType)
+                                    : rule.selectorType === 'any'
+                                      ? `All ${getPluralResourceName(rule.resourceType)}`
+                                      : `${getPluralResourceName(rule.resourceType)} ${rule.selectorType === 'literal' ? 'matching' : 'starting with'}: "${rule.selectorValue}"`;
+                                return text.charAt(0).toUpperCase() + text.slice(1);
```
malinskibeniamin, line 200: I think we have capitalize() function

frontend/src/components/pages/acls/new-acl/AclUpdatePage.tsx
```diff
@@ -0,0 +1,112 @@
...
+  // Fetch existing ACL data
+  const { data, isLoading } = useGetAclsByPrincipal(`User:${aclName}`);
+
+  const { applyUpdates } = useUpdateAclMutation();
+
+  const updateAclMutation =
+    (actualRules: Rule[], sharedConfig: SharedConfig) => async (_: string, _2: string, rules: Rule[]) => {
+      await applyUpdates(actualRules, sharedConfig, rules);
+
+      // TODO: handle partial failures
```
malinskibeniamin, line 54: Do you mean we are executing multiple mutations, and some might fail?

frontend/tests/console/pages/RolePage.ts
```diff
@@ -0,0 +1,132 @@
...
+  async deleteMembership(usernames: string[]) {
+    for (const username of usernames) {
+      // Find the member element and its delete button
+      const deleteButton = this.page.getByTestId(`remove-user-${username}-button`);
+
+      // Click the delete button
+      await deleteButton.click();
+
+      // Wait for deletion to complete
+      await this.page.waitForTimeout(500);
```
malinskibeniamin, line 116: Ideally we should refrain from the timeout pattern, because it might pass locally, but take longer for CI. You can group an action like clicking an element with an expected element change.

frontend/tests/console/pages/RolePage.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * Validate that a specific member exists in the role
+   * @param username The username to check for
+   */
+  async validateMemberExists(username: string) {
+    // The member should appear in the matching users card
+    const memberElement = this.page.locator('.text-left').filter({ hasText: username });
+    await memberElement.waitFor({ state: 'visible', timeout: 5000 });
```
malinskibeniamin, line 90: Most of the timeouts should be revisited at another PR, we should get the test passing without them to ensure no flakiness.

frontend/src/react-query/api/acl.tsx
```diff
@@ -374,3 +385,149 @@ export const useCreateACLMutation = () => {
...
+    invalid,
+  };
+};
+
+export const useDeleteAclMutation = (
+  transportOptions?: UseMutationOptions<typeof DeleteACLsRequestSchema, typeof DeleteACLsResponseSchema>,
+) => {
+  const { invalid } = useInvalidateAclsList();
+  return useMutation(deleteACLs, {
+    onSettled: async (_, error) => {
```
malinskibeniamin, line 451: Why not just onError?

frontend/src/react-query/api/acl.tsx
```diff
@@ -374,3 +385,149 @@ export const useCreateACLMutation = () => {
...
+// New ACL implementation
+
+interface ACLWithId extends CreateACLRequest {
+  id: string;
+}
+
+const useInvalidateAclsList = () => {
+  const queryClient = useQueryClient();
+
+  const invalid = async () => {
```
malinskibeniamin, line 432: Maybe let's call it invalidate(), invalid() sounds like we are verifying something.

frontend/src/components/routes.tsx
```diff
@@ -426,12 +429,16 @@ export const APP_ROUTES: IRouteEntry[] = [
...
+  MakeRoute<{}>('/security/acls/create', AclCreatePage, 'Create ACL'),
+  MakeRoute<{}>('/security/acls/:aclName/update', AclUpdatePage, 'Update ACL'),
+  MakeRoute<{}>('/security/acls/:aclName/details', AclDetailPage, 'ACL details'),
+
   MakeRoute<{}>('/security/users/create', UserCreatePage, 'Security'),
   MakeRoute<{ userName: string }>('/security/users/:userName/details', UserDetailsPage, 'Security'),
 
   MakeRoute<{}>('/security/roles/create', RoleCreatePage, 'Security'),
-  MakeRoute<{ roleName: string }>('/security/roles/:roleName/details', RoleDetailsPage, 'Security'),
-  MakeRoute<{ roleName: string }>('/security/roles/:roleName/edit', RoleEditPage, 'Security'),
```
malinskibeniamin, line 434: Do we need to add a redirect from /edit to /update for backwards compatibility?

jvorcak (approved):

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
```diff
@@ -0,0 +1,273 @@
...
+        <div className=" py-4">
+          <div className=" mx-auto">
+            <div className="flex items-center justify-between">
+              <div className="flex items-center gap-4">
+                <div>
+                  <p className="text-sm text-gray-600">View the created ACL configuration</p>
+                </div>
+              </div>
+              <Button
+                className="bg-gray-900 hover:bg-gray-800 text-white"
```
malinskibeniamin, line 104: We should use variants for the button. That way it will be consistent.

frontend/src/components/pages/acls/new-acl/ACLDetails.tsx
```diff
@@ -0,0 +1,273 @@
...
+              <Card size="full" className={`${isSimpleView ? 'hidden' : ''}`}>
+                <CardHeader className="pb-4">
+                  <CardTitle data-testid={'share-config-title'} className="text-lg font-medium text-gray-900">
+                    Shared configuration
+                  </CardTitle>
+                </CardHeader>
+                <CardContent className="space-y-4">
+                  <div className="grid grid-cols-2 gap-6">
+                    <div>
+                      <div className="text-sm font-medium text-gray-700 mb-1">Principal</div>
```
malinskibeniamin, line 132: typography prose - <h3> automatically apply the proper styling for HTML tags.

frontend/src/components/pages/acls/Acl.List.tsx
```diff
@@ -593,24 +603,49 @@ const RolesTab = observer(() => {
...
+      status: 'success',
+      description: (
+        <Text as="span">
+          Deleted ACLs for <CodeEl>{principal}</CodeEl>
```
malinskibeniamin, line 633: With the new design, you should be able to just do `<code>` because typography prose css file should automatically handle how all HTML tags related to displaying some text content are already styled accordingly.

PR #1878, opened 2025-08-20
Feature/sr acls: Backend Work

r-vasquez:
## PR Description: Backend Support for Schema Registry ACLs

This PR introduces the backend support for **Schema Registry ACLs** in Console in the current ACLService of the Dataplane API.

### Key Changes  
- **Migration to `common-go/rpsr`**  
  - Replaces usage of `franz-go/sr` with the new `rpsr` wrapper.  
  - Maintains full compatibility with existing methods while adding new functionality for Schema Registry ACLs.  

- **Schema Registry ACL Management**  
  - Added support for **Create**, **List** and **Delete** operations on Schema Registry ACLs.  
  - Introduced new error codes and parsing logic using the latest `franz-go/sr` error definitions.  
  - Improved request dispatching for ACL operations across different services.  

- **Schema Registry ACL Probe**  
  - Implemented `CheckSchemaRegistryACLSupport` to determine cluster compatibility.  
  - Uses a validation-first ACL Create request to distinguish between:  
    - **404** → older clusters or clusters without SR ACL support.  
    - **403** → unsupported license.

- **Protocol & Validation Updates**  
  - Allowed `CreateACL` to accept an empty resource name for `Registry`, mirroring `Cluster` behavior.  
  - Relaxed principal validation to support Redpanda roles and custom prefixes.  
  - Documented validation rules for `CreateACL`.

- **Testing & Reliability Improvements**  
  - Fixed stranded resource cleanup for ACLs (ensuring `Hosts` are included).  
  - Added integration tests and cleaned up older tests.


proto/redpanda/api/dataplane/v1/acl.proto
```diff
@@ -150,8 +155,7 @@ message CreateACLRequest {
   // authorizer, you must include the prefix "User:" with the user name.
   string principal = 4 [
     (google.api.field_behavior) = REQUIRED,
-    (buf.validate.field).required = true,
-    (buf.validate.field).string.prefix = "User:"
```
weeco, line 154: NIT: We couuld require the string to contain a colon. I liked that we would indicate that a principal type prefix is necessary.

andresaristizabal: something like `(buf.validate.field).string.pattern = ".*:.*"` ?

r-vasquez: I added this https://github.com/redpanda-data/console/pull/1878/commits/c76223603a8dbb35faf08272e71b02c13aae4ad0

Which validates and displays a better error message than just `invalid field`, I noticed that multiple colons are allowed, so, technically speaking: `User:foo:bar` is allowed.

PR #1879, opened 2025-08-21
frontend: upgrade to biome 2.2.0 and fix errors

malinskibeniamin:
We want to match all the UIs to use the same biome linter setup as much as possible, so that the context switching is easier and developer experience is improved.


weeco (approved):

jvorcak (approved):

malinskibeniamin:
Closing in favor of other setups.

PR #1881, opened 2025-08-21
frontend: migrate sidebar for console in self-hosted mode

malinskibeniamin:
Fix for [AI-111](https://redpandadata.atlassian.net/browse/AI-111)
Screenshot 2025-08-22 at 00 55 44
Screenshot 2025-08-22 at 00 56 01
Screenshot 2025-08-22 at 00 56 09
Screenshot 2025-08-22 at 00 56 21
Screenshot 2025-08-22 at 00 56 35

https://github.com/user-attachments/assets/1a25fe3f-0fd9-4d2d-9835-b384e7214ef3

[AI-111]: https://redpandadata.atlassian.net/browse/AI-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/components/console-sidebar.tsx
+++ b/frontend/src/components/console-sidebar.tsx
@@ -1,11 +1,10 @@
 'use client';
 
-import { observer } from 'mobx-react';
-import { Link, Link as RouterLink, useLocation } from 'react-router-dom';
-import { ChevronsLeft, ChevronsRight } from 'lucide-react';
-
 import { useColorMode } from '@redpanda-data/ui';
 import { useBooleanFlagValue } from 'custom-feature-flag-provider';
+import { ChevronsLeft, ChevronsRight } from 'lucide-react';
+import { observer } from 'mobx-react';
+import { Link, Link as RouterLink, useLocation } from 'react-router-dom';
 
 import redpandaIconColor from '../assets/logos/redpanda-icon-color.svg';
 import redpandaIconWhite from '../assets/logos/redpanda-icon-white.svg';
@@ -38,19 +37,19 @@ const ConsoleSidebar = observer(() => {
   const APP_ROUTES_WITHOUT_AI_AGENTS = APP_ROUTES.filter((route) => !route.path.startsWith('/agents'));
   const FINAL_APP_ROUTES = isAiAgentsEnabled ? APP_ROUTES : APP_ROUTES_WITHOUT_AI_AGENTS;
   const sidebarItems = createVisibleSidebarItems(FINAL_APP_ROUTES);
-  
+
   // Helper function to determine if a sidebar item is active
   const isItemActive = (itemPath: string) => {
     // Exact match for overview
     if (itemPath === '/overview') {
       return location.pathname === '/overview' || location.pathname === '/';
     }
-    
+
     // For other routes, check if current path starts with the item path
     // This handles both exact matches and sub-routes
     return location.pathname.startsWith(itemPath);
   };
-  
+
   console.log(sidebarItems);
 
   return (
@@ -60,7 +59,7 @@ const ConsoleSidebar = observer(() => {
         <SidebarContent>
           {/* Nav Main */}
           <SidebarGroup>
-          <SidebarMenu>
+            <SidebarMenu>
               {sidebarItems.map((item) => (
                 <Collapsible
                   key={item.title as string}
@@ -73,11 +72,7 @@ const ConsoleSidebar = observer(() => {
                       asChild={!item.isDisabled}
                       disabled={item.isDisabled}
                       isActive={!item.isDisabled && isItemActive(item.to)}
-                      tooltip={
-                        item.isDisabled && item.disabledText
-                          ? item.disabledText
-                          : (item.title as string)
-                      }
+                      tooltip={item.isDisabled && item.disabledText ? item.disabledText : (item.title as string)}
                     >
                       {item.isDisabled ? (
                         <div className="flex items-center gap-2">
@@ -115,51 +110,50 @@ const ConsoleSidebarHeader = () => {
   const { colorMode } = useColorMode(); // Chakra UI color mode
   const { state } = useSidebar();
 
-return (
-  <SidebarHeader>
-  <div className={`flex mt-[0.875rem] items-center ${state === 'expanded' ? 'px-2' : 'justify-center'}`}>
-    <RouterLink to="/" className="flex items-center">
-      {colorMode === 'light' && (
-        <img
-          src={state === 'collapsed' ? redpandaIconColor : redpandaTextColor}
-          alt="Redpanda"
-          className={state === 'collapsed' ? 'h-6 w-6' : 'h-6'}
-        />
-      )}
-      {colorMode === 'dark' && (
-        <img
-          src={state === 'collapsed' ? redpandaIconWhite : redpandaTextWhite}
-          alt="Redpanda"
-          className={state === 'collapsed' ? 'h-6 w-6' : 'h-6'}
-        />
-      )}
-    </RouterLink>
-  </div>
-</SidebarHeader>
-)
-}
+  return (
+    <SidebarHeader>
+      <div className={`flex mt-[0.875rem] items-center ${state === 'expanded' ? 'px-2' : 'justify-center'}`}>
+        <RouterLink to="/" className="flex items-center">
+          {colorMode === 'light' && (
+            <img
+              src={state === 'collapsed' ? redpandaIconColor : redpandaTextColor}
+              alt="Redpanda"
+              className={state === 'collapsed' ? 'h-6 w-6' : 'h-6'}
+            />
+          )}
+          {colorMode === 'dark' && (
+            <img
+              src={state === 'collapsed' ? redpandaIconWhite : redpandaTextWhite}
+              alt="Redpanda"
+              className={state === 'collapsed' ? 'h-6 w-6' : 'h-6'}
+            />
+          )}
+        </RouterLink>
+      </div>
+    </SidebarHeader>
+  );
+};
 
 const ConsoleSidebarFooter = () => {
   const { state, toggleSidebar } = useSidebar();
 
   return (
     <SidebarFooter>
-    <SidebarMenu>
-    <UserProfile />
-      {/* Collapse/Expand Sidebar */}
-      <SidebarMenuItem>
-        <SidebarMenuButton
-          tooltip={state === 'collapsed' ? 'Expand sidebar' : 'Collapse sidebar'}
-          onClick={toggleSidebar}
-        >
-          {state === 'collapsed' ? <ChevronsRight className="size-4" /> : <ChevronsLeft className="size-4" />}
-          <span>{state === 'collapsed' ? 'Expand sidebar' : 'Collapse sidebar'}</span>
-        </SidebarMenuButton>
-      </SidebarMenuItem>
-    </SidebarMenu>
-  </SidebarFooter>
-  )
-}
-
+      <SidebarMenu>
+        <UserProfile />
+        {/* Collapse/Expand Sidebar */}
+        <SidebarMenuItem>
+          <SidebarMenuButton
+            tooltip={state === 'collapsed' ? 'Expand sidebar' : 'Collapse sidebar'}
+            onClick={toggleSidebar}
+          >
+            {state === 'collapsed' ? <ChevronsRight className="size-4" /> : <ChevronsLeft className="size-4" />}
+            <span>{state === 'collapsed' ? 'Expand sidebar' : 'Collapse sidebar'}</span>
+          </SidebarMenuButton>
+        </SidebarMenuItem>
+      </SidebarMenu>
+    </SidebarFooter>
+  );
+};
 
 export default ConsoleSidebar;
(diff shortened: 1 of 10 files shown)


frontend/src/components/console-sidebar.tsx
```diff
@@ -0,0 +1,159 @@
...
+    if (itemPath === '/overview') {
+      return location.pathname === '/overview' || location.pathname === '/';
+    }
+
+    // For other routes, check if current path starts with the item path
+    // This handles both exact matches and sub-routes
+    return location.pathname.startsWith(itemPath);
+  };
+
+  console.log(sidebarItems);
```
eblairmckee: ope

frontend/src/components/console-sidebar.tsx
```diff
@@ -0,0 +1,159 @@
...
+              ))}
+            </SidebarMenu>
+          </SidebarGroup>
+          {/* Nav Main */}
+        </SidebarContent>
+        <ConsoleSidebarFooter />
+        <SidebarRail />
+      </Sidebar>
+
+      <SidebarInset>
```
eblairmckee, line 111: I was wondering about this composition... `inset` isn't well documented, what does that mean as opposed toe `sidebar` and `floating`?

frontend/src/components/console-sidebar.tsx
```diff
@@ -0,0 +1,159 @@
...
+  );
+});
+
+const ConsoleSidebarHeader = () => {
+  const { colorMode } = useColorMode(); // Chakra UI color mode
+  const { state } = useSidebar();
+
+  return (
+    <SidebarHeader>
+      <div className={`flex mt-[0.875rem] items-center ${state === 'expanded' ? 'px-2' : 'justify-center'}`}>
```
eblairmckee, line 126: where did you get this value from? `mt-[0.875rem]`?

frontend/src/components/console-sidebar.tsx
```diff
@@ -0,0 +1,159 @@
...
+              className={state === 'collapsed' ? 'h-6 w-6' : 'h-6'}
+            />
+          )}
+          {colorMode === 'dark' && (
+            <img
+              src={state === 'collapsed' ? redpandaIconWhite : redpandaTextWhite}
+              alt="Redpanda"
+              className={state === 'collapsed' ? 'h-6 w-6' : 'h-6'}
+            />
+          )}
```
eblairmckee: side note: this makes me feel like we should have a ui-registry component for Image that takes a `darkModeSrc` and `lightModeSrc`

frontend/src/components/layout/Header.tsx
```diff
@@ -9,16 +9,28 @@
  * by the Apache License, Version 2.0
  */
 
-import { Badge, Box, Breadcrumbs, Button, ColorModeSwitch, CopyButton, Flex, Text } from '@redpanda-data/ui';
+import { Badge, Box, Button, ColorModeSwitch, CopyButton, Flex, Text } from '@redpanda-data/ui';
```
eblairmckee, line 12: 🔥 love seeing how easy these migrations can be just by changing an import!!!!!!! 🔥 🔥 🔥

frontend/src/components/layout/Header.tsx
```diff
@@ -9,16 +9,28 @@
...
+import { Home } from 'lucide-react';
 import { computed } from 'mobx';
 import { observer } from 'mobx-react';
+import React from 'react';
```
eblairmckee, line 16: nit: is this a lint rule to require importing react? I did notice some components would type error when imported into consumer repos because they were missing this. for some reason I thought react 18 didn't require it anymore?

frontend/src/components/layout/Header.tsx
```diff
@@ -9,16 +9,28 @@
  * by the Apache License, Version 2.0
  */
 
-import { Badge, Box, Breadcrumbs, Button, ColorModeSwitch, CopyButton, Flex, Text } from '@redpanda-data/ui';
+import { Badge, Box, Button, ColorModeSwitch, CopyButton, Flex, Text } from '@redpanda-data/ui';
+import { Home } from 'lucide-react';
```
eblairmckee, line 13: any chance we could replace these imports with ui-registry components?

frontend/src/components/layout/Header.tsx
```diff
@@ -9,16 +9,28 @@
...
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  BreadcrumbLink,
+  BreadcrumbList,
+  BreadcrumbPage,
+  BreadcrumbSeparator,
+} from '../redpanda-ui/components/breadcrumb';
+import { Separator } from '../redpanda-ui/components/separator';
+import { SidebarTrigger } from '../redpanda-ui/components/sidebar';
```
eblairmckee, line 33: at first glance I wondered... given all the separate imports I'm wondering if we should make a barrel file for the new ui-registry components so we can easily swap out `@redpanda-data/ui` with our new local registry instances. but that would require extending the fumadocs cli though 😬 to generate a barrel file.. given we're trying to eliminate maintenance burden.. i retract the thought.

frontend/src/components/layout/Header.tsx
```diff
@@ -42,19 +54,39 @@ const AppPageHeader = observer(() => {
...
-            showHomeIcon={false}
-            items={breadcrumbItems.map((x) => ({
-              name: x.title,
-              heading: x.heading,
-              to: x.linkTo,
-            }))}
-          />
-        )}
-      </Flex>
+      <header className="flex h-16 shrink-0 items-center gap-2 transition-[width,height] ease-linear group-has-[[data-collapsible=icon]]/sidebar-wrapper:h-12">
```
eblairmckee: this makes me think we should have a BreadcumbHeader component..... would be nice to have something we've added our opinionated styles to that people can easily import. and given we're leaning into MCPs would hope claude would know to use this...

eblairmckee: then again, given we're whitegloving all these sidebar implementations in consumer repos, 🤷‍♀️ , may not be worth it. would only matter if we're anticipating new redpanda apps beings built

eblairmckee (approved):
most of my comments are more philosophical wrt the future of the ui-registry (with exception to the console log) but only other comment I'd make is the new sidebar is much more dense (gaps between sidebar items) than the previous one was. wondering if we should lean into the new design or increase our gap to look more like the old version? 
I would only suggest this for "layout" like components that could cause CLS when moving between embedded/etc. ways of using cloud ui/console.

frontend/src/components/console-sidebar.tsx
malinskibeniamin: I played around with what we had in the old chakra sidebar, this is the setting

malinskibeniamin: I played around with what we had in the old chakra sidebar, this is the setting

frontend/src/components/console-sidebar.tsx
malinskibeniamin: Indeed, but for now I actually need to rely on the chakra UI theme instead

frontend/src/components/layout/Header.tsx
malinskibeniamin: Sure, just didn't want to overdo it initially with too many changes.

frontend/src/components/layout/Header.tsx
malinskibeniamin: It's not about react 18, it's a tsconfig setting - you can just between react-jsx, preserve etc. we can fix it but would need then to clean up our imports across all apps.

frontend/src/components/layout/Header.tsx
malinskibeniamin: I think it would confuse people because they wouldn't know how to determine what's new and already migrated. I think this is ok

frontend/src/components/routes.tsx
```diff
@@ -397,7 +398,7 @@ export const APP_ROUTES: IRouteEntry[] = [
     '/knowledgebases',
     KnowledgeBaseList,
     'Knowledge Bases',
-    BookOpenIcon,
+    Brain,
```
malinskibeniamin: Will update to book icon

frontend/src/components/routes.tsx
```diff
@@ -379,7 +380,7 @@ export const APP_ROUTES: IRouteEntry[] = [
     '/agents',
     AgentListPage,
     'AI Agents',
-    HiOutlinePuzzlePiece,
+    Workflow,
```
malinskibeniamin: Need to add group prop to make these items render nested items.

frontend/src/components/layout/Header.tsx
```diff
@@ -42,19 +54,39 @@ const AppPageHeader = observer(() => {
...
+        <div className="flex items-center gap-2 px-4">
+          <SidebarTrigger className="-ml-1" />
+          <Separator orientation="vertical" className="mr-2 h-4" />
+          {!isEmbedded() && (
```
malinskibeniamin, line 74: Need to move the whole breadcrumbs including home icon under isEmbedded

frontend/src/components/layout/Header.tsx
jvorcak: Should we stick to the current icon set we use everywhere? We did some work to utilise all icons to be from react-icons where we prefer to have one / max two sets. I think it'd be better for consistency + we wouldn't need another package.

frontend/src/components/layout/Header.tsx
```diff
@@ -42,19 +54,39 @@ const AppPageHeader = observer(() => {
...
-            items={breadcrumbItems.map((x) => ({
-              name: x.title,
-              heading: x.heading,
-              to: x.linkTo,
-            }))}
-          />
-        )}
-      </Flex>
+      <header className="flex h-16 shrink-0 items-center gap-2 transition-[width,height] ease-linear group-has-[[data-collapsible=icon]]/sidebar-wrapper:h-12">
+        <div className="flex items-center gap-2 px-4">
```
jvorcak: Do we want this px-4?
```suggestion
        <div className="flex items-center gap-2">
```

Without it - 
Screenshot 2025-08-22 at 9 43 47
Screenshot 2025-08-22 at 9 43 30

With px-4
Screenshot 2025-08-22 at 9 44 13
Screenshot 2025-08-22 at 9 44 07

frontend/src/components/redpanda-ui/components/breadcrumb.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { Slot as SlotPrimitive } from 'radix-ui';
+import React from 'react';
+
+import { cn } from '../lib/utils';
+
+function Breadcrumb({ testId, ...props }: React.ComponentProps<'nav'> & { testId?: string }) {
+  return <nav aria-label="breadcrumb" data-slot="breadcrumb" data-testid={testId} {...props} />;
+}
+
+function BreadcrumbList({ className, testId, ...props }: React.ComponentProps<'ol'> & { testId?: string }) {
```
jvorcak, line 11: question: maybe these would belong to a UI-registry in the future?

jvorcak (approved):
Really nice work, I like it and it also solves the bug where users use it on a smaller screen. I've only added nits.

frontend/src/components/redpanda-ui/components/breadcrumb.tsx
malinskibeniamin: It is actually from the registry :) It's pulled using fumadocs CLI.

malinskibeniamin:
@jvorcak  https://github.com/redpanda-data/console/pull/1881#discussion_r2292971757 right now we use:
- hero icons
- chakra icons
- react icons MD and IO subset

I want to use lucide icons, they are a superset of all the above. I already used the lucide icons everywhere, never found any issues, very high quality (works with dark mode too, good resolution, 10 variants of each icon etc.) If you are ok with this, I can fix all the icons across all the apps to use lucide. easy task for an LLM

jvorcak:
> @jvorcak [#1881 (comment)](https://github.com/redpanda-data/console/pull/1881#discussion_r2292971757) right now we use:
> 
> * hero icons
> * chakra icons
> * react icons MD and IO subset
> 
> I want to use lucide icons, they are a superset of all the above. I already used the lucide icons everywhere, never found any issues, very high quality (works with dark mode too, good resolution, 10 variants of each icon etc.) If you are ok with this, I can fix all the icons across all the apps to use lucide. easy task for an LLM

I'm totally fine with it, I just wanted it to be consistent :)

frontend/src/components/console-sidebar.tsx
malinskibeniamin: same with some funky font size like 12.6 inter, I won't question it, just trying to avoid layout shift if we have some short time with discrepancies between sidebar in console and cloud

frontend/src/components/console-sidebar.tsx
malinskibeniamin: I just followed shadcn docs, I think the idea is you should only ever have 1 `<main>` tag and also the inset variant of the sidebar can be used, I would rather keep it like this in case we decide to use inset variant in the future. It's just a `<main>` HTML tag with special styling

frontend/src/components/redpanda-ui/components/breadcrumb.tsx
malinskibeniamin: actually this came from registry originally :)

PR #1883, opened 2025-08-22
frontend: set up design system

malinskibeniamin:
We needed to make sure we use the new `theme.css` from the registry, and added some more components in advance of the feature work.


frontend/bun.lock
```diff
@@ -68,6 +69,7 @@
         "tailwind-merge": "^3.3.1",
         "tailwindcss": "^4.1.12",
         "uuid": "^11.1.0",
+        "vaul": "^1.1.2",
```
malinskibeniamin, line 72: some dependencies that our components would need regardless - not a big deal if we have them in already.

frontend/package.json
```diff
@@ -51,6 +51,7 @@
     "@monaco-editor/react": "^4.7.0",
     "@primer/octicons-react": "^19.15.2",
     "@redpanda-data/ui": "^4.2.0",
+    "@stepperize/react": "^5.1.6",
```
eblairmckee, line 54: any ideal why this and vaul got added?

frontend/bun.lock
eblairmckee: oh. that's why.

frontend/package.json
malinskibeniamin: let me remove

PR #1889, opened 2025-08-26
[2.8] UI: Custom debug bundle form improvements

r-vasquez:
Backport of https://github.com/redpanda-data/console/pull/1888

This PR introduces two changes to our custom form in the debug bundle:

1. Added units to fields that were previously missing them (size and duration).
2. The form now comes pre-filled with default values, making it easier for customers who only need to change a single field.

image


eblairmckee (approved):
LGTM

PR #1896, opened 2025-08-28
feat: add Kafka connection and client metrics

c-julin:
## Summary
Adds Prometheus metrics to monitor Kafka broker connections and active client instances for improved observability.

## Key Changes
- **New metrics**: `kafka_open_connections` gauge tracks broker connections
- **New metrics**: `kafka_active_clients` gauge tracks client instances  
- **Connection tracking**: Automatic increment/decrement via broker connect/disconnect hooks
- **Client tracking**: Exported functions for client lifecycle management
- **Error handling**: Enhanced client creation with proper error propagation

This provides better visibility into Kafka connection health and resource usage patterns.

### DIFF
--- a/backend/pkg/factory/kafka/client_hooks.go
+++ b/backend/pkg/factory/kafka/client_hooks.go
@@ -160,7 +160,7 @@ func IncrementClientCount() {
 	}
 }
 
-// DecrementClientCount decrements the active client count metric  
+// DecrementClientCount decrements the active client count metric
 func DecrementClientCount() {
 	if promActiveClients != nil {
 		promActiveClients.Dec()
--- a/backend/pkg/factory/kafka/client_hooks.go
+++ b/backend/pkg/factory/kafka/client_hooks.go
@@ -38,7 +38,7 @@ type clientHooks struct {
 	requestsReceivedCount prometheus.Counter
 	bytesReceived         prometheus.Counter
 
-	openConnections prometheus.Gauge
+	openConnections *prometheus.GaugeVec
 }
 
 var (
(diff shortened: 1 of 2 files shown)


backend/pkg/factory/kafka/client_hooks.go
```diff
@@ -130,3 +152,17 @@ func (c clientHooks) OnBrokerWrite(_ kgo.BrokerMetadata, _ int16, bytesWritten i
...
+		promActiveClients.Inc()
+	}
+}
+
+// DecrementClientCount decrements the active client count metric
+func DecrementClientCount() {
+	if promActiveClients != nil {
+		promActiveClients.Dec()
+	}
+}
```
weeco: Do we need those nil checks and therefore the wrapper functions rather than calling Inc() Dec() directly on the prometheus metrics like we do with the other hooks?

Unless you create hooks without using the constructor it should be impossible that the prom metrics are not initialized yet.

backend/pkg/factory/kafka/client_hooks.go
```diff
@@ -75,6 +79,20 @@ func newClientHooks(logger *slog.Logger, metricsNamespace string) *clientHooks {
 			Subsystem: "kafka",
 			Name:      "received_bytes",
 		})
+
+		promOpenConnections = promauto.NewGauge(prometheus.GaugeOpts{
```
weeco: I'm wondering if we should make it GaugeVec and add the broker id as a dimension. If we have a single client and a 3 node cluster, we often expect 3 connections as a minimum. This could be confusing.

backend/pkg/factory/kafka/client_hooks.go
c-julin: Yeah i was trying something with claude here to see if we can keep track of number of clients but the cache go package has no callback to decrease the counter. Do you think its worth adding?

weeco: I think this specific factory implementation only ever creates a single client so that's fine - but the other factory implementations (closed source) it's more dynamic.

PR #1897, opened 2025-08-29
Improved heap analytics

jvorcak:


backend/pkg/api/handle_frontend.go
```diff
@@ -45,7 +45,7 @@ func (api *API) handleFrontendIndex() http.HandlerFunc {
 
 	// Add analytics_enabled feature if analytics is enabled
 	if api.Cfg.Analytics.Enabled {
-		enabledFeatures = append(enabledFeatures, "analytics_enabled")
+		enabledFeatures = append(enabledFeatures, "ANALYTICS")
```
jvorcak, line 48: I've changed the naming so that we're consistent with other flags.

frontend/public/index.html
```diff
@@ -76,11 +76,11 @@
                     heap.load("<%= HEAP_APP_ID %>");
                 `;
                 document.head.appendChild(heapScript);
-            }
+            }   
         };
         
-        // Load analytics scripts immediately since we have the config
-        window.loadAnalyticsScripts();
+        // Load analytics scripts after the page is fully loaded
```
jvorcak, line 82: in dev mode, we need to wait for ENABLED_FEATURES to be loaded.

weeco (approved):

PR #1899, opened 2025-08-29
feat: support custom Prometheus registry for Kafka metrics

c-julin:
## Summary
- Enables injection of custom Prometheus registry for Kafka client metrics
- Replaces global registry usage with per-registry metric management
- Maintains all existing metrics functionality including active client tracking

## Changes
- Modified `client_hooks.go` to support per-registry metrics instead of global promauto
- Updated `CachedClientProvider` to accept and use injected Prometheus registry
- Fixed `/admin/metrics` endpoint to use injected registry
- Updated test files to pass registry parameter

This allows enterprise repos to inject their own Prometheus registry while preserving all Kafka metrics collection.

### DIFF
--- a/backend/pkg/api/routes.go
+++ b/backend/pkg/api/routes.go
@@ -23,6 +23,7 @@ import (
 	chimiddleware "github.com/go-chi/chi/v5/middleware"
 	"github.com/go-chi/cors"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	commoninterceptor "github.com/redpanda-data/common-go/api/interceptor"
 	"github.com/redpanda-data/common-go/api/metrics"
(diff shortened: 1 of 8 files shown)


weeco (approved):

backend/pkg/api/routes.go
```diff
@@ -546,7 +547,7 @@ func (api *API) routes() *chi.Mux {
 		// Private routes - these should only be accessible from within Kubernetes or a protected ingress
 		router.Group(func(r chi.Router) {
 			api.Hooks.Route.ConfigInternalRouter(r)
-			r.Handle("/admin/metrics", promhttp.Handler())
+			r.Handle("/admin/metrics", promhttp.HandlerFor(api.PrometheusRegistry.(*prometheus.Registry), promhttp.HandlerOpts{}))
```
weeco, line 550: I don't have a compiler at hand with this branch at the moment, but is the type assertion necessary here?

c-julin: we use the interface in the constructor which probably isn't necessary would be simpler to use the type. can update this

weeco: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#HandlerFor

This suggests that it also accepts the interface though :D

c-julin: Different interfaces Registerer and Gatherer i think its better to just use the type which implements both interfaces. instead we store it as a registerer and then use it as both.

PR #1904, opened 2025-09-04
feat: add List topic configuration and get documentation service endpoints

c-julin:
## Summary

Adds two new topic service endpoints to improve API efficiency and provide better topic management capabilities:

- **ListTopicsConfigurations** - Optimized endpoint that only calls DescribeConfigs 
- **GetTopicDocumentations** - documentation retrieval for a topic

## Key Changes

- Remove metadata dependency from `ListTopicsConfigurations` for better performance
- Implement documentation state enum (NOT_CONFIGURED, NOT_EXISTENT, AVAILABLE)
- Move response mapping logic to mapper classes for better separation of concerns

## gRPC Service Methods

- `ListTopicsConfigurations` - List configurations for multiple topics
- `GetTopicDocumentations` - Get documentation for topic

### DIFF
--- a/backend/pkg/api/connect/service/topic/v1/mapper.go
+++ b/backend/pkg/api/connect/service/topic/v1/mapper.go
@@ -238,7 +238,7 @@ func (*mapper) setTopicConfigurationsResourceToKafka(req *v1.SetTopicConfigurati
 	return kafkaReq
 }
 
-func (*mapper) topicDocumentationToProto(_ string, docs *console.TopicDocumentation) *v1.GetTopicDocumentationResponse {
+func (*mapper) topicDocumentationToProto(docs *console.TopicDocumentation) *v1.GetTopicDocumentationResponse {
 	result := &v1.GetTopicDocumentationResponse{}
 
 	if !docs.IsEnabled {
--- a/backend/pkg/api/connect/service/topic/v1/service.go
+++ b/backend/pkg/api/connect/service/topic/v1/service.go
@@ -562,7 +562,7 @@ func (s *Service) ListTopicsConfigurations(ctx context.Context, req *connect.Req
 // GetTopicDocumentation gets documentation for a specific topic.
 func (s *Service) GetTopicDocumentation(_ context.Context, req *connect.Request[v1.GetTopicDocumentationRequest]) (*connect.Response[v1.GetTopicDocumentationResponse], error) {
 	docs := s.consoleSvc.GetTopicDocumentation(req.Msg.TopicName)
-	protoDoc := s.mapper.topicDocumentationToProto(req.Msg.TopicName, docs)
+	protoDoc := s.mapper.topicDocumentationToProto(docs)
 
 	return connect.NewResponse(protoDoc), nil
 }
--- a/backend/pkg/console/servicer.go
+++ b/backend/pkg/console/servicer.go
@@ -86,8 +86,6 @@ type Servicer interface {
 	DescribeClientQuotas(ctx context.Context, req *kmsg.DescribeClientQuotasRequest) (*kmsg.DescribeClientQuotasResponse, error)
 	// AlterClientQuotas proxies the request/response for altering client quotas via the Kafka API.
 	AlterClientQuotas(ctx context.Context, req *kmsg.AlterClientQuotasRequest) (*kmsg.AlterClientQuotasResponse, error)
-	// DescribeLogDirs proxies the request/response for describing log directories via the Kafka API.
-	DescribeLogDirs(ctx context.Context, req *kmsg.DescribeLogDirsRequest) (*kmsg.DescribeLogDirsResponse, error)
 }
 
 // SchemaRegistryServicer is the interface for schema registry servicer


proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -400,6 +498,68 @@ service TopicService {
       api: API_KAFKA
     };
   }
+  rpc ListTopicsConfigurations(ListTopicsConfigurationsRequest) returns (ListTopicsConfigurationsResponse) {
+    option (google.api.http) = {get: "/v1/topics/configurations"};
```
weeco: We can't use these paths, they would clash with a potential `/v1/topics/{topicName}` path.

In these cases we follow https://google.aip.dev/231 . So something like: `/v1/topics:batchGetConfigurations`.

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -322,6 +322,104 @@ message AddPartitionsToTopicsResponse {
...
+  repeated string topic_names = 1 [
+    (buf.validate.field).required = true,
+    (buf.validate.field).repeated.min_items = 1,
+    (buf.validate.field).repeated.items.string.min_len = 1,
+    (buf.validate.field).repeated.items.string.max_len = 249,
+    (buf.validate.field).repeated.items.string.pattern = "^[a-zA-Z0-9._\\-]*$"
+  ];
+}
+
+message ListTopicsConfigurationsResponse {
```
weeco, line 337: We should add pagination for List / Bulk endpoints

backend/pkg/api/connect/service/topic/v1/service.go
```diff
@@ -511,3 +511,104 @@ func (s *Service) SetPartitionsToTopics(ctx context.Context, req *connect.Reques
...
+		configResp, err := s.consoleSvc.DescribeConfigs(ctx, &configReq)
+		if err != nil {
+			s.logger.WarnContext(ctx, "failed to describe topic configs", slog.Any("error", err))
+			// Continue without configurations
+			configuredTopics = append(configuredTopics, &v1.ListTopicsConfigurationsResponse_TopicWithConfigurations{
+				Name:           topicName,
+				Configurations: []*v1.Topic_Configuration{},
+			})
+			continue
+		}
```
weeco: Why are we doing this? Usually we don't want to create endpoints that end up in many requests under the hood. But this would be the case here. If you request 50 topics this is **50 sequential round trips** to Kafka.

However, the Kafka API actually supports requesting configs in bulk. We should be using this instead of creating requests sequentially in a loop.

backend/pkg/api/connect/service/topic/v1/service.go
```diff
@@ -511,3 +511,104 @@ func (s *Service) SetPartitionsToTopics(ctx context.Context, req *connect.Reques
...
+
+		// Make the request
+		configResp, err := s.consoleSvc.DescribeConfigs(ctx, &configReq)
+		if err != nil {
+			s.logger.WarnContext(ctx, "failed to describe topic configs", slog.Any("error", err))
+			// Continue without configurations
+			configuredTopics = append(configuredTopics, &v1.ListTopicsConfigurationsResponse_TopicWithConfigurations{
+				Name:           topicName,
+				Configurations: []*v1.Topic_Configuration{},
+			})
```
weeco: That's also weird. Errors wouldn't be propagated back to the user

backend/pkg/api/connect/service/topic/v1/service.go
```diff
@@ -511,3 +511,104 @@ func (s *Service) SetPartitionsToTopics(ctx context.Context, req *connect.Reques
...
+			continue
+		}
+
+		// Convert configurations
+		var configurations []*v1.Topic_Configuration
+		if len(configResp.Resources) > 0 {
+			configs, err := s.mapper.describeTopicConfigsToProto(configResp.Resources[0].Configs)
```
weeco: This probably also have a resource specific error code that we must check.

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -400,6 +498,68 @@ service TopicService {
       api: API_KAFKA
     };
   }
+  rpc ListTopicsConfigurations(ListTopicsConfigurationsRequest) returns (ListTopicsConfigurationsResponse) {
+    option (google.api.http) = {get: "/v1/topics/configurations"};
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "List topics with configurations"
+      description: "List topics with their configurations, partition count and replication factor. Optional: filter based on topic name."
```
weeco: partition count and replication factor aren't included here. The "filter" is not really a filter but more a selector and thus not optional I guess?

backend/pkg/api/connect/service/topic/v1/mapper.go
```diff
@@ -219,3 +220,83 @@ func (*mapper) setTopicConfigurationsResourceToKafka(req *v1.SetTopicConfigurati
...
+
+	return logDirs
+}
+
+func (*mapper) topicDocumentationToProto(topicName string, docs *console.TopicDocumentation) *v1.ListTopicDocumentationsResponse_TopicDocumentation {
+	result := &v1.ListTopicDocumentationsResponse_TopicDocumentation{
+		TopicName: topicName,
+	}
+
+	if !docs.IsEnabled {
```
weeco, line 244: I think docs can be nil, haven't seen a check for it?

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -400,6 +498,68 @@ service TopicService {
...
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_KAFKA
+    };
+  }
+
+  rpc ListTopicDocumentations(ListTopicDocumentationsRequest) returns (ListTopicDocumentationsResponse) {
```
weeco: Do we need this RPC? Seems weird to have a bulk endpoint for this. I think only a per topic endpoint is needed

backend/pkg/api/connect/service/topic/v1/service.go
```diff
@@ -511,3 +512,101 @@ func (s *Service) SetPartitionsToTopics(ctx context.Context, req *connect.Reques
...
+			connect.CodeInternal,
+			err,
+			apierrors.NewErrorInfo(v1.Reason_REASON_KAFKA_API_ERROR.String(), apierrors.KeyValsFromKafkaError(err)...),
+		)
+	}
+
+	// Process each resource in the response (1:1 with requested topics)
+	for _, resource := range configResp.Resources {
+		// Check for inner Kafka error
+		if err = kerr.ErrorForCode(resource.ErrorCode); err != nil {
```
c-julin, line 536: Would be open to feedback here, should i modify the list topics response so it sends partial responses and collects the errors separately. Same later if the mapper fails should we send back a partial response

proto/redpanda/api/dataplane/v1/topic.proto
c-julin: As we're using the topic name as a selector, should I still have pagination? My thinking is if a user has requested the configuration of 50 topics would they then be expected to supply a pagination token, the validation above means at least one topic name is required.

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -322,6 +322,112 @@ message AddPartitionsToTopicsResponse {
...
+    // Topic name.
+    string name = 1;
+    // Total size in bytes that this topic takes up across all log directories.
+    int64 total_size = 2;
+    // Total number of partitions for this topic across all log directories.
+    int64 total_partitions = 3;
+  }
+
+  repeated LogDir log_dirs = 1;
+  repeated TopicSummary topic_summaries = 2;
```
c-julin: Same here would like feedback, not sure what the right thing here is, as there is processing it feels like a backend responsibility but also feels odd to ListLogDirs endpoint, should this be a separate endpoint? or is it okay to have this logic in the frontend?

proto/redpanda/api/dataplane/v1/topic.proto
weeco: I see your point, with the request they already control the response size. Interesting case. I'd limit the the number of topics that can be requested though. I don't think it will go well if someone requests 1000+ topics' configs.

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -400,6 +500,77 @@ service TopicService {
...
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_KAFKA
+    };
+  }
+
+  rpc ListLogDirs(ListLogDirsRequest) returns (ListLogDirsResponse) {
```
weeco: I'd rename to `ListTopicsLogDirs()` and updating the proto messages for the request and response types. We aggregate the information by topic, and this way we avoid ambiguous naming if we happen to introduce some LogDirs requests for the brokers at some point.

proto/redpanda/api/dataplane/v1/topic.proto
weeco: Yes this contains quite some logic, also for de-duplicating the shared storage log dirs (e.g. s3). It makes sense to me to do this on the backend which is more familiar with how Kafka / Redpanda works generally. I'd split it into separate RPCs - see other comment on the service rpc.

proto/redpanda/api/dataplane/v1/topic.proto
weeco: Three thoughts here:

1. I think a cluster wide `ListLogDirs` rpc would be good to have in a cluster/broker service. We don't have one yet afiak.
2. It's fair to have a `ListTopicLogDirs` rpc where a single topic's logdirs can be requested. This rpc is quite a low-level one and we don't surface that information anywhere yet but generally good to have. This should simplify the API usage.
3. Currently the response combines two distinct data views (detailed hierarchical data and the cross-directory aggregates). The one seems more relevant for operational debugging and the other is used in our UI for monitoring / user facing metrics. My preference would be to split them into two separate RPCs.

Something like this:

```proto
  rpc ListTopicLogDirs(ListTopicLogDirsRequest) returns (ListTopicLogDirsResponse);
  rpc GetTopicStorageSummary(GetTopicStorageSummaryRequest) returns (GetTopicStorageSummaryResponse);
```

A couple challenges with the `GetTopicStorageSummary` which generally tries to answer how much data each topic has. I hope we can re-use existing implementations which considered these challenges:

1. We always have to ask all brokers for their log dir information. The slowest broker determines the response and it can happen that there's one broker which times out or is unavailable (e.g. rolling upgrades for large clusters can take an hour or so). We must ensure it's still working fine if just one broker is not answering, and at the same time transport the information that one broker's response is missing. You can find quite some logic for this in the current handler implementation.
2. We have to de-duplicate shared storage log dirs.

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -400,6 +500,77 @@ service TopicService {
       api: API_KAFKA
     };
   }
+  rpc ListTopicsConfigurations(ListTopicsConfigurationsRequest) returns (ListTopicsConfigurationsResponse) {
```
weeco, line 450: Generally not a fan of the bulk APIs, because using them is quite hard. The whole response can fail, or all the individual sub parts can fail. Wondering if we need this RPC?

proto/redpanda/api/dataplane/v1/topic.proto
weeco: I thought about the API. We have to come up with a proto design that can handle:

1. Top-level errors (cluster authorization failures)
2. Directory-level errors (individual disk failures within a broker)
3. Topic-level implicit failures (topics missing from directories where they should be)
4. Partition-level implicit failures (partitions missing from topics)
5. Broker communication failures (timeouts, network issues)

The challenge is to capture this granularity without making the API overwhelming.

My proposal:

```proto
  rpc GetTopicStorageSummary(GetTopicStorageSummaryRequest) returns (GetTopicStorageSummaryResponse);

  message GetTopicStorageSummaryRequest {
    // Topics to get storage summaries for
    repeated string topic_names = 1 [
      (buf.validate.field).repeated.min_items = 1,
      (buf.validate.field).repeated.max_items = 100,
      (buf.validate.field).repeated.items.string.min_len = 1,
      (buf.validate.field).repeated.items.string.max_len = 249,
      (buf.validate.field).repeated.items.string.pattern = "^[a-zA-Z0-9._\\-]*$"
    ];

    // Timeout for broker queries (default: 6s)
    optional int32 timeout_ms = 2 [(buf.validate.field).int32 = {gte: 1000, lte: 30000}];
  }

  message GetTopicStorageSummaryResponse {
    enum ResponseStatus {
      RESPONSE_STATUS_UNSPECIFIED = 0;
      RESPONSE_STATUS_SUCCESS = 1;           // All expected data retrieved
      RESPONSE_STATUS_PARTIAL_SUCCESS = 2;   // Some data missing but aggregation possible
      RESPONSE_STATUS_INSUFFICIENT_DATA = 3; // Too much data missing for reliable aggregation
      RESPONSE_STATUS_FAILED = 4;            // No data available (authorization, etc.)
    }

    // Overall response status
    ResponseStatus status = 1;

    // Per-topic aggregated storage summaries
    repeated TopicStorageSummary topics = 2;

    // Simple cluster coverage info
    int32 total_brokers = 3;
    int32 responding_brokers = 4;
  }

  message TopicStorageSummary {
    enum DataQuality {
      DATA_QUALITY_UNSPECIFIED = 0;
      DATA_QUALITY_COMPLETE = 1;      // All expected brokers/directories responded
      DATA_QUALITY_ESTIMATED = 2;     // Some data missing, but good estimates possible
      DATA_QUALITY_UNRELIABLE = 3;    // Significant data missing, estimates uncertain
      DATA_QUALITY_UNAVAILABLE = 4;   // No data available for this topic
    }

    message MissingDataInfo {
      // Broker ID that was expected to provide data but didn't
      int32 broker_id = 1;

      // Specific partitions missing (empty array = all expected partitions from this broker)
      repeated int32 affected_partitions = 2;

      // Reason why data is missing from this broker
      string reason = 3;
  }

    message StorageMetrics {
      // Total bytes including all replicas (physical storage consumed)
      int64 total_bytes_with_replication = 1;

      // Logical topic size (without counting replicas - actual data size)
      int64 total_bytes_without_replication = 2;

      // Local disk storage (may include replicas depending on rf and broker distribution)
      int64 local_storage_bytes = 3;

      // Remote storage like S3 (deduplicated across brokers)
      int64 remote_storage_bytes = 4;
    }

    string name = 1;

    // Storage metrics with clear replication semantics
    optional StorageMetrics storage = 2;
    optional int32 partition_count = 3;
    optional int32 replication_factor = 4;

    // Data reliability
    DataQuality quality = 5;

    // What data is missing from brokers that should have it
    repeated MissingDataInfo missing_data = 6;
  }
```


Response would look like this for different scenarios then:

**1. Perfect scenario with clear replication semantics:**

```json
  {
    "status": "RESPONSE_STATUS_SUCCESS",
    "topics": [{
      "name": "user-events",
      "storage": {
        "total_bytes_with_replication": 3221225472,    // 3GB total (1GB × rf=3)
        "total_bytes_without_replication": 1073741824, // 1GB logical topic size
        "local_storage_bytes": 2147483648,             // 2GB on local disks
        "remote_storage_bytes": 1073741824             // 1GB in S3 (deduplicated)
      },
      "partition_count": 12,
      "replication_factor": 3,
      "quality": "DATA_QUALITY_COMPLETE",
      "missing_data": []
    }]
  }
```


**2. Partial failure - only report brokers that should have had data:**
```json
  {
    "status": "RESPONSE_STATUS_PARTIAL_SUCCESS",
    "topics": [{
      "name": "user-events",
      "storage": {
        "total_bytes_with_replication": 2147483648,    // Estimated (missing 1GB from broker 2)
        "total_bytes_without_replication": 1073741824, // Still accurate (can calculate from available replicas)
        "local_storage_bytes": 1073741824,             // Estimated (missing broker 2's local data)
        "remote_storage_bytes": 1073741824             // Accurate (S3 data from other brokers)
      },
      "partition_count": 12,
      "replication_factor": 3,
      "quality": "DATA_QUALITY_ESTIMATED",
      "missing_data": [
        {
          "broker_id": 2,
          "affected_partitions": [],  // All expected partitions from this broker
          "reason": "broker timeout after 30s"
        }
      ]
    }]
  }
```

c-julin:
Have removed the log dirs from this PR and simplified it to only list topic configurations and get documentation

proto/gen/openapi/openapi.yaml
```diff
@@ -5508,6 +5588,46 @@ paths:
...
+              type: string
+            type: array
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ListTopicsConfigurationsResponse'
+          description: OK
+        "401":
```
malinskibeniamin, line 5610: Should there also be `404`? Maybe in case the topic name does not match any existing topics.

PR #1907, opened 2025-09-08
Fix: ACL page, allow all with schema registry acl

andresaristizabal:
**ACL Resource Type Handling:**

- Changed the logic in `CreateACL.tsx` to always set the default resource type to `ResourceTypeTopic` when cluster rules exist, removing the previous check for subject rules.
- Updated the list of resource types to only include `ResourceTypeSubject` and `ResourceTypeSchemaRegistry` if `schemaRegistryEnabled` is true, preventing them from being included by default.

https://github.com/user-attachments/assets/507f5b25-a0b1-464d-adb4-db1241a23a72

### DIFF
--- a/frontend/tests/console/acl.spec.ts
+++ b/frontend/tests/console/acl.spec.ts
@@ -90,59 +90,6 @@ test.describe('ACL Creation', () => {
             DESCRIBE: OperationTypeAllow,
           },
         } as Rule,
-        {
-          id: 4,
-          resourceType: ResourceTypeSubject,
-          mode: ModeCustom,
-          selectorType: ResourcePatternTypeLiteral,
-          selectorValue: '*',
-          operations: {
-            READ: OperationTypeAllow,
-            WRITE: OperationTypeAllow,
-          },
-        } as Rule,
-        {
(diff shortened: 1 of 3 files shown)


frontend/tests/console/acl.spec.ts
```diff
@@ -1009,162 +887,12 @@ test.describe('ACL Disable Rules Validation', () => {
...
 
     await test.step('Verify ACL creation was successful', async () => {
-      await aclPage.validateRulesCount(3);
+      await aclPage.validateRulesCount(2);
       await aclPage.validateSharedConfig();
     });
   });
 });
 
-test.describe('Role membership', () => {
```
malinskibeniamin, line 1018: Did we remove this spec intentionally?

andresaristizabal: Yes, those tests should run only for `console-enterprice` https://github.com/redpanda-data/console/blob/master/frontend/tests/console-enterprise/acl.spec.ts#L1018

I didn't know about this folder when I wrote those tests.

nicolaferraro (approved):

PR #1910, opened 2025-09-09
front-end: show errors on create/update acls

andresaristizabal:
This pull request improves error handling and user feedback when creating or updating ACLs and roles. Previously, partial failures (where some operations failed but others succeeded) were not surfaced to the user. Now, errors from failed operations are collected and displayed as toast notifications, while successful operations still show a success message. The changes also refactor the API layer to return detailed results about which operations succeeded or failed.

https://github.com/user-attachments/assets/15134230-4431-4607-8e9e-712f55a90074

### DIFF
--- a/frontend/src/react-query/api/acl.tsx
+++ b/frontend/src/react-query/api/acl.tsx
@@ -497,8 +497,15 @@ export const useUpdateAclMutation = () => {
       ),
     );
 
-    await Promise.all([...createResults, ...deleteResults]);
+    const allResults = await Promise.allSettled([...createResults, ...deleteResults]);
+    const errs = new Map<number, ConnectError>();
+    const rejected = allResults.filter((result): result is PromiseRejectedResult => result.status === 'rejected');
+    rejected.forEach((result) => {
+      const r = result.reason as ConnectError;
+      errs.set(r.code, r);
+    });
     await invalid();
+    return { errors: errs.values().toArray(), created: rejected.length < allResults.length };
   };
 
   return { applyUpdates };
@@ -526,8 +533,15 @@ export const useCreateAcls = () => {
   const { invalid } = useInvalidateAclsList();
 
   const createAcls = async (acls: CreateACLRequest[]) => {
-    await Promise.all(acls.map((r) => createACLMutation(r)));
+    const results = await Promise.allSettled(acls.map((r) => createACLMutation(r)));
+    const errs = new Map<number, ConnectError>();
+    const rejected = results.filter((result): result is PromiseRejectedResult => result.status === 'rejected');
+    rejected.forEach((result) => {
+      const r = result.reason as ConnectError;
+      errs.set(r.code, r);
+    });
     await invalid();
+    return { errors: errs.values().toArray(), created: rejected.length < results.length };
   };
   return { createAcls };
 };
--- a/frontend/src/components/pages/acls/new-acl/ACL.model.tsx
+++ b/frontend/src/components/pages/acls/new-acl/ACL.model.tsx
@@ -7,6 +7,8 @@ import {
   type ListACLsResponse,
 } from 'protogen/redpanda/api/dataplane/v1/acl_pb';
 import type { ReactNode } from 'react';
+import type { ConnectError } from '@connectrpc/connect';
+import type { UseToastOptions } from '@redpanda-data/ui';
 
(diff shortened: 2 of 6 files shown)


frontend/src/react-query/api/acl.tsx
```diff
@@ -526,8 +533,15 @@ export const useCreateAcls = () => {
...
 
   const createAcls = async (acls: CreateACLRequest[]) => {
-    await Promise.all(acls.map((r) => createACLMutation(r)));
+    const results = await Promise.allSettled(acls.map((r) => createACLMutation(r)));
+    const errs = new Map<number, ConnectError>();
+    const rejected = results.filter((result): result is PromiseRejectedResult => result.status === 'rejected');
+    rejected.forEach((result) => {
+      const r = result.reason as ConnectError;
+      errs.set(r.code, r);
+    });
```
malinskibeniamin, line 542: Right now we would show multiple toasts and if there is an issue with the creation of some ACLs, it would show both success and fail states at the same time. Is it possible we show a warning toast instead, explaining that some of the ACLs were not created properly? Showing both success and failure toasts at the same time may appear confusing.

malinskibeniamin: TLDR: for AI agents which consist of several pipelines, if any of the pipelines would not create successfully, the agent which was a UI-only entity would just report "degraded" status. Here we are hoping all ACLs are also created successfully, as there is no dry run I am aware of. Do we want to agree on a common pattern here that would let us establish a best practice for errors while calling multiple mutations at once? @eblairmckee

andresaristizabal: makes complete sense

eblairmckee: Oh, that's a really good call out. I'm encountering the same issue with the onboarding wizard. The pattern I'm leaning towards would be to `Promise.allSettled`, show toasts for successful mutations, but keep the form populated and show errors on the forms that caused errors.

eblairmckee: `useForm` handles it pretty nicely. that way you get inline error messaging for specific fields, or for an entire form. keeps the user on the page, and localizes the error rather than showing an ephemeral toast with an error log.

eblairmckee: I'm also partial to the "dry run" concept... does react-query have any kind of 'dry run' functionality baked into mutations? would make failing fast a lot easier

andresaristizabal:
@malinskibeniamin what do you think about it? I added a special case for error on create schema registry and but not error in kafka


Uploading Screen Recording 2025-09-09 at 4.26.37 PM.mov…

eblairmckee:
> @malinskibeniamin @eblairmckee what do you think about it? I added a special case for error on create schema registry and but not error in kafka
> 
>  Screen.Recording.2025-09-09.at.4.26.37.PM.mov

@andresaristizabal is there not a mechanism to fail on the actual form group that had the failure? atm the error messaging doesn't pinpoint which rule failed, which doesn't actually help the end user. wondering why we're not using the new UI registry Form + useForm? that seems to handle validation and individual form errors really nicely (will share a PR with what I'm doing in console for the onboarding wizard that shows this tomorrow). 
Atm, we're relying solely on logging errors, when I feel like we should be relying on the form itself to surface error's next to the form fields themselves. 

for example:
Screenshot 2025-09-09 at 6 32 45 PM
I haven't implemented parallel form submissions yet, but this is a good example of how you can fail fast on submission for a single form field using useForm + Form from ui registry

frontend/src/react-query/api/acl.tsx
malinskibeniamin: We would need a dry run to happen on the backend, however for the React Hook Form itself there is a setting that lets you either show errors on touch/update of the field vs on the submission of the entire form. I think we should be consistent between forms, but we can change it to provide errors earlier, before the user completes the entire form.

weeco: Makes sense, but we should show at least one of the returned errors. It's not unlikely that all the returned errors are the same, so as long as the user sees the reason for one of the errors it will likely be helpful. Errors could be: Authorization failed, cluster controller unavailable etc.

andresaristizabal: Let's go with this approach, and we can follow this work on a new ticket

PR #1914, opened 2025-09-10
frontend: restore delete consumer group offsets ability

malinskibeniamin:
Fix for [UX-455](https://redpandadata.atlassian.net/browse/UX-455).

When we were originally working on the consumer group offsets migration to Chakra UI, we used the new ConfirmModal which lets us add a trigger to any DOM element, in this case it was a delete button. The original change was done [here](https://github.com/redpanda-data/console/pull/1639). However, by doing so, we unfortunately had a regression where we lost the functionality to delete individual offsets, not just the entire group.

This is now fixed, and the original look & feel was restored.

A follow-up PR will be needed to ensure this UI can be refactored to use the new dialog/modal/UI registry. A new subtask was added in [UX-481](https://redpandadata.atlassian.net/browse/UX-481).

Thanks @c-julin for spending some time pairing on this.

### DIFF
--- a/frontend/src/components/pages/consumers/Group.Details.tsx
+++ b/frontend/src/components/pages/consumers/Group.Details.tsx
@@ -156,14 +156,9 @@ class GroupDetails extends PageComponent<{ groupId: string }> {
           <Button variant="outline" onClick={() => this.editGroup()} disabledReason={cannotEditGroupReason(group)}>
             Edit Group
           </Button>
-          <DeleteOffsetsModal
-            group={group}
-            mode={this.deletingMode}
-            offsets={this.deletingOffsets}
-            onClose={() => (this.deletingOffsets = null)}
-            onInit={() => this.deleteGroup()}
-            disabledReason={cannotDeleteGroupReason(group)}
-          />
+          <Button variant="outline" onClick={() => this.deleteGroup()} disabledReason={cannotDeleteGroupReason(group)}>
+            Delete Group
+          </Button>
         </Flex>
         {/* Statistics Card */}
         {uiSettings.consumerGroupDetails.showStatisticsBar && (
@@ -217,6 +212,14 @@ class GroupDetails extends PageComponent<{ groupId: string }> {
           onClose={() => (this.edittingOffsets = null)}
           initialTopic={this.editedTopic}
           initialPartition={this.editedPartition}
+        />
+        <DeleteOffsetsModal
+          group={group}
+          mode={this.deletingMode}
+          offsets={this.deletingOffsets}
+          onClose={() => (this.deletingOffsets = null)}
+          onInit={() => this.deleteGroup()}
+          disabledReason={cannotDeleteGroupReason(group)}
         />
       </PageContent>
     );
--- a/frontend/src/components/pages/consumers/Modals.tsx
+++ b/frontend/src/components/pages/consumers/Modals.tsx
@@ -9,12 +9,11 @@
  * by the Apache License, Version 2.0
  */
 
-import { ChevronLeftIcon, ChevronRightIcon, SkipIcon } from '@primer/octicons-react';
+import { ChevronLeftIcon, ChevronRightIcon, SkipIcon, TrashIcon } from '@primer/octicons-react';
 import {
   Accordion,
   Box,
   Button,
-  ConfirmItemDeleteModal,
   createStandaloneToast,
   DataTable,
   Flex,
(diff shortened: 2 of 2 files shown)


frontend/src/components/pages/consumers/Modals.tsx
```diff
@@ -834,134 +833,159 @@ export class DeleteOffsetsModal extends Component<{
...
-        }}
-      >
-        <Box>
-          {visible && (
-            <Box>
-              {mode === 'group' && (
-                <Box>
-                  <Box borderRadius="md" mb={4}>
-                    <Text>This action will delete the following consumer group:</Text>
+      <Modal isOpen={visible} onClose={this.props.onClose}>
```
malinskibeniamin, line 836: @c-julin the key here is the "visible" prop, it was not possible to keep using the `ConfirmItemDeleteModal` component because it only lets us modify the visibility using the `trigger` component, but here we want to render the modal from 2 different places (group or individual offset). We can still update the modal content body accordingly depending on what the modal is used for.

CC @jvorcak it would be great if you could test it out just to be safe, thank you!

frontend/src/components/pages/consumers/Group.Details.tsx
```diff
@@ -156,14 +156,9 @@ class GroupDetails extends PageComponent<{ groupId: string }> {
...
-            group={group}
-            mode={this.deletingMode}
-            offsets={this.deletingOffsets}
-            onClose={() => (this.deletingOffsets = null)}
-            onInit={() => this.deleteGroup()}
-            disabledReason={cannotDeleteGroupReason(group)}
-          />
+          <Button variant="outline" onClick={() => this.deleteGroup()} disabledReason={cannotDeleteGroupReason(group)}>
+            Delete Group
+          </Button>
```
eblairmckee, line 161: nit for ui registry... given we're adding state just so we can pass things to the modal based on the group.. would make more sense to have something like `modal.open` in a callback which takes `group` as an arg. would remove dependency on `useState`

frontend/src/components/pages/consumers/Modals.tsx
eblairmckee: this relates a bit to my earlier comment... I feel like we should consider programmatic rendering of modals via a hook (in the future), like `{open, close} = useModal()`

frontend/src/components/pages/consumers/Modals.tsx
```diff
@@ -834,134 +833,159 @@ export class DeleteOffsetsModal extends Component<{
...
+                  width: '64px',
+                  height: '64px',
+                  padding: '16px',
+                  background: '#F53649',
+                  borderRadius: '50%',
+                  display: 'flex',
+                  alignItems: 'center',
+                  justifyContent: 'center',
+                }}
+              >
```
eblairmckee, line 853: any reason why we're not using tailwind classnames?

frontend/src/components/pages/consumers/Modals.tsx
```diff
@@ -834,134 +833,159 @@ export class DeleteOffsetsModal extends Component<{
...
+                {/* @ts-ignore */}
+                <TrashIcon color="white" />
+              </div>
+              <Box>
```
eblairmckee, line 857: nit: I know we're not migating Modal -> UI registry dialog at the moment, but I'd expect us to use divs and UI registry components here rather than the old Box

frontend/src/components/pages/consumers/Modals.tsx
malinskibeniamin: We are not using UI registry here. I just had enough time to reestablish the previous code and make it work again.

frontend/src/components/pages/consumers/Modals.tsx
malinskibeniamin: I think the 1st PR should be to unblock the customer and fix functionality, and we can worry about the rest in due course once we are working on the project to migrate to use the new registry underway.

frontend/src/components/pages/consumers/Modals.tsx
malinskibeniamin: I think that makes sense. In Chakra UI there was a hook called `useDisclosure()` which did just that.

frontend/src/components/pages/consumers/Group.Details.tsx
malinskibeniamin: Should we just add `useDisclosure()` for each modal like Chakra does? That way we can toggle visibility of each modal/dialog easily from anywhere.

PR #1919, opened 2025-09-12
AI Agent proto

tomasz-sadura:


proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,663 @@
...
+  map<string, string> redpanda_mcp_tools = 5 [
+    (buf.validate.field).map.keys.string = {
+      pattern: "^[A-Za-z0-9_-]+$"
+      max_len: 64
+    },
+    (buf.validate.field).map.values.string = {
+      pattern: "^[A-Za-z0-9-_/]+$"
+      max_len: 64
+    },
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Map of Redpanda MCP tools to their IDs that this AI agent can use."}
```
malinskibeniamin: Shouldn't the tools also have description, and each parameter can also have its own description?

malinskibeniamin: I think we may need a `Tool` message and `ToolParameter` message

tomasz-sadura: yea, I think this would need to be extended. It's `redpanda_mcp_servers` (wrongly named `redpanda_mcp_tools`) actually right now.
Could you expand why we would need description field and what you mean by parameters?
Could you also expand what the `Tool` and `ToolParameter` should be?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.min_len = 1,
+      (buf.validate.field).string.max_len = 128
+    ];
+
+    // Reference to the API key secret for authenticating with the LLM provider.
+    string api_key_secret_ref = 3 [
+      (google.api.field_behavior) = REQUIRED,
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.pattern = "^[A-Z${}.][A-Z0-9_${}.]*$",
```
weeco: `(buf.validate.field).string.pattern = "^\\$\\{secrets\\.[A-Za-z0-9_]+\\}$"`

NIT: I think this pattern would be stricter and therefore less room for misusing the API

Used in: https://github.com/redpanda-data/console/blob/3680f37b23fcbd16baa8b72040096a11843b046c/proto/redpanda/api/dataplane/v1alpha3/knowledge_base.proto#L43

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+
+  // AI Agent ID.
+  string id = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_/]+$"
+  ];
+
+  // User-friendly AI agent name.
+  string name = 2 [
```
weeco:
```suggestion
  string display_name = 2 [
```

Just for consistency with pipelines and knowledge bases

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+    type: "redpanda.api.dataplane.v1alpha3/AIAgent"
+    singular: "ai_agent"
+    plural: "ai_agents"
+  };
+
+  // AI Agent ID.
+  string id = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_/]+$"
```
weeco: NIT: We could add: `(google.api.field_behavior) = OUTPUT_ONLY,`. Given it's not a shared resource for create and get responses unsure how helpful that is. Probably also immutable.

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+
+  // Optional AI agent description.
+  string description = 3 [(buf.validate.field).string.max_len = 256];
+
+  // System Prompt for the AI agent behavior.
+  string system_prompt = 4 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.min_len = 10,
+    (buf.validate.field).string.max_len = 8192,
```
weeco:
```suggestion
    (buf.validate.field).string.max_len = 16384,
```

The system prompts are often quite verbose.

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_EDIT
+      api: API_CONSOLE
+    };
+  }
+
+  // StopAIAgent stops a specific AI agent.
+  rpc StopAIAgent(StopAIAgentRequest) returns (StopAIAgentResponse) {
+    option (google.api.http) = {post: "/v1alpha3/ai-agents/{id}:stop"};
```
weeco:
```suggestion
    option (google.api.http) = {put: "/v1alpha3/ai-agents/{id}:stop"};
```

We decided to go with put for stopping pipelines: https://github.com/redpanda-data/console/blob/3680f37b23fcbd16baa8b72040096a11843b046c/proto/redpanda/api/dataplane/v1/pipeline.proto#L267

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+          description: "Not Found"
+          schema: {
+            json_schema: {ref: ".google.rpc.Status"}
+          }
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_EDIT
+      api: API_CONSOLE
```
weeco: Probably should add a new API type for all the agent rpcs. For knowledgebase and pipelines we use `API_REDPANDA_CONNECT`. Can be done at a later point as well since we aren't doing granular authZ yet.

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+      lte: 32768
+    }];
+
+    // Response format type.
+    enum ResponseFormat {
+      RESPONSE_FORMAT_UNSPECIFIED = 0;
+      RESPONSE_FORMAT_TEXT = 1;
+      RESPONSE_FORMAT_JSON = 2;
+      RESPONSE_FORMAT_JSON_OBJECT = 3;
+    }
```
weeco: We have this enum (almost identical) twice. Once inside `LLMSettings` and once at the `AIAgent` level. I guess by accident. I think this belongs to the agent level and not LLM settings? It seems this one is unused, so probably here by accident?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+      (buf.validate.field).required = true,
+      (buf.validate.field).string.pattern = "^[A-Z${}.][A-Z0-9_${}.]*$",
+      (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Reference to a secret containing the API key for the LLM provider."}
+    ];
+
+    // Optional temperature setting for response randomness (0.0 to 2.0).
+    optional double temperature = 4 [(buf.validate.field).double = {
+      gte: 0.0
+      lte: 2.0
+    }];
```
weeco: Some models (e.g. GPT-5) do not support the temperature setting and then we would run into issues. I would probably just remove this setting and use the model defaults

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+
+    // Optional maximum tokens for responses.
+    optional int32 max_tokens = 5 [(buf.validate.field).int32 = {
+      gte: 1
+      lte: 32768
```
weeco: We should go as high as the model specific limits, so probably up to 2M or so?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+  ];
+
+  // LLM configuration settings for the AI agent.
+  LLMSettings llm_settings = 5 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
+  ];
+
+  // Map of MCP servers that this AI agent can use.
+  map<string, MCPServer> mcp_servers = 6 [
```
weeco: I prefer to just use repeat MCPServer and make the key a part of the MCPServer message then. I prefer it because:
1.  I was wondering what the key of this map is for. As part of the MCPServer we could give it an explicit name.
2. The key as part of the message allows us to do better validation and we can add whatever annotations we prefer.

Do we need to use a map / guarantee some key uniqueness? (We could still achieve a unique value across a slice with validation though).

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+
+  // Optional response format specification.
+  optional ResponseFormat response_format = 8 [(buf.validate.field).enum = {defined_only: true}];
+
+  // Optional JSON schema for structured responses.
+  // Only used when response_format is RESPONSE_FORMAT_JSON.
+  optional string response_schema = 9 [
+    (buf.validate.field).string.max_len = 8192,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "JSON schema defining the structure of expected responses when using JSON response format."}
+  ];
```
weeco: Could imagine refactoring this into a one_of pattern, so that we can express the field dependency better. What do you think?

```proto
  oneof response_configuration {
    TextResponseConfig text_response = 8;
    JsonResponseConfig json_response = 9;
  }

  message TextResponseConfig {
    // Could add text-specific options later
  }

  message JsonResponseConfig {
    // Optional JSON schema for structured responses
    optional string schema = 1 [(buf.validate.field).string.max_len = 8192];
  }
```

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+    // The AI agent is stopped and in paused state.
+    STATE_STOPPED = 4;
+    // The AI agent encountered an error.
+    STATE_ERROR = 5;
+  }
+
+  // MCP transport types.
+  enum MCPTransport {
+    MCP_TRANSPORT_UNSPECIFIED = 0;
+    MCP_TRANSPORT_REDPANDA_MCP_SERVER = 1;
```
weeco: Do we require different input / extra information or why do we need to distinguish between streamable/sse and redpanda I wonder? What changes if this is set?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,695 @@
...
+    MCP_TRANSPORT_STREAMABLE = 2;
+    MCP_TRANSPORT_SSE = 3;
+  }
+
+  // MCP server configuration.
+  message MCPServer {
```
weeco, line 114: Do we need to add something for the authorization to our Redpanda hosted MCP servers which are OAuth protected?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
tomasz-sadura: Yea, you can update a single map item with updatemask, while you cannot update a single list item it seems...

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
tomasz-sadura: For Redpanda we can do wiring within the cluster. It's true we can determine Redpanda by the URL, so basically we could simplify and remove the `REDPANDA_MCP_SERVER` option. Adding @birdayz for his opinion whether he sees any need to explicitly highlight Redpanda MCP Servers here

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
tomasz-sadura: maybe later. For now I see it as passing the authorization header from the agent request. You can use `pass_headers` for this.

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,684 @@
...
+  string system_prompt = 4 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.min_len = 10,
+    (buf.validate.field).string.max_len = 16384,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "System instructions that define how the AI agent should behave and respond."}
+  ];
+
+  // LLM configuration settings for the AI agent.
+  LLMSettings llm_settings = 5 [
```
birdayz: should we have a shared type for this?

we'll need provider + model + credentials combo more often i suppose.

we have https://github.com/redpanda-data/console/pull/1919/files#diff-b0a5439bbd6fd803547132a9976f6e639baed20911cae690b6dbf978fa0ea0ce for KB.

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,684 @@
...
+      (google.api.field_behavior) = REQUIRED,
+      (buf.validate.field).required = true,
+      (buf.validate.field).enum = {
+        defined_only: true
+        not_in: [0]
+      }
+    ];
+
+    // URL of the MCP server.
+    string server_url = 2 [
```
birdayz: do we need arbitrary servers at first?

can we start with a simple variant, that just works with our remote mcp servers?

e.g. repeated string mcp_server_ids.
and we config it under the hood.

maybe only open up to ther MCPs later. we just build the fully managed experience for now with our mcp server.
then we don't need things like pass headers, headers.. 

for Oct14, we need to support usage of the "entire thing". for truly remote servers, we will most likely need more elaborate config with oidc dynamic client registration etc, i'd rather skip it until we truly focus on it

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,684 @@
...
+            pattern: "^[A-Za-z0-9-_]+$"
+            max_len: 64
+          }
+        }
+      },
+      (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "List of header names from the request to pass through to the MCP server (e.g., 'Authorization', 'X-User-ID')."}
+    ];
+
+    // Optional timeout for requests to the MCP server in seconds.
+    optional int32 timeout = 5 [
```
birdayz: do we need this knob?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
weeco: In the end we always have to configure the agents with these information anyways no? In fact I was wondering whether we wouldn't need the oidc client id & secret for our own agents and whether this is missing. I think arbitrary remote MCPs would be helpful, the majority supports apitokens besides OAuth FWIW

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
weeco: Okay interesting, I'd clarify the purpose of the map key then. I assume it's the mcp server id.

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
tomasz-sadura: makes sense if required props are different. Done

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
tomasz-sadura: what about we support Redpanda only for now. I commented out the more generic MCPServers section. I see no issue with supporting them (with API key) once we support Redpanda MCP, as the code to call proper transport will be basically the same? So it's a small change and some testing then

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
tomasz-sadura: probably not at this stage

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,684 @@
...
+  // URL to connect to the AI agent endpoint.
+  string url = 11 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  // Response configuration.
+  oneof response_configuration {
+    // Text response configuration.
+    TextResponseConfig text_response = 12;
+
+    // JSON response configuration.
+    JsonResponseConfig json_response = 13;
```
birdayz: how well would this work with Gemini and Claude ? if it's provider specific and not 100% generic, we have to move it into the OpenAI LLM block

all providers support some sort of structured output, but is this API field identical for all?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,684 @@
...
+    STATE_STOPPED = 4;
+    // The AI agent encountered an error.
+    STATE_ERROR = 5;
+  }
+
+  // MCP transport types.
+  enum MCPTransport {
+    MCP_TRANSPORT_UNSPECIFIED = 0;
+    MCP_TRANSPORT_STREAMABLE = 1;
+    MCP_TRANSPORT_SSE = 2;
```
birdayz: i think we can skip this one, if we support only redpanda cloud remote mcp for now, we do not need this sort of configurability

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
birdayz: i suggest to go this way. only support redpanda cloud managed remote mcp, and expand later. Minimal effort, most streamlined Milestone 0 (we must build on top of Remote MCP which was just shipped)

i'd throw out the commented out code and add it back when we need it (dead code is bad code)

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -0,0 +1,684 @@
...
+    (buf.validate.field).required = true
+  ];
+  string model = 7 [
+    (google.api.field_behavior) = REQUIRED,
+    (google.api.field_behavior) = IMMUTABLE,
+    (buf.validate.field).required = true
+  ];
+
+  // List of Redpanda MCP server IDs that this AI agent can use.
+  repeated string mcp_server_ids = 8;
```
birdayz: hm, i think we need a message type value for this, so we can e.g. include/exclude tools.
then we can also easier evolve this type further, e.g. if we really introduce support for non-managed MCPs

@weeco mentioned this, some frameworks allow plugging an mcp, but you can provide an exclusion, inclusion regex for tool names.

what about

message MCPServer {
  string id = 1;
  included_tools_pattern = 2; // Default to including all / .*
  excluded_tools_pattern = 3; // Default to empty / disabled
  // Anything else?
}

might as well do map for fieldmask tbh

weeco: What I have seen was the filter on the consumer side (agent that can configure the include/exclude)

birdayz: this is the consumer side, no? (list of tools the agent has available)

birdayz (approved):
approving to unblock. i think, further design discussions will arise, but we will be in a better position to discuss these if we have deployed a prototype already and wired it up with the agent.

tomasz-sadura:
ok, makes sense. I cleanup up a little:
- use map from string to MCPServer object with string id only for now
- remove response_format altogether for now

PR #1942, opened 2025-09-28
frontend: replace vitest with bun test

malinskibeniamin:
We are going to work on our test coverage, so it was a good opportunity to move from `vitest` to `bun:test`.

Bun test is extremely fast and scales well. We want to introduce a good integration test suite for our Console frontend without making CI feel sluggish.

**BEFORE (9s)**

```sh
 Test Files  13 passed (13)
      Tests  68 passed | 1 skipped (69)
   Start at  19:22:24
   Duration  9.52s (transform 3.81s, setup 1.04s, collect 61.15s, tests 2.01s, environment 5.90s, prepare 722ms)

 HTML  Report is generated
       You can run npx vite preview --outDir html to see the test results.
```
---

**AFTER (2s)**

```sh
$ bun run test
$ TZ=GMT bun test --concurrent src/**/*.test.{ts,tsx} src/components/**/*.test.{ts,tsx} src/components/pages/**/*.test.{ts,tsx} src/components/pages/topics/**/*.test.{ts,tsx}
bun test v1.2.23 (cf136713)

src/utils/utils.test.ts:
✓ substringWithEllipsis > returns the original string if its length is less than maxLength [0.53ms]
✓ substringWithEllipsis > returns the original string if its length is equal to maxLength [0.01ms]
✓ substringWithEllipsis > handles cases where maxLength is less than 3 [0.02ms]
✓ substringWithEllipsis > returns an empty string with ellipsis if maxLength is 0
✓ substringWithEllipsis > correctly handles an empty input string

src/utils/protobuf-reflection.test.ts:
✓ Protobuf Field Reflection > should detect field behaviors for KnowledgeBase [1.23ms]
✓ Protobuf Field Reflection > should detect field behaviors for KnowledgeBaseUpdate [0.39ms]
✓ Protobuf Field Reflection > should detect field behaviors for nested Postgres schema [0.07ms]
Field 'nonExistentField' not found in message schema 'KnowledgeBase'
Field 'nonExistentField' not found in message schema 'KnowledgeBase'
Field 'nonExistentField' not found in message schema 'KnowledgeBase'
Field 'nonExistentField' not found in message schema 'KnowledgeBase'
✓ Protobuf Field Reflection > should handle non-existent fields gracefully [1.32ms]

src/hooks/usePaginationParams.test.ts:
✓ usePaginationParams > returns default values when URL parameters are absent [4.15ms]
✓ usePaginationParams > parses pageSize and pageIndex from URL parameters [0.56ms]
✓ usePaginationParams > uses defaultPageSize when pageSize is not in URL [0.28ms]
✓ usePaginationParams > adjusts pageIndex if it exceeds calculated totalPages [0.37ms]

src/components/pages/secrets/delete-secret-modal.test.tsx:
✓ DeleteSecretModal > should let the user delete a secret after entering confirmation text [74.36ms]
✓ DeleteSecretModal > should show a warning if the secret is in use [77.85ms]

src/components/pages/secrets/create-secret-modal.test.tsx:
✓ CreateSecretModal > should let the user create a secret [112.53ms]

src/components/pages/secrets/update-secret-modal.test.tsx:
✓ UpdateSecretModal > should let the user update an existing secret with new value and labels [52.55ms]
✓ UpdateSecretModal > should show a warning if the secret is in use [46.15ms]

src/components/pages/knowledgebase/regex-matching.test.ts:
✓ Topic Regex Matching > should match confluence with conf.* [0.43ms]
✓ Topic Regex Matching > should match multiple topics starting with conf [0.04ms]
✓ Topic Regex Matching > should match topics containing conf with .*conf.* [0.02ms]
✓ Topic Regex Matching > should match exact topic names [0.02ms]
✓ Topic Regex Matching > should handle invalid regex by falling back to substring [0.03ms]
✓ Topic Regex Matching > should return empty array for empty pattern [0.01ms]
✓ Topic Regex Matching > should handle complex regex patterns [0.02ms]
✓ Topic Regex Matching > should handle the exact case mentioned: conf.* matching confluence [0.03ms]

src/components/pages/topics/TopicConfiguration.test.tsx:
{"NODE_ENV":"test","GIT_SHA":"abc123","GIT_REF":"dev-local","BUILD_TIMESTAMP":"32503680000","DEV_HINT":"true"}
✓ TopicConfiguration > renders groups in the correct order [18.30ms]

src/components/pages/topics/DeleteRecordsModal/DeleteRecordsModal.test.tsx:
✓ DeleteRecordsModal > renders all expected elements in step 1 [28.63ms]

src/components/pages/topics/Tab.Acl/AclList.test.tsx:
Pipeline Service state does not matter. Showing sidebar link.
Creating Table Instance...
✓ AclList > renders an empty table when no data is present [5.88ms]
Creating Table Instance...
✓ AclList > a table with one entry [2.92ms]
Creating Table Instance...
✓ AclList > informs user about missing permission to view ACLs [2.11ms]
Creating Table Instance...
✓ AclList > informs user about missing authorizer config in Kafka cluster [2.17ms]

 32 pass
 0 fail
 79 expect() calls
Ran 32 tests across 10 files. [2.35s]
```


frontend/playwright.config.ts
```diff
@@ -23,7 +23,7 @@ export default defineConfig({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
```
sago2k8, line 26: [blocker/question]: why  increasing the workers now ?

frontend/src/react-query/api/acl.tsx
```diff
@@ -511,10 +511,11 @@ export const useUpdateAclMutation = () => {
   return { applyUpdates };
 };
 
-export const useGetAclsByPrincipal = <T = AclDetail>(
+// biome-ignore lint/complexity/useArrowFunction: we need a function because of typescript complaining
+export const useGetAclsByPrincipal = function <T = AclDetail>(
   principal: string,
   transformFn?: (aclList: ListACLsResponse) => T,
-) => {
+ ) {
```
sago2k8: [minor]: I think  you can use a named function in here

frontend/src/utils/protobuf-reflection.test.ts
```diff
@@ -42,7 +43,7 @@ describe('Protobuf Field Reflection', () => {
   test('should detect field behaviors for KnowledgeBaseUpdate', () => {
     debugFieldBehaviors(KnowledgeBaseUpdateSchema);
 
-    const _metadata = getMessageFieldMetadata(KnowledgeBaseUpdateSchema);
+    getMessageFieldMetadata(KnowledgeBaseUpdateSchema);
```
sago2k8, line 46: [blocker]: is this function doing side-effects?, if so I would prefer to refactor that if not this is a void invocation.

.github/workflows/frontend-verify.yml
```diff
@@ -2,6 +2,8 @@
 name: "Frontend Lint & Compile check"
 on:
   push:
+    branches:
```
sago2k8, line 5: [blocker]: don't we want to trigger CI for every branch?

frontend/bun.lock
```diff
@@ -38,6 +38,7 @@
         "cmdk": "^1.1.1",
         "dexie": "^4.0.11",
         "dexie-react-hooks": "^1.1.7",
+        "dotenv": "^17.2.2",
```
sago2k8: I don't think this is necessary, I would rather find it confusing, no ?

Ref: https://bun.com/docs/runtime/env

frontend/playwright.enterprise.config.ts
sago2k8, line 1: why is there a different config ? can we add an issue to review this ?

sago2k8 (commented):
Left couple of comments.

frontend/playwright.enterprise.config.ts
malinskibeniamin: We have different test config for playwright enterprise edition - happy to refactor in a follow-up PR

frontend/bun.lock
malinskibeniamin: We need this because playwright requires it, after upgrading the playwright package it seems we need to explicitly add dotenv, otherwise the env vars cannot be populated.

.github/workflows/frontend-verify.yml
malinskibeniamin: We are trigger CI for every branch that has an open PR or any commit that gets merged into the master branch.

This change ensures that we are going to only run CI jobs once though, previously there were 2 CI job runs for no reason.

frontend/src/utils/protobuf-reflection.test.ts
malinskibeniamin: This is just a test

frontend/playwright.config.ts
malinskibeniamin: This makes the CI tests for Playwright faster, since the machine we use for GH Actions can easily accommodate at least 4 workers to run tests in parallel. All of our e2e tests are isolated, so we can run them in parallel.

frontend/bunfig.toml
```diff
@@ -2,3 +2,19 @@
...
 print = "yarn"
+
+[test]
+preload = ["./happydom.ts", "./testing-library.ts"]
+timeout = 15000
+coveragePathIgnorePatterns = [
+  "**/*.spec.ts",
+  "**/*.test.ts",
+  "src/utils/**",
+  "src/components/redpanda-ui/**",
```
malinskibeniamin, line 13: We are ignoring these components because we test them in the UI registry.

frontend/package.json
```diff
@@ -17,18 +17,15 @@
...
-    "linter": "biome ci src/*",
-    "lint:check": "biome ci src/*",
+    "lint:fix": "biome check src/* --write --unsafe",
+    "lint": "biome ci src/*",
     "format:check": "biome format src/*",
     "format": "biome format src/* --write",
     "load-git-hooks": "cd .. && git config core.hooksPath \"./.git-hooks\"",
-    "analyze": "RSDOCTOR=true npm run build"
+    "analyze": "RSDOCTOR=true npm run build",
+    "postinstall": "node scripts/fix-remark-emoji.js"
```
malinskibeniamin, line 30: We need the `remark-emoji` package to be installed with the correct version, hence the `postinstall` script.

frontend/package.json
```diff
@@ -17,18 +17,15 @@
     "e2e-test": "playwright test tests/console/ -c playwright.config.ts",
     "e2e-test-enterprise": "playwright test tests/console-enterprise/ -c playwright.enterprise.config.ts",
     "e2e-test:ui": "playwright test tests/console/ -c playwright.config.ts --ui",
-    "test": "TZ=GMT vitest run",
-    "test:coverage": "TZ=GMT vitest run --coverage",
-    "test:ui": "TZ=GMT vitest --ui",
-    "test:watch": "TZ=GMT vitest --watch --ui",
```
eblairmckee, line 23: nit: should probably document the bun flags for these somewhere or make aliases for them here, eg: `"test:watch": "bun test --watch"`

PR #1949, opened 2025-10-02
frontend: move back to vitest

malinskibeniamin:
Unfortunately the following errors were blocking us from fully committing to `bun test` runner:
- Filters do not work well (need to specify each path, it's not recursive)
- Tests related to Monaco editor (ESM packages do not play along nicely, and `bunfig.toml` offers limited customization options)
- Tests related to react routers.


frontend/vitest.config.mts
```diff
@@ -0,0 +1,64 @@
...
+      envCompatible({ prefix: ENV_PREFIX }),
+      tsconfigPaths({
+        ignoreConfigErrors: true,
+      }),
+      yaml(),
+    ],
+    resolve: {
+      preserveSymlinks: true,
+    },
+  };
```
malinskibeniamin, line 63: Specifically this setup is hard to achieve in `bunfig.toml`.

PR #1950, opened 2025-10-02
Schema registry - Remove MobX

jvorcak:


frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,189 @@
...
+      });
+
+      if (!response.ok) {
+        throw new Error('Failed to update subject compatibility mode');
+      }
+
+      return response.json();
+    },
+    onSuccess: (_, { subjectName }) => {
+      queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'subjects', subjectName, 'details'] });
```
malinskibeniamin: Should we add `exact: false`?

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,189 @@
...
+      });
+
+      if (!response.ok) {
+        throw new Error('Failed to update global compatibility mode');
+      }
+
+      return response.json();
+    },
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'compatibility'] });
```
malinskibeniamin: Same as other comment - do we want to invalidate on each key with `exact: false`, or do we only invalidate when all keys match?

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,189 @@
...
+        method: 'PUT',
+        headers: {
+          Authorization: `Bearer ${config.jwt}`,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({ compatibility: mode }),
+      });
+
+      if (!response.ok) {
+        throw new Error('Failed to update global compatibility mode');
```
malinskibeniamin: Can we use `ConnectError.from(error)` and then formatErrorMessage with gRPC? I know it's RESTful API this time but we can still format our errors in the same way, I believe it would work

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,189 @@
...
+        },
+      );
+
+      if (!response.ok) {
+        throw new Error(`Failed to fetch schema details for ${subjectName}`);
+      }
+
+      return response.json();
+    },
+    staleTime: 30000, // Cache for 30 seconds
```
malinskibeniamin: Let's use react query utils (create 1 if it doesn't exist) and make the stale time consistent, just create some SHORT/MEDIUM/LONG stale times to reuse across services/endpoints. Then we can adjust in the future as needed

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,189 @@
...
+    refetchOnMount: true,
+    enabled: options?.enabled !== false,
```
malinskibeniamin: I think you can handle `subjectName !== ''` here

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,189 @@
...
+        return null;
+      }
+
+      return data.compatibility ?? null;
+    },
+    refetchOnMount: true,
+  });
+};
+
+export const useSchemaDetailsQuery = (subjectName: string, options?: { enabled?: boolean }) => {
```
malinskibeniamin: You can make `subjectName` optional so then we don't need `|| ''` fallback.

Also I would suggest always using an object so that argument order does not matter.

frontend/src/components/pages/schemas/EditCompatibility.tsx
```diff
@@ -44,97 +51,110 @@ function renderNotConfigured() {
...
-  }
 
-  refreshData(force?: boolean) {
-    api.refreshSchemaCompatibilityConfig();
-    api.refreshSchemaMode();
-    const subjectName = this.props.subjectName ? decodeURIComponent(this.props.subjectName) : undefined;
+    appGlobal.onRefresh = () => {
+      refetchMode();
+      refetchCompatibility();
+      if (subjectName) refetchDetails();
```
malinskibeniamin: I think this is the part I don't quite get - shouldn't we rely on the tanstack query cache, or do we want to force refetch at all times? Just feels like we can remove the refresh logic when MobX decorators don't require it anymore.

frontend/src/react-query/api/schema-registry.tsx
```diff
@@ -0,0 +1,469 @@
...
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'compatibility'], exact: false });
+    },
+    onError: (error) => {
+      const connectError = ConnectError.from(error);
+      return formatToastErrorMessageGRPC({
+        error: connectError,
+        action: 'update',
+        entity: 'global compatibility mode',
+      });
```
malinskibeniamin, line 148: 👏

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,469 @@
...
+    onSuccess: async () => {
+      // Only invalidate subjects list, not mode/compatibility
+      await queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'subjects'], exact: false });
+    },
+    onError: (error) => {
+      const connectError = ConnectError.from(error);
+      return formatToastErrorMessageGRPC({
+        error: connectError,
+        action: 'delete',
+        entity: 'schema',
```
malinskibeniamin: schema subject?

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,469 @@
...
+      return formatToastErrorMessageGRPC({
+        error: connectError,
+        action: 'update',
+        entity: 'subject compatibility mode',
+      });
+    },
+  });
+};
+
+export const useDeleteSchemaMutation = () => {
```
malinskibeniamin: `useDeleteSchemaSubjectMutation`?

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,469 @@
+import { ConnectError } from '@connectrpc/connect';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
```
malinskibeniamin: Maybe explicitly state and add alias that this is tanstack, since we usually may not notice upon 1st sight that this is not Connect/RPC realm, so something like what we do for `frontend/src/react-query/api/topic.tsx`
```tsx
import {
  useQueryClient,
  useMutation as useTanstackMutation,
  useQuery as useTanstackQuery,
} from '@tanstack/react-query';
```

frontend/src/react-query/api/schema-registry.tsx
```diff
@@ -0,0 +1,469 @@
...
+    queryFn: async () => {
+      const response = await fetch(
+        `${config.restBasePath}/schema-registry/subjects/${encodeURIComponent(subjectName)}/versions/all`,
+        {
+          method: 'GET',
+          headers: {
+            Authorization: `Bearer ${config.jwt}`,
+          },
+        }
+      );
```
malinskibeniamin, line 106: If it's getting repetitive, let's create `authenticatedFetch` with 
```tsx
          headers: {
            Authorization: `Bearer ${config.jwt}`,
          },
```
already added

frontend/src/components/pages/schemas/EditCompatibility.tsx
jvorcak: I think this is fixed now

frontend/src/react-query/api/remote-mcp.tsx
```diff
@@ -262,12 +262,11 @@ export const createMCPClientWithSession = async (
   // Create StreamableHTTP transport for HTTP endpoints
   const transport = new StreamableHTTPClientTransport(new URL(serverUrl), {
     fetch: async (input, init) => {
-      const response = await fetch(input, {
+      const response = await authenticatedFetch(input, {
```
malinskibeniamin: I understand the change, but was it intentional to do it in this PR?

jvorcak: I think it was a reaction to this comment https://github.com/redpanda-data/console/pull/1950#discussion_r2414434880

frontend/src/components/pages/schemas/edit-compatibility.tsx
```diff
@@ -44,122 +53,141 @@ function renderNotConfigured() {
...
-    const subjectName = this.props.subjectName ? decodeURIComponent(this.props.subjectName) : undefined;
+  return (
+    <PageContent>
+      <EditSchemaCompatibility
+        onClose={() => {
+          if (subjectName) {
+            appGlobal.historyReplace(`/schema-registry/subjects/${encodeURIComponent(subjectName)}`);
+          } else {
+            appGlobal.historyReplace('/schema-registry');
+          }
```
malinskibeniamin, line 104: Why can't we just do `navigate`?

frontend/src/react-query/api/schema-registry.tsx
malinskibeniamin, line 1: I would call it `schema-registry.tsx`

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+      if (!response.ok) {
+        const errorText = await response.text();
+        throw new Error(errorText || 'Failed to update subject compatibility mode');
+      }
+
+      return response.json();
+    },
+    onSuccess: (_, { subjectName }) => {
+      queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'subjects', subjectName, 'details'], exact: false });
```
malinskibeniamin: Don't we need to await this?

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,460 @@
...
+        const errorText = await response.text();
+        throw new Error(errorText || 'Failed to create schema');
+      }
+
+      return response.json();
+    },
+    onSuccess: (_, { subjectName }) => {
+      queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'subjects'], exact: false });
+      queryClient.invalidateQueries({ queryKey: ['schemaRegistry', 'subjects', subjectName, 'details'], exact: false });
+    },
```
malinskibeniamin: Don't we need to await this?

frontend/src/react-query/api/schema-registry.tsx
```diff
@@ -0,0 +1,460 @@
...
+        error: connectError,
+        action: 'create',
+        entity: 'schema',
+      });
+    },
+  });
+};
+
+export const useValidateSchemaMutation = () =>
+  useTanstackMutation<
```
malinskibeniamin, line 315: Let's handle the `onError` callback here like in the examples above

frontend/src/react-query/api/schema.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+      return response.json();
+    },
+  });
+
+export const useSchemaReferencedByQuery = (subjectName: string, version: number, options?: { enabled?: boolean }) => {
+  return useTanstackQuery<{ schemaId: number; error?: string; usages: { subject: string; version: number }[] }[]>({
+    queryKey: ['schemaRegistry', 'subjects', subjectName, 'versions', version, 'referencedby'],
+    queryFn: async () => {
+      const response = await fetch(
```
malinskibeniamin: Do you need `authenticatedFetch` here as well?

frontend/src/utils/authenticated-fetch.ts
```diff
@@ -0,0 +1,42 @@
...
+ * });
+ * ```
+ */
+export function authenticatedFetch(url: RequestInfo | URL, options?: RequestInit): Promise<Response> {
+  const headers = new Headers(options?.headers);
+
+  // Add Authorization header if JWT is available
+  if (config.jwt) {
+    headers.set('Authorization', `Bearer ${config.jwt}`);
+  }
```
malinskibeniamin: We also have `config.fetch`, are we meant to use it?

frontend/src/react-query/api/schema.tsx
jvorcak: I don't think so, I think this fn is not awaited and there is nothing after invalidateQueries in the `onSuccess`. But I've added it just to be on a safe side.

frontend/src/react-query/api/schema.tsx
jvorcak: The same here as well, I think that in react query - onSuccess can be async, but RQ does not wait for it before considering the mutation “successful”. But I've added it just to be sure.

PR #1952, opened 2025-10-03
feat: exposed ConnectTiles via federated module for use in serverless onboarding experience in Cloudv2

eblairmckee:
exposes the ConnectTiles component via federated modules, as well as component types

### DIFF
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -22,6 +22,7 @@
         "@heroicons/react": "^1.0.6",
         "@hookform/resolvers": "^4.1.3",
         "@modelcontextprotocol/sdk": "^1.18.2",
+        "@module-federation/typescript": "^3.1.5",
         "@monaco-editor/react": "^4.7.0",
         "@primer/octicons-react": "^19.18.0",
         "@redpanda-data/ui": "^4.2.0",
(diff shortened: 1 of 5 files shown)


sago2k8 (approved):
LGTM

frontend/bun.lock
malinskibeniamin, line 1: Let's update `yarn.lock` too

frontend/module-federation.js
```diff
@@ -1,12 +1,14 @@
...
 
-module.exports = {
+const federationConfig = {
   filename: 'embedded.js',
   name: 'rp_console',
 
   exposes: {
     './EmbeddedApp': './src/EmbeddedApp.tsx',
     './injectApp': './src/injectApp.tsx',
+    './connect-tiles': './src/components/pages/rp-connect/onboarding/connect-tiles.tsx',
```
malinskibeniamin, line 10: Do you think we should have a special directory for components exposed in this way? We may be doing a refactor with the file hierarchy/naming convention later and I want to make sure this continues to work.

eblairmckee: hrm, not sure what value that adds, given this file is the single source of truth for what's exposed. but I'm not opposed

frontend/rsbuild.config.ts
```diff
@@ -27,7 +26,6 @@ export default defineConfig({
     pluginYaml(),
     pluginModuleFederation({
       ...moduleFederationConfig,
-      dts: false, // Disable DTS generation to avoid type resolution issues
```
eblairmckee, line 30: this was fixed when I found the missing exported component types

PR #1954, opened 2025-10-03
Increase timeout and use filtering in tests

graham-rp:
The e2e enterprise tests were timing out after the default (30s) timeout, and, on several retries, some of the values we were looking for were getting pushed to the second page of results.

This bumps the timeout to 60s and enters style principal name in the filter box before checking that t in the filter box before checking that the entry exists.

### DIFF
--- a/frontend/tests/console/pages/ACLPage.ts
+++ b/frontend/tests/console/pages/ACLPage.ts
@@ -447,6 +447,7 @@ export class ACLPage {
   async validateListItem(host: string, principal: string) {
     await this.gotoList();
     // Validate that the ACL list item is visible with correct host and principal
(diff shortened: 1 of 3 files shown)


frontend/tests/console/pages/ACLPage.ts
```diff
@@ -447,6 +447,7 @@ export class ACLPage {
   async validateListItem(host: string, principal: string) {
     await this.gotoList();
     // Validate that the ACL list item is visible with correct host and principal
+    await this.page.getByTestId('search-field-input').fill(principal);
```
andresaristizabal, line 450: Nice catch !!

PR #1956, opened 2025-10-06
frontend: Fix MCP inspector null response handling

malinskibeniamin:
As reported by @birdayz for GCP BigQuery processor template, if the text response is `"null"` as a string, we actually attempt to parse it with `JSON.parse(text)`, but it will end up returning `null` due to how JavaScript works
```typescript
JSON.parse("null") === null)
```
and we don't usually render `null` by itself in React.
Now we fall back to the actual response verbatim without parsing if it equates to `null`.
Screenshot 2025-10-06 at 15 33 58

### DIFF
--- a/frontend/src/components/pages/mcp-servers/details/remote-mcp-inspector-tab.tsx
+++ b/frontend/src/components/pages/mcp-servers/details/remote-mcp-inspector-tab.tsx
@@ -363,8 +363,11 @@ export const RemoteMCPInspectorTab = () => {
       serverToolResponse.content.length > 0
     ) {
       const firstContentItem = serverToolResponse.content[0];
-      if (firstContentItem.type === 'text' && firstContentItem.text) {
+      if (firstContentItem.type === 'text' && typeof firstContentItem.text === 'string') {
         // Try to parse text content as JSON for JSONView
+        if (firstContentItem.text === 'null') {
+          return firstContentItem.text;
+        }
         try {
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/mcp-servers/details/remote-mcp-inspector-tab.tsx
```diff
@@ -363,8 +363,11 @@ export const RemoteMCPInspectorTab = () => {
       serverToolResponse.content.length > 0
     ) {
       const firstContentItem = serverToolResponse.content[0];
-      if (firstContentItem.type === 'text' && firstContentItem.text) {
+      if (firstContentItem.type === 'text' && typeof firstContentItem.text === 'string') {
```
malinskibeniamin, line 366: It's best to double check the text is definitely a string even if the type reports it already.

frontend/src/components/pages/mcp-servers/details/remote-mcp-inspector-tab.tsx
```diff
@@ -363,8 +363,11 @@ export const RemoteMCPInspectorTab = () => {
...
         // Try to parse text content as JSON for JSONView
+        if (firstContentItem.text === 'null') {
+          return firstContentItem.text;
+        }
```
malinskibeniamin, line 370: JavaScript being JavaScript

PR #1961, opened 2025-10-07
feat: add shadowlink dataplane API support

c-julin:
## Summary

Add shadowlink dataplane API support and resolve proto namespace conflicts with BSR packages.

## Changes

- Add shadowlink service to admin API client
- Import shadowlink proto types from redpanda/core
- Split shadowlink APIs into console and dataplane packages
- Configure buf to use BSR import paths for redpandadata/core module
- Create separate buf.gen.remote.yaml for remote dependency generation
- Resolve proto namespace conflicts preventing duplicate proto registration

## Technical Details

The changes fix a runtime proto registration conflict where `redpanda/pbgen/options.proto` was being registered twice:
- Once from BSR module: `buf.build/gen/go/redpandadata/core/protocolbuffers/go`
- Once from local generation: `github.com/redpanda-data/console/backend/pkg/protogen`

Solution: Configure buf's managed mode to use BSR package paths for `redpandadata/core` imports, and generate remote dependencies to a separate `protogen-remote` directory.


buf.yaml
```diff
@@ -9,6 +9,7 @@ deps:
   - buf.build/googleapis/googleapis
   - buf.build/grpc-ecosystem/grpc-gateway:v2.24.0
   - buf.build/redpandadata/common
+  - buf.build/redpandadata/core
```
c-julin, line 12: Had to do buf dep update which has updated protovalidate which has gone into v1, the linter is claiming its a breaking change but i dont think it should be breaking as its just a validation rule. we can fix it on an old version but seems like a good thing to do to move it to v1.0.0

backend/pkg/factory/redpanda/redpanda.go
```diff
@@ -165,4 +167,14 @@ type AdminAPIClient interface {
...
+
+	// ShadowLinkService returns a client for the ShadowLinkService of the Admin API V2.
+	// This provides access to shadow link operations:
+	//   - CreateShadowLink: Creates a new shadow link
+	//   - DeleteShadowLink: Deletes an existing shadow link
+	//   - GetShadowLink: Retrieves details of a specific shadow link
+	//   - ListShadowLinks: Lists all shadow links
+	//   - UpdateShadowLink: Updates an existing shadow link
+	//   - FailOver: Fails over a shadow link or single shadow topic
+	ShadowLinkService(opts ...connect.ClientOption) adminv2connect.ShadowLinkServiceClient
```
c-julin, line 179: This implementation i might change, if anyone has any ideas/opinions happy to hear it out

proto/redpanda/api/console/v1alpha1/shadowlink.proto
```diff
@@ -0,0 +1,137 @@
...
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_EDIT
+      api: API_REDPANDA_ADMIN
+    };
+  }
+
+  rpc GetShadowLink(core.admin.v2.GetShadowLinkRequest) returns (GetShadowLinkResponse) {
+    option (google.api.http) = {get: "/v1alpha1/shadowlink/{name}"};
```
malinskibeniamin, line 57: why not id?

proto/redpanda/api/console/v1alpha1/shadowlink.proto
malinskibeniamin, line 1: will we need GetShadowLinkByCluster (or vice versa) to track the original cluster/fallback?

proto/redpanda/api/console/v1alpha1/shadowlink.proto
c-julin: rp admin api gets by shadowlink name there is a seperate shadowlink id which we cant get by yet

proto/redpanda/api/console/v1alpha1/shadowlink.proto
c-julin: I think this is something we'll need to add once multiple shadowlinks will be supported, as of now its just the one shadowlink so GetByCluster would return the same. 
From the source cluster we cant retrieve shadowlinked clusters, all operations are from the destination cluster

proto/redpanda/api/dataplane/v1alpha3/shadowlink.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+
+package redpanda.api.dataplane.v1alpha3;
+
+import "google/api/annotations.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "redpanda/api/auth/v1/authorization.proto";
+import "redpanda/core/admin/v2/shadow_link.proto";
+
+message FailOverResponse {}
```
malinskibeniamin, line 10: Intentionally left empty?

proto/redpanda/api/console/v1alpha1/shadowlink.proto
```diff
@@ -0,0 +1,137 @@
...
+
+message GetShadowLinkResponse {
+  core.admin.v2.ShadowLink shadow_link = 1;
+}
+
+message ListShadowLinksResponse {
+  repeated core.admin.v2.ShadowLink shadow_links = 1;
+}
+
+message UpdateShadowLinkResponse {}
```
malinskibeniamin, line 22: shouldn't we return the updated shadow link here?

proto/redpanda/api/dataplane/v1alpha3/shadowlink.proto
c-julin: It is, as current response from core is to return back the full shadowlink configuration, the failover operation happens eventually and only way of getting status is by getting shadowlink

proto/redpanda/api/console/v1alpha1/shadowlink.proto
c-julin: same as failover to get latest status it would be better to get shadowlink, in my enhanced endpoitn where i simplify what a shadowlink i will return the shadowlink with the high level state instead of the full configuration and status

proto/redpanda/api/console/v1alpha1/shadowlink.proto
```diff
@@ -0,0 +1,137 @@
...
+  core.admin.v2.ShadowLink shadow_link = 1;
+}
+
+message GetShadowLinkResponse {
+  core.admin.v2.ShadowLink shadow_link = 1;
+}
+
+message ListShadowLinksResponse {
+  repeated core.admin.v2.ShadowLink shadow_links = 1;
+}
```
malinskibeniamin, line 20: Shouldn't this be paginated, or are we leaving it as there will only be 1-1 mapping?

malinskibeniamin (approved):
https://github.com/redpanda-data/console/pull/1961/files#r2444777849 1 comment

proto/redpanda/api/console/v1alpha1/shadowlink.proto
c-julin: leaving it for now in the mvp guise as its 1-1 but will add it in, in my enhanced version of the api

proto/redpanda/api/console/v1alpha1/shadowlink.proto
```diff
@@ -0,0 +1,137 @@
...
+        value: {
+          description: "Shadow link created"
+          schema: {
+            json_schema: {ref: ".redpanda.api.console.v1alpha1.CreateShadowLinkResponse"}
+          }
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_EDIT
```
weeco, line 51: @c-julin Possibly this should be an admin operation given the existing scheme of permissions

c-julin: Updating it on my new branch

PR #1962, opened 2025-10-07
Loosen validation on secret update values

graham-rp:
In order to support updating labels/scopes on secrets without needing to set a new value, we need to remove some validation from the proto definition. This is the middle commit between a console-enterprise [commit](https://github.com/redpanda-data/console-enterprise/pull/638) to support metadata-only updates and an update to the zod validation on console's frontend.


proto/redpanda/api/dataplane/v1/secret.proto
```diff
@@ -182,8 +182,7 @@ message UpdateSecretRequest {
   // The secret data. Must be Base64-encoded.
   bytes secret_data = 4 [
     (google.api.field_behavior) = INPUT_ONLY,
-    (google.api.field_behavior) = REQUIRED,
-    (buf.validate.field).bytes.min_len = 1,
+    (buf.validate.field).bytes.min_len = 0,
```
graham-rp, line 186: This is the only non-generated change

malinskibeniamin (approved):

weeco (approved):

proto/redpanda/api/dataplane/v1/secret.proto
weeco: NIT: We could have changed to `(google.api.field_behavior) = OPTIONAL,`

PR #1969, opened 2025-10-09
frontend: introduce controlplane transport

malinskibeniamin:
**This PR:**
- ensures Console has a way of calling controlplane services with proper type safety instead of calling fetch API directly
- provides service account service via MobX as well as `connect-query` hooks
- provides role binding service via MobX as well as `connect-query` hooks
- sets npm registry properly to be able to install buf registry dependencies (see `.npmrc`)
- renames existing transport to `dataplaneTransport` to distinguish from `controlplaneTransport`
- bumps `@bufbuild` dependencies to ensure compatibility with Redpanda Connect buf registry items (they required `@bufbuild/protobuf` @ v2.7.0+)
- adds transports to config for easier use (although this global config might go away 1 day so it is just done for ease of use)
- registers some controlplane service clients

**Disclaimer:**
- `getControlplanePath` needs improving, it should ideally come from Cloud UI so it does not need to be inferred from `window.location.hostname`
- when building features, one must be careful to ensure the components that require controlplane API need to be wrapped in either a `TransportProvider` with `controlplaneTransport` or use MobX equivalent transport.
- this change has increased bundle size a lot, namely:

**BEFORE**
Total:   38559.1 kB   13277.8 kB (gzip)

**AFTER**
Total:   40490.8 kB   13618.7 kB (gzip)

Screenshot 2025-10-10 at 00 09 56
Screenshot 2025-10-10 at 00 10 07

### DIFF
--- a/frontend/src/react-query/api/controlplane/service-account.tsx
+++ b/frontend/src/react-query/api/controlplane/service-account.tsx
@@ -1,0 +1,189 @@
+import {
+  type ListRoleBindingsRequest_Filter,
+  ListRoleBindingsRequestSchema,
+} from '@buf/redpandadata_cloud.bufbuild_es/redpanda/api/iam/v1/role_binding_pb';
+import {
+  type GetServiceAccountCredentialsRequest,
+  type GetServiceAccountCredentialsResponse,
+  type GetServiceAccountRequest,
+  type GetServiceAccountResponse,
+  type ListServiceAccountsRequest,
+  ListServiceAccountsRequestSchema,
+  type ListServiceAccountsResponse,
+  ServiceAccountService,
+} from '@buf/redpandadata_cloud.bufbuild_es/redpanda/api/iam/v1/service_account_pb';
+import { listRoleBindings } from '@buf/redpandadata_cloud.connectrpc_query-es/redpanda/api/iam/v1/role_binding-RoleBindingService_connectquery';
+import {
+  createServiceAccount,
+  getServiceAccount,
+  getServiceAccountCredentials,
+  listServiceAccounts,
+  rotateServiceAccountSecret,
+  updateServiceAccount,
+} from '@buf/redpandadata_cloud.connectrpc_query-es/redpanda/api/iam/v1/service_account-ServiceAccountService_connectquery';
+import { create } from '@bufbuild/protobuf';
+import type { GenMessage } from '@bufbuild/protobuf/codegenv2';
+import { Code } from '@connectrpc/connect';
+import { createConnectQueryKey, useMutation, useQuery } from '@connectrpc/connect-query';
+import { useQueryClient } from '@tanstack/react-query';
+import {
+  MAX_PAGE_SIZE,
+  MEDIUM_LIVED_CACHE_STALE_TIME,
+  NO_LIVED_CACHE_STALE_TIME,
+  type QueryOptions,
+} from 'react-query/react-query.utils';
+import { formatToastErrorMessageGRPC } from 'utils/toast.utils';
+
+export const useListServiceAccountsQuery = (
+  input?: ListServiceAccountsRequest,
+  options?: QueryOptions<GenMessage<ListServiceAccountsRequest>, ListServiceAccountsResponse>
+) => {
+  const listServiceAccountsRequest = create(ListServiceAccountsRequestSchema, {
+    pageSize: input?.pageSize ?? MAX_PAGE_SIZE,
+    pageToken: input?.pageToken ?? '',
+  });
+
+  return useQuery(listServiceAccounts, listServiceAccountsRequest, {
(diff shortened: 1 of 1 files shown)


frontend/package.json
```diff
@@ -36,8 +36,9 @@
   "dependencies": {
     "@autoform/react": "^4.0.0",
     "@autoform/zod": "^5.0.0",
-    "@bufbuild/protobuf": "^2.5.2",
-    "@bufbuild/protoc-gen-es": "^2.5.2",
+    "@buf/redpandadata_cloud.connectrpc_query-es": "^2.2.0-20251003155837-1268bf48ede8.1",
```
malinskibeniamin, line 39: We should have a dependabot bumping this regularly to ensure that any services that change have the most up-to-date version of the proto exposed to the console frontend.

frontend/src/react-query/api/controlplane/role-binding.tsx
malinskibeniamin, line 1: did not include delete mutations intentionally.

frontend/.npmrc
malinskibeniamin, line 1: we need this change to be able to properly install `@buf/redpandadata_cloud.connectrpc_query-es`

frontend/package.json
```diff
@@ -36,8 +36,9 @@
...
+    "@bufbuild/protobuf": "^2.9.0",
```
andresaristizabal, line 40: Do you need to update this?

frontend/src/config.ts
```diff
@@ -54,6 +58,22 @@ const getRestBasePath = (overrideUrl?: string) => overrideUrl ?? DEFAULT_API_BAS
 
 export const getGrpcBasePath = (overrideUrl?: string) => overrideUrl ?? getBasePath();
 
+const getControlplanePath = () => {
+  // TODO: Make this part of config from Cloud UI
+  if (window.location.hostname.includes('main')) {
```
andresaristizabal: we can avoid this by pass url from cloud-ui

frontend/src/config.ts
```diff
@@ -54,6 +58,22 @@ const getRestBasePath = (overrideUrl?: string) => overrideUrl ?? DEFAULT_API_BAS
...
+  if (window.location.hostname.includes('preprod')) {
+    return 'https://api.ppd.cloud.redpanda.com';
+  }
+  if (window.location.hostname.includes('cloud.redpanda.com')) {
+    return 'https://api.cloud.redpanda.com';
+  }
+  return 'https://api.ign.cloud.redpanda.com';
+};
+
+export const getControlplaneBasePath = (overrideUrl?: string) => overrideUrl ?? getControlplanePath();
```
andresaristizabal: avoid this, you always call this function with non-value, we should prevent generate controlplane url logic and crash if cloud-ui doesn't provide the URL.

frontend/src/react-query/api/controlplane/role-binding.tsx
andresaristizabal, line 1: Where did you set the control plane transport?

frontend/package.json
malinskibeniamin: Yes. We bump @bufbuild dependencies to ensure compatibility with Redpanda Connect buf registry items (they required @bufbuild/protobuf @ v2.7.0+)

frontend/src/config.ts
malinskibeniamin: Yes, that's the long-term solution, but for now I have this as a temporary measure

frontend/src/react-query/api/controlplane/role-binding.tsx
malinskibeniamin: I didn't - it would have to happen in a component that's currently using this service, of which currently there is none.

frontend/src/config.ts
malinskibeniamin: Done in https://github.com/redpanda-data/cloudv2/pull/23478

frontend/src/react-query/api/controlplane/service-account.tsx
```diff
@@ -0,0 +1,189 @@
...
+        }),
+        exact: false,
+      });
+    },
+    onError: (error) => {
+      if (error.code === Code.PermissionDenied) {
+        return;
+      }
+
+      return formatToastErrorMessageGRPC({
```
jvorcak, line 121: nit: I'm not a huge fan of tight coupling toast notification logic with the hooks. I think it should be up to a controller/component to handle error in their own way.

malinskibeniamin: There is no toast here. Just a message formatting

frontend/.npmrc
sago2k8: I would not add it in here but rather in the bunconfig

PR #1970, opened 2025-10-09
feat: improved schema to derive types directly from rp-connect-schema-full.json

eblairmckee:
This had been bugging me.. basically rewrote the schema parser to: 
1. use types derived from schema itself (previously was hard coded and borrowed from the `node-editor` branch)
2. improved the merge logic when you're adding connectors to an existing yaml config (uses yaml.parseDocument instead of stringified yaml)
3. separated yaml helper into their own file and added tests

### DIFF
--- a/frontend/src/components/pages/rp-connect/utils/yaml.test.ts
+++ b/frontend/src/components/pages/rp-connect/utils/yaml.test.ts
@@ -1,13 +1,9 @@
-import { beforeEach, describe, expect, test } from 'vitest';
+import { describe, expect, test } from 'vitest';
 
 import { getBuiltInComponents, schemaToConfig } from './schema';
 import { configToYaml, mergeConnectConfigs } from './yaml';
 
 describe('yaml utils for creating connect configs', () => {
-  beforeEach(() => {
-    // Clear session storage before each test
-    sessionStorage.clear();
-  });
   describe('YAML spacing for merged components', () => {
     test('should add newline between root-level items when adding cache', () => {
       const builtInComponents = getBuiltInComponents();


frontend/src/components/pages/rp-connect/utils/yaml.ts
```diff
@@ -253,3 +180,113 @@ export const addRootSpacing = (yamlString: string): string => {
...
+      mergeRateLimitResource(doc, newConfigObject);
+      break;
+    case 'root':
+      mergeRootComponent(doc, newConfigObject);
+      break;
+    case 'scanner':
+      mergeScanner(doc, newConfigObject);
+      break;
+    case 'unknown':
+      mergeRootComponent(doc, newConfigObject);
```
malinskibeniamin, line 218: Where do we handle input/output component types?

malinskibeniamin: Or buffer/tracer etc for that matter

frontend/src/components/pages/rp-connect/utils/yaml.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      const newConfig = schemaToConfig(processorSpec, false);
+      if (!newConfig) {
+        throw new Error('Failed to generate processor config');
+      }
+
+      const mergedDoc = mergeConnectConfigs(existingYaml, newConfig);
+      const yamlString = configToYaml(mergedDoc);
+
+      // Should have newlines between root-level keys
+      expect(yamlString).toContain('input:\n  stdin:\n    codec: lines\n\npipeline:');
```
malinskibeniamin, line 55: Consider using snapshots for this test, I think it makes sense

frontend/src/components/pages/rp-connect/utils/yaml.test.ts
```diff
@@ -0,0 +1,62 @@
+import { beforeEach, describe, expect, test } from 'vitest';
+
+import { getBuiltInComponents, schemaToConfig } from './schema';
+import { configToYaml, mergeConnectConfigs } from './yaml';
+
+describe('yaml utils for creating connect configs', () => {
+  beforeEach(() => {
+    // Clear session storage before each test
+    sessionStorage.clear();
```
malinskibeniamin: Why do we need to?

frontend/src/components/pages/rp-connect/utils/yaml.ts
eblairmckee: input/output are handled by the default case

eblairmckee: actually all are default
Screenshot 2025-10-09 at 4 58 01 PM

eblairmckee: so technically root should fall under default as well

frontend/src/components/pages/rp-connect/utils/yaml.test.ts
eblairmckee: hrm. is that a vitest feature?

frontend/src/components/pages/rp-connect/utils/yaml.test.ts
eblairmckee: ah, can remove

frontend/src/components/pages/rp-connect/utils/yaml.ts
eblairmckee: oh wait. nvm this is telling you where to inject a component, so if it's not within another configuration, it is root.

PR #1972, opened 2025-10-10
Integrated Mo's Feedback for the RPCN user journey in the new onboarding experience

eblairmckee:
Key Changes

##  1. UX Improvements

  - Cancel Button: Added to wizard footer (left side, next to Previous) essentially like a Dialog "close" button
    - Uses onCancel prop if provided, otherwise navigates to /connect-clusters
  - Previous Button: Added chevron icon for better visual hierarchy
  - Custom Component: Added "Custom" tile to skip wizard and go directly to YAML editor (Mo request, to fulfill superuser needs who may want to create a hand-written yaml config and supports legacy UX)
  - Session Reset Hook: Created useResetWizardSessionStorage to centralize cleanup logic

##  2. Session Storage Management Fix

  Problem: Session storage was updated immediately when clicking a connector tile, causing stale data to persist when users clicked "Custom" or navigated back. This was an anti-pattern and inconsistent with the rest of the wizard steps.

  Solution:
  - Session storage now only updates after validation passes when "Next" is clicked
  - Listed `onChange` callback to wizard level, given we want to know what changed in the serverless experience (in cloud). Instead, we now expose the validated form value from the connect-tiles at the wizard level, and surface that to serverless cloud via a `useRef` handler (like every other step)
  - ConnectTiles Form state resets properly when controlled (default values correctly reset based on props) now that it's a controlled form 

##  3. Minor Fixes & Refinements

  - Choicebox: Minor styling/behavior improvements
  - Typography: Small adjustments
  - Schema Utils: Enhanced component handling
  - Step Components: Cleaned up add-topic and add-user step implementations
  - Pipelines Pages: Minor refinements to create/list pages
  - Module federation dependencies: was missing clsx and motion

### DIFF
--- a/frontend/module-federation.js
+++ b/frontend/module-federation.js
@@ -8,7 +8,6 @@ module.exports = {
     './EmbeddedApp': './src/embedded-app.tsx',
     './injectApp': './src/inject-app.tsx',
     './connect-tiles': './src/components/pages/rp-connect/onboarding/connect-tiles.tsx',
-    './onboarding-wizard': './src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx',
     './config': './src/config.ts',
   },
 
@@ -48,25 +47,13 @@ module.exports = {
       singleton: false,
       requiredVersion: deps['tailwind-merge'],
     },
-    mobx: {
-      singleton: true,
-      requiredVersion: deps.mobx,
-    },
-    '@tanstack/react-query': {
-      singleton: true,
-      requiredVersion: deps['@tanstack/react-query'],
-    },
-    'react-router-dom': {
-      singleton: true,
-      requiredVersion: deps['react-router-dom'],
-    },
-    'framer-motion': {
+    motion: {
       singleton: false,
-      requiredVersion: deps['framer-motion'],
+      requiredVersion: deps.motion,
     },
-    '@bufbuild/protobuf': {
-      singleton: true,
-      requiredVersion: deps['@bufbuild/protobuf'],
+    clsx: {
+      singleton: false,
+      requiredVersion: deps.clsx,
     },
   },
 };


frontend/module-federation.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
     './EmbeddedApp': './src/embedded-app.tsx',
     './injectApp': './src/inject-app.tsx',
     './connect-tiles': './src/components/pages/rp-connect/onboarding/connect-tiles.tsx',
+    './onboarding-wizard': './src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx',
```
eblairmckee: we're going to be using the wizard directly in the new serverless onboarding experience, so exposing it via federated modules

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
```diff
@@ -239,14 +239,14 @@ export const AddTopicStep = forwardRef<BaseStepRef, AddTopicStepProps>(({ topicL
         <CardTitle>
           <Heading level={2}>Select a topic to send data to</Heading>
         </CardTitle>
-        <CardDescription>
+        <CardDescription className="mt-4">
```
eblairmckee, line 242: reduced CLS when jumping between steps

frontend/src/components/pages/rp-connect/onboarding/connect-tiles.tsx
```diff
@@ -16,28 +16,54 @@ import { Label } from 'components/redpanda-ui/components/label';
...
 import { getAllCategories } from '../utils/categories';
 import { getBuiltInComponents } from '../utils/schema';
 
+const getComponentSummary = (component: ConnectComponentSpec, componentTypeFilter?: ConnectComponentType[]) => {
+  if (component.name === 'redpanda' && componentTypeFilter?.includes('input')) {
+    return 'Add data to a topic on this cluster';
+  }
+  if (component.name === 'redpanda' && componentTypeFilter?.includes('output')) {
+    return 'Read data from a topic on this cluster';
+  }
```
eblairmckee, line 39: schema returns a different description than I'd like for new users, so manually overriding. hopefully this is the only case for that.

frontend/src/components/pages/rp-connect/onboarding/connect-tiles.tsx
```diff
@@ -86,6 +116,14 @@ const searchComponents = (
       return a.name.localeCompare(b.name);
     })
     .filter((component) => {
+      // Always show custom component regardless of type filter
+      if (
+        component.name === CUSTOM_COMPONENT_NAME &&
+        (filters?.types?.includes('input') || filters?.types?.includes('output'))
+      ) {
+        return true;
+      }
```
eblairmckee, line 125: we always want this 'custom' component to show first, essentially it skips the entire wizard and makes it more accessible to power users who want to write their own configs

frontend/src/components/pages/rp-connect/onboarding/connect-tiles.tsx
```diff
@@ -140,13 +179,16 @@ export const ConnectTiles = forwardRef<BaseStepRef, ConnectTilesProps>(
       className,
       tileWrapperClassName,
       title,
+      handleSkip: handleSkipProp,
```
eblairmckee, line 182: custom skip logic for cloud vs console

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -83,46 +97,42 @@ export const ConnectOnboardingWizard = () => {
...
-    (connectionName: string, connectionType: ConnectComponentType) => {
-      setPersistedConnector({
-        output: { connectionName, connectionType },
-        ...(persistedConnector.input && { input: persistedConnector.input }),
-      });
-    },
-    [setPersistedConnector, persistedConnector]
-  );
-
-  const handleNext = async (methods: Steps) => {
```
eblairmckee, line 106: we're now using the same form submission logic as every other step, where you need to make a selection before proceeding. but we've also surfaced a "skip" button which allows you to proceed to the next step with submission

frontend/src/components/redpanda-ui/components/choicebox.tsx
```diff
@@ -18,13 +18,12 @@ export type ChoiceboxItemProps = RadioGroupPrimitive.RadioGroupItemProps & { tes
...
-  <RadioGroupPrimitive.Item {...props}>
+  <RadioGroupPrimitive.Item {...props} className="group">
     <Card
       size={size}
       className={cn(
-        'flex cursor-pointer flex-row items-start justify-between rounded-md p-4 shadow-none transition-all border-2 border-solid !border-border text-left',
-        '[&[data-state="checked"]]:border-selected',
-        '[&[data-state="checked"]]:bg-selected/5',
+        'flex cursor-pointer flex-row items-start justify-between rounded-md p-4 shadow-none transition-all border-2 border-solid text-left',
+        'group-data-[state=checked]:!border-selected',
```
eblairmckee, line 26: this was borked, fixed selected ui treatment to match OG one

frontend/src/components/redpanda-ui/components/typography.tsx
```diff
@@ -28,7 +28,7 @@ const headingVariants = cva('font-semibold !leading-none', {
 const textVariants = cva('leading-[1.5]', {
   variants: {
     variant: {
-      default: '',
+      default: 'text-[1rem]',
```
eblairmckee, line 31: default was insanely small. this will affect other UI's, cc: @malinskibeniamin and @andresaristizabal who are likely using this default. it's much more accessible now

frontend/module-federation.js
```diff
@@ -47,5 +48,25 @@ module.exports = {
       singleton: false,
       requiredVersion: deps['tailwind-merge'],
     },
+    mobx: {
```
malinskibeniamin: Why do we need to share mobx with cloud UI? We don't use mobx there

frontend/module-federation.js
```diff
@@ -47,5 +48,25 @@ module.exports = {
...
+    },
+    'react-router-dom': {
+      singleton: true,
+      requiredVersion: deps['react-router-dom'],
+    },
+    'framer-motion': {
+      singleton: false,
+      requiredVersion: deps['framer-motion'],
+    },
+    '@bufbuild/protobuf': {
```
malinskibeniamin: I'm quite worried about this one. In cloud UI we use protobuf v1, and in console we use protobuf v2. Why is this necessary?

frontend/module-federation.js
```diff
@@ -47,5 +48,25 @@ module.exports = {
       singleton: false,
       requiredVersion: deps['tailwind-merge'],
     },
+    mobx: {
+      singleton: true,
+      requiredVersion: deps.mobx,
+    },
+    '@tanstack/react-query': {
```
malinskibeniamin: Similar to protobuf comment - we have connect query and tamarack query at different versions compared to cloud UI. Ideally we should invest time into upgrading in cloud UI.

frontend/src/components/redpanda-ui/components/choicebox.tsx
malinskibeniamin: Let's add ticket for UI registry

frontend/src/components/redpanda-ui/components/typography.tsx
malinskibeniamin: Let's add ticket for UI registry

frontend/src/components/pages/rp-connect/hooks/use-reset-wizard-session-storage.tsx
```diff
@@ -0,0 +1,19 @@
+import { useSessionStorage } from 'hooks/use-session-storage';
+import { CONNECT_WIZARD_CONNECTOR_KEY, CONNECT_WIZARD_TOPIC_KEY, CONNECT_WIZARD_USER_KEY } from 'state/connect/state';
+
+import type { AddTopicFormData, AddUserFormData, ConnectTilesFormData } from '../types/wizard';
+
+export const useResetWizardSessionStorage = () => {
```
malinskibeniamin, line 6: Is this how we share data with cloud UI? Do you think we should expose zustand store from cloud UI one day? We would then be able to use callbacks from there

frontend/module-federation.js
eblairmckee: I'm actually not entirely sure if we need to expose every dependency of the wizard or not.... given this is the first time we've exposed a component via console module federation (rather than using the injector) and my lack of module federation expertise --who knows. but I will start a thread in team-ux to see if anyone else knows

frontend/src/components/pages/rp-connect/hooks/use-reset-wizard-session-storage.tsx
eblairmckee: hrm, that's not a bad idea... I did something like that via the `Console.tsx` component in cloud-ui + `injector` and it worked nicely. I'll add it to the follow up work

frontend/module-federation.js
eblairmckee: since the last step of the wizard is actually the `pipelines-create` component, which uses a few observables, we do

frontend/module-federation.js
eblairmckee: removing this for now, will rethink... we should never expose pipelines-create via federated modules

frontend/src/components/redpanda-ui/components/choicebox.tsx
eblairmckee: have an action item for myself in a followup to upstream changes from this branch as well as the 2 preceding it into ui-registry

PR #1984, opened 2025-10-15
wip: try fix update fieldmask

birdayz:


tomasz-sadura:
I'm not sure what the PR is about. It don't see any changes related to field_mask as the title says, it's in draft also. I would wait for @birdayz

PR #1986, opened 2025-10-16
Refactor/topic listing

jvorcak:


frontend/src/components/pages/topics/Tab.Messages/common/delete-records-menu-item.tsx
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright 2022 Redpanda Data, Inc.
```
malinskibeniamin: 2025

frontend/src/components/pages/topics/Tab.Messages/types.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright 2022 Redpanda Data, Inc.
```
malinskibeniamin: 2025

frontend/src/components/pages/topics/Tab.Messages/preview-settings.tsx
```diff
@@ -185,151 +184,139 @@ const globHelp = (
...
-
-            <OptionGroup
-              label="Wrapping"
-              onChange={(e) => (uiState.topicSettings.previewDisplayMode = e)}
-              options={{ 'Single Line': 'single', Wrap: 'wrap', Rows: 'rows' }}
-              value={uiState.topicSettings.previewDisplayMode}
-            />
-          </div>
+      <div style={{ marginTop: '1em' }}>
+        <h3 style={{ marginBottom: '0.5em' }}>Settings</h3>
```
malinskibeniamin, line 132: Why not use typography from Chakra (if we are keeping it here)

frontend/src/components/pages/topics/Tab.Messages/preview-settings.tsx
```diff
@@ -185,151 +184,139 @@ const globHelp = (
...
+                  variant="outline"
+                >
+                  Add entry...
+                </Button>
+              </div>
+            )}
+          </Droppable>
+        </DragDropContext>
+
+        {/* <CustomTagList tags={uiState.topicSettings.previewTags} allCurrentKeys={this.props.allCurrentKeys} /> */}
```
malinskibeniamin: Should this remain commented out?

frontend/src/components/pages/topics/Tab.Messages/preview-settings.tsx
```diff
@@ -185,151 +184,139 @@ const globHelp = (
...
-  messageSearch: MessageSearch;
-}> {
-  @computed.struct get allCurrentKeys() {
-    // @ts-expect-error perhaps we still need this due to MobX behavior?
-    const _unused = this.props.messageSearch.messages.length;
-    return getAllMessageKeys(this.props.messageSearch.messages)
-      .map((p) => p.propertyName)
-      .distinct();
-  }
+export const PreviewSettings = observer(({ messageSearch }: { messageSearch: MessageSearch }) => {
```
malinskibeniamin: Looks like we haven't moved away from MobX pattern yet?

frontend/src/components/pages/topics/Tab.Messages/message-display/payload-component.tsx
```diff
@@ -0,0 +1,227 @@
...
+    case 29:
+      return 'GS';
+    case 30:
+      return 'RS';
+    case 31:
+      return 'US';
+    default:
+      return '';
+  }
+}
```
malinskibeniamin: We could move this to a utility function for easier unit testing

frontend/src/components/pages/topics/Tab.Messages/message-display/message-meta-data.tsx
```diff
@@ -0,0 +1,54 @@
...
+    Partition: msg.partitionID,
+    Offset: numberToThousandsString(msg.offset),
+    Key: msg.key.isPayloadNull ? 'Null' : `${titleCase(msg.key.encoding)} (${prettyBytes(msg.key.size)})`,
+    Value: msg.value.isPayloadNull
+      ? 'Null'
+      : `${titleCase(msg.value.encoding)} (${msg.value.schemaId > 0 ? `${msg.value.schemaId} / ` : ''}${prettyBytes(msg.value.size)})`,
+    Headers: msg.headers.length > 0 ? `${msg.headers.length}` : 'No headers set',
+    Compression: msg.compression,
+    Transactional: msg.isTransactional ? 'true' : 'false',
+    // "Producer ID": "(msg.producerId)",
```
malinskibeniamin: Should this be kept?

frontend/src/components/pages/topics/Tab.Messages/message-display/message-key-preview.tsx
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  if (sequentialChars.length > 0) {
+    elements.push(sequentialChars);
+  }
+
+  return elements;
+}
+
+function getControlCharacterName(code: number): string {
```
malinskibeniamin: I think this may be duplicated, but could also be a utility function

frontend/src/components/pages/topics/Tab.Messages/dialogs/save-messages-dialog.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+    const saveMessages = () => {
+      if (!messages) {
+        return;
+      }
+
+      const cleaned = cleanMessages(messages);
+
+      // biome-ignore lint/suspicious/noConsole: intentional console usage
+      console.log(`saving cleaned messages; messages: ${messages.length}`);
```
malinskibeniamin: let's remove this?

frontend/src/components/pages/topics/Tab.Messages/preview-settings.tsx
jvorcak: It'll be hard to do in a first step, we first need to minimize the usage of it's state and only after the last one disappears we can remove the `observer`. I have a plan on this, I'll be happy to explain on a call if needed.

frontend/src/components/pages/topics/Tab.Messages/message-display/payload-component.tsx
jvorcak: We moved it to helpers.tsx for now. We will need to do this in phases as well. For now it sits in a separate file, we can always refactor/reorganise later on after we see we have no regressions and no MobX state left.

frontend/src/components/pages/topics/Tab.Messages/preview-settings.tsx
jvorcak: I try to avoid doing markup changes in this PR, and mostly focus on state so that the scope is minimal (it's too big anyway)

frontend/src/components/pages/topics/Tab.Messages/message-display/message-meta-data.tsx
jvorcak: It's the same in master branch - https://github.com/redpanda-data/console/blob/d510f5e5dd34de378ae56aa404cd433a344a89ab/frontend/src/components/pages/topics/Tab.Messages/index.tsx#L2057

frontend/src/stores/topic-settings-store.ts
```diff
@@ -0,0 +1,444 @@
...
+  consumerPageSize: 20,
+  partitionPageSize: 20,
+  aclPageSize: 20,
+  produceRecordEncoding: PayloadEncoding.TEXT,
+  produceRecordCompression: CompressionType.SNAPPY,
+  quickSearch: '',
+  ...overrides,
+});
+
+export const useTopicSettingsStore = create<TopicSettingsStore>()(
```
malinskibeniamin, line 156: I think you should use https://zustand.docs.pmnd.rs/middlewares/devtools devtools, I explained more in @eblairmckee's PR [here](https://github.com/redpanda-data/console/pull/1990#discussion_r2447446571)

frontend/src/stores/topic-settings-store.ts
```diff
@@ -0,0 +1,444 @@
...
+              return {
+                state: {
+                  perTopicSettings: fullSettings.perTopicSettings,
+                },
+                version: 1,
+              };
+            }
+          } catch (error) {
+            // biome-ignore lint/suspicious/noConsole: intentional console usage for debugging
+            console.error('Error parsing uiSettings-v3:', error);
```
malinskibeniamin: Can we handle it in some other way, or use `console.warn` or `console.debug`?

frontend/src/stores/topic-settings-store.ts
```diff
@@ -0,0 +1,444 @@
...
+    .then(({ uiSettings }) => {
+      // Update MobX's perTopicSettings with Zustand's data
+      // This keeps them in sync so MobX's autorun doesn't overwrite Zustand's changes
+      uiSettings.perTopicSettings.splice(0, uiSettings.perTopicSettings.length, ...state.perTopicSettings);
+    })
+    .catch((error) => {
+      // biome-ignore lint/suspicious/noConsole: intentional console usage for debugging sync errors
+      console.error('Failed to sync Zustand to MobX:', error);
+    });
+});
```
malinskibeniamin: What is the worst side effect this may have? The state being overwritten?

frontend/src/components/pages/topics/Tab.Messages/message-display/message-meta-data.tsx
malinskibeniamin: I understand it may be the same, but are we meant to expose this to the user in some way? If not, let's just delet the comment.

frontend/src/stores/topic-settings-store.ts
```diff
@@ -0,0 +1,444 @@
...
+      },
+
+      getSorting: (topicName: string) => {
+        const topic = get().perTopicSettings.find((t) => t.topicName === topicName);
+        return topic?.searchParams.sorting ?? DEFAULT_SORTING;
+      },
+
+      setSearchParams: (topicName: string, searchParams: Partial<TopicSearchParams>) => {
+        set((state) => {
+          const existingIndex = state.perTopicSettings.findIndex((t) => t.topicName === topicName);
```
eblairmckee: any reason why we use `get().<SELECTOR>` vs `state.<SELECTOR>` interchangeably?

eblairmckee:
@jvorcak I added a few [zustand utils](https://github.com/redpanda-data/console/blob/master/frontend/src/utils/store.ts) in cloud and console to make managing persisted state easier. I'm only using sessionStorage in my work, but we should probably add localStorage utils too. 
I also added a [zustand vitest mock](https://github.com/redpanda-data/console/blob/master/frontend/__mocks__/zustand.ts), which you can use in unit tests. 
And updated [Claude](https://github.com/redpanda-data/console/pull/1990/files#diff-e15e11ddea7a1bee8e30c75ad5e191005f7c811fccfc7fe5201f90bdb7b3aa12R328) directions to reflect these changes (not sure if you're using an agent to do any migrations, but may help?)

frontend/src/stores/topic-settings-store.ts
jvorcak: Let me try to remove this and test carefully. In theory it should be fine w/o it.

frontend/src/stores/topic-settings-store.ts
malinskibeniamin: I see the comment got resolved but I think we still need the `devtools` wrapper. It helps us with debugging via Redux DevTools plugin, it's compatible with it.

frontend/src/components/pages/topics/Tab.Messages/dialogs/save-messages-dialog.tsx
```diff
@@ -0,0 +1,245 @@
...
+        link.click();
+      } else if (format === 'csv') {
+        const csvContent = convertToCSV(cleaned as TopicMessage[]);
+        const link = document.createElement('a');
+        const file = new Blob([csvContent], { type: 'text/csv' });
+        link.href = URL.createObjectURL(file);
+        link.download = 'messages.csv';
+        document.body.appendChild(link); // required in firefox
+        link.click();
+      }
```
malinskibeniamin, line 200: I think this should become a utility function where we can download as CSV or JSON

jvorcak: while I agree, I think we should do it as part of different PR. The scope of this is already too big and this is something that was there for a while (not introduced in this PR)

frontend/src/stores/topic-settings-store.ts
jvorcak: done

frontend/src/stores/topic-settings-store.ts
jvorcak: done

malinskibeniamin (approved):

PR #1989, opened 2025-10-20
fix: improved yaml comments for new RPCN UX

eblairmckee:
## after
Screenshot 2025-10-20 at 11 23 18 AM
Screenshot 2025-10-20 at 11 23 40 AM

## before
Screenshot 2025-10-20 at 11 26 24 AM
Screenshot 2025-10-20 at 11 26 47 AM

### DIFF
--- a/frontend/src/components/pages/rp-connect/types/constants.ts
+++ b/frontend/src/components/pages/rp-connect/types/constants.ts
@@ -18,7 +18,15 @@ export const REDPANDA_TOPIC_AND_USER_COMPONENTS = [
  * Fields that are critical for connection and should always be shown
  * even when they have defaults
  */
-export const CRITICAL_CONNECTION_FIELDS = new Set(['addresses', 'seed_brokers', 'topics', 'topic', 'brokers']);
+export const CRITICAL_CONNECTION_FIELDS = new Set([
+  'addresses',
+  'seed_brokers',
+  'topics',
+  'topic',
+  'brokers',
+  'tls',
+  'consumer_group',
+]);
 
 /**
  * Configuration object fields that should be hidden for REDPANDA_SECRET_COMPONENTS
--- a/frontend/src/components/pages/rp-connect/utils/schema.test.ts
+++ b/frontend/src/components/pages/rp-connect/utils/schema.test.ts
@@ -788,7 +788,8 @@ describe('generateDefaultValue', () => {
       const builtInComponents = getBuiltInComponents();
       const kafkaOutput = builtInComponents.find((c) => c.name === 'kafka' && c.type === 'output');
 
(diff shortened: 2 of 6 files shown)


frontend/src/components/pages/rp-connect/types/constants.ts
```diff
@@ -18,7 +18,15 @@ export const REDPANDA_TOPIC_AND_USER_COMPONENTS = [
...
  */
-export const CRITICAL_CONNECTION_FIELDS = new Set(['addresses', 'seed_brokers', 'topics', 'topic', 'brokers']);
+export const CRITICAL_CONNECTION_FIELDS = new Set([
+  'addresses',
+  'seed_brokers',
+  'topics',
+  'topic',
+  'brokers',
+  'tls',
+  'consumer_group',
```
eblairmckee, line 28: after talking with connect team, these 2 fields are not-explicitly-required for redpanda and kafka components.

frontend/src/components/pages/rp-connect/utils/schema.test.ts
```diff
@@ -788,7 +788,8 @@ describe('generateDefaultValue', () => {
       const builtInComponents = getBuiltInComponents();
       const kafkaOutput = builtInComponents.find((c) => c.name === 'kafka' && c.type === 'output');
 
-      const config = schemaToConfig(kafkaOutput, false) as Record<string, any>;
+      const result = schemaToConfig(kafkaOutput, false);
+      const config = result?.config as Record<string, any>;
```
andresaristizabal, line 792:
```suggestion
      const config = result?.config as Record<string, any>;
      const config = result?.config ?? new Record<string,any>;
```
what do you think handle the possible undefined in config by creating empty Record

frontend/src/components/pages/rp-connect/utils/schema.test.ts
```diff
@@ -825,7 +826,8 @@ describe('generateDefaultValue', () => {
       const builtInComponents = getBuiltInComponents();
       const redpandaInput = builtInComponents.find((c) => c.name === 'redpanda' && c.type === 'input');
 
-      const config = schemaToConfig(redpandaInput, false) as Record<string, any>;
+      const result = schemaToConfig(redpandaInput, false);
+      const config = result?.config as Record<string, any>;
```
andresaristizabal, line 830: same here, we should handle cases for empty or undefined cases

frontend/src/components/pages/rp-connect/utils/schema.test.ts
eblairmckee: In the test? I'd just expect the test to fail. since we want to test the `schemaToConfig` is able to generate a component config for a given schema, we wouldn't want to have a fallback result in the unit test.

PR #1997, opened 2025-10-22
Lazy load schema registry in a publish topic page

jvorcak:
https://www.loom.com/share/b94469c8f83d4fd180c981a5c3d84620

When testing, we need to make sure there is no backwards compatibility issue when sending data that are currently supported (JSON, etc).


malinskibeniamin (approved):
approved now that recent releases are over.

PR #1998, opened 2025-10-23
feat: added event analytics tracking to onboarding wizards

eblairmckee:
Console implementation related to https://github.com/redpanda-data/cloudv2/pull/23597

- uses embedded console pass through of new analytics event handlers
- implemented in both the serverless onboarding wizard and api onboarding wizard
- added "get started serverless user" flag event data to track users that came from the serverless dashboard entry point

TODO:

I'm still waiting on Heap SSO access, so I'm unable to verify that this is working correctly outside watching 200 network requests

### DIFF
--- a/frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
+++ b/frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
@@ -20,7 +20,6 @@ import {
 } from 'state/onboarding-wizard-store';
 import { uiState } from 'state/ui-state';
 import {
-  isAnalyticsEnabled,
   rpcnOnboardingWizardExistingTopicChangeEvent,
   rpcnOnboardingWizardExistingUserChangeEvent,
   rpcnOnboardingWizardInputConnectorChangeEvent,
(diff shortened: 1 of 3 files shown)


frontend/src/embedded-app.tsx
```diff
@@ -70,12 +71,20 @@ export interface EmbeddedProps extends SetConfigArguments {
    */
   isConsoleReadyToMount?: boolean;
   /**
-   * LaunchDarkly feature flags to be used in console UI when in embedded mode.
+   * Callback for Console to track analytics events
+   * @param eventName - The name of the event
+   * @param eventData - Optional additional event data
    */
-  featureFlags?: Record<string, boolean>;
```
malinskibeniamin, line 75: [blocker] How come `featureFlags` got removed?

eblairmckee: because it was duplicated, it's already extended from `SetConfigArguments`

eblairmckee: Screenshot 2025-10-23 at 8 12 32 AM

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -142,6 +153,14 @@ export const ConnectOnboardingWizard = ({
         const connectionName = result?.data?.connectionName;
         const connectionType = result?.data?.connectionType;
 
+        if (isAnalyticsEnabled()) {
+          captureUserEvent?.(rpcnOnboardingWizardInputConnectorChangeEvent, {
+            connectionName,
+            connectionType,
+            isServerlessGetStartedUser,
+          });
+        }
```
malinskibeniamin: Why should we wrap every `captureUserEvent` in `isAnalyticsEnabled`? Can't we do this check in the Analytics provider, so that we avoid duplication? That way it can fail silently if there is no analytics context.

PR #2004, opened 2025-10-24
Some fixes for a2a

birdayz:
# A2A Inspector: Tool Collapse and Artifact Deduplication

This fixes two critical UX bugs in the A2A chat inspector and adds comprehensive test coverage.

## Tool Block Collapse Behavior

Tools were stuck open - users couldn't collapse them. The original implementation used uncontrolled `defaultOpen`, which broke when we needed auto-collapse behavior.

**Solution:** Controlled `open` state synced to `isLastBlock`. Tools spawn open, auto-collapse when no longer last, auto-open when they become last. User can still toggle manually, but `isLastBlock` changes override.

Implementation in `tool-block.tsx`: `useState(true)` + `useEffect` syncing to `isLastBlock`.

## Artifact Duplication

Artifacts appeared twice: once as artifact block (correct), once as text message with sparkles icon (wrong).

**Root cause:** The Vercel AI SDK processes raw `artifact-update` events AND emits their content as `text-delta` events. Two scenarios exist:

1. **Text-delta BEFORE artifact-update**: Agent streams "Creating chart..." followed by chart markdown, then artifact-update arrives
2. **Text-delta AFTER artifact-update**: Artifact-update arrives, THEN SDK emits duplicate text-delta with artifact content

**Solution:** Two-phase deduplication:

In `handleArtifactUpdateEvent`: Check if `activeTextBlock` matches or ends with artifact content. Discard if exact match, strip if suffix match, keep if legitimate pre-artifact text.

In `handleTextDeltaEvent`: Check if text-delta matches most recent artifact. Skip if duplicate.

This preserves legitimate text like "Creating your chart..." while eliminating duplicate artifact markdown.

## Status Update Blocks

Re-applied from previous work: Shows task state transitions (working → completed) as discrete blocks. Detects state changes via `previousTaskState` tracking.

## Test Coverage

Added 7 unit tests in `event-handlers.test.ts`:
- Tool request/response filtering from status messages
- Artifact-update kind filtering
- Exact duplicate detection (text-delta → artifact-update)
- Legitimate text preservation with artifact stripping
- Reverse order (artifact-update → text-delta)

All tests green.

## Files Changed

- `tool-block.tsx`: Controlled collapse with isLastBlock sync
- `status-update-block.tsx`: New component for state transitions
- `chat-message.tsx`: Render status-update blocks, pass isLastBlock to tools
- `event-handlers.ts`: Artifact deduplication logic, status transitions
- `streaming-types.ts`: Add previousTaskState field
- `event-handlers.test.ts`: Comprehensive test suite (334 lines)

+554 lines, -61 lines

### DIFF
--- a/frontend/src/components/pages/agents/details/a2a/chat/hooks/streaming-types.ts
+++ b/frontend/src/components/pages/agents/details/a2a/chat/hooks/streaming-types.ts
@@ -9,8 +9,9 @@
  * by the Apache License, Version 2.0
  */
 
-import type { ChatMessage, ContentBlock } from '../types';
 import type { TaskStateProps } from 'components/ai-elements/task';
+
+import type { ChatMessage, ContentBlock } from '../types';
 
 /**
  * Artifact data structure from raw events


frontend/.gitignore
```diff
@@ -158,3 +158,4 @@ dist
 
 # Gemini CLI
 .gemini/settings.json
+bun.lockb
```
malinskibeniamin: Let's remove this gitignore maybe, we shouldn't use binary locks from bun

Also you may want to run `bun upgrade`

frontend/src/components/pages/agents/details/a2a/chat/hooks/event-handlers.ts
```diff
@@ -377,6 +443,29 @@ export const handleTextDeltaEvent = (
...
+    }
+
+    // If appending would create exact artifact match, skip
+    if (state.activeTextBlock?.type === 'text') {
+      const wouldBe = state.activeTextBlock.text + textDelta;
+      if (wouldBe === artifactText) {
+        return;
+      }
+    }
+  }
```
malinskibeniamin, line 467: As you can see we have a bunch of frontend logic - I would probably keep it to some extent just to make it easier to parse the streamed text, but maybe going forward (in future milestones) we should work on the agent providing some other types of streamed data, for example "reasoning"/"chain-of-thought" etc. Then this code wouldn't be necessary for the most part

frontend/src/components/pages/agents/details/a2a/chat/hooks/streaming-types.ts
```diff
@@ -152,6 +152,10 @@ export type StreamingState = {
    * Latest task state from status updates
    */
   capturedTaskState: ChatMessage['taskState'] | undefined;
+  /**
+   * Previous task state for detecting state transitions
+   */
+  previousTaskState: string | undefined;
```
malinskibeniamin: [nit] task state could technically be `'submitted' | 'working' | 'input-required' | 'completed' | 'canceled' | 'failed' | 'rejected' | 'auth-required' | 'unknown'`, can be taken from `TaskStateProps['state']`

PR #2005, opened 2025-10-24
feat: added improved AI agent guidelines for creating UIs

eblairmckee:
guidlelines were outdated, fixed within UI registry docs: https://github.com/redpanda-data/ui-registry/pull/30 and implemented in consumer repos

### DIFF
--- a/frontend/.claude/UI.md
+++ b/frontend/.claude/UI.md
@@ -1,0 +1,45 @@
+Whenever a user wants to build a UI or requests code examples or documentation for the Redpanda UI Registry (or design system or UI library):
+
+- Invoke MCP: use context7 /websites/redpanda-ui-registry_netlify_app
+- Then apply the Best practices list below.
+
+This file provides guidance for building UIs.
+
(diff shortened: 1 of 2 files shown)


frontend/.claude/UI.md
```diff
@@ -0,0 +1,45 @@
...
+8. Validate that tests pass
+9. Follow consumer repo instructions to build app and ensure no build errors
+10. Run dev server if possible and check for runtime errors
+
+# React Best Practices
+
+- Always use functional React components, never class-based
+- Never cast or type variables as `any`, instead deduce the correct type
+- Don't leave comments or `console.log`s in generated code, keep it clean and production-ready
+- `forwardRef` components when applicable
```
malinskibeniamin, line 24: I wouldn't suggest `forwardRef` actually - it's getting deprecated in React 19 so we would have to refactor afterwards, maybe just remove this line

eblairmckee: for now I think it's pretty important given some of the form submission patterns I'm using. I don't feel like adding more would incur too much tech debt, tbh.

PR #2008, opened 2025-10-30
[UX-605] Add ListConnections to dataplane API

graham-rp:
This adds an endpoint at /v1/monitoring/connections that returns information on open and recently closed kafka connections across the cluster.

Notes:
* An integration test is currently skipped until 2025.3.1 is released
* Integration tests now use a common image (2025.2.10)


backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+			Port:            conn.Source.Port,
+			Id:              conn.ClientId,
+			SoftwareName:    conn.ClientSoftwareName,
+			SoftwareVersion: conn.ClientSoftwareVersion,
+		},
+		Group: &v1.GroupInfo{
+			Id:         conn.GroupId,
+			InstanceId: conn.GroupInstanceId,
+			MemberId:   conn.GroupMemberId,
+		},
```
graham-rp: RFC: these are currently separate in the rpk output, but figured it makes more sense to group this way

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+message ConnectionClient {
+  string ip = 1;
+  uint32 port = 2;
+  string id = 3;
+  string software_name = 4;
+  string software_version = 5;
+}
+
+// Everything is up by one (produce is actually 0, fetch -> 1, etc)
+enum KafkaAPI {
```
graham-rp, line 90: Any thoughts about throwing this in a common place? Seems like we'll use it more than once

backend/pkg/testutil/constants.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2023 Redpanda Data, Inc.
```
c-julin, line 1: 2025  :D 
 dont know how i feel about testutil pkg for a constant thats used only in the api integration tests currently :/ maybe its the name, testconfig more appropriate?

would this be better as an env variable with fallback?

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+		2:  v1.KafkaAPI_KAFKA_API_OFFSETS,
+		3:  v1.KafkaAPI_KAFKA_API_METADATA,
+		4:  v1.KafkaAPI_KAFKA_API_LEADER_AND_ISR,
+		5:  v1.KafkaAPI_KAFKA_API_STOP_REPLICA,
+		6:  v1.KafkaAPI_KAFKA_API_UPDATE_METADATA,
+		7:  v1.KafkaAPI_KAFKA_API_CONTROLLED_SHUTDOWN,
+		8:  v1.KafkaAPI_KAFKA_API_OFFSET_COMMIT,
+		9:  v1.KafkaAPI_KAFKA_API_OFFSET_FETCH,
+		10: v1.KafkaAPI_KAFKA_API_GROUP_COORDINATOR,
+		12: v1.KafkaAPI_KAFKA_API_JOIN_GROUP,
```
c-julin, line 92: skipped a number here, so i think everything from here down is off by 1

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if !ok {
+		return v1.KafkaAPI_KAFKA_API_UNSPECIFIED
+	}
+	return val
+}
+
+func adminConnectionToConnect(conn *adminv2.KafkaConnection) *v1.Connection {
+	// Convert the current requests into our format
+	currentRequests := make([]*v1.ActiveRequests_Request, len(conn.InFlightRequests.SampledInFlightRequests))
```
c-julin: i believe these are all pointers in proto-go  conn.GetInFlightRequests().GetSampledInFlightRequests() to be panic safe

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+	for key, val := range map[string]string{
+		"source.ip_address":                  req.IpAddress,
+		"client_id":                          req.ClientId,
+		"client_software_name":               req.ClientSoftwareName,
+		"client_software_version":            req.ClientSoftwareVersion,
+		"group_id":                           req.GroupId,
+		"authentication_info.user_principal": req.User,
+	} {
+		if val != "" {
+			clauses = append(clauses, fmt.Sprintf("%s = %q", key, val))
```
c-julin, line 238: TIL :D

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	limit := defaultLimit
+	if req.Msg.Limit > 0 {
+		limit = req.Msg.Limit
+	}
+
+	resp, err := adminClient.ClusterService().ListKafkaConnections(ctx, &connect.Request[adminv2.ListKafkaConnectionsRequest]{
+		Msg: &adminv2.ListKafkaConnectionsRequest{
+			Filter:   strings.Join(buildFilterClauses(req.Msg), " AND "),
```
c-julin: what about filter raw do you need to add a check before building the filter clauses? also think it would be cleaner to do the join in the build function then you can return a string and handle filter raw there

backend/pkg/api/connect/service/monitoring/v1/service.go
c-julin: agreed, i think its cleaner

proto/redpanda/api/dataplane/v1/monitoring.proto
c-julin: :old-man-yells-at-protobuf: yeah its a shame. We do have a common.proto, feel like we need a kafka.proto . 
Slight nit on this, i didn't actually know what this enum was for without clicking into the core proto as it says kafka api then gives an enum

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+			OrderBy:  req.Msg.OrderBy,
+			PageSize: int32(limit),
+		},
+	})
+	if err != nil {
+		return nil, apierrors.NewConnectError(
+			connect.CodeInternal,
+			fmt.Errorf("failed to fetch connections: %w", err),
+			apierrors.NewErrorInfo("REASON_REDPANDA_ADMIN_API_ERROR"),
+		)
```
weeco: We have helpers for admin api, kafka api errors etc, e.g.

```go
return nil, apierrors.NewConnectErrorFromRedpandaAdminAPIError(err, "")
```

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	adminClient, err := s.redpandaClientFactory.GetRedpandaAPIClient(ctx)
+	if err != nil {
+		return nil, apierrors.NewConnectError(
+			connect.CodeInternal,
+			fmt.Errorf("failed to get admin API client: %w", err),
+			apierrors.NewErrorInfo("REASON_REDPANDA_ADMIN_API_ERROR"),
+		)
+	}
```
weeco, line 289: Currently the pattern across the codebase is to not wrap this into a specific connect error:

```go
redpandaCl, err := s.redpandaClientProvider.GetRedpandaAPIClient(ctx)
	if err != nil {
		return nil, err
	}
```

We try to use `REASON_REDPANDA_ADMIN_API_ERROR` for errors returned by the admin api, but a failure in the client factory would be an issue with Console. If this reason is set the consuming side could be looking for more details for this specific error (e.g. for the Kafka API we attach error codes & names).

proto/redpanda/api/dataplane/v1/monitoring.proto
weeco: We could start a kafka/api package here https://github.com/redpanda-data/common-go/tree/main/proto if this shall be re-used across rpk and console. It seems this is only a subset of way more Kafka API keys (~100) though. The Apache Kafka repository has the source of truth (JSOn definitions), but I find the one in franz-go easier to find: https://github.com/twmb/franz-go/tree/aea970d4d0d2cc93b0cbe21dd9c5cca35045d49a/generate/definitions

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+  core.admin.v2.KafkaConnectionState state = 1;
+  string ip_address = 2;
+  string client_id = 3;
+  string client_software_name = 4;
+  string client_software_version = 5;
+  string group_id = 6;
+  string user = 7;
+  int64 idle_ms = 8 [(buf.validate.field).int64.gte = 0];
+
+  string filter_raw = 9;
```
weeco: Are these all filters? If so we had our own convention by putting them into a filter object so that it's clearer.

The filter_raw is then an XOR to the other props I guess? Possibly worth to put it at the top by number since we'll likely add more filters in the course of time? Just to avoid something like:

```proto
  int64 idle_ms = 8 [(buf.validate.field).int64.gte = 0];

  // Filter raw is special, deserves a proper comment IMO
  string filter_raw = 9;

  string some_new_filter_added_later = 10;
```

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+  bool has_more_requests = 2;
+}
+
+message GroupInfo {
+  string id = 1;
+  string instance_id = 2;
+  string member_id = 3;
+}
+
+message Connection {
```
weeco: Wondering if we should embed them into the response object. The scope of generic names such as "Connection" is quite large given the package size.

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
+syntax = "proto3";
```
weeco, line 1: For all new properties I think we should comment them more (also for the openapi spec which feeds the docs). I think could be done with AI when pointing to the right sources. Some fields / objects def deserve a comment for more explanation.

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+import "buf/validate/validate.proto";
+import "google/api/annotations.proto";
+import "google/api/field_info.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "redpanda/api/auth/v1/authorization.proto";
+import "redpanda/core/admin/v2/kafka_connections.proto";
+
+message ListConnectionsRequest {
```
weeco, line 36: We should add more validation annotations on the request fields. Some have more known / easy to define bounds I guess. I think the admin api v2 does not use protovalidate and doesn't return as rich errors as we can here?

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+	if !ok {
+		return v1.KafkaAPI_KAFKA_API_UNSPECIFIED
+	}
+	return val
+}
+
+func adminConnectionToConnect(conn *adminv2.KafkaConnection) *v1.Connection {
+	// Convert the current requests into our format
+	currentRequests := make([]*v1.ActiveRequests_Request, len(conn.InFlightRequests.SampledInFlightRequests))
+	for i, req := range conn.InFlightRequests.SampledInFlightRequests {
```
weeco: Wondering if `InFlightRequests` is guaranteed to be non-nil

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+		NodeId:             conn.NodeId,
+		ShardId:            conn.ShardId,
+		Uid:                conn.Uid,
+		State:              conn.State,
+		Authentication:     conn.AuthenticationInfo,
+		OpenTime:           conn.OpenTime,
+		CloseTime:          conn.CloseTime,
+		ConnectionDuration: durationpb.New(duration),
+		IdleDuration:       conn.IdleDuration,
+		TlsEnabled:         conn.TlsInfo.Enabled,
```
weeco: TlsInfo also always non nil?

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+			Id:         conn.GroupId,
+			InstanceId: conn.GroupInstanceId,
+			MemberId:   conn.GroupMemberId,
+		},
+		ListenerName:    conn.ListenerName,
+		TransactionalId: conn.TransactionalId,
+		ApiVersions:     apiVersions,
+		ActiveRequests: &v1.ActiveRequests{
+			Requests:        currentRequests,
+			HasMoreRequests: conn.InFlightRequests.HasMoreRequests,
```
weeco: InflightRequests also worth double confirming if this is never nil or we will immediately have panics. Couldn't inspect the adminv2 types from the github PR view, sorry :D

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+  string ip_address = 2;
+  string client_id = 3;
+  string client_software_name = 4;
+  string client_software_version = 5;
+  string group_id = 6;
+  string user = 7;
+  int64 idle_ms = 8 [(buf.validate.field).int64.gte = 0];
+
+  string filter_raw = 9;
+  string order_by = 10;
```
weeco: Not clear how this can be used, needs a comment, possibly some validation too.

proto/redpanda/api/dataplane/v1/monitoring.proto
weeco: Also wondering if we should actually expose it, given this would be inconsistent with the other API endpoints where we aren't doing this.

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,207 @@
...
+	redpandaClientFactory redpandafactory.ClientFactory
+}
+
+// NewService instantiates a new Service
+func NewService(cfg *config.Config, logger *slog.Logger, redpandaClientFactory redpandafactory.ClientFactory) *Service {
+	return &Service{cfg: cfg, logger: logger, redpandaClientFactory: redpandaClientFactory}
+}
+
+func kafkaAPIFromKey(key int32) v1.KafkaAPI {
+	val, ok := map[int32]v1.KafkaAPI{
```
weeco, line 80: NIT: Should probably move to package level to avoid the allocation of the entire map for every call.

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+  KAFKA_API_SYNC_GROUP = 15;
+  KAFKA_API_DESCRIBE_GROUPS = 16;
+  KAFKA_API_LIST_GROUPS = 17;
+  KAFKA_API_SASL_HANDSHAKE = 18;
+  KAFKA_API_API_VERSIONS = 19;
+  KAFKA_API_CREATE_TOPICS = 20;
+  KAFKA_API_DELETE_TOPICS = 21;
+}
+
+message APIVersion {
```
weeco: NIT: Maybe worth adding the actual key of the kafka api too? May also be helpful when we run into kafka api keys that haven't been added yet to the protos. In these cases we would see for example:

```
KafkaAPI: KAFKA_API_UNSPECIFIED
version: 1
```

proto/redpanda/api/dataplane/v1/monitoring.proto
graham-rp: Yep! This isn't exhaustive, so we'd definitely want the full list if we're defining a canonical source. I'll put together a definition in a sec

proto/redpanda/api/dataplane/v1/monitoring.proto
birdayz: agree, the package is not about redpanda in specific, and a type "Connection" is too broad.

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: great catch!

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: Good call - I think this probably gets moved to common-go, so I'll keep that in mind.

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: filter_raw vs the other ones gets handled in the proto, but the more I think about it, the more I think we probably shouldn't expose it: https://github.com/redpanda-data/console/pull/2008/files#diff-a3ba38ddd8fadc94f7229d8908ba0c00c78a00f32b7ce4a047966ca41377b70eR28

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: I think all of the not-nil checks were based on stuff in practice, but we probably shouldn't depend on that

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: good call - nested definitely _isn't_ safe, but I've added a test and some guards

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+  option (buf.validate.message).cel = {
+    id: "filter_raw_excludes_others"
+    message: "filter_raw must not be present alongside other filtering mechanisms"
+    expression:
+      "!(has(this.filter_raw) && ("
+      "has(this.ip_address) || has(this.client_id) || has(this.client_software_name) ||"
+      "has(this.client_software_version) || has(this.group_id) || has(this.user) || has(this.idle_ms)"
+      "))"
+  };
+}
```
sago2k8, line 87:
```suggestion
message ListConnectionsRequest {
message Filter {
  core.admin.v2.KafkaConnectionState state = 1;
  string ip_address = 2;
  string client_id = 3;
  string client_software_name = 4;
  string client_software_version = 5;
  string group_id = 6;
  string user = 7;
  int64 idle_ms = 8 [(buf.validate.field).int64.gte = 0];
  string raw = 9;  // maybe we can call it expression or something similar.
  option (buf.validate.message).cel = {
    id: "filter_raw_excludes_others"
    message: "filter_raw must not be present alongside other filtering mechanisms"
    expression:
      "!(has(this.filter_raw) && ("
      "has(this.ip_address) || has(this.client_id) || has(this.client_software_name) ||"
      "has(this.client_software_version) || has(this.group_id) || has(this.user) || has(this.idle_ms)"
      "))"
  };
  }
  Filter filter = 1;
  string order_by = 2;
  uint32 limit = 3;

}
```

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,145 @@
...
+  string id = 1;
+  string instance_id = 2;
+  string member_id = 3;
+}
+
+message Connection {
+  int32 node_id = 1;
+  uint32 shard_id = 2;
+  string uid = 3 [(google.api.field_info).format = UUID4];
+  core.admin.v2.KafkaConnectionState state = 4;
```
sago2k8:
```suggestion
KafkaConnectionState state = 4;
```
could we implement our own version of this type, for the public api.

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: update: we do have to support it (including the xor)

PR #2009, opened 2025-10-30
Various UI/UX improvements to onboarding wizards

eblairmckee:
better loading UX, fixed loads of hydration issues from session storage, improved stepper design, cleaned up yaml editor synchronization (will demo later)

### DIFF
--- a/frontend/src/components/pages/rp-connect/onboarding/connect-tile.tsx
+++ b/frontend/src/components/pages/rp-connect/onboarding/connect-tile.tsx
@@ -65,7 +65,7 @@ export const ConnectTile = ({
       {/* padding right to compensate for the absolute position of the logo */}
       <div className="relative flex h-full w-full items-center gap-2 pr-8">
         <div className="flex flex-col gap-1">
-          <InlineCode className="truncate bg-background font-semibold text-md">{component.name}</InlineCode>
+          <InlineCode className="truncate bg-background px-0 py-0 font-semibold text-md">{component.name}</InlineCode>
           <span>
             {component.status && component.status !== 'stable' && component.name !== 'redpanda' && (
               <Badge size="sm" variant="gray">


frontend/src/components/pages/overview/api-connect-wizard.tsx
```diff
@@ -137,16 +134,6 @@ export const APIConnectWizard = () => {
...
-  const { data: topicList } = useLegacyListTopicsQuery(create(ListTopicsRequestSchema, {}), {
-    hideInternalTopics: true,
-    staleTime: LONG_LIVED_CACHE_STALE_TIME,
-    refetchOnWindowFocus: false,
-  });
-  const { data: usersList } = useListUsersQuery(undefined, {
-    staleTime: LONG_LIVED_CACHE_STALE_TIME,
-    refetchOnWindowFocus: false,
-  });
-
```
eblairmckee, line 149: moved these into their respective steps so they get refetched when mounted

frontend/src/components/pages/overview/api-connect-wizard.tsx
```diff
@@ -231,23 +218,27 @@ export const APIConnectWizard = () => {
...
+                      <APIWizardStepper.Step
+                        key={step.id}
+                        of={step.id}
+                        onClick={() => {
+                          if (step.id === APIWizardStep.ADD_DATA) {
+                            navigate('/get-started?type=input');
+                            window.location.reload();
+                          } else {
+                            methods.goTo(step.id);
+                          }
```
eblairmckee, line 230: added logic like this in a few places, where if we've come from "get started" (cloud) we should go back there and hard refresh

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
```diff
@@ -225,78 +234,81 @@ export const AddTopicStep = forwardRef<BaseStepRef<AddTopicFormData>, AddTopicSt
...
+                      disabled={topicOptions.length === 0}
+                      id={CreatableSelectionOptions.EXISTING}
+                      value={CreatableSelectionOptions.EXISTING}
+                    >
+                      Existing
+                    </ToggleGroupItem>
+                    <ToggleGroupItem id={CreatableSelectionOptions.CREATE} value={CreatableSelectionOptions.CREATE}>
+                      New
+                    </ToggleGroupItem>
+                  </ToggleGroup>
```
eblairmckee, line 252: Screenshot 2025-10-30 at 10 23 48 AM

frontend/src/components/pages/overview/api-connect-wizard.tsx
```diff
@@ -202,7 +189,6 @@ export const APIConnectWizard = () => {
     });
   }, []);
 
-  // biome-ignore lint/correctness/useExhaustiveDependencies: shouldn't need to be re-run for any dependency changes
```
malinskibeniamin, line 205: 👏

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -76,7 +74,17 @@ export const ConnectOnboardingWizard = ({
...
+  // Manually trigger rehydration if store hasn't hydrated yet
+  // This handles the case where we navigate back to the wizard after reset
+  useEffect(() => {
+    const store = useOnboardingWizardDataStore;
+    const hasHydrated = store.getState()._hasHydrated;
+
+    if (!hasHydrated) {
+      store.persist.rehydrate();
+    }
+  }, []);
```
eblairmckee, line 86: this has been a nightmare of an experience keeping everything synced and hydrated appropriately. lots of quick fixes in this round, but I'd like to rethink this in the future

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -160,6 +158,17 @@ export const ConnectOnboardingWizard = ({
...
+            connectionName,
+            connectionType,
+            showOptionalFields: false,
+            existingYaml: useOnboardingYamlContentStore.getState().yamlContent,
+          });
+
+          if (yamlContent) {
+            useOnboardingYamlContentStore.getState().setYamlContent({ yamlContent });
+          }
+
```
eblairmckee, line 171: moved yaml content into zustand state, so we can update it as we add more connectors, user data, etc.
removed a lot of yaml update logic from `pipelines-create` and moved into wizard step submission.

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -291,7 +314,17 @@ export const ConnectOnboardingWizard = ({
...
+                        icon={
+                          wizardStepDefinitions.findIndex((s) => s.id === step.id) <
+                          wizardStepDefinitions.findIndex((s) => s.id === methods.current.id) ? (
+                            <CheckIcon className="text-white" size={16} />
+                          ) : undefined
+                        }
+                        key={step.id}
+                        of={step.id}
+                        onClick={() => methods.goTo(step.id)}
+                      >
```
eblairmckee, line 327: I didn't add this to `stepper` because I wasn't sure if we wanted this to be the default treatment or not for when a step is checked... sometimes user may want to see previous step numbers?

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -382,7 +413,7 @@ export const ConnectOnboardingWizard = ({
                   )}
                   {!methods.isLast && (
                     <Button disabled={isCurrentStepLoading} onClick={() => handleNext(methods)}>
-                      {isCurrentStepLoading ? 'Loading...' : 'Next'} <ChevronRightIcon />
+                      {isCurrentStepLoading ? <Spinner /> : 'Next'} <ChevronRightIcon />
```
eblairmckee, line 416: going to rethink this... it doesn't look great with CLS. may set a hard width.

frontend/src/components/pages/rp-connect/pipelines-create.tsx
```diff
@@ -460,8 +448,60 @@ export const PipelineEditor = observer(
...
+    // On initial mount in serverless mode, generate YAML from persisted connectors
+    // Wait for hydration to complete before initializing
+    // biome-ignore lint/correctness/useExhaustiveDependencies: Only runs once after hydration, ref prevents re-initialization
+    useEffect(() => {
+      const shouldInitialize = enableRpcnTiles && isServerlessMode && hasHydrated && !hasInitializedServerless.current;
+
+      if (!shouldInitialize) {
+        return;
+      }
+
```
eblairmckee, line 474: if user is coming directly from cloud (get started) and has input/output selected, we need to generate a config. this is kinda of janky.. might make a hook just for this, so it's a little more clear what's going on

frontend/src/components/pages/overview/api-connect-wizard.tsx
```diff
@@ -231,23 +218,27 @@ export const APIConnectWizard = () => {
...
                   <APIWizardStepper.Navigation>
                     {apiWizardStepDefinitions.map((step) => (
-                      <APIWizardStepper.Step key={step.id} of={step.id} onClick={() => methods.goTo(step.id)}>
+                      <APIWizardStepper.Step
+                        key={step.id}
+                        of={step.id}
+                        onClick={() => {
+                          if (step.id === APIWizardStep.ADD_DATA) {
+                            navigate('/get-started?type=input');
+                            window.location.reload();
```
malinskibeniamin: Why do we need to do a hard reload here? Let's add a comment, this should 1 day be removed.

frontend/src/components/pages/overview/api-connect-wizard.tsx
```diff
@@ -257,7 +248,18 @@ export const APIConnectWizard = () => {
               <APIWizardStepper.Controls className="justify-between">
                 <div className="flex gap-2">
                   {!(methods.isFirst || methods.isLast) && (
-                    <Button onClick={methods.prev} type="button" variant="secondary">
+                    <Button
+                      onClick={
+                        methods.current.id === APIWizardStep.ADD_TOPIC
+                          ? () => {
+                              navigate('/get-started?type=input');
+                              window.location.reload();
```
malinskibeniamin: Same as https://github.com/redpanda-data/console/pull/2009/files#r2479193189

frontend/src/components/redpanda-ui/components/button.tsx
```diff
@@ -21,6 +21,7 @@ const buttonVariants = cva(
         ghost: 'hover:bg-accent hover:text-accent-foreground dark:hover:bg-accent/50',
         link: 'text-primary underline-offset-4 hover:underline',
         dashed: 'border-dashed border-2 border-primary',
+        secondaryOutline: 'border !border-secondary text-secondary bg-background shadow-xs hover:bg-secondary/10',
```
malinskibeniamin, line 24: Need to port the change to UI registry

frontend/src/components/pages/rp-connect/pipelines-create.tsx
```diff
@@ -495,10 +535,31 @@ export const PipelineEditor = observer(
...
+          if (debouncedSyncTimerRef.current) {
+            clearTimeout(debouncedSyncTimerRef.current);
+          }
+
+          // Set new timer for 500ms debounce
+          debouncedSyncTimerRef.current = setTimeout(() => {
+            useOnboardingYamlContentStore.getState().setYamlContent({
+              yamlContent: newValue,
+            });
+          }, 500);
```
malinskibeniamin: Any chance we can avoid `setTimeout` here? Maybe let's wrap it in a `useDebounce` hook? 

```tsx

/**
 */

    (...args: Parameters<T>) => {
      if (timeoutRef.current) {
      }

        callback(...args);
      }, delay);
    },
    [callback, delay]
  );

};

```

malinskibeniamin: We can also do it for subsequent PRs

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
malinskibeniamin: I think it is to do with the fact we should:
- not overly rely on global store if possible (I understand in this case it's inevitable)
- have a better provider hoisted at the root of each app which initializes/provides data at the start so that there's no need to worry about hydrating it depending on where the user left off

frontend/src/components/pages/rp-connect/onboarding/add-user-step.tsx
```diff
@@ -50,25 +46,27 @@ import {
...
   topicName?: string;
 }
 
 export const AddUserStep = forwardRef<BaseStepRef<AddUserFormData>, AddUserStepProps & MotionProps>(
-  ({ usersList, defaultUsername, defaultSaslMechanism, topicName, ...motionProps }, ref) => {
+  ({ defaultUsername, defaultSaslMechanism, topicName, ...motionProps }, ref) => {
+    const { data: usersList } = useListUsersQuery(undefined, {
+      refetchOnMount: 'always',
+      staleTime: NO_LIVED_CACHE_STALE_TIME,
+      refetchOnWindowFocus: true,
```
malinskibeniamin: This will be very aggressive, are we sure we don't want to use short/medium cache stale time? It will make the users data refreshing a lot, but I don't think it's that likely the user is being added every few seconds.

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
```diff
@@ -43,17 +47,22 @@ import {
...
-  topicList: Topic[] | undefined;
   defaultTopicName?: string;
 }
 
 export const AddTopicStep = forwardRef<BaseStepRef<AddTopicFormData>, AddTopicStepProps & MotionProps>(
-  ({ topicList, defaultTopicName, ...motionProps }, ref) => {
+  ({ defaultTopicName, ...motionProps }, ref) => {
     const [showAdvancedSettings, setShowAdvancedSettings] = useState(false);
 
+    const { data: topicList } = useLegacyListTopicsQuery(create(ListTopicsRequestSchema, {}), {
```
malinskibeniamin: Good improvement making it happen inside this component

frontend/src/components/pages/overview/api-connect-wizard.tsx
malinskibeniamin: The hard refresh part needs a `FIXME` comment or similar.

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
eblairmckee: I love the idea of your second bullet point. Will very gladly implement that or something similar in a follow. Hoping to be able to share zustand store keys or the store as well via that provider. I did a previous attempt of this, but it collided with mobx and crashed console.

frontend/src/components/pages/overview/api-connect-wizard.tsx
eblairmckee: added comment. not sure why but it 404's whenever you don't hard reload. my thought was because we're going from console (the api connect wizard) back to cloud.

frontend/src/components/redpanda-ui/components/button.tsx
eblairmckee: i have an ongoing branch with these changes

frontend/src/components/pages/rp-connect/pipelines-create.tsx
eblairmckee: was actually really surprised we don't have a debounce hook... will add

frontend/src/components/pages/rp-connect/onboarding/add-user-step.tsx
eblairmckee: few reasons for this... and actually, the aggressive approach isn't really solving it either surprisingly... 
1. if you have a separate tab open and create a user via security > acl's it actually uses a different API than this, so the cache does not get invalidated at the add-user-step level. so if you go back to your original tab, where you have the wizard, it won't refetch to get the user you just added. 
2. same for topics, different API

I'm thinking instead I'll trigger a refetch whenever you open the "existing users" select. (same for topics) and remove all the extra options on the query. that way we always get fresh data on-demand.

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
eblairmckee: few reasons that led to that... originally had it at the parent level because: 
1. I like controlled components (personal preference) and makes them reusable. but given this component already performs side-effects, was incongruous to a truly presentation component
2. seemed to solve some hydration issues... but now that we have a new implementation it's no longer relevant

PR #2010, opened 2025-10-31
feat: fixed hydration issues for onboarding wizards, added consumer group field

eblairmckee:
Context: 
I had used an agent to migrate the query calls in `onboarding-wizard` into each individual step in #2009 however, Claude did a terrible job (and I did a worse one at reviewing) and realized it had added undesired params to query calls (making them fetch too often, etc) as well as added a `useEffect` to refetch ACLs 😞 this reverts all of that and improves both how we fetch data in the new Connect UX, as well as how we refetch the data. 

I kept query calls within their own steps, and invalidate the cache whenever we mutate data. Also added cache invalidation whenever you open a Select that displays existing data (this ensures that if you make a change in a separate tab, it'll show up here, eg: adding a new user in a different tab/window)

Also added a form field for specifying a consumer group. This is relevant when the input is a redpanda component, as you need to either specify an explicit topic partition or a consumer group. Specifically for new users, felt like an easy win to introduce this option field and whether user selects an existing consumer group or creates a new one we will add the appropriate ACLs to that user. 

 See demo: 
https://www.loom.com/share/d39ffecb9d02480ebd349dd6a1c20178

Screenshot 2025-11-03 at 8 31 04 AM
Screenshot 2025-11-03 at 8 33 26 AM

### DIFF
--- a/frontend/src/components/pages/overview/api-connect-wizard.tsx
+++ b/frontend/src/components/pages/overview/api-connect-wizard.tsx
@@ -1,22 +1,18 @@
-import { create } from '@bufbuild/protobuf';
 import { TransportProvider } from '@connectrpc/connect-query';
 import { Markdown } from '@redpanda-data/ui';
 import PageContent from 'components/misc/page-content';
 import { Button } from 'components/redpanda-ui/components/button';
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from 'components/redpanda-ui/components/card';
+import { Spinner } from 'components/redpanda-ui/components/spinner';
 import { defineStepper } from 'components/redpanda-ui/components/stepper';
 import { Heading } from 'components/redpanda-ui/components/typography';
 import { config } from 'config';
 import { useControlplaneTransport } from 'hooks/use-controlplane-transport';
-import { ChevronLeftIcon } from 'lucide-react';
+import { ChevronLeftIcon, ChevronRightIcon } from 'lucide-react';
 import { runInAction } from 'mobx';
-import { ListTopicsRequestSchema } from 'protogen/redpanda/api/dataplane/v1/topic_pb';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useGetOnboardingCodeSnippetQuery } from 'react-query/api/onboarding';
 import { useGetServerlessClusterQuery } from 'react-query/api/serverless';
-import { useLegacyListTopicsQuery } from 'react-query/api/topic';
-import { useListUsersQuery } from 'react-query/api/user';
-import { LONG_LIVED_CACHE_STALE_TIME } from 'react-query/react-query.utils';
 import { useNavigate } from 'react-router-dom';
 import { useAPIWizardStore } from 'state/api-wizard-store';
 import { uiState } from 'state/ui-state';
@@ -139,17 +135,6 @@ export const APIConnectWizard = () => {
   const addTopicStepRef = useRef<BaseStepRef<AddTopicFormData>>(null);
   const addUserStepRef = useRef<BaseStepRef<AddUserFormData>>(null);
 
-  const { data: usersList, refetch: refetchUsers } = useListUsersQuery(undefined, {
-    staleTime: LONG_LIVED_CACHE_STALE_TIME,
-    refetchOnWindowFocus: false,
-  });
-
-  const { data: topicList, refetch: refetchTopics } = useLegacyListTopicsQuery(create(ListTopicsRequestSchema, {}), {
-    hideInternalTopics: true,
-    staleTime: LONG_LIVED_CACHE_STALE_TIME,
-    refetchOnWindowFocus: false,
-  });
-
   const handleNext = async (methods: APIWizardStepperSteps) => {
     switch (methods.current.id) {
       case APIWizardStep.ADD_TOPIC: {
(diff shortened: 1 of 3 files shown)


frontend/src/components/pages/rp-connect/onboarding/add-user-step.tsx
```diff
@@ -236,6 +233,10 @@ export const AddUserStep = forwardRef<BaseStepRef<AddUserFormData>, AddUserStepP
                                 {...field}
                                 className="w-[300px]"
                                 disabled={isLoading}
+                                onChange={(value) => {
+                                  field.onChange(value);
+                                  refetchACLs();
+                                }}
```
eblairmckee, line 351: we not only refetch ACLs when you make a selection

frontend/src/components/pages/rp-connect/onboarding/add-user-step.tsx
```diff
@@ -236,6 +233,10 @@ export const AddUserStep = forwardRef<BaseStepRef<AddUserFormData>, AddUserStepP
...
                                 onOpen={() => refetchUsers()}
                                 options={userOptions}
```
eblairmckee: I had added something similar to both comboboxes in add-topic and add-user steps yesterday, this follows same pattern

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -74,6 +79,17 @@ export const ConnectOnboardingWizard = ({
     [persistedInputConnectionName]
   );
 
+  const { data: usersList, refetch: refetchUsers } = useListUsersQuery(undefined, {
+    staleTime: LONG_LIVED_CACHE_STALE_TIME,
+    refetchOnWindowFocus: false,
+  });
```
eblairmckee: what's annoying is if you have a separate tab open and create a user from the Security route, it won't ever get picked up here because the other UI doesn't use this react query key, so the cache won't get invalidated. But the current workaround is you just reopen the Combobox menu and it'll refetch. Not ideal, but given this is an edge case, deprioritizing for right now.

malinskibeniamin: Just FYI, you can use the same query key as the cloudv2 controlplane, since we have the same API/endpoint available. Alternatively, you could construct the query key by yourself by using the `createConnectQueryKey` method.

frontend/src/components/pages/rp-connect/onboarding/add-user-step.tsx
malinskibeniamin: I would suggest using invalidate whenever possible. If the component needs information about ACLs, it will refetch it on its own where necessary.

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -74,6 +79,17 @@ export const ConnectOnboardingWizard = ({
...
+
+  const { data: topicList, refetch: refetchTopics } = useLegacyListTopicsQuery(create(ListTopicsRequestSchema, {}), {
+    hideInternalTopics: true,
+    staleTime: LONG_LIVED_CACHE_STALE_TIME,
+    refetchOnWindowFocus: false,
+  });
```
malinskibeniamin: Similar to another comment, it's best to invalidate the query in my opinion, but this will also work.

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
eblairmckee: oh, this isn't a concern with cloudv2 control plane parity, more parity with the Security > User/ACLs forms which seem to use a REST API (although it's very hard to tell from the source code given it's obscured by the mobx middleware). Would actually appreciate some more context there, wrt what API(s) we're using for various user/ACL read/write.

frontend/src/components/pages/overview/api-connect-wizard.tsx
```diff
@@ -280,8 +281,18 @@ export const APIConnectWizard = () => {
...
+                      onClick={() => handleNext(methods)}
+                    >
+                      {isCurrentStepLoading ? (
+                        <Spinner />
+                      ) : (
+                        <>
+                          Next <ChevronRightIcon />
+                        </>
+                      )}
                     </Button>
```
eblairmckee, line 296: matches connect wizard. 
only annoying this is loading isn't in sync, which is odd because it's the exact same implementation as the connect wizard. will address in follow.

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -64,25 +64,22 @@ export const ConnectOnboardingWizard = ({
...
-  // Manually trigger rehydration if store hasn't hydrated yet
-  // This handles the case where we navigate back to the wizard after reset
   useEffect(() => {
     const store = useOnboardingWizardDataStore;
-    const hasHydrated = store.getState()._hasHydrated;
-
-    if (!hasHydrated) {
-      store.persist.rehydrate();
-    }
+    store.persist.rehydrate();
```
eblairmckee, line 82: something about the `_hasHydrated` state was flakey. forcing rehydration on component mount is a much cleaner fix.

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
malinskibeniamin: We should be using gRPC exclusively now. We should invest time to moving the REST APIs to gRPC/connect where possible. Not having the same connect query keys can be alleviated by using `createConnectQueryKey` function to store query keys for legacy REST tanstack react query calls (for topics, acls etc). Then you will invalidate query cache properly

malinskibeniamin: That being said, it's not for this PR, it's just for future reference

PR #2011, opened 2025-11-03
Add ListComponents endpoint

tomasz-sadura:


proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,20 @@ message GetPipelineServiceConfigSchemaResponse {
...
+  // Examples: "inputs|outputs", "processors", ".*" (for all)
+  string component_type_regex = 1;
+
+  // List of component fields to skip in the response.
+  repeated string skip_fields = 2;
+}
+
+message GetComponentListResponse {
+  // JSON string containing the lists of available components.
+  string components_json = 1;
```
malinskibeniamin: How much more information can we infer here from the proto, is it just a stringifed JSON? Can it be a more specific map of processors, inputs, outputs and other component types?

We are going to be using `protovalidate` in the future and this would be a great addition

tomasz-sadura: I can return a more structured response if that's helpful

malinskibeniamin: @eblairmckee let us know what else you need, to make your life on the UI easier, I think having better types coming back from frontend will mean you have less heavy lifting/transformations to do.

tomasz-sadura: updated the return schema, does it look ok?

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,103 @@ message GetPipelineServiceConfigSchemaResponse {
   string config_schema = 1;
 }
 
+message GetComponentListRequest {
+  // Regex pattern to filter component types.
+  // Examples: "inputs|outputs", "processors", ".*" (for all)
+  string component_type_regex = 1;
```
birdayz: i suggest to follow the existing pattern for filters: putting into a filter struct: https://github.com/redpanda-data/console/blob/master/proto/redpanda/api/dataplane/v1/topic.proto#L120

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,103 @@ message GetPipelineServiceConfigSchemaResponse {
...
+
+  // List of component fields to skip in the response.
+  repeated string skip_fields = 2;
```
birdayz: this sounds like a read_mask, https://github.com/redpanda-data/cloudv2/blob/main/proto/descriptors/redpanda/iam/v1/user_service.proto#L125

https://google.aip.dev/157

you can then use our fieldmask library in pkg to strip fields not covered by read mask

tomasz-sadura: the idea here was to allow to skip examples which weight a lot. The config can be nested, I think we are not able to build fieldmask to filter it on all levels, e.g. (don't have `examples` here but we can look at `description`):
```
{
  "components": {
    "inputs": [
      {
        "name": "amqp_0_9",
        "type": "input",
        "status": "stable",
        "summary": "Connects to an AMQP (0.91) queue. AMQP is a messaging protocol used by various message brokers, including RabbitMQ.",
        "description": "xxx",
        "categories": [
          "Services"
        ],
        "config": {
          "type": "object",
          "kind": "scalar",
          "children": [
            {
              "name": "urls",
              "type": "string",
              "kind": "array",
              "description": "A list of URLs to connect to. The first URL to successfully establish a connection will be used until the connection is closed. If an item of the list contains commas it will be expanded into multiple URLs.",
              "version": "3.58.0"
            },
            {
              "name": "queue",
              "type": "string",
              "kind": "scalar",
              "description": "An AMQP queue to consume from."
            },
            {
              "name": "queue_declare",
              "type": "object",
              "kind": "scalar",
              "description": "Allows you to passively declare the target queue. If the queue already exists then the declaration passively verifies that they match the target fields.",
              "children": [
                {
                  "name": "enabled",
                  "type": "bool",
                  "kind": "scalar",
                  "description": "Whether to enable queue declaration."
                },
...

Comment (tomasz-sadura) [reply proto/redpanda/api/dataplane/v1/pipeline.proto]: it doesn't look right:
```
["inputs.name", "inputs.type", "inputs.kind", 
"inputs.config.name", "inputs.config.type", "inputs.config.kind", 
"inputs.config.children.name", "inputs.config.children.type", "inputs.config.children.kind", 
"inputs.config.children.children.name", "inputs.config.children.children.type", "inputs.config.children.children.kind", 
...]
```

Comment (tomasz-sadura) [reply proto/redpanda/api/dataplane/v1/pipeline.proto]: I checked the examples actually does not weight so much, so I don't think we need to complicate here. Still users can use read_mask

File: proto/redpanda/api/dataplane/v1/pipeline.proto (hunk 2 above)
Comment (tomasz-sadura) [reply proto/redpanda/api/dataplane/v1/pipeline.proto]: I added read_mask, it covers per-component filtering as well

File: proto/redpanda/api/dataplane/v1/pipeline.proto [hunk 4]
```diff
@@ -650,6 +672,98 @@ message GetPipelineServiceConfigSchemaResponse {
...
+  ParamSpec params = 5;
+  repeated string examples = 6;
+  string version = 7;
+}
+
+// BloblangMethodSpec represents a Bloblang method specification.
+message BloblangMethodSpec {
+  string name = 1;
+  string description = 2;
+  string status = 3;
```
malinskibeniamin, line 763: Shouldn't this be an enum instead of a string?

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,98 @@ message GetPipelineServiceConfigSchemaResponse {
...
+
+message GetComponentListResponse {
+  // Structured component list data instead of JSON string.
+  ComponentList components = 1;
+}
+
+// ComponentList contains the structured component data.
+message ComponentList {
+  string version = 1;
+  string date = 2;
```
malinskibeniamin: shouldn't this be a Timestamp? or is it just stringified date

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -353,6 +353,28 @@ service PipelineService {
     };
   }
 
+  // GetComponentList returns a JSON containing lists of allowed components based on component type filter.
+  rpc GetComponentList(GetComponentListRequest) returns (GetComponentListResponse) {
```
malinskibeniamin: [nit] I think `ListComponents` would be better? Usually we reserve `Get` prefix for obtaining 1 item

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,98 @@ message GetPipelineServiceConfigSchemaResponse {
...
+}
+
+message GetComponentListResponse {
+  // Structured component list data instead of JSON string.
+  ComponentList components = 1;
+}
+
+// ComponentList contains the structured component data.
+message ComponentList {
+  string version = 1;
```
malinskibeniamin: Is there any extra validation that could be set at the proto level? For example I imagine this would be a semver regex

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,98 @@ message GetPipelineServiceConfigSchemaResponse {
...
+  string type = 2;
+  string kind = 3;
+  string description = 4;
+  bool advanced = 5;
+  bool deprecated = 6;
+  bool optional = 7;
+  string default_value = 8;
+  repeated string examples = 9;
+  repeated FieldSpec children = 10;
+  map<string, string> annotated_options = 11;
```
malinskibeniamin: [question] Is this just a metadata with any arbitrary key value items?

proto/redpanda/api/dataplane/v1/pipeline.proto
```diff
@@ -650,6 +672,98 @@ message GetPipelineServiceConfigSchemaResponse {
...
+  string description = 5;
+  repeated string categories = 6;
+  string version = 7;
+  FieldSpec config = 8;
+  repeated string examples = 9;
+  string footnotes = 10;
+}
+
+// FieldSpec represents a field specification.
+message FieldSpec {
```
eblairmckee, line 726: can confirm this includes all fields requested, thank you!

proto/redpanda/api/dataplane/v1/pipeline.proto
tomasz-sadura: looks like date is always empty, I can skip it

proto/redpanda/api/dataplane/v1/pipeline.proto
tomasz-sadura: this is how it looks originally:
```
                "annotated_options": [
                  [
                    "auto",
                    "EXPERIMENTAL: Attempts to derive a codec for each file based on information such as the extension. For example, a .tar.gz file would be c
onsumed with the `gzip/tar` codec. Defaults to all-bytes."
                  ],
                  [
                    "all-bytes",
                    "Consume the entire file as a single binary message."
                  ],
```

I will use repeated `AnnotatedOption` for it:
```
message AnnotatedOption {
  string value = 1;
  string description = 2;
}
```

malinskibeniamin (approved):

PR #2014, opened 2025-11-05
serde: add Buf Schema Registry (BSR) protobuf support

sbhuiyan-buf:
This commit adds support for deserializing protobuf messages that use plain protobuf payloads but use Buf Schema Registry (BSR). Unlike Confluent Schema Registry which embeds schema information in the message payload, BSR stores schema metadata in Kafka record headers.

The BSR wire format uses two headers to identify the schema:
- buf.registry.value.schema.message: The fully-qualified message type
- buf.registry.value.schema.commit: The schema version/commit hash

The implementation includes:

Backend Changes:
- New BSRClient interface and HTTP client for fetching schemas from BSR
- ProtobufBSRSerde for deserializing BSR-encoded messages
- Configuration support for BSR endpoint URL and authentication token
- Integration with existing serde service to automatically detect BSR messages
- Comprehensive unit and integration tests

Frontend Changes:
- Added PAYLOAD_ENCODING_PROTOBUF_BSR enum value
- Updated message preview components to display BSR encoding
- Extended proto definitions and TypeScript interfaces

The BSR serde is prioritized before other protobuf serdes in the deserialization pipeline since BSR messages are identifiable by their specific headers, allowing for efficient format detection.

All existing tests pass and new tests verify BSR deserialization works correctly with dynamic protobuf messages.


backend/BSR_INTEGRATION.md
```diff
@@ -0,0 +1,190 @@
+# Buf Schema Registry (BSR) Integration
```
weeco: Let's remove this file or embed it into the existing protobuf serde docs here: https://github.com/redpanda-data/console/blob/master/docs/features/protobuf.md

backend/pkg/bsr/client.go
```diff
@@ -0,0 +1,270 @@
...
+				return nil, entry.err
+			}
+		} else if time.Since(entry.cachedAt) < cacheTTL {
+			// Return cached success
+			return entry.messageDesc, nil
+		}
+	}
+
+	// Fetch from BSR
+	_, messageDesc, err := c.fetchFromBSR(ctx, messageName, commit)
```
weeco: Without something like singleflight (or my favorite caching library which can also collapse concurrent requests this is potentially called many times in parallel: https://github.com/twmb/go-cache

This is kind of a race condition, so hard to say how often it happens in practice, but I think it's likely to hit because deserializing records works concurrently. So n records using the same schema, will ask for this schema and then all of these goroutines will try to fetch the schema until it's populated in the cache for the next goroutine.

backend/pkg/bsr/client.go
```diff
@@ -0,0 +1,270 @@
...
+// Client is a wrapper around the Buf Schema Registry API for fetching protobuf descriptors.
+type Client struct {
+	cfg    config.BSR
+	logger *slog.Logger
+
+	// Connect RPC client for FileDescriptorSetService
+	fileDescriptorSetClient modulev1connect.FileDescriptorSetServiceClient
+
+	// Cache for file descriptor sets to avoid repeated API calls
+	cache      map[cacheKey]*cacheEntry
```
weeco: NIT: Cache grows without unbounds / no eviction policy. Not sure if schemas can be invalidated/removed or also occupy to much space. We commonly use twmb/go-cache which can also evict cache entries.

backend/pkg/config/config.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	Kafka          Kafka        `yaml:"kafka"`
 	Serde          Serde        `yaml:"serde"`
 	SchemaRegistry Schema       `yaml:"schemaRegistry"`
+	BSR            BSR          `yaml:"bsr"`
```
weeco: This config should live inside the `Serde` struct. Possibly changing the yaml key to `protobufBsr` instead of just `bsr`

backend/pkg/serde/protobuf_bsr.go
```diff
@@ -0,0 +1,148 @@
...
+	messageDescriptor, err := d.bsrClient.GetMessageDescriptor(ctx, messageName, commit)
+	if err != nil {
+		return &RecordPayload{}, fmt.Errorf("failed to get message descriptor from BSR: %w", err)
+	}
+
+	// Create dynamic protobuf message and unmarshal
+	protoMessage := dynamicpb.NewMessage(messageDescriptor)
+	err = proto.Unmarshal(payload, protoMessage)
+	if err != nil {
+		return nil, fmt.Errorf("unable to unmarshal protobuf encoded record: %w", err)
```
weeco: Apologies for this footgun, it's not idiomatic at all to do this, but currently it's required or will cause issues on the caller side. May address this in a different PR.

```suggestion
		return &RecordPayload{}, fmt.Errorf("unable to unmarshal protobuf encoded record: %w", err)
```

backend/pkg/serde/protobuf_bsr.go
```diff
@@ -0,0 +1,148 @@
...
+	// Marshal proto message into JSON
+	o := protojson.MarshalOptions{
+		UseProtoNames:     false, // use lowerCamelCase
+		EmitUnpopulated:   true,
+		EmitDefaultValues: true,
+		Resolver:          files.AsResolver(), // linker.Files.AsResolver() returns protojson.Resolver
+	}
+	jsonBytes, err := o.Marshal(protoMessage)
+	if err != nil {
+		return nil, fmt.Errorf("unable to marshal protobuf message as JSON: %w", err)
```
weeco: Apologies for this footgun, it's not idiomatic at all to do this, but currently it's required or will cause issues on the caller side. May address this in a different PR.

```suggestion
		return &RecordPayload{}, fmt.Errorf("unable to marshal protobuf message as JSON: %w", err)
```

sbhuiyan-buf:
>i haven't completed the review on the serde and the client itself, but I was wondering whether it would make sense to use https://github.com/bufbuild/bsr-kafka-serde-go/tree/main to reduce the maintenance burden.

The redpanda console is a bit different in that it is needing to use the BSR client to get the filedescriptor and use it to then deserialize to JSON to display on a UI. The bsr-kafka-serde-go is hiding that client for its own use to deserialize to the strict application types that application developers are using themselves

weeco:
Thanks for the updates @sbhuiyan-buf , approved and merged.

PR #2016, opened 2025-11-05
Move JS filters to Session storage

jvorcak:
https://redpanda.zoom.us/clips/share/Qlro4VR9RWyzLJnb33pLaA


frontend/src/utils/topic-filters-session.ts
```diff
@@ -0,0 +1,80 @@
...
+interface TopicFiltersMap {
+  [topicName: string]: FilterEntry[];
+}
+
+/**
+ * Get all filters from sessionStorage
+ */
+function getAllFilters(): TopicFiltersMap {
+  try {
+    const stored = sessionStorage.getItem(SESSION_STORAGE_KEY);
```
sago2k8, line 25: Is sessionStorage a global variable ?

frontend/src/utils/topic-filters-session.ts
sago2k8, line 1: could we validate the session storage works in secure environments, I assume it does. But I would like to double check

frontend/src/components/pages/topics/Tab.Messages/common/message-search-filter-bar.tsx
```diff
@@ -10,21 +10,24 @@
...
 import { Box, GridItem, Tag, TagCloseButton, TagLabel } from '@redpanda-data/ui';
-import { observer } from 'mobx-react';
 import type { FC } from 'react';
 import { MdOutlineSettings } from 'react-icons/md';
 
 import type { FilterEntry } from '../../../../../state/ui';
-import { uiState } from '../../../../../state/ui-state';
 
-export const MessageSearchFilterBar: FC<{ onEdit: (filter: FilterEntry) => void }> = observer(({ onEdit }) => {
-  const settings = uiState.topicSettings.searchParams;
```
sago2k8, line 21: I understand the change was to replace the global state by sessions storage, could we use zustand to manage this global state and use the persistence layer https://redpandadata.slack.com/archives/C063X3UEWCA/p1762753166216419

frontend/src/utils/topic-filters-session.ts
jvorcak: It's a globally accessible object we can rely on

frontend/src/components/pages/topics/Tab.Messages/common/message-search-filter-bar.tsx
jvorcak: I don't follow, what is the advantage of this? Session storage is persisted per reloads and only gets erased once user close the tab. Having zustand in between does't really make sense. UX doesn't want to have any kind of persistance here.

frontend/src/utils/topic-filters-session.ts
jvorcak: what do you mean by secure env?

sago2k8: I meant https, I think the browser might require and extra config. it's possible that the `sessionStorage` object handles this for you.

frontend/src/components/pages/topics/Tab.Messages/common/message-search-filter-bar.tsx
sago2k8: Yes my point is if I want to use a global state that happens to be stored in Session Storage, so maybe if there is a known pattern of handling this global state it will be easier to maintain as opposed to a solution only for this page.

PR #2018, opened 2025-11-07
[UX-605] Client connections proxy endpoint

graham-rp:
This implements a straightforward proxy to the adminv2 endpoint rather than reformatting the output like in #2008.

### DIFF
--- a/backend/pkg/api/connect/service/monitoring/v1/service.go
+++ b/backend/pkg/api/connect/service/monitoring/v1/service.go
@@ -12,6 +12,7 @@ package monitoring
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 
 	adminv2 "buf.build/gen/go/redpandadata/core/protocolbuffers/go/redpanda/core/admin/v2"
@@ -50,7 +51,11 @@ func (s *Service) ListKafkaConnections(ctx context.Context, req *connect.Request
 
 	resp, err := adminClient.ClusterService().ListKafkaConnections(ctx, req)
 	if err != nil {
-		return nil, apierrors.NewConnectErrorFromRedpandaAdminAPIError(err, "")
+		// If we've got a valid connect-go err, return it
+		if cerr := new(connect.Error); errors.As(err, &cerr) {
+			return nil, cerr
+		}
+		return nil, apierrors.NewConnectError(connect.CodeInternal, err, nil, nil)
 	}
 
 	return &connect.Response[adminv2.ListKafkaConnectionsResponse]{Msg: resp.Msg}, nil
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -5,8 +5,8 @@ go 1.25.1
 require (
 	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.36.10-20250912141014-52f32327d4b0.1
 	buf.build/gen/go/redpandadata/common/protocolbuffers/go v1.36.10-20250904135917-9feeb2588236.1
-	buf.build/gen/go/redpandadata/core/connectrpc/go v1.19.1-20251106113042-049a26f0436d.2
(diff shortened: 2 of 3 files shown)


proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,39 @@
...
+
+service MonitoringService {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_tag) = {
+    name: "Monitoring"
+    description: "Monitoring operations for Redpanda."
+  };
+
+  rpc ListKafkaConnections(core.admin.v2.ListKafkaConnectionsRequest) returns (core.admin.v2.ListKafkaConnectionsResponse) {
+    option (google.api.http) = {
+      post: "/v1/monitoring/connections"
```
weeco: Should be a GET I assume? Also would possibly further rename to match the original name in case we add more in the future. So something like `/v1/monitoring/kafka/connections` maybe?

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,63 @@
...
+
+	adminClient, err := s.redpandaClientFactory.GetRedpandaAPIClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if req.Msg.PageSize == 0 {
+		req.Msg.PageSize = defaultLimit
+	}
+
```
pgellert: I think I would prefer not to set a default limit here unless we see a strong reason for it. RP Core already sets a default limit of 1000, rpk also sets a default limit. I don't see a reason why the cloud API would need an independent limit.

proto/redpanda/api/dataplane/v1/monitoring.proto
graham-rp: Good call - GET was a bit much when we were looking at expanding the request params, but now it's back to three, so that should be okay.

backend/pkg/api/connect/service/monitoring/v1/service.go
graham-rp: Fair enough. The thinking was that it would violate least astonishment, but that may be more for paginated endpoints.

backend/pkg/api/connect/service/monitoring/v1/service.go
```diff
@@ -0,0 +1,57 @@
...
+	adminClient, err := s.redpandaClientFactory.GetRedpandaAPIClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := adminClient.ClusterService().ListKafkaConnections(ctx, req)
+	if err != nil {
+		return nil, apierrors.NewConnectErrorFromRedpandaAdminAPIError(err, "")
```
weeco: Sorry I missed this previously, but I think this helper for errors won't work for the admin apiv2. Maybe need to double check

proto/redpanda/api/dataplane/v1/monitoring.proto
```diff
@@ -0,0 +1,36 @@
...
+        value: {
+          description: "Recent connections"
+          schema: {
+            json_schema: {ref: ".redpanda.core.admin.v2.ListKafkaConnectionsResponse"}
+          }
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_ADMIN
```
weeco, line 32: Should it be admin?

weeco (approved):

proto/redpanda/api/dataplane/v1/monitoring.proto
graham-rp: oh, good call. I had initially figured it would be, but maybe that's not necessarily true

PR #2021, opened 2025-11-12
frontend: update dependencies and tests

malinskibeniamin:
This PR:
- updates dependencies for AI, Rspack, TypeScript, ConnectRPC, Vitest/Playwright test runners.
- refactors Vitest tests to be split between unit/integration with separate configs for each.
- uses matrix sharding strategy to speed up Vitest integration test suite
- introduces caching for installes dependencies

### DIFF
--- a/frontend/CLAUDE.md
+++ b/frontend/CLAUDE.md
@@ -1,189 +1,233 @@
 # CLAUDE.md
 
-This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+Guidance for Claude Code when working with this repository.
 
 When creating user interfaces, refer to @.claude/UI.md
 
 ## Project Overview
 
-This is the frontend for Redpanda Console, a web application for managing Kafka/Redpanda clusters. Built with React 18.3 + TypeScript using Rsbuild (Rspack-based bundler).
+Frontend for Redpanda Console - web app for managing Kafka/Redpanda clusters. React 18.3 + TypeScript with Rsbuild (Rspack).
 
 ## Commands
 
 ### Package Management
-
-**IMPORTANT**: Always use `bun` as the package manager. After installing packages, run `bun i --yarn` to generate `yarn.lock` for Snyk security tracking.
+**CRITICAL**: Use `bun` as package manager. Run `bun i --yarn` after installing packages (generates `yarn.lock` for Snyk).
 
 ```bash
-bun install              # Install dependencies
-bun i --yarn             # Generate yarn.lock for Snyk tracking
-bun add <package>        # Add a new dependency
-bun add -d <package>     # Add a dev dependency
+bun install                    # Install dependencies
+bun i --yarn                   # Generate yarn.lock
+bun add <package>              # Add dependency
+bun add -d <package>           # Add dev dependency
 ```
 
 ### Development
-
 ```bash
-bun start                # Start dev server on http://localhost:3000
-bun start2 --port=3004   # Start with additional features enabled
+bun start                      # Dev server http://localhost:3000
+bun start2 --port=3004         # With additional features (SSO, REASSIGN_PARTITIONS)
 ```
-
-Dev server proxies API requests to `http://localhost:9090` for `/api`, `/redpanda.api`, `/auth`, `/logout`.
+Dev server proxies `/api`, `/redpanda.api`, `/auth`, `/logout` to `http://localhost:9090`.
 
 ### Testing
-
-**Three-Tier Testing Pyramid:**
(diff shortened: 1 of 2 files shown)


.github/workflows/frontend-verify.yml
```diff
@@ -6,107 +6,235 @@ on:
...
-      - uses: oven-sh/setup-bun@v2
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+      - name: Cache dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.bun/install/cache
+            frontend/node_modules
+          key: ${{ runner.os }}-bun-frontend-${{ hashFiles('frontend/bun.lock') }}
```
eblairmckee, line 38: nice

.github/workflows/frontend-verify.yml
```diff
@@ -6,107 +6,235 @@ on:
...
+      - uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: integration-test-results-${{ matrix.shard }}
+          path: frontend/test-results
+          retention-days: 2
+
   e2e-test:
-    needs: "lint-and-compile"
+    needs: ["lint", "type-check", "build"]
```
eblairmckee, line 175: we do all of these steps earlier, is this sequential? or in parallel?

frontend/CLAUDE.md
```diff
@@ -120,760 +144,193 @@ bun run analyze          # Build with bundle analyzer (RSDOCTOR=true)
...
 ├── hooks/                  # Custom React hooks
 ├── utils/                  # Utility functions
-├── protogen/               # Auto-generated protobuf code (DO NOT EDIT)
+├── protogen/               # Auto-generated protobuf (DO NOT EDIT)
 └── config.ts              # App configuration
 ```
 
 ### Component Guidelines
 
-**Philosophy: Separation of Business Logic and Styling**
```
eblairmckee, line 177: why did we delete all of this?

frontend/src/app.tsx
```diff
@@ -113,10 +109,6 @@ const App = () => {
                 </RequireAuth>
               </ErrorBoundary>
               <ReactQueryDevtools initialIsOpen={process.env.NODE_ENV !== 'production' && developerView} />
-              <StagewiseToolbar
```
eblairmckee, line 116: did we ever use this?

malinskibeniamin: I have but it now became worse because it's pretty much unusable without being a paid version.

frontend/CLAUDE.md
malinskibeniamin: I had a warning with `CLAUDE.md` size being too large and affecting performance. I asked it to rewrite in a concise way without losing existing business logic.

.github/workflows/frontend-verify.yml
malinskibeniamin: We previously had only 1 job doing everything. Now there will be a separate worker for each job. This means e2e test will only run after all 3 of the above jobs have completed (since e2e tests are more expensive)

frontend/CLAUDE.md
eblairmckee: why don't you just make a separate file... we did that with some React best practices here https://github.com/redpanda-data/console/blob/master/frontend/.claude/UI.md

frontend/package.json
```diff
@@ -139,8 +139,6 @@
     "@rsbuild/plugin-yaml": "^1.0.3",
     "@rsdoctor/rspack-plugin": "^1.3.0",
     "@rspack/cli": "^1.5.7",
-    "@stagewise-plugins/react": "^0.4.7",
-    "@stagewise/toolbar-react": "^0.4.7",
```
andresaristizabal, line 144: I didn't know we have this

frontend/CLAUDE.md
andresaristizabal, line 1: Nice!! thank you for adding this

frontend/CLAUDE.md
andresaristizabal: Everything is going to be contextualised. I agree with Ben that we need to keep this compact.

andresaristizabal (approved):

frontend/CLAUDE.md
malinskibeniamin: I made it more concise, now all business logic is preserved but without hitting the threshold/limit for context size

frontend/package.json
malinskibeniamin: Yes but it's now a paid tool pretty much so I removed it.

PR #2027, opened 2025-11-17
 fix(messages): respect backend payload encoding to prevent precision loss

jvorcak:
https://redpanda.zoom.us/clips/share/m7-He5vtRn-NbzrutEwq-Q

Screenshot 2025-11-17 at 21 51 07

Screenshot 2025-11-17 at 21 51 15


frontend/src/state/backend-api.ts
```diff
@@ -2928,15 +2932,16 @@ export function createMessageSearch() {
...
                 m.key.normalizedPayload = key?.normalizedPayload;
 
+                // Only parse JSON payloads with JSONBigInt to preserve large integer precision
                 try {
-                  m.key.payload = JSON.parse(keyPayload);
+                  m.key.payload = JSONBigInt.parse(keyPayload);
                 } catch {
-                  // no op - payload may not be JSON
+                  // no op - payload may not be valid JSON
                 }
```
weeco, line 2940: NIT: We could exclude all encodings that aren't expected to be rendered as JSON so that we don't need to try to parse it. And the same for value encoding if you think this is helpful. This code is somewhat resource intensive and also on the hot path (2x for each sent record), so maybe worth the optimization?


```tsx
  /**
   * Raw text/binary encodings that should not be parsed as JSON.
   * All other encodings are structured formats that the backend converts to JSON.
   */
  const RAW_ENCODINGS = new Set([
    'text',
    'utf8WithControlChars',
    'binary',
    'uint',
    'null'
  ]);

  // Helper function for readability
  const shouldParseAsJson = (encoding: string) => !RAW_ENCODINGS.has(encoding);

  // Parse structured payloads as JSON to preserve large integer precision
  if (shouldParseAsJson(m.key.encoding)) {
    try {
      m.key.payload = JSONBigInt.parse(keyPayload);
    } catch (err) {
      // Payload isn't valid JSON - fallback to raw string
      if (IsDev) {
        console.warn('Failed to parse payload as JSON:', {
          encoding: m.key.encoding,
          error: err,
          payloadPreview: keyPayload.substring(0, 100)
        });
      }
    }
  }
```

frontend/src/state/backend-api.ts
```diff
@@ -22,6 +22,10 @@ import {
   isBakedInTrial,
   prettyLicenseType,
 } from 'components/license/license-utils';
+import JSONBigIntFactory from 'json-bigint';
```
malinskibeniamin, line 25: Is it possible for us to refrain from using this library?

Let's just do it ourselves, something such as
```
const parseWithBigInt = (jsonString, bigNumChecker) =>
    JSON.parse(
        enquoteBigNumber(jsonString, bigNumChecker),
        (key, value) =>
            !isNaN(value) && bigNumChecker(value)
                ? BigInt(value)
                : value
    )
```
I think as long as we have a way to extract the number and get a `BigInt` parsed that should suffice.

malinskibeniamin (approved):

frontend/src/state/backend-api.ts
jvorcak: this introduced some corner-cases, but I'll try to investigate this separately.

PR #2028, opened 2025-11-17
fix: remove ignored branches from frontend verification workflow

andresaristizabal:
This pull request makes a small update to the workflow trigger configuration in `.github/workflows/frontend-verify.yml`. The change removes the list of branches to ignore for workflow runs, simplifying the workflow configuration.


graham-rp (approved):
+1 as soon as biome's happy

PR #2031, opened 2025-11-18
feat[UX-664]: add detailed schema compatibility error messages

c-julin:
## Summary

Previously when protobuf schema validation failed due to compatibility issues, Console only returned a generic boolean (`isCompatible: false`) without explaining why the schema was incompatible. This made it difficult for users to understand what breaking changes were preventing schema updates.

## Changes

**Backend:**
- Added `schemaRegValidationError` struct with `errorType` and `description` fields
- Enabled verbose mode for schema registry compatibility checks using `sr.Verbose` parameter
- Implemented JSON parsing for schema registry messages (handles invalid JSON with unquoted keys)
- Returns structured error details in validation response

**Frontend:**
- Updated TypeScript interfaces to include structured compatibility error object
- Enhanced validation error modal to display formatted error messages
- Error type labels converted from snake_case to human-readable format (e.g., `FIELD_SCALAR_KIND_CHANGED` → `FIELD SCALAR KIND CHANGED`)
- Hides error details section when no meaningful error information is available

## Example

When changing a field type from `string` to `int32`, users now see:

**Error: FIELD SCALAR KIND CHANGED**  
The kind of a SCALAR field at path '#/Car/2' in the new schema does not match its kind in the old schema
Screenshot 2025-11-18 at 10 47 02

Instead of just: `Compatibility issues found`

### DIFF
--- a/backend/pkg/console/schema_registry.go
+++ b/backend/pkg/console/schema_registry.go
@@ -611,7 +611,6 @@ func (s *Service) parseCompatibilityError(messages []string) schemaRegValidation
 	// Iterate through all messages to extract errorType and description
 	data := schemaRegValidationError{}
 	for _, msg := range messages {
-
 		// Schema registry may return invalid JSON with unquoted keys like: {errorType:"...", description:"..."}
 		// Use regex to quote unquoted keys: word: becomes "word":
 		fixedMsg := regexp.MustCompile(`(\w+):`).ReplaceAllString(msg, `"$1":`)
--- a/frontend/src/components/pages/schemas/modals.tsx
+++ b/frontend/src/components/pages/schemas/modals.tsx
@@ -150,12 +150,15 @@ export function openInfoModal(p: {
   });
 }
 
-export function openValidationErrorsModal(result: {
-  isValid: boolean;
-  errorDetails?: string | undefined;
-  isCompatible?: boolean | undefined;
-  compatibilityError?: { errorType: string; description: string } | undefined;
-}) {
+export function openValidationErrorsModal(
+  result: {
+    isValid: boolean;
+    errorDetails?: string | undefined;
+    isCompatible?: boolean | undefined;
+    compatibilityError?: { errorType: string; description: string } | undefined;
+  },
+  onClose?: () => void
+) {
   const { isValid, errorDetails, isCompatible, compatibilityError } = result;
 
   const compatBox = (() => {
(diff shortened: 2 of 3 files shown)


backend/pkg/console/schema_registry.go
```diff
@@ -589,8 +590,46 @@ type SchemaRegistrySchemaValidation struct {
...
+		err := json.Unmarshal([]byte(fixedMsg), &data)
+		if err != nil {
+			continue
+		}
+
+		// Stop once we have either error type or Description we can exit
+		if data.ErrorType != "" || data.Description != "" {
+			break
+		}
+	}
```
weeco, line 631: Wtf for real or is just generated code? I can't imagine the schema registry returning such weird responses, if it does we should add some tests of actual outputs. 

We may use a regex instead to quote unquoted keys, if schema registry really returns weird stuff. If this is an actual issue from the schema registry we should double check that the Redpanda implementation does the same as the reference implementation.

backend/pkg/console/schema_registry.go
```diff
@@ -589,8 +590,46 @@ type SchemaRegistrySchemaValidation struct {
...
+		// Schema registry returns invalid JSON with unquoted keys like: {errorType:"...", description:"..."}
+		// We need to fix this by adding quotes around the keys
+		// Replace: {key: with {"key":
+		fixedMsg := strings.ReplaceAll(msg, "{", "{\"")
+		fixedMsg = strings.ReplaceAll(fixedMsg, ":", "\":")
+		fixedMsg = strings.ReplaceAll(fixedMsg, ", ", ", \"")
+
+		err := json.Unmarshal([]byte(fixedMsg), &data)
+		if err != nil {
+			continue
```
weeco, line 624: I think we should log this issue here then, could be interesting.

backend/pkg/console/schema_registry.go
c-julin: `{
  "is_compatible": false,
  "messages": [
    "{errorType:\"FIELD_SCALAR_KIND_CHANGED\", description:\"The kind of a SCALAR field at path '#/Car/2' in the new schema does not match its kind in the old schema\"}",
    "{oldSchemaVersion: 1}",
    "{oldSchema: 'syntax = \"proto3\";\n\nmessage Car {\n  string make = 1;\n  string model = 2;\n  int32 year = 3;\n}\n\n'}",
    "{compatibility: 'BACKWARD'}"
  ]
}`

This is output from the client call https://github.com/twmb/franz-go/blob/master/pkg/sr/api.go#L749, can update it so it tries to unmarshal, if it fails then we regex

PR #2039, opened 2025-11-20
[UX-678]: fixed new connect UX for kafka connect users

### LINKED ISSUE #1: Add readiness & liveness check
We don't have any endpoints to check readiness & liveness yet.

### LINKED ISSUE #2: Add prometheus metrics
There are no prometheus metrics yet. Interesting metrics can be:
- Request latency bucket (where request-type is a label)
- Number of incoming requests
- Bytes / IN / OUT - (helps to understand how busy the backend is - could be pretty excessive due to large messages or high throughput of messages)

### LINKED ISSUE #3: Create helm chart
Create helm chart for easy deployment into Kubernetes clusters

eblairmckee:
Jira: https://redpandadata.atlassian.net/browse/UX-678

When `enableRpcnTiles` feature flag was true users could no longer access KafkaConnect from the Connect Overview route _when they didn't have any RPCN pipelines created_. The motivation for this decision was to send users directly to the RPCN wizard if they did not have any RPCN pipelines.  

The fix now checks if kafka connect is enabled for the cluster, and if it is, we no longer redirect users to the wizard. Only if kafka connect is _disabled_ AND there are no existing RPCN pipelines will we redirect them to the wizard. 

  ---
  Behavior Matrix: Before vs After

  | Scenario                 | enableRpcnTiles | isKafkaConnectEnabled | Pipelines | MASTER                           | CURRENT                     |
  |--------------------------|-----------------|-----------------------|-----------|----------------------------------|-----------------------------|
  | Incident Case            | ✅ true          | ✅ true                | 0         | 🔴 Redirect to wizard (BLOCKED!) | ✅ Shows both tabs           |
  | New user flow            | ✅ true          | ❌ false               | 0         | 🔴 Redirect to wizard            | ✅ Redirect to wizard        |
  | Existing RP Connect user | ✅ true          | ❌ false               | >0        | Shows both tabs (confusing)      | ✅ Shows only RP Connect tab |
  | Legacy mode              | ❌ false         | ✅ true                | any       | Shows both tabs                  | ✅ Shows both tabs           |
  | Legacy mode              | ❌ false         | ❌ false               | any       | Shows both tabs (confusing)      | ✅ Shows only RP Connect tab |

  ---
  The Three Fixes Explained

  Fix #1: Prevent Blocking Kafka Connect Users

  // BEFORE: Blocks everyone with new UX enabled and no pipelines
  if (isFeatureFlagEnabled('enableRpcnTiles') && pipelinesApi.pipelines?.length === 0)

  // AFTER: Only redirect users who can't use Kafka Connect
  if (isFeatureFlagEnabled('enableRpcnTiles') &&
      pipelinesApi.pipelines?.length === 0 &&
      !this.props.isKafkaConnectEnabled)
  Result: Kafka Connect users now access their tab even with enableRpcnTiles = true

Screenshot 2025-11-20 at 12 53 59 PM

☝️ kafka connect is enabled and enableRpcnTiles is true

  Fix #2: Hide Kafka Connect tabs when not enabled

  // BEFORE: Only checks serverless
  if (isServerless())

  // AFTER: Checks actual availability
  if (isServerless() || !this.props.isKafkaConnectEnabled)
  Result: No confusing "Kafka Connect" tab when the feature isn't available

☝️ edge case that wasn't previously accounted for when on BYOC or Dedicated cluster that _does not_ have kafka connect enabled

  Fix #3: Context-Aware Messaging

  // BEFORE: Based on feature flag state
  {!isFeatureFlagEnabled('enableRpcnTiles') && <Text>Two ways...</Text>}

Screenshot 2025-11-20 at 1 13 49 PM

  // AFTER: Based on actual feature availability
  {this.props.isKafkaConnectEnabled && <Text>Two ways...</Text>}
  Result: "Two ways" message only shows when both options actually exist

Screenshot 2025-11-20 at 1 06 02 PM

  ---
  Incident Resolution

  Before: Kafka Connect customers with enableRpcnTiles enabled would land on /connect-clusters and immediately get redirected to /rp-connect/wizard, unable to access their Kafka Connect
  connectors through the normal UI flow.

  After: Kafka Connect customers see both tabs and can use either Redpanda Connect or Kafka Connect, regardless of feature flag state. The redirect only happens for users who genuinely
  should use the wizard (no Kafka Connect, no pipelines).

### DIFF
--- a/frontend/src/components/pages/connect/overview.tsx
+++ b/frontend/src/components/pages/connect/overview.tsx
@@ -11,12 +11,12 @@
 
 import { create } from '@bufbuild/protobuf';
 import { Badge, Box, DataTable, Stack, Tooltip } from '@redpanda-data/ui';
-import { useQuery } from '@tanstack/react-query';
 import ErrorResult from 'components/misc/error-result';
 import { Link, Text } from 'components/redpanda-ui/components/typography';
 import { WaitingRedpanda } from 'components/redpanda-ui/components/waiting-redpanda';
 import { observer, useLocalObservable } from 'mobx-react';
 import { Component, type FunctionComponent } from 'react';
+import { useKafkaConnectConnectorsQuery } from 'react-query/api/kafka-connect';
 import { Navigate, useLocation } from 'react-router-dom';
 
 import {
@@ -29,16 +29,11 @@ import {
   TaskState,
   TasksColumn,
 } from './helper';
-import { config, isFeatureFlagEnabled, isServerless } from '../../../config';
+import { isFeatureFlagEnabled, isServerless } from '../../../config';
 import { ListSecretScopesRequestSchema } from '../../../protogen/redpanda/api/dataplane/v1/secret_pb';
 import { appGlobal } from '../../../state/app-global';
 import { api, pipelinesApi, rpcnSecretManagerApi } from '../../../state/backend-api';
-import type {
-  ClusterConnectorInfo,
-  ClusterConnectors,
-  ClusterConnectorTaskInfo,
-  KafkaConnectors,
-} from '../../../state/rest-interfaces';
+import type { ClusterConnectorInfo, ClusterConnectors, ClusterConnectorTaskInfo } from '../../../state/rest-interfaces';
 import { Features } from '../../../state/supported-features';
 import { uiSettings } from '../../../state/ui';
 import { Code, DefaultSkeleton } from '../../../utils/tsx-utils';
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/connect/overview.tsx
```diff
@@ -74,15 +82,38 @@ const getDefaultView = (defaultView: string): { initialTab: ConnectView; redpand
...
+        headers: [['Content-Type', 'application/json']],
+      });
+
+      if (!response.ok) {
+        throw new Error('Failed to fetch kafka connect connectors');
+      }
+
+      return response.json();
+    },
+  });
```
malinskibeniamin: Let's use a dedicated custom React hook for this, maybe call it `kafka-connect.tsx`

PR #2040, opened 2025-11-21
UX-637 e2e tests with testcontainers

jvorcak:


frontend/tests/auth.setup.ts
```diff
@@ -3,19 +3,17 @@ import { expect, test as setup } from '@playwright/test';
...
+  // Wait for login form to be visible
+  await expect(page.getByTestId('auth-username-input')).toBeVisible({ timeout: 30000 });
 
   await page.getByTestId('auth-username-input').fill('e2euser');
   await page.getByTestId('auth-password-input').fill('very-secret');
   await page.getByTestId('auth-submit').click();
 
-  await expect(page.getByTestId('versionTitle')).toBeVisible();
+  // Wait for successful login - version title appears in footer
+  await expect(page.getByTestId('versionTitle')).toBeVisible({ timeout: 30000 });
```
malinskibeniamin: Can use `waitFor` instead of `timeout`?

frontend/tests/auth.setup.ts
```diff
@@ -3,19 +3,17 @@ import { expect, test as setup } from '@playwright/test';
...
-  await page.goto('/');
+  await page.goto('/', { waitUntil: 'networkidle' });
 
-  // check if there is an error modal, closed it if needed
-  // const errorOKButton = page.getByTestId('login-error__ok-button')
-  // if(errorOKButton) {
-  //     await errorOKButton.click()
-  // }
+  // Wait for login form to be visible
+  await expect(page.getByTestId('auth-username-input')).toBeVisible({ timeout: 30000 });
```
malinskibeniamin: Same as below, use `waitFor` instead of timeout

frontend/tests/auth.setup.ts
jvorcak: Done

frontend/tests/auth.setup.ts
jvorcak: Done

frontend/biome.jsonc
```diff
@@ -875,11 +875,16 @@
...
       "linter": {
         "rules": {
           "suspicious": {
-            "noConsole": "off"
+            "noConsole": "off",
+            "useAwait": "off",
+            "noEmptyBlockStatements": "off"
+          },
+          "nursery": {
+            "noShadow": "off"
```
malinskibeniamin: I don't think we should touch these rules - just ignore them for testcontainers tests if you need to.

jvorcak: You're absolutely right, I forgot to remove this before my push. Thank you!

frontend/tests/global-setup.mjs
```diff
@@ -0,0 +1,493 @@
...
+  const network = await new Network().start();
+  state.networkId = network.getId();
+  state.network = network;
+  console.log(`✓ Network created: ${state.networkId}`);
+  return network;
+}
+
+async function startRedpandaContainer(network, state) {
+  console.log('Starting Redpanda container...');
+  const redpanda = await new GenericContainer('redpandadata/redpanda:v25.2.1')
```
sago2k8, line 50: [minor]: There is a redpanda tescontainer we could use https://github.com/testcontainers/testcontainers-node/tree/main/packages/modules/redpanda. 

if there is something that does not work for your use case I am happy to extend the test container module.

frontend/tests/global-setup.mjs
```diff
@@ -0,0 +1,493 @@
...
+    .withNetwork(network)
+    .withNetworkAliases('redpanda')
+    .withExposedPorts(
+      { container: 19_092, host: 19_092 },
```
sago2k8, line 54: [question]: why do you need to expose ports ? could we keep only the internal docker docker network ?

sago2k8: I see why now, 👍

frontend/tests/global-setup.mjs
```diff
@@ -0,0 +1,493 @@
...
+    console.error(`[${prefix}] Process error:`, error);
+  });
+  processInstance.on('exit', (code) => {
+    if (code !== 0 && code !== null) {
+      console.error(`[${prefix}] Process exited with code ${code}`);
+    }
+  });
+}
+
+async function startBackendServer(isEnterprise, state) {
```
sago2k8: [question]: could we run this process inside a container also ? we could write a dockerfile or better something like the following.
```suggestion
// test/helpers/ko-builder.ts
import { execSync } from 'child_process';

export function ensureKo() {
  try {
    execSync('ko version', { stdio: 'ignore' });
  } catch {
    throw new Error(
      'ko not found. Install via Nix: nix profile install nixpkgs#ko'
    );
  }
}

export function buildGoImage(importPath: string, workDir: string): string {
  ensureKo();
  
  return execSync(`ko build --local --bare ${importPath}`, {
    cwd: workDir,
    encoding: 'utf-8',
    env: {
      ...process.env,
      KO_DOCKER_REPO: 'ko.local',
    },
  }).trim();
}

async function startBackendServer(isEnterprise, state) {
```

frontend/tests/global-setup.mjs
```diff
@@ -0,0 +1,493 @@
...
+    throw new Error('Failed to start backend process - no PID assigned');
+  }
+
+  state.backendPid = backendProcess.pid.toString();
+  console.log(`Backend started with PID: ${state.backendPid}`);
+
+  setupProcessStreaming(backendProcess, 'BACKEND');
+}
+
+async function startFrontendServer(isEnterprise, state) {
```
sago2k8: [minor]: can we make go serve the frontend, so it's more close to what we have today ?

sago2k8 (commented):
Left couple of comments.

frontend/tests/README.md
malinskibeniamin, line 1: Can we also add a claude SKILL for generating these tests?

malinskibeniamin: Can be in another PR

jvorcak: @malinskibeniamin I have already implemented this in another PR https://playwright.dev/docs/test-agents
It'll be a follow up after we merge this

frontend/tests/global-setup.mjs
```diff
@@ -0,0 +1,493 @@
...
+    .withExposedPorts(
+      { container: 19_092, host: 19_092 },
+      { container: 18_081, host: 18_081 },
+      { container: 18_082, host: 18_082 },
+      { container: 9644, host: 19_644 }
+    )
+    .withCommand([
+      'redpanda',
+      'start',
+      '--smp',
```
r-vasquez, line 62: There is a flag: `--mode dev-container` which should speed up the start process and was made for this particular case. If we go with the Redpanda testcontainer module, this is already in there (https://github.com/testcontainers/testcontainers-node/blob/main/packages/modules/redpanda/src/redpanda-container.ts#L60)

frontend/tests/global-setup.mjs
jvorcak: @sago2k8 I've tried that initially and I had problems with configuring SASL. That's why I used GenericContainer instead. Now I tried that again and I still cant' get it working. If we want to use Redpanda Testcontainer, I could use some help with this.

frontend/tests/global-setup.mjs
jvorcak: Thanks, I've added that until we implement RP testcontainer in another PR

frontend/tests/global-setup.mjs
jvorcak: done

frontend/tests/global-setup.mjs
jvorcak: done

PR #2043, opened 2025-11-23
Connect now uses new ListComponentsRequestSchema to fetch components + UI improvements

eblairmckee:
This PR addresses a vast majority of [bug bash reports](https://docs.google.com/document/d/1fZA97hzzBocp3UxsMLuyvNPKndvCFYNylNH65U8r2uQ/edit?tab=t.vlq492r3dzvf#heading=h.9iuifoqfmvdr), specifically the "major" ones. 

Please see l[atest comment](https://github.com/redpanda-data/console/pull/2043/#issuecomment-3584255314) for most recent changes and how we're better guarding against known regressions (and potential future ones due to this entirely new UX that's dependent on the backend)

# key changes

## replaces hard coded benthos json schemas with new backend query

This did introduce some breaking changes, which affected how we parse the schema, however, it results in eliminated hundreds of lines of code where we had to coerce the schema to meet our expectations discovered during bug bashing. this is a much more simplified approach, and we can use the proto types instead of inferring types from the schema. 

However, if the backend schema changes, and parser is unable to parse the schema it won't show any RPCN components (it will still show "custom" and "additional components", though, ensuring connect is always usable)

### component schema changes (ComponentSpec)
```typescript
{
  name: string;
  type: string;
  status: ComponentStatus;  // previously a string
  summary: string;
  description: string;
  categories: string[];
  version: string;  // ✅ NEW at component level
  config?: FieldSpec;
  examples: string[];  // ✅ NEW - full YAML examples
  footnotes: string;  // ✅ NEW
  plugin: boolean,  // ❌ REMOVED
}
```

### field schema changes (FieldSpec)
```typescript
{
  name: string;
  type: string;
  kind: string;
  description: string;
  advanced: boolean;  // 🔄 RENAMED from is_advanced
  deprecated: boolean;  // ✅ NEW
  optional: boolean;  // ✅ NEW
  defaultValue: string;  // 🔄 RENAMED from default
  is_secret: boolean,  // ❌ REMOVED
  scrubber: string,  // ❌ REMOVED
  linter: string,  // ❌ REMOVED
  version: string,  // ❌ REMOVED (moved to component)
  annotated_options: string[],  // ❌ REMOVED
  options: string[],  // ❌ REMOVED
  no_dynamic: boolean,  // ❌ REMOVED
  scalars_to_literal: boolean,  // ❌ REMOVED
  examples: string[];
  children: FieldSpec[];
}
```

## less opinionated defaults
this was a pretty contentious topic, and throughout the early iterations kept flip-flopping on whether we should provide default values for certain fields on `redpanda` components. those are almost all gone (brokers still gets a default) and we're relying solely on the schema for what fields to surface based on if they are not optional, and not advanced. 

Screenshot 2025-11-23 at 4 38 03 PM

### follow up
We should look at what endpoint the docs site uses, because currently it appears to use the old schema, which will be very confusing to users if they try to use the new UI (see schema changes above). As well as what's considered "common" vs "advanced" (very different in new schema). 

## better linting
We're now linting (debounce) on yaml editor content changes, and validating against the new backend API (same as the one where we pull component schemas from). 

Once I fix the monaco intellisense UI, we'll be able to get autocompletion based on the connect components definitions and properties. 

### lint ui
Improved styling, however, will be completely revisiting this in the next iteration (form-based UI) so minimal changes for now. 

## new Connect UI is decoupled from legacy
ripped out all new connect UI from the old and it now lives in its own directory. new UI is conditionally rendered when create/edit/view pages are mounted if `enableRpcnTiles` feature flag is true. 

## better wizard step validation
"Next" button is disabled until current wizard step is valid/populated. Given all steps are skippable, this makes it more clear, as previously Next button was always enabled regardless of form state. 

Also fixed the "create" buttons to be more responsive to query loading states to eliminate this bug: 
Screenshot 2025-11-23 at 4 38 06 PM

## better secrets detection
Screenshot 2025-11-23 at 4 34 00 PM

when user references a secret variable that doesn't exist, we suggest creating one, and automatically convert invalid formatting into correct format (screaming snake case) and replace the invalid secret variable with the correct one after user creates the secret. 

better loading state when adding secrets. previously when you clicked "create" _both_ 'missing secrets' and 'create new secret' submit buttons would load, which was confusing. now only the relevant one loads, and they are nested within their own individual `Form` components. 

## better error handling when updating yaml
previously when adding a new input/output/processor if existing yaml was invalid,  it would convert it to json 😬 now we just surface a toast that you need to fix the yaml. 

Screenshot 2025-11-23 at 4 37 15 PM

## swapped out tiles in "top 12" based on alex's feedback
Screenshot 2025-11-23 at 4 46 27 PM
Screenshot 2025-11-23 at 4 47 15 PM
Screenshot 2025-11-23 at 4 47 20 PM

### DIFF
--- a/frontend/src/components/pages/rp-connect/pipeline/index.tsx
+++ b/frontend/src/components/pages/rp-connect/pipeline/index.tsx
@@ -10,7 +10,6 @@
  */
 
 import { create } from '@bufbuild/protobuf';
-import { ConnectError } from '@connectrpc/connect';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { Alert, AlertDescription } from 'components/redpanda-ui/components/alert';
 import { Button } from 'components/redpanda-ui/components/button';
@@ -63,9 +62,10 @@ interface FooterProps {
   onSave?: () => void;
   onCancel: () => void;
   isSaving?: boolean;
+  disabled?: boolean;
 }
 
-const Footer = memo(({ mode, onSave, onCancel, isSaving }: FooterProps) => {
+const Footer = memo(({ mode, onSave, onCancel, isSaving, disabled }: FooterProps) => {
   if (mode === 'view') {
     return (
       <div className="flex items-center justify-between gap-2 border-t pt-4">
@@ -81,7 +81,7 @@ const Footer = memo(({ mode, onSave, onCancel, isSaving }: FooterProps) => {
       <Button disabled={isSaving} onClick={onCancel} variant="outline">
         Cancel
       </Button>
-      <Button className="min-w-[70px]" disabled={isSaving} onClick={onSave}>
+      <Button className="min-w-[70px]" disabled={isSaving || disabled} onClick={onSave}>
         {mode === 'create' ? 'Create Pipeline' : 'Update Pipeline'}
         {isSaving && <Spinner size="sm" />}
       </Button>
@@ -384,7 +384,7 @@ export default function PipelinePage() {
           setLintHints(extractLintHintsFromError(err));
           toast.error(
             formatToastErrorMessageGRPC({
-              error: ConnectError.from(err),
+              error: err,
               action: 'create',
               entity: 'pipeline',
             })
@@ -426,7 +426,7 @@ export default function PipelinePage() {
           setLintHints(extractLintHintsFromError(err));
           toast.error(
             formatToastErrorMessageGRPC({
-              error: ConnectError.from(err),
+              error: err,
               action: 'update',
               entity: 'pipeline',
             })
@@ -481,12 +481,18 @@ export default function PipelinePage() {
           </Alert>
         </div>
       )}
-      <Footer isSaving={isSaving} mode={mode} onCancel={handleCancel} onSave={handleSave} />
+      <Footer
+        disabled={Object.keys(lintHints).length > 0}
+        isSaving={isSaving}
+        mode={mode}
+        onCancel={handleCancel}
+        onSave={handleSave}
+      />
     </>
   );
 
   return (
-    <div className="flex w-full gap-4">
+    <div className="grid grid-cols-[minmax(auto,_950px)_260px] gap-4">
       <div className="flex flex-1 flex-col gap-4">
         {mode === 'view' && pipelineId && (
           <Toolbar pipelineId={pipelineId} pipelineName={form.getValues('name')} pipelineState={pipeline?.state} />
--- a/frontend/src/components/ui/lint-hint/lint-hint-list.tsx
+++ b/frontend/src/components/ui/lint-hint/lint-hint-list.tsx
@@ -8,8 +8,9 @@
  * by the Apache License, Version 2.0
  */
 
+import { SimpleCodeBlock } from 'components/redpanda-ui/components/code-block';
 import { Spinner } from 'components/redpanda-ui/components/spinner';
-import { Pre, Text } from 'components/redpanda-ui/components/typography';
+import { Text } from 'components/redpanda-ui/components/typography';
 import { cn } from 'components/redpanda-ui/lib/utils';
 import { PencilRuler } from 'lucide-react';
 import type { LintHint } from 'protogen/redpanda/api/common/v1/linthint_pb';
@@ -40,13 +41,9 @@ export const LintHintList: React.FC<LintHintListProps> = memo(({ className, lint
           {Object.entries(lintHints).map(([toolName, hint]) => (
             <div className="flex-col flex gap-1.5" key={toolName}>
               {hint.line > 0 ? (
-                  <Pre variant="dense">
-                    Line {hint.line}, Col {hint.column}: {hint.hint}
-                  </Pre>
+                  <SimpleCodeBlock code={`Line ${hint.line}, Col ${hint.column}: ${hint.hint}`} width="full" className="my-0"/>
               ) : (
-                <Pre variant="dense">
-                  {hint.hint}
-                </Pre>
+                <SimpleCodeBlock code={hint.hint} width="full" className="my-0"/>
               )}
               {hint.lintType && (
                 <Text className="font-medium text-gray-500 text-xs uppercase tracking-wide">{hint.lintType}</Text>
--- a/frontend/src/components/pages/rp-connect/errors.tsx
+++ b/frontend/src/components/pages/rp-connect/errors.tsx
@@ -1,6 +1,6 @@
 import { create } from '@bufbuild/protobuf';
 import { ConnectError } from '@connectrpc/connect';
-import { Text } from '@redpanda-data/ui';
+import { Text } from 'components/redpanda-ui/components/typography';
 
 import { type LintHint, LintHintSchema } from '../../../protogen/redpanda/api/common/v1/linthint_pb';
 
@@ -10,7 +10,7 @@ import { type LintHint, LintHintSchema } from '../../../protogen/redpanda/api/co
  *
  * Key differences from formatPipelineError:
  * - Returns structured LintHint objects (for LintResults component)
- * - formatPipelineError returns JSX (for legacy toast display)
+ * - formatPipelineError returns JSX (for toast erroring messaging)
  *
  * Leverages ConnectError properties:
  * - err.code: gRPC status code (e.g., "invalid_argument", "not_found")
--- a/frontend/src/components/pages/rp-connect/onboarding/create-pipeline-sidebar.tsx
+++ b/frontend/src/components/pages/rp-connect/onboarding/create-pipeline-sidebar.tsx
@@ -1,6 +1,5 @@
 import { useDisclosure } from '@redpanda-data/ui';
 import { Skeleton, SkeletonGroup } from 'components/redpanda-ui/components/skeleton';
-import { cn } from 'components/redpanda-ui/lib/utils';
 import type { editor } from 'monaco-editor';
 import type { ComponentList } from 'protogen/redpanda/api/dataplane/v1/pipeline_pb';
 import { memo, useCallback, useMemo, useState } from 'react';
@@ -13,8 +12,6 @@ import { AddSecretsCard } from './add-secrets-card';
 import type { ConnectComponentType } from '../types/schema';
 import { parseSchema } from '../utils/schema';
 import { getConnectTemplate } from '../utils/yaml';
-
-const sidebarClassNames = 'w-[260px]';
 
 type CreatePipelineSidebarProps = {
   editorInstance: editor.IStandaloneCodeEditor | null;
@@ -67,7 +64,7 @@ export const CreatePipelineSidebar = memo(
     );
 
     if (editorInstance === null) {
-      return <div className={sidebarClassNames} />;
+      return null;
     }
 
     const handleConnectorTypeChange = (connectorType: ConnectComponentType) => {
@@ -77,7 +74,7 @@ export const CreatePipelineSidebar = memo(
 
     if (isComponentListLoading) {
       return (
-        <div className={cn(sidebarClassNames, 'flex flex-col gap-4')}>
+        <div className="flex flex-col gap-4">
           <SkeletonGroup direction="vertical" spacing="default">
             <Skeleton variant="heading" width="full" />
             <Skeleton className="h-10 w-full" />
@@ -87,7 +84,7 @@ export const CreatePipelineSidebar = memo(
     }
 
     return (
-      <div className={cn(sidebarClassNames, 'flex flex-col gap-4')}>
+      <div className="flex flex-col gap-4">
         <AddConnectorsCard
           editorContent={editorContent}
           hasInput={hasInput}
--- a/frontend/src/components/redpanda-ui/components/code-block.tsx
+++ b/frontend/src/components/redpanda-ui/components/code-block.tsx
@@ -174,6 +174,10 @@ interface SimpleCodeBlockProps {
   maxHeight?: 'sm' | 'default' | 'lg' | 'none';
   onCopy?: () => void;
   className?: string;
+  /**
+   * Class name for the code block
+   */
+  codeClassName?: string;
   testId?: string;
 }
 
@@ -188,6 +192,7 @@ export const SimpleCodeBlock = ({
   maxHeight = 'default',
   onCopy,
   className,
+  codeClassName,
   testId,
 }: SimpleCodeBlockProps) => {
   return (
@@ -203,7 +208,7 @@ export const SimpleCodeBlock = ({
       testId={testId}
     >
       <Pre>
-        <code className={language}>{code}</code>
+        <code className={cn(language, 'break-all wrap-break-word whitespace-pre-wrap', codeClassName)}>{code}</code>
       </Pre>
     </CodeBlock>
   );


frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -405,7 +446,11 @@ export const ConnectOnboardingWizard = ({
                   </Button>
                 )}
                 {!methods.isLast && (
-                  <Button className="min-w-[70px]" disabled={isSubmitting} onClick={() => handleNext(methods)}>
+                  <Button
+                    className="min-w-[70px]"
```
eblairmckee, line 455: gets rid of CLS when loading spinner shows

frontend/src/components/pages/rp-connect/pipeline/details.tsx
```diff
@@ -0,0 +1,139 @@
...
+                  <Text variant="muted">One compute unit = 0.1 CPU and 400 MB memory</Text>
+                </div>
+              </AccordionContent>
+            </AccordionItem>
+          </Accordion>
+        </CardContent>
+      </Card>
+    );
+  }
+
```
eblairmckee, line 75: View mode

Screenshot 2025-11-23 at 4 08 34 PM

frontend/src/components/pages/rp-connect/pipeline/details.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return (
+    <Card size="full">
```
eblairmckee, line 77: Edit mode

Screenshot 2025-11-23 at 4 10 19 PM

frontend/src/react-query/api/pipeline.tsx
```diff
@@ -182,6 +189,13 @@ export const useStopPipelineMutation = () => {
           cardinality: 'infinite',
         }),
       });
+      await queryClient.invalidateQueries({
```
malinskibeniamin, line 192: I think it was deleted before because some people preferred to have the invalidation happen in the component layer, but let's stick to the pattern we established before and keep it tied to the hook.

frontend/src/react-query/api/connect.tsx
```diff
@@ -0,0 +1,61 @@
...
+  });
+};
+
+export const useListComponentsQuery = (
+  options?: QueryOptions<GenMessage<ListComponentsResponse>, ListComponentsResponse>
+) => {
+  const listComponentsRequest = create(ListComponentsRequestSchema, {});
+
+  return useQuery(listComponents, listComponentsRequest, {
+    staleTime: 1000 * 60 * 60, // 1 hour
```
malinskibeniamin: Let's use constants for stale time

frontend/src/react-query/api/connect.tsx
```diff
@@ -0,0 +1,61 @@
...
+    gcTime: 1000 * 60 * 60 * 24, // 24 hours
```
malinskibeniamin: I don't think we need garbage collection time in react query v5+ and above it should be mostly handled for us.

frontend/src/react-query/api/connect.tsx
```diff
@@ -0,0 +1,61 @@
...
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    retry: 3,
```
malinskibeniamin: Should we update query client instead?

frontend/src/react-query/api/connect.tsx
```diff
@@ -0,0 +1,61 @@
...
+  options?: QueryOptions<GenMessage<GetPipelineServiceConfigSchemaResponse>, GetPipelineServiceConfigSchemaResponse>
+) => {
+  const request = create(GetPipelineServiceConfigSchemaRequestSchema, {});
+
+  return useQuery(getPipelineServiceConfigSchema, request, {
+    staleTime: 1000 * 60 * 60, // 1 hour
+    gcTime: 1000 * 60 * 60 * 24, // 24 hours
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    retry: 3,
```
malinskibeniamin: Same as below, let's either raise it to the query client level or use proper constants

frontend/src/hooks/use-debounced-value.ts
malinskibeniamin, line 1: We already had a hook called `use-debounce.ts`

```ts

/**
 */

    (...args: Parameters<T>) => {
      if (timeoutRef.current) {
      }

        callback(...args);
      }, delay);
    },
    [callback, delay]
  );

};

```

malinskibeniamin: I like your approach too. No ref required

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -220,31 +281,31 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
             </CardTitle>
           </CardHeader>
           <CardContent className="space-y-4">
-            <Alert className="border-amber-200 bg-amber-50">
-              <AlertTriangle className="h-4 w-4 text-amber-600" />
-              <AlertDescription className="text-amber-800">
+            <Alert variant="destructive">
+              <AlertTriangle className="h-4 w-4" />
+              <AlertDescription>
```
malinskibeniamin, line 286: 👏

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -166,9 +213,23 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
+    const normalizedName = normalizeSecretName(data.name);
+    
+    // Check if this secret already exists
+    if (existingSecretsSet.has(normalizedName) || createdSecretsSet.has(normalizedName)) {
+      const errorMessage = `Secret "${normalizedName}" already exists`;
+      if (onError) {
+        onError([errorMessage]);
+      } else {
+        toast.error(errorMessage);
+      }
```
malinskibeniamin, line 226: Should we show a toast error even if `onError` callback exists?

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -111,14 +143,23 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
-    const secretEntries = Object.entries(data);
+    // Filter out secrets that have already been created
+    const secretEntries = Object.entries(data).filter(([normalizedSecretName]) => {
+      return !(existingSecretsSet.has(normalizedSecretName) || createdSecretsSet.has(normalizedSecretName));
+    });
+
+    if (secretEntries.length === 0) {
+      toast.info('All secrets have already been created');
+      return;
+    }
```
malinskibeniamin, line 154: Is it required?

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -86,19 +87,50 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
-    (secret) => !(existingSecrets.includes(secret) || createdSecrets.includes(secret))
-  );
+  /**
+   * Normalizes secret name to uppercase and validates characters
+   * Secrets are always stored in uppercase
+   */
+  const normalizeSecretName = (name: string): string => {
+    // Convert to uppercase and replace invalid characters with underscores
+    return name.toUpperCase().replace(/[^A-Z0-9_]/g, '_');
+  };
```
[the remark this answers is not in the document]
malinskibeniamin: I think we could have the regex called `ALPHANUMERIC_WITH_HYPHENS = new RegExp(/[^A-Z0-9_]/g);`

frontend/src/components/ui/lint-hint/lint-hint-list.tsx
```diff
@@ -8,44 +8,50 @@
...
+            <div className="flex-col flex gap-1.5" key={toolName}>
               {hint.line > 0 ? (
                 <div className="flex flex-col gap-1">
-                  <Text className="font-medium text-gray-600 text-xs">
+                  <Pre variant="dense">
                     Line {hint.line}, Col {hint.column}
-                  </Text>
-                  <Text className="rounded border bg-white px-2 py-1 font-mono text-gray-800 text-sm leading-relaxed">
+                  </Pre>
+                  <Pre variant="dense">
```
malinskibeniamin: 👏  Maybe we should also use `DynamicCodeBlock`?

frontend/src/components/pages/rp-connect/pipelines-create.tsx
```diff
@@ -583,27 +398,17 @@ export const PipelineEditor = observer(
...
-                        }}
-                      />
-                    ))}
+                  {!p.isDisabled && (
+                    <QuickActions
+                      editorInstance={editorInstance}
+                      resetAutocompleteSecrets={() => {
+                        if (secretAutocomplete && monaco) {
+                          secretAutocomplete.dispose();
+                          void registerSecretsAutocomplete(monaco, setSecretAutocomplete);
```
malinskibeniamin: Should it be awaited instead of `void`'s fire-and-forget behavior?

frontend/src/components/pages/rp-connect/pipelines-create.tsx
```diff
@@ -252,17 +218,13 @@ class RpConnectPipelinesCreate extends PageComponent<{}> {
...
-          } else {
-            this.lintResults = extractLintHintsFromError(err);
-          }
+          toast({
+            status: 'error',
+            duration: null,
+            isClosable: true,
+            title: 'Failed to create pipeline',
+            description: formatPipelineError(err),
+          });
```
malinskibeniamin: Let's try to create a `ConnectError` with `ConnectError.from(error)`, if possible, and follow the format function. the description could be kept intact with `formatPipelineError`

frontend/src/components/pages/rp-connect/pipelines-create.tsx
```diff
@@ -225,25 +196,20 @@ class RpConnectPipelinesCreate extends PageComponent<{}> {
...
+          const retUnits = cpuToTasks(r.response?.pipeline?.resources?.cpuShares);
+          if (retUnits && this.tasks !== retUnits) {
             toast({
-              status: 'success',
-              duration: 4000,
+              status: 'warning',
+              duration: 6000,
               isClosable: false,
-              title: 'Pipeline created',
+              title: `Pipeline has been resized to use ${retUnits} compute units`,
```
malinskibeniamin: I think the duration should be established at the root level.

frontend/src/components/pages/rp-connect/utils/yaml.ts
```diff
@@ -428,29 +429,37 @@ export const configToYaml = (
     let yamlString = yamlStringify(doc, yamlConfig);
     yamlString = addRootSpacing(yamlString);
     return yamlString;
-  } catch (_error) {
-    return JSON.stringify(configObject, null, 2);
+  } catch (error) {
+    const description = error instanceof Error ? error.message : String(error);
+    toast.error('Failed to generate YAML', {
+      description,
+    });
```
malinskibeniamin: Shall we try to parse it to a connect error and use the format function for the toast messages to remain consistent?

frontend/src/components/pages/rp-connect/utils/yaml.ts
```diff
@@ -271,22 +271,23 @@ export const mergeConnectConfigs = (
...
-  // }
 
-  mergeByComponentType(componentType, doc, newConfigObject);
+  try {
+    mergeByComponentType(componentType, doc, newConfigObject);
+  } catch (error) {
+    const description = error instanceof Error ? error.message : String(error);
+    toast.error('Cannot merge: Merge operation failed', {
+      description,
+    });
```
malinskibeniamin: Same as previous comments regarding toasts

frontend/src/components/pages/rp-connect/utils/use-pipeline-mode.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { useLocation, useParams } from 'react-router-dom';
```
malinskibeniamin, line 12: If we were to update our router, do you think there is any way to abstract this to make it less painful? Just wondering. Nothing actionable

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+      });
+
+      startMutation.mutate(startRequest, {
+        onSuccess: () => {
+          toast({
+            status: 'success',
+            title: 'Pipeline started',
+            duration: 4000,
+            isClosable: false,
+          });
```
malinskibeniamin: Do we need to do this? I think the state transition should happen automatically, so the user would see the pipeline already starting

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+        },
+        onError: (err) => {
+          toast({
+            status: 'error',
+            title: 'Failed to start pipeline',
+            description: formatPipelineError(err),
+            duration: null,
+            isClosable: true,
+          });
```
malinskibeniamin: Let's use format function

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+        },
+        onError: (err) => {
+          toast({
+            status: 'error',
+            title: 'Failed to stop pipeline',
+            description: formatPipelineError(err),
+            duration: null,
+            isClosable: true,
+          });
```
malinskibeniamin: Let's use format function

malinskibeniamin: See other comments about toasts

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const isRunning = useMemo(() => pipelineState === PipelineState.RUNNING, [pipelineState]);
+  const isTransitioning = useMemo(
+    () => pipelineState === PipelineState.STARTING || pipelineState === PipelineState.STOPPING,
+    [pipelineState]
+  );
+  const isActionLoading = useMemo(
+    () => startMutation.isPending || stopMutation.isPending || deleteMutation.isPending,
+    [startMutation.isPending, stopMutation.isPending, deleteMutation.isPending]
+  );
+  const isLoading = useMemo(() => isActionLoading || isTransitioning, [isActionLoading, isTransitioning]);
```
malinskibeniamin: There is a lot of memoization here, 3 questions:
- what was the original issue? was it not performing well without `useMemo`?
- was it still functional prior to `useMemo`?
- maybe we could do some react compiler prioritization to avoid having to think about memoizing things manually?

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+import { Button } from 'components/redpanda-ui/components/button';
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from 'components/redpanda-ui/components/dialog';
+import { Group } from 'components/redpanda-ui/components/group';
```
malinskibeniamin, line 15: I think this component is deprecated now, can we use `ButtonGroup`?

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+ * Use of this software is governed by the Business Source License
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { create } from '@bufbuild/protobuf';
+import { useToast } from '@redpanda-data/ui';
```
malinskibeniamin: Let's avoid using `@redpanda-data/ui`

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -0,0 +1,415 @@
...
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { create } from '@bufbuild/protobuf';
+import { zodResolver } from '@hookform/resolvers/zod';
+import { useToast } from '@redpanda-data/ui';
```
malinskibeniamin: Same as before - can we use sonner toast component instead of `@redpanda-data/ui`

frontend/src/components/pages/rp-connect/pipeline/footer.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <div className="flex items-center justify-end gap-2 border-t pt-4">
+      <Button disabled={isSaving} onClick={handleCancel} variant="outline">
+        Cancel
+      </Button>
+      <Button className="min-w-[70px]" disabled={isSaving} onClick={onSave}>
+        {mode === 'create' ? 'Create Pipeline' : 'Update Pipeline'}
+        {isSaving && <Spinner size="sm" />}
+      </Button>
+    </div>
```
malinskibeniamin: Let's use `ButtonGroup` component

frontend/src/components/pages/rp-connect/pipeline/details.tsx
```diff
@@ -0,0 +1,139 @@
...
+              <AccordionContent className="space-y-4">
+                <div className="space-y-2">
+                  <Label>Description</Label>
+                  <Text>{description || '-'}</Text>
+                </div>
+
+                <div className="space-y-2">
+                  <Label>Compute Units: {computeUnits}</Label>
+                  <Text variant="muted">One compute unit = 0.1 CPU and 400 MB memory</Text>
+                </div>
```
malinskibeniamin, line 67: I don't know if we still want to show this, @birdayz would you be able to advise if this is still relevant for the backend? Similar question for all other AI features

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
malinskibeniamin: It's still the same CSS, did I miss anything?

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -48,14 +50,27 @@ export const ConnectOnboardingWizard = ({
...
       type: 'custom',
-      plugin: false,
+      status: 0,
+      summary: 'Build your own pipeline from scratch.',
+      description: '',
+      categories: [],
+      version: '',
+      examples: [],
+      footnotes: '',
+      $typeName: 'redpanda.api.dataplane.v1.ComponentSpec',
```
malinskibeniamin: Can we avoid using `$typeName`? It's internal to the generate proto connect. Maybe we could use `create()` and use the schema?

frontend/src/components/pages/rp-connect/onboarding/connect-tile.tsx
```diff
@@ -6,9 +6,11 @@ import { InlineCode, Text } from 'components/redpanda-ui/components/typography';
 import { cn } from 'components/redpanda-ui/lib/utils';
 import { CheckIcon, Waypoints } from 'lucide-react';
 import { AnimatePresence, type MotionProps, motion } from 'motion/react';
+import { ComponentStatus } from 'protogen/redpanda/api/dataplane/v1/pipeline_pb';
```
malinskibeniamin, line 9: ❤️

frontend/src/components/pages/rp-connect/onboarding/add-secrets-card.tsx
```diff
@@ -85,14 +106,48 @@ export const AddSecretsCard = ({
...
+      }
+
+      const model = editorInstance.getModel();
+      if (!model) {
+        return;
+      }
+
+      // Find and replace all occurrences of ${secrets.oldName} with ${secrets.newName}
+      const oldPattern = `\${secrets.${oldName}}`;
+      const newPattern = `\${secrets.${newName}}`;
```
malinskibeniamin, line 123: We may need to biome ignore these 2 lines because sometimes it messes up/tries to update this pattern

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+
+```bash
+# Single component
+yes | bunx @fumadocs/cli add --dir https://redpanda-ui-registry.netlify.app/r button
+
+# Multiple components (space-separated)
+yes | bunx @fumadocs/cli add --dir https://redpanda-ui-registry.netlify.app/r card dialog form
+
+# Then generate yarn.lock
+bun i --yarn
```
malinskibeniamin: `bun i && bun i --yarn` just for the sake of completeness, since we don't actually use `yarn.lock` for anything other than Snyk

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+  test('should call onSubmit when button clicked', () => {
+    const onSubmit = vi.fn();
+    render(<MyComponent onSubmit={onSubmit} />);
+
+    fireEvent.click(screen.getByRole('button'));
+
+    expect(onSubmit).toHaveBeenCalled();
+  });
+});
+```
```
malinskibeniamin, line 156: I would replace it with grpc router usage, checking if clicking a button to submit a form would trigger the mutation for example.

frontend/src/components/ui/yaml/yaml-editor.tsx
malinskibeniamin, line 1: Feedback from codex
```
- [P2] Reconfigure YAML editor when schema data arrives — frontend/src/components/ui/yaml/yaml-editor.tsx:76-129
    The new YAML editor tries to consume the backend schema by building monacoYamlOptions from the schema prop, but configureMonacoYaml is only executed inside the beforeMount callback (lines 76‑129).
  beforeMount fires once right before Monaco is instantiated, so the editor locks in whichever schema exists during the initial render. In PipelinePage the schema is fetched asynchronously via
  useGetPipelineServiceConfigSchemaQuery, so the first render always uses the fallback schema and the later response is ignored. As a result validation/completions never switch to the service schema,
  defeating the purpose of the change. Please re-run configureMonacoYaml whenever schema changes (e.g. in an effect that disposes and re‑configures the language service or by remounting the editor once
  the schema is ready).
```

frontend/src/components/pages/rp-connect/pipeline/index.tsx
malinskibeniamin, line 1: Feedback from codex:
```
- [P2] Preserve pipeline type tag when creating pipelines — frontend/src/components/pages/rp-connect/pipeline/index.tsx:234-248
    The new create flow (lines 234‑248) builds the PipelineCreateSchema request without including any tags. The legacy MobX page (pipelines-create.tsx lines 181‑194) always set
  tags.__redpanda_cloud_pipeline_type = 'pipeline' so downstream filters and billing could distinguish Connect pipelines from agents. With the feature flag enabled, pipelines created through
  PipelinePage will miss that tag and therefore won’t show up for consumers that rely on it. Please set the same tag (and merge any user‑supplied tags if needed) in the new request before sending it to
  the backend.
```

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+**Styling:**
+
+```typescript
+// ✅ CORRECT: Use component variants
+<Button variant="primary" size="lg">Click</Button>
+
+// ✅ CORRECT: Wrap for spacing
+<div className="mt-4">
+  <Button variant="primary">Click</Button>
+</div>
```
malinskibeniamin, line 78: Let's tweak it to suggest using `ButtonGroup`

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+
+- Define prop interfaces
+- Never use `any` - infer correct types
+- Use generics for collections
+
+**Performance:**
+
+- Hoist static content outside component
+- Use `useMemo` for expensive computations
+- Use `memo` for components receiving props
```
malinskibeniamin: Let's add section saying `useMemo` should only be used to improve performance, not to make things work initially.

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+
+### 5. Validation
+
+```bash
+# Run in order
+bun run type:check    # TypeScript errors
+bun run test          # All tests
+bun run lint          # Code quality
+bun run build         # Production build
+bun run start         # Dev server - check browser
```
malinskibeniamin: I would say `bun run start2 --port=3004`

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+- ✓ No runtime errors in browser
+
+## Testing Commands
+
+```bash
+bun run test                 # All tests (unit + integration)
+bun run test:unit            # Unit tests only (.test.ts)
+bun run test:integration     # Integration tests only (.test.tsx)
+bun run test:watch           # Watch mode
+bun run test:ui              # Interactive UI
```
malinskibeniamin: I think we should not let it run UI mode because it will get confused, it spawns an electron app

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+- Test loading and error states
+- Verify state updates
+
+## Quick Reference
+
+**Check before installing:**
+
+```bash
+# Check if component exists
+ls src/components/redpanda-ui/components/
```
malinskibeniamin, line 296: Maybe say if file exists and then `ls src/components/redpanda-ui/` because we may also have `hooks` or other directories.

frontend/.claude/skills/frontend-developer/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+
+```bash
+yes | bunx @fumadocs/cli add --dir https://redpanda-ui-registry.netlify.app/r \
+  button card dialog form input label select
+```
+
+**After package install:**
+
+```bash
+bun i --yarn  # ALWAYS run this
```
malinskibeniamin:
```suggestion
bun i && bun i --yarn
```

frontend/src/components/pages/rp-connect/onboarding/add-secrets-card.tsx
```diff
@@ -85,14 +106,48 @@ export const AddSecretsCard = ({
     },
     [editorInstance]
   );
+
+  const handleSecretsCreated = useCallback(() => {
+    refetchSecrets();
+    setIsSecretsDialogOpen(false);
+  }, [refetchSecrets]);
```
malinskibeniamin: Let's invalidate the relevant connect query cache keys instead, no need to have refetch triggered manually, WDYT?

frontend/src/components/pages/rp-connect/onboarding/connect-tile.tsx
```diff
@@ -67,11 +69,14 @@ export const ConnectTile = ({
...
           <span>
-            {component.status && component.status !== 'stable' && component.name !== 'redpanda' && (
-              <Badge size="sm" variant="gray">
-                {component.status}
-              </Badge>
-            )}
+            {(component.status === ComponentStatus.BETA ||
+              component.status === ComponentStatus.EXPERIMENTAL ||
+              component.status === ComponentStatus.DEPRECATED) &&
+              component.name !== 'redpanda' && (
```
malinskibeniamin, line 75: Is this to prevent the `beta` status badge on redpanda component?

frontend/src/components/pages/rp-connect/onboarding/connect-tiles.tsx
```diff
@@ -13,19 +13,21 @@ import { Form, FormControl, FormField, FormItem, FormMessage } from 'components/
 import { Input, InputStart } from 'components/redpanda-ui/components/input';
 import { Label } from 'components/redpanda-ui/components/label';
 import { SimpleMultiSelect } from 'components/redpanda-ui/components/multi-select';
+import { Skeleton, SkeletonGroup } from 'components/redpanda-ui/components/skeleton';
```
malinskibeniamin, line 16: 👏

frontend/src/components/pages/rp-connect/pipeline/details.tsx
malinskibeniamin: CC @birdayz should we rinse & repeat for AI features too

frontend/src/components/pages/rp-connect/pipeline/footer.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onCancel?: () => void;
+  isSaving?: boolean;
+}
+
+export const Footer = memo(({ mode, onSave, onCancel, isSaving }: FooterProps) => {
+  const navigate = useNavigate();
+
+  const handleCancel = useCallback(() => {
+    onCancel?.();
+    navigate(-1);
```
malinskibeniamin: Do we have access to `goBack` in this version of `react-router-dom`?

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+}
+
+export const Toolbar = memo(({ pipelineId, pipelineName, pipelineState }: ToolbarProps) => {
+  const navigate = useNavigate();
+  const toast = useToast();
+  const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+
+  const deleteMutation = useDeletePipelineMutation();
+  const startMutation = useStartPipelineMutation();
+  const stopMutation = useStopPipelineMutation();
```
malinskibeniamin: Shouldn't we destructure these mutations to `[mutation, isPending]`? Right now it looks like we don't consider what happens when the mutation is ongoing.

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+    <>
+      <div className="mb-4 flex items-center justify-between border-b pb-4">
+        <Button
+          disabled={isLoading}
+          // biome-ignore lint/style/noNestedTernary: because I said so
+          icon={isLoading ? <Spinner size="sm" /> : isRunning ? <Pause /> : <Play />}
+          onClick={handleStartStop}
+          variant="secondary"
+        >
+          {/** biome-ignore lint/style/noNestedTernary: because I said so */}
```
malinskibeniamin: I have previously created `ai-agent-toggle-button.tsx` and `remote-mcp-toggle-button.tsx` because there is a lot of logic here and it's easier to test if it's a separate component, maybe we could have a similar case here for the `pipeline-toggle-button.tsx`

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+          <DialogFooter>
+            <Button onClick={() => setIsDeleteDialogOpen(false)} variant="outline">
+              Cancel
+            </Button>
+            <Button disabled={deleteMutation.isPending} onClick={handleDeleteConfirm} variant="destructive">
+              {deleteMutation.isPending ? 'Deleting...' : 'Delete'}
+            </Button>
+          </DialogFooter>
+        </DialogContent>
+      </Dialog>
```
malinskibeniamin: Use `frontend/src/components/ui/delete-resource-alert-dialog.tsx`

frontend/src/components/pages/rp-connect/pipeline/details.tsx
birdayz: nice! let's do that

birdayz: Well, can we also have an input in addition?

For this use case a direct numeric input would be more practical. Users typically have a specific number in mind rather than needing to explore a range. Having to drag the slider to find an exact value adds unnecessary friction compared to just typing the number directly.

Could we add a text input field alongside the slider, or make the slider value editable?

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
eblairmckee: I added this in like 3 other places, too. idk why I picked this line to leave the comment, lol

frontend/src/components/pages/rp-connect/pipeline/details.tsx
eblairmckee: Oooh, good point. I'll add a number input next to the slider.

frontend/src/components/ui/secret/quick-add-secrets.tsx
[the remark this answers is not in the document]
eblairmckee: added a new `frontend/src/utils/regex.ts` with this, since it didn't exist already

frontend/src/react-query/api/pipeline.tsx
eblairmckee: +1

frontend/src/react-query/api/connect.tsx
eblairmckee: Hrm, I'm not sure what you mean by this.. can you clarify?

frontend/src/components/ui/secret/quick-add-secrets.tsx
eblairmckee: wanted to make it more custom, given this is a reusable component that we use in multiple features. more like a design system component API. in some cases I assume we may not want to show a toast. eg: if it's in a modal, and we don't want to dismiss modal, we may want to surface an alert.

frontend/src/components/ui/lint-hint/lint-hint-list.tsx
eblairmckee: this is a very temporary fix, as in v2 of this I'm actually going to have a logger and completely different lint interface

frontend/src/components/pages/rp-connect/pipelines-create.tsx
eblairmckee: this is also coming in v2, where I completely rewrite error and lint handling/UI. this is just a temp fix.

frontend/src/components/pages/rp-connect/pipeline/details.tsx
eblairmckee: Screenshot 2025-11-26 at 2 00 54 PM
:chef-kiss:

frontend/src/react-query/api/connect.tsx
eblairmckee: nvm figured it out. didn't realize we had a default

frontend/src/hooks/use-debounced-value.ts
eblairmckee: where does it live, though? in cloud? not in this repo. and this is identifcal to the one I added. note there's useDebounce and useDebouncedValue

frontend/src/components/ui/secret/quick-add-secrets.tsx
eblairmckee: nope, removed the toast but kept early return.

frontend/src/components/pages/rp-connect/pipelines-create.tsx
eblairmckee: fwiw, I'll be rewrite all of the toast, alert, and error messaging in the very near future.

frontend/src/components/pages/rp-connect/pipelines-create.tsx
eblairmckee: also this is legacy code, let's not focus on this page, given it's basically getting reverted (with minor improvements) to old UI)

eblairmckee: ended up using ConnectError.from but still using legacy formatPipelineError because it formats lint errors, and int he legacy UI we show those in toasts

frontend/src/components/pages/rp-connect/pipelines-create.tsx
eblairmckee: removed all for now since it defaults to 5k

eblairmckee: and replaced with sonner

frontend/src/components/pages/rp-connect/utils/yaml.ts
eblairmckee: that doesn't apply when we're translating a typescript object into yaml

eblairmckee: but I'll use formatToastErrorMessageGRPC plus ConnectError.from since it does seem to parse errors nicely. just weird that it's coupled to connect..

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
eblairmckee: can do that later when we migrate all of them

frontend/src/components/pages/rp-connect/pipeline/footer.tsx
eblairmckee: again, anti pattern

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
eblairmckee: it's required on the schema. it's just for mocking anyways

frontend/src/components/ui/yaml/yaml-editor.tsx
eblairmckee: how did you get codex to review the PR? did you checkout the branch and ask it about specific files?

frontend/.claude/skills/frontend-developer/SKILL.md
eblairmckee: only when you want them to be attached, I actually need to add that as a variant. because the whole idea of an unattached button group is to have same gap everywhere. that Group component is actually a lot better at this and more reusable.

eblairmckee: we can talk about it later, though

frontend/src/components/pages/rp-connect/onboarding/connect-tile.tsx
eblairmckee: yes. there's an open item I have to resolve this once David Yu is back, we're stil deciding what components are beta/experimental and if we even want to show them in this UX

frontend/src/components/pages/rp-connect/pipeline/footer.tsx
eblairmckee: apparently not

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
eblairmckee: why? I ended up memoizing the props instead.

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
eblairmckee: that's a weird pattern to include the trigger in something called a dialog.... left it as is for now, but we should rethink the naming. I also added a triggerVariant because I need to render it as a button and not a dropdown menu item

frontend/src/components/pages/rp-connect/pipeline/index.tsx
eblairmckee: oh, nice! added

frontend/src/components/ui/yaml/yaml-editor.tsx
eblairmckee: done! thanks codex

frontend/.claude/skills/frontend-developer/SKILL.md
malinskibeniamin, line 1: If that's ok I will commit this separately to our repo so we can start using it sooner

frontend/src/react-query/api/connect.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return useQuery(listComponents, listComponentsRequest, {
+    staleTime: LONG_LIVED_CACHE_STALE_TIME,
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    ...options,
+  });
+};
+
+export const useLintPipelineConfigQuery = (
```
malinskibeniamin, line 49: Please check if this hook is also in the `pipeline.tsx` file, maybe we can remove 1 or the other.

frontend/src/hooks/use-debounced-value.ts
malinskibeniamin: Thank you, I think we only had it in cloudv2, indeed

frontend/src/components/ui/yaml/yaml-editor.tsx
malinskibeniamin: In Codex CLI there is a built-in `/review` command which lets you point the commit/branch to compare against the baseline such as `main` and just run through any potential issues.

frontend/src/components/redpanda-ui/components/spinner.tsx
malinskibeniamin, line 1: Have we already moved it to the UI registry as well?

eblairmckee: no. i do an upstream dump every few weeks

frontend/src/react-query/api/connect.tsx
eblairmckee: It's not, we use a different lint endpoint with the new grpc endpoint

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
andresaristizabal: I agree with Ben, please use create from protobuf, it helps to make correct default values in a deep object, or, is there a big reason to don't use?

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
andresaristizabal: I think `because I said so` is not a enoght tecnical reason 😄 , also claude can solve it really quicky, lets avoid biome ignore if there isn't a good reason

eblairmckee: totally valid, moved to being prop driven in a useMemo instead
Screenshot 2025-12-01 at 6 39 40 PM

frontend/.claude/skills/frontend-developer/SKILL.md
eblairmckee: tbh, in an ideal world, I'd love to have a repo just for claude: skills, commands, etc. that we can pull/update into consuming repos similar to the ui-registry (or we can explore pulling into ui-registry given these are all frontend focused). let's discuss when you're back

frontend/.claude/skills/frontend-developer/SKILL.md
eblairmckee: ended up moving these changes into a separate PR as it's gotten too noisy in here.

eblairmckee: https://github.com/redpanda-data/console/pull/2057

PR #2045, opened 2025-11-24
Fix tool call error handling and add execution duration

birdayz:
## What

Tool execution errors now properly display with error state and show the actual error message. Added execution duration tracking for completed and failed tool calls.

## Why

Tool failures were incorrectly shown as "completed" with green badges, making debugging impossible. Users couldn't see what actually went wrong when tools failed.

## Implementation details

- `event-handlers.ts`: Detect error field in tool_response, set state to 'output-error', capture end timestamp
- `tool.tsx`: Format and display duration (ms for <1s, seconds otherwise), fix badge structure to match task badges
- `types.ts`: Add endTimestamp field to tool ContentBlock
- `chat-message.tsx`: Wire through endTimestamp to ToolBlock
- `tool-block.tsx`: Calculate duration from start/end timestamps

Error display uses same CodeBlock styling as normal output (no red text/background).
Duration shows next to status badge for completed/failed tools only.

Tool error display

### DIFF
--- a/frontend/src/components/pages/agents/details/a2a/chat/hooks/event-handlers.ts
+++ b/frontend/src/components/pages/agents/details/a2a/chat/hooks/event-handlers.ts
@@ -182,8 +182,14 @@ const processToolResponse = (state: StreamingState, data: Record<string, unknown
   if (existingToolBlock && existingToolBlock.type === 'tool') {
     const hasError = 'error' in data && data.error;
     existingToolBlock.state = hasError ? 'output-error' : 'output-available';
-    existingToolBlock.output = hasError ? undefined : 'result' in data ? data.result : undefined;
-    existingToolBlock.errorText = hasError ? (data.error as string) : undefined;
+
+    if (hasError) {
+      existingToolBlock.output = undefined;
+      existingToolBlock.errorText = data.error as string;
+    } else {
+      existingToolBlock.output = 'result' in data ? data.result : undefined;
+      existingToolBlock.errorText = undefined;
+    }
   }
 };
 


frontend/src/components/ai-elements/tool.tsx
```diff
@@ -195,15 +221,8 @@ export const ToolOutput = ({
...
-        className={cn(
-          "rounded-md text-xs [&_table]:w-full",
-          errorText
-            ? "bg-destructive/10 text-destructive"
-            : "bg-muted/50 text-foreground"
-        )}
-      >
-        {errorText && <div className="break-words p-3">{errorText}</div>}
+      <div className={cn("rounded-md", errorText ? "bg-muted/50" : "bg-muted/50")}>
+        {errorText && <CodeBlock code={errorText} language="text" />}
```
malinskibeniamin: Can you use `DynamicCodeBlock` instead? It will have better syntax highlighting.

frontend/src/components/ai-elements/tool.tsx
```diff
@@ -78,11 +91,19 @@ export const ToolHeader = ({
...
 }: ToolHeaderProps) => {
   const toolName = title ?? type.split("-").slice(1).join("-");
   const textToCopy = toolCallId ? `${toolName} (${toolCallId})` : toolName;
 
+  const formatDuration = (ms: number): string => {
+    if (ms < 1000) {
+      return `${ms}ms`;
+    }
+    return `${(ms / 1000).toFixed(2)}s`;
+  };
```
malinskibeniamin, line 105: This will need refactoring later but it's ok for now.

malinskibeniamin (approved):

frontend/src/components/ai-elements/tool.tsx
birdayz: Switched to DynamicCodeBlock for better syntax highlighting.

PR #2046, opened 2025-11-24
Add context usage tracking to AI agent chat

birdayz:
## What
Add context usage widget to AI agent chat input that displays token usage and cost metrics.

## Why
Users need visibility into:
- Current context window usage (percentage filled)
- Cumulative token counts across conversation
- Estimated costs for input/output tokens

### Context Usage Display
- Widget positioned in chat input footer, left of "Clear context" button
- Shows current context window usage (e.g., "2% / 5.4K / 272K")
- Hover card displays detailed breakdown:
  - Input tokens with cost
  - Output tokens with cost
  - Reasoning tokens
  - Cached tokens

### Data Flow
1. Usage metadata captured from streaming responses via `response-metadata` events
2. Stored in database alongside messages for persistence across refreshes
3. Cumulative totals calculated across all messages in conversation
4. Latest request provides context window data (max_input_tokens, input_tokens)

### Performance Optimizations
- `useMemo` for expensive calculations (modelId construction, usage object)
- `React.memo` on ChatInput to prevent re-renders during typing
- `useCallback` for stable callback references
- Only recalculates when actual values change

### Dynamic Model/Provider Detection
- Extracts provider from `agent.provider.provider.case` (e.g., "openai")
- Extracts model from `agent.model` (e.g., "gpt-5")
- Constructs `modelId` as `"provider:model"` for tokenlens cost calculations
- Supports future providers without code changes

### UI/UX Details
- Hover card positioned above trigger to stay visible
- Always shows breakdown even with 0 tokens (displays "$0.00")
- Context cleared via "Clear context" button (renamed from "Clear history")
- Initial display shows 0% usage before first message

### Files Changed
**Core functionality:**
- `ai-agent-chat.tsx` - Usage calculation with memoization
- `chat-input.tsx` - Context widget integration with memo
- `types.ts` - UsageMetadata type with cumulative fields

**Streaming & persistence:**
- `use-message-streaming.ts` - Pass usage to DB updates
- `database-operations.ts` - Store usage field
- `message-converter.ts` - Load usage from DB
- `chat-db.ts` - Database schema

**UI component:**
- `context.tsx` - Remove conditional hiding when tokens are 0

## References
- Uses `@tokenlens/vercel-ai` for cost calculations
- Context widget from `@ai-elements/context`

### DIFF
--- a/frontend/src/components/pages/agents/details/a2a/chat/ai-agent-chat.tsx
+++ b/frontend/src/components/pages/agents/details/a2a/chat/ai-agent-chat.tsx
@@ -94,7 +94,7 @@ export const AIAgentChat = ({ agent }: AIAgentChatProps) => {
         </div>
       )}
 
-      <Conversation className="flex-1 min-h-0" initial="instant" resize="instant">
+      <Conversation className="min-h-0 flex-1" initial="instant" resize="instant">
         <ConversationContent>
           {isLoadingHistory && (
             <div className="flex h-full items-center justify-center">


frontend/src/components/pages/agents/details/a2a/chat/components/chat-input.tsx
```diff
@@ -75,20 +112,35 @@ export const ChatInput = ({
...
-                  <PromptInputModelSelectItem value={model}>
-                    <AIAgentModel model={model} size="sm" />
+                  <PromptInputModelSelectItem value={agent.model}>
+                    <AIAgentModel model={agent.model} size="sm" />
                   </PromptInputModelSelectItem>
                 </PromptInputModelSelectContent>
               </PromptInputModelSelect>
             )}
+            <Context
+              maxTokens={usage.max_input_tokens || 272_000}
```
malinskibeniamin: 272k is only for some models, should we just not show the context at all when `max_input_tokens === undefined`

frontend/package.json
```diff
@@ -61,6 +61,9 @@
     "@modelcontextprotocol/sdk": "^1.21.1",
     "@monaco-editor/react": "^4.7.0",
     "@primer/octicons-react": "^19.20.0",
+    "@radix-ui/react-hover-card": "^1.1.15",
+    "@radix-ui/react-progress": "^1.1.8",
+    "@radix-ui/react-slot": "^1.2.4",
```
malinskibeniamin: These packages can be replaced by updating the imports to just call `radix-ui`

context-usage-hover.png
malinskibeniamin, line 1: Let's remove this screenshot

frontend/src/components/ui/progress.tsx
malinskibeniamin, line 1: This component does exist in our registry, let's fetch it from https://redpanda-ui-registry.netlify.app/docs/progress instead.
```sh
bun x --bun @fumadocs/cli add --dir https://redpanda-ui-registry.netlify.app/r progress
```

frontend/src/components/ui/hover-card.tsx
malinskibeniamin, line 1: Let's remove this component in favor of our Redpanda UI one
```sh
bun x --bun @fumadocs/cli add --dir https://redpanda-ui-registry.netlify.app/r hover-card
```

Can be done in another PR.

frontend/src/components/ui/button.tsx
malinskibeniamin, line 1: Let's use Redpanda UI button instead.

frontend/src/components/pages/agents/details/a2a/chat/ai-agent-chat.tsx
```diff
@@ -40,6 +40,71 @@ export const AIAgentChat = ({ agent }: AIAgentChatProps) => {
...
+    }
+  }, [messages, handleCancelTask]);
+
+  const handleSubmitMessage = useCallback(
+    (message: Parameters<typeof handleSubmit>[0], event: React.FormEvent) => {
+      event.preventDefault();
+      void handleSubmit(message);
+    },
+    [handleSubmit]
+  );
```
malinskibeniamin: Not quite sure if I like this. Seems very overengineered around this part. I think we can get rid of it and only keep the `useMemo` bit, that should suffice. It feels like claude tried to throw everything at its disposal at this problem and I believe this doesn't really do anything

malinskibeniamin: If we want to avoid keystrokes rerendering anything, let's debounce the input instead.

malinskibeniamin: Example `use-debounce.tsx` hook
```tsx

/**
 */

    (...args: Parameters<T>) => {
      if (timeoutRef.current) {
      }

        callback(...args);
      }, delay);
    },
    [callback, delay]
  );

};

```

frontend/package.json
malinskibeniamin: Once you remove the new imports from the `ui` directory, we will be able to safely delete these radix UI dependencies and rerun `bun i && bun i --yarn`

context-usage-hover.png
birdayz: LOL indeed.

frontend/src/components/pages/agents/details/a2a/chat/components/chat-input.tsx
birdayz: Fixed. Context widget now only renders when max_input_tokens is provided by the model - no more hardcoded 272K fallback.

frontend/package.json
birdayz: Done. Switched to Redpanda UI registry components (button, progress, hover-card), removed individual @radix-ui/* packages, and reran bun i && bun i --yarn.

frontend/src/components/pages/agents/details/a2a/chat/ai-agent-chat.tsx
birdayz: Removed the useCallback wrappers - inlined the callbacks directly. Kept the useMemo for usage calculation.

frontend/src/components/ui/button.tsx
birdayz: Already done - this file was removed and we're using the Redpanda UI button component.

frontend/src/components/ai-elements/message.tsx
```diff
@@ -115,12 +116,22 @@ export const MessageTimestamp = ({
   );
 };
 
+// Format token count with compact notation (e.g., "1.2K", "5.6M")
+function formatTokenCount(tokens: number): string {
+  return new Intl.NumberFormat('en-US', {
+    notation: 'compact',
+    maximumFractionDigits: 1,
+  }).format(tokens);
+}
```
malinskibeniamin: Can we use `tokenlens` package to parse token count etc? https://www.npmjs.com/package/tokenlens

frontend/src/components/ai-elements/message.tsx
```diff
@@ -169,6 +184,25 @@ export const MessageMetadata = ({
...
+                  )}
+                  {outputTokens && outputTokens > 0 && (
+                    <span className="flex items-center gap-1">
+                      <ArrowDownIcon className="size-3" />
+                      {formatTokenCount(outputTokens)}
+                    </span>
+                  )}
+                </div>
+              </div>
+            )}
```
malinskibeniamin: Maybe let's create a component called `MessageTokenSection`, I would like to keep this somewhat separate

frontend/src/components/ai-elements/prompt-input.tsx
```diff
@@ -20,7 +20,7 @@ import {
   HoverCard,
   HoverCardContent,
   HoverCardTrigger,
-} from "components/ui/hover-card";
+} from "components/redpanda-ui/components/hover-card";
```
malinskibeniamin, line 23: 👍

frontend/src/components/pages/agents/details/a2a/chat/ai-agent-chat.tsx
```diff
@@ -40,8 +41,94 @@ export const AIAgentChat = ({ agent }: AIAgentChatProps) => {
...
+    return {
+      input_tokens: 0,
+      output_tokens: 0,
+      total_tokens: 0,
+      cumulativeInputTokens: 0,
+      cumulativeOutputTokens: 0,
+      cumulativeReasoningTokens: 0,
+      cumulativeCachedTokens: 0,
+    };
+  }, [messages]);
```
malinskibeniamin: Can we extract this to its own hook such as `useModelContextUsage()` or similar?

frontend/src/components/pages/agents/details/a2a/chat/components/chat-input.tsx
```diff
@@ -48,21 +61,45 @@ type ChatInputProps = {
...
+      cachedInputTokens: usage.cumulativeCachedTokens,
+      totalTokens: usage.cumulativeInputTokens + usage.cumulativeOutputTokens,
+    }),
+    [
+      usage.cumulativeInputTokens,
+      usage.cumulativeOutputTokens,
+      usage.cumulativeReasoningTokens,
+      usage.cumulativeCachedTokens,
+    ]
+  );
```
malinskibeniamin: Maybe if we have a dedicated hook for this, we can also reuse here

frontend/src/components/pages/agents/details/a2a/chat/components/chat-input.tsx
```diff
@@ -75,20 +112,37 @@ export const ChatInput = ({
...
-                  <PromptInputModelSelectItem value={model}>
-                    <AIAgentModel model={model} size="sm" />
+                  <PromptInputModelSelectItem value={agent.model}>
+                    <AIAgentModel model={agent.model} size="sm" />
                   </PromptInputModelSelectItem>
                 </PromptInputModelSelectContent>
               </PromptInputModelSelect>
             )}
+            <Context
+              maxTokens={usage.max_input_tokens || 1}
```
malinskibeniamin: don't fall back to `1`, I would say if `usage.max_input_tokens === undefined` just don't render `Context` component.

frontend/src/components/pages/agents/details/a2a/chat/components/chat-message.tsx
```diff
@@ -126,17 +132,72 @@ export const ChatMessage = ({ message, isLoading: _isLoading }: ChatMessageProps
...
+              <MessageBody>
+                {preTaskElements}
+                {taskElements}
+              </MessageBody>
+              <MessageMetadata
+                contextId={message.contextId}
+                from={message.role}
+                inputTokens={shouldShowTokens ? message.usage?.input_tokens : undefined}
+                messageId={message.id}
+                outputTokens={shouldShowTokens ? message.usage?.output_tokens : undefined}
```
malinskibeniamin: Instead of prop drilling I would just pass the entire `message` object at this point and simplify the props significantly, then you can handle the message `usage` property in metadata

frontend/src/components/pages/agents/details/a2a/chat/components/message-blocks/task-status-update-block.tsx
```diff
@@ -12,26 +12,38 @@ import { Artifact, ArtifactContent, ArtifactHeader, ArtifactTitle } from 'compon
...
  * Unified component for task status updates and agent messages
  * Shows state badge (only if actual state change), collapsible message text (if present)
  */
+// Format token count with compact notation (e.g., "1.2K", "5.6M")
+function formatTokenCount(tokens: number): string {
+  return new Intl.NumberFormat('en-US', {
+    notation: 'compact',
+    maximumFractionDigits: 1,
+  }).format(tokens);
+}
```
malinskibeniamin: I think this is duplicated, let's encapsulate in a utility function

frontend/src/components/pages/agents/details/a2a/chat/components/message-blocks/task-status-update-block.tsx
```diff
@@ -67,6 +81,25 @@ export const TaskStatusUpdateBlock = ({
             <span className="font-mono">{messageId}</span>
           </div>
         )}
+        {hasTokens && (
+          <div className="flex items-center gap-1.5">
+            <span className="font-medium">tokens:</span>
```
malinskibeniamin: use Typography component

frontend/src/components/redpanda-ui/lib/utils.ts
```diff
@@ -4,3 +4,7 @@ import { twMerge } from 'tailwind-merge';
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
+
+export type SharedProps = {
+  testId?: string;
+};
```
malinskibeniamin: This will get overridden - let's keep this file clean as-is, we can move the shared props elsewhere.

frontend/src/components/pages/agents/details/a2a/chat/components/chat-input.tsx
birdayz: Fixed - now rendering Context component only when max_input_tokens is defined.

frontend/src/components/redpanda-ui/lib/utils.ts
birdayz: Moved SharedProps to separate shared-types.ts file to avoid being overridden by registry updates.

frontend/src/components/pages/agents/details/a2a/chat/components/message-blocks/task-status-update-block.tsx
birdayz: Replaced span elements with Text component for consistent typography.

frontend/src/components/pages/agents/details/a2a/chat/components/chat-message.tsx
birdayz: Refactored to pass the entire message object instead of individual props. This eliminates prop drilling and simplifies the component interface.

frontend/src/components/ai-elements/message.tsx
birdayz: The formatTokenCount function here is just for displaying already-computed token counts in a compact format (1200 → "1.2K"). The tokenlens package is for estimating/counting tokens in text before making LLM API calls, which is a different use case. We already receive the token counts from the backend API responses, so we only need number formatting here.

frontend/src/components/pages/agents/details/a2a/chat/components/message-blocks/task-status-update-block.tsx
birdayz: Extracted the duplicated formatTokenCount function to utils/format-token-count.ts for reuse across both files.

frontend/src/components/pages/agents/details/a2a/chat/components/chat-input.tsx
birdayz: Created a dedicated useContextUsage hook to encapsulate the context usage transformation. This makes the logic reusable and keeps the component cleaner.

frontend/src/components/pages/agents/details/a2a/chat/ai-agent-chat.tsx
birdayz: Extracted to useCumulativeUsage hook. This encapsulates the logic for calculating cumulative token usage from messages and makes it reusable.

frontend/src/components/ai-elements/message.tsx
birdayz: Extracted the token display to a separate MessageTokenSection component to keep it isolated from the metadata layout logic.

PR #2050, opened 2025-11-27
Add ServiceAccount field to MCP v1 API

alenkacz:
Changes:
- Add ServiceAccount message to MCPServer in v1/mcp.proto
- Add service_account field to MCPServer, MCPServerCreate, and MCPServerUpdate


proto/redpanda/api/dataplane/v1/mcp.proto
```diff
@@ -407,6 +424,11 @@ message MCPServerCreate {
       string: {pattern: "^([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]*)$"}
     }
   }];
+
+  MCPServer.ServiceAccount service_account = 6 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true
+  ];
```
malinskibeniamin, line 431: This could be a breaking change I believe, should we make it optional for now and I can adjust the UI?

alenkacz: this is v1 api, it's not used yet, I think the goal is to migrate to it and let old versions degrade to run the mcp without tracing

birdayz (approved):

PR #2051, opened 2025-11-27
UX-720 configure playwright test agents & claude e2e skill

jvorcak:
We need to merge this one first https://github.com/redpanda-data/console/pull/2040


frontend/.claude/agents/playwright-test-generator.md
```diff
@@ -0,0 +1,59 @@
...
+tools: Glob, Grep, Read, mcp__playwright-test__browser_click, mcp__playwright-test__browser_drag, mcp__playwright-test__browser_evaluate, mcp__playwright-test__browser_file_upload, mcp__playwright-test__browser_handle_dialog, mcp__playwright-test__browser_hover, mcp__playwright-test__browser_navigate, mcp__playwright-test__browser_press_key, mcp__playwright-test__browser_select_option, mcp__playwright-test__browser_snapshot, mcp__playwright-test__browser_type, mcp__playwright-test__browser_verify_element_visible, mcp__playwright-test__browser_verify_list_visible, mcp__playwright-test__browser_verify_text_visible, mcp__playwright-test__browser_verify_value, mcp__playwright-test__browser_wait_for, mcp__playwright-test__generator_read_log, mcp__playwright-test__generator_setup_page, mcp__playwright-test__generator_write_test
```
malinskibeniamin, line 4: Will having so many tools exposed to the agent make it burn more context window tokens or get confusing or is this intentional?

frontend/.claude/agents/playwright-test-generator.md
```diff
@@ -0,0 +1,59 @@
...
+
+   ```ts file=add-valid-todo.spec.ts
+   // spec: specs/plan.md
+   // seed: tests/seed.spec.ts
+
+   test.describe('Adding New Todos', () => {
+     test('Add Valid Todo', async { page } => {
+       // 1. Click in the "What needs to be done?" input field
+       await page.click(...);
+
```
malinskibeniamin, line 54: I would add `test.step` so it also generates the logs instead of comments.

frontend/playwright.config.ts
```diff
@@ -52,6 +52,8 @@ export default defineConfig({
       name: 'chromium',
       use: {
         ...devices['Desktop Chrome'],
+        // Grant clipboard permissions for tests
+        permissions: ['clipboard-read', 'clipboard-write'],
```
malinskibeniamin, line 56: Keep in mind this would only work in Chrome, Firefox has different set of permissions regarding clipboard.

frontend/.claude/agents/playwright-test-generator.md
jvorcak: This was generated using this docs https://playwright.dev/docs/test-agents. I think most of these tools are used so that the `planner` agent can do its job.

frontend/tests/seed.spec.ts
malinskibeniamin, line 1: Great idea

frontend/playwright.config.ts
jvorcak: 👍 for now, I guess it's OK if we add coverage for Chrome only. That will reduce the number of regressions we have dramatically. Adding other platforms can always be done later on.

frontend/tests/schema.utils.ts
```diff
@@ -0,0 +1,177 @@
...
+  keyOrValue?: 'KEY' | 'VALUE';
+}
+
+export const createSchema = async (
+  page: Page,
+  { subjectName, schemaFormat = 'AVRO', schemaText, strategy = 'CUSTOM', keyOrValue }: CreateSchemaOptions
+) => {
+  return await test.step(`Create schema: ${subjectName}`, async () => {
+    await page.goto('/schema-registry');
+    await page.getByRole('button', { name: 'Create new schema' }).click();
```
malinskibeniamin, line 17: Can we add test ID or make the agent more likely to find the correct element to add `data-testid` attribute itself if it's missing?

malinskibeniamin: `getByRole` is usually making the test run a bit longer

frontend/tests/console/utils/topic-page.ts
```diff
@@ -0,0 +1,287 @@
...
+    await this.page.getByRole('button', { name: 'Cancel' }).click();
+  }
+
+  async verifyCreateButtonDisabled() {
+    await expect(this.page.getByTestId('onOk-button')).toBeDisabled();
+  }
+
+  async verifyCreateButtonEnabled() {
+    await expect(this.page.getByTestId('onOk-button')).toBeEnabled();
+  }
```
malinskibeniamin, line 111: This feels very verbose/overengineered, I imagine it was claude generated, is this the pattern we are meant to use or should we tweak it?

frontend/tests/console/utils/security-page.ts
```diff
@@ -0,0 +1,88 @@
+/** biome-ignore-all lint/performance/useTopLevelRegex: this is a test */
```
malinskibeniamin, line 1: Maybe we exclude the biome config from running against these playwright test files

frontend/tests/console/utils/security-page.ts
malinskibeniamin, line 1: should we use `kebab-case` instead of `PascalCase`

frontend/tests/schema.utils.ts
jvorcak: I think you're right, however I always liked when e2e which simulate the real user also go after a11y and test it as a side effect. Testing using roles does make sense to some extent. But since I've told Claude rules to add testids before testing, I think I should be consistent and add it here as well.

frontend/tests/console/utils/RolePage.ts
```diff
@@ -128,4 +128,51 @@ export class RolePage extends ACLPage {
...
+      await this.page.waitForURL('/security/roles/create', {
+        waitUntil: 'domcontentloaded',
+      });
+      await this.page.getByLabel('Role name').fill(roleName);
+      await this.page.getByTestId('roles-allow-all-operations').click({
+        force: true,
+      });
+      await this.page.getByRole('button').getByText('Create').click({
+        force: true,
+      });
```
malinskibeniamin: Usually if we force the click, it means there is something obstructing the view and you click "through" another element that is hiding the given DOM element you are targeting. Do we need to use force in the tests in this case?

frontend/tests/console/utils/topic-page.ts
jvorcak: I like the abstraction of using Pages, it somehow complies with DRY principle and I think it's suitable here. WDYT? We use these methods (or will use them) from multiple test suites.

frontend/tests/console/utils/debug-bundle-page.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  async verifyBrokerError(brokerId: number, expectedMessage?: string) {
+    const errorLabel = this.page.getByTestId(`broker-${brokerId}-error-label`);
+    await expect(errorLabel).toBeVisible();
+
+    if (expectedMessage) {
+      const errorMessage = this.page.getByTestId(`broker-${brokerId}-error-message`);
+      await expect(errorMessage).toHaveText(expectedMessage);
+    }
+  }
```
malinskibeniamin, line 344: I would love it if each of these self-contained functions could be wrapped in a `test.step` at some level, could be within each function or elsewhere.

frontend/tests/console/utils/debug-bundle-page.ts
```diff
@@ -0,0 +1,357 @@
...
+    return progressLink.isVisible({ timeout: 2000 }).catch(() => false);
+  }
+
+  async waitForBundleCompletion(timeout = 60_000) {
+    await Promise.race([
+      this.page.getByText(/complete|success|ready/i).waitFor({ timeout }),
+      this.page.getByText(/download/i).waitFor({ timeout }),
+      this.page.getByText(/error|failed/i).waitFor({ timeout }),
+    ]);
+  }
```
malinskibeniamin, line 269: Are we happy with just 1 of these promises resolving/succeeding?

frontend/tests/console/utils/debug-bundle-page.ts
```diff
@@ -0,0 +1,357 @@
...
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  /**
+   * Navigation methods
+   */
+  async goto() {
+    await this.page.goto('/debug-bundle');
+    await expect(this.page.getByRole('heading', { name: /debug bundle/i })).toBeVisible();
```
malinskibeniamin, line 20: Shouldn't we `waitForURL` instead and check `.includes('/debug-bundle')` instead?

frontend/tests/console/topics/topic-navigation.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+        'Compaction',
+        'Replication',
+        'Tiered Storage',
+        'Write Caching',
+        'Iceberg',
+        'Schema Registry and Validation',
+        'Message Handling',
+        'Compression',
+        'Storage Internals',
+      ];
```
malinskibeniamin: Do we have some wiggle room, for example if there is a new group provided then it would not break the test?

frontend/tests/console/topics/topic-messages-production.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      await valueEditor.click();
+      await page.keyboard.insertText(messageValue);
+
+      // Produce message
+      await page.getByTestId('produce-button').click();
+
+      // Verify message was produced - page should show the message
+      await expect(page.getByRole('cell', { name: new RegExp(messageValue, 'i') }).first()).toBeVisible({
+        timeout: 10_000,
+      });
```
malinskibeniamin, line 55: This seems unstable, with the timeout and getting cell role. I think we need a test ID here

frontend/tests/console/topics/topic-messages-production.spec.ts
```diff
@@ -0,0 +1,254 @@
+// spec: specs/topics.md
+// seed: tests/seed.spec.ts
+
+import { expect, test } from '@playwright/test';
+
+import { TopicPage } from '../utils/TopicPage';
+
+test.use({
+  permissions: ['clipboard-write', 'clipboard-read'],
+});
```
malinskibeniamin: We should also make sure we only run this test on Chromium browsers since otherwise the permissions (WebKit/Firefox) would break the test.

frontend/tests/console/utils/debug-bundle-page.ts
jvorcak: I'm more used to verify the contents of the page. The page can e.g. not render for some reason and URL will be correct.

frontend/tests/console/topics/topic-messages-filtering.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+    await topicPage.createTopic(topicName);
+
+    await test.step('View empty topic', async () => {
+      await page.goto(`/topics/${topicName}`);
+
+      // Should show empty state message
+      await expect(page.getByText(/No messages|empty/i).first()).toBeVisible({ timeout: 5000 });
+
+      // Produce button should still be available
+      await expect(page.getByRole('button', { name: /produce/i }).first()).toBeVisible({ timeout: 5000 });
```
malinskibeniamin, line 177: Is the timeout mainly because of ephemeral elements appearing/fading out of view?

frontend/tests/console/topics/topic-messages-production.spec.ts
jvorcak: We need to improve a11y and testability of our topics in general. But it's a wider topic.

frontend/tests/console/topics/topic-messages-filtering.spec.ts
malinskibeniamin: I think in Console we should establish the same approach as in Cloud UI playwright tests - if you adda PLAYWRIGHT e2e test flag, it will make the toasts never disappear etc.

frontend/tests/console/topics/topic-messages-filtering.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+
+      // Look for offset filter input
+      const offsetInput = page.getByPlaceholder(/offset/i);
+      if (await offsetInput.isVisible()) {
+        // Set start offset to 1 (skip first message)
+        await offsetInput.fill('1');
+        await page.keyboard.press('Enter');
+
+        // Wait for filtering
+        await page.waitForTimeout(1000);
```
malinskibeniamin, line 123: Is there any request happening here? Can we `waitForRequest` instead?

frontend/tests/console/topics/topic-messages-actions.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+// seed: tests/seed.spec.ts
+
+import { expect, test } from '@playwright/test';
+
+import fs from 'node:fs';
+import { TopicPage } from '../utils/TopicPage';
+
+test.use({
+  permissions: ['clipboard-write', 'clipboard-read'],
+});
```
malinskibeniamin: Same as other test - I think we should only run it in chromium, there is a way to add specify in playwright that this test only can run on a specific browser

malinskibeniamin (approved):

frontend/tests/console/topics/topic-messages-actions.spec.ts
jvorcak: This is already specified in a PW config, including Chrome device specifics.

PR #2054, opened 2025-12-01
refactor: use core proto ShadowLinkTaskStatus in console and dataplane APIs

c-julin:
## Summary
- Removes duplicate `TaskState` enum and `ShadowLinkTaskStatus` message definitions from console and dataplane protos
- Updates both console (`v1alpha1`) and dataplane (`v1alpha3`) shadowlink protos to use the shared `core.admin.v2.ShadowLinkTaskStatus` type
- Updates buf dependency `buf.build/redpandadata/core` to latest version
- Updates frontend imports and field accessors to match the core proto structure (`shard` instead of `shardId`)

## Test plan
- [x] Proto generation passes (`task proto:generate`)
- [x] Frontend linting passes (`bun run lint`)
- [x] Backend linting passes (`task backend:lint`)


.github/workflows/buf.yml
```diff
@@ -44,15 +44,16 @@ jobs:
...
           format: true
-          breaking: true
-          push: false  # Only validate, don't push to registry
-          # Compare against the default branch so the whole PR is checked
-          breaking_against: https://github.com/redpanda-data/console.git#branch=master
+          breaking: false
+          push: false
+      - uses: bufbuild/buf-setup-action@v1
+      - name: Buf – breaking (excluding imports)
+        run: buf breaking --against 'https://github.com/redpanda-data/console.git#branch=master' --exclude-imports
```
c-julin: adding this in but can be temporary  but because of the core breaking changes this rule fails

r-vasquez: You can also add a label to this PR and introduce a breaking change: https://buf.build/docs/bsr/ci-cd/github-actions/#skip-breaking-change-detection-using-labels, which can help us in the future as we introduce more changes from alpha/beta versions.

c-julin: great suggestion, have updated it :D

PR #2055, opened 2025-12-01
Use MCP v1 api and allow specifying service account

alenkacz:
Claude made this, I tested it. It supports both v1alpha3 and v1 apis.
https://redpanda.zoom.us/clips/share/n339UHtaT7e_HRFYVV9IGg

### DIFF
--- a/frontend/src/components/pages/agents/details/ai-agent-configuration-tab.tsx
+++ b/frontend/src/components/pages/agents/details/ai-agent-configuration-tab.tsx
@@ -43,10 +43,9 @@ import {
   AIAgentUpdateSchema,
   UpdateAIAgentRequestSchema,
 } from 'protogen/redpanda/api/dataplane/v1alpha3/ai_agent_pb';
-import type { MCPServer } from 'protogen/redpanda/api/dataplane/v1alpha3/mcp_pb';
 import { useCallback, useMemo, useState } from 'react';
 import { useGetAIAgentQuery, useUpdateAIAgentMutation } from 'react-query/api/ai-agent';
-import { useListMCPServersQuery } from 'react-query/api/remote-mcp';
+import { type MCPServer, useListMCPServersQuery } from 'react-query/api/remote-mcp';
 import { useListSecretsQuery } from 'react-query/api/secret';
 import { useParams } from 'react-router-dom';
 import { toast } from 'sonner';
--- a/frontend/src/components/pages/mcp-servers/create/remote-mcp-create-page.tsx
+++ b/frontend/src/components/pages/mcp-servers/create/remote-mcp-create-page.tsx
@@ -28,15 +28,7 @@ import { ExpandedYamlDialog } from 'components/ui/yaml/expanded-yaml-dialog';
 import { useYamlLabelSync } from 'components/ui/yaml/use-yaml-label-sync';
 import { config, isFeatureFlagEnabled } from 'config';
 import { ArrowLeft, FileText, Hammer, Loader2 } from 'lucide-react';
-import {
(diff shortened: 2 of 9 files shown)


birdayz:
Review from Claude, i hope it helps.

# PR Review: Use MCP v1 API and allow specifying service account

## Summary

Works for the happy path. Leaks resources on failure. Service account caching is broken. Using tags for system metadata is asking for trouble.

## Critical Bugs

### 1. Broken Service Account Caching (remote-mcp-create-page.tsx:202-207)

```typescript
const createServiceAccountIfNeeded = async (serverName: string) => {
  if (serviceAccountInfo) {
    return serviceAccountInfo; // Wrong - returns cached account regardless of form changes
  }
```

**Problem:** User creates server, hits validation error, changes service account name, resubmits. Gets the *old* service account with *new* tags. State is now corrupted.

**Fix:** Clear cache on form changes or remove caching entirely.

### 2. Resource Leak on Failure (remote-mcp-create-page.tsx:283-349)

Service account and secret are created *before* the MCP server. If `createServer()` fails (validation, network, whatever), you've leaked both resources. No cleanup, no rollback.

```typescript
const serviceAccountResult = await createServiceAccountIfNeeded(values.displayName);
// ... create secret, store IDs in tags
await createServer(...); // Fails here - account and secret already exist
```

**Fix:** Either make the backend handle this atomically or implement proper rollback in the frontend. Pick one.

### 3. Name Collision (remote-mcp-create-page.tsx:108-112)

```typescript
const sanitizedServerName = displayName.toLowerCase().replace(/[^a-z0-9-]/g, '-');
const generatedName = `${clusterType}-${config.clusterId}-mcp-${sanitizedServerName}-sa`;
```

"test", "Test", "test-server" all generate the same name. Second creation fails with a cryptic error.

**Fix:** Add timestamp or UUID suffix. Or validate uniqueness before form submission.

## Design Issues

### 4. Tags as System Metadata (remote-mcp-create-page.tsx:318-319)

```typescript
tagsMap.service_account_id = serviceAccountId;
tagsMap.secret_id = secretName;
```

You're storing required system state in user-editable tags. User deletes the tag, association is lost. Tags are for filtering and organization, not for required relationships.

**Fix:** Add proper fields to the MCP server schema for service account references.

### 5. Misleading Feature Flag

`enableMcpServiceAccount` both enables service accounts *and* switches from v1alpha3 to v1 API. That's two unrelated concerns under one flag.

Either rename it to something like `useMcpV1Api` or split it into two flags.

### 6. No Error Handling for Secret Creation (service-account-selector.tsx:118-129)

```typescript
await createSecret(
  create(CreateSecretRequestSchemaConsole, { ... })
);
// No try-catch, no error handling
```

Service account exists, secret creation fails silently. System is now inconsistent - account without credentials.

**Fix:** Wrap in try-catch, rollback service account creation on failure.

## Minor Issues

- Form validation always checks `serviceAccountName` errors even when flag is disabled (use-metadata-validation.ts:35)
- Loading state shows generic "Creating..." - should indicate what's being created
- ServiceAccountSection displays even if account was deleted (remote-mcp-configuration-tab.tsx:943)
- Type exports mix v1 and v1alpha3 inconsistently (remote-mcp.tsx:64-73)

## Testing

Zero tests. At minimum need:
- Service account creation failure path
- MCP server creation failure path
- Name collision handling
- Form validation with flag on/off

## Verdict

Fix bugs 1-3 before merging. Design issues should be addressed but aren't blockers for initial testing behind a feature flag.

The dual API version support is clean. Service account creation flow makes sense. But the error handling is incomplete and the resource lifecycle management is broken.

PR #2061, opened 2025-12-04
feat: add unified hooks for shadow link GET/EDIT with controlplane fallback

andresaristizabal:
## Summary
- Add `useGetShadowLinkUnified` hook with fallback to controlplane when dataplane fails
- Add `useEditShadowLink` hook for editing via controlplane in embedded mode
- Add `UnifiedShadowLink` types and transformation functions
- Add controlplane hooks (list, get by ID/name, update)
- Update details page components to use UnifiedShadowLink type
- Add warning banner when showing partial data from controlplane
- Add dataUnavailable state to tasks table for graceful degradation

## Test plan
- [ ] Verify shadow link details page loads correctly in both embedded and non-embedded modes
- [ ] Test fallback behavior when dataplane fails but controlplane succeeds (warning banner appears)
- [ ] Verify edit page uses controlplane API in embedded mode
- [ ] Run existing tests pass

### DIFF
--- a/frontend/src/components/constants.ts
+++ b/frontend/src/components/constants.ts
@@ -16,4 +16,5 @@ export const FEATURE_FLAGS = {
   enableAiAgentsInspectorInConsole: false,
   enableAiAgentsInConsoleServerless: false,
   enableMcpServiceAccount: false,
+  shadowlinkCloudUi: false,
 };
--- a/frontend/src/components/routes.tsx
+++ b/frontend/src/components/routes.tsx
@@ -596,7 +596,12 @@ export const APP_ROUTES: IRouteEntry[] = [
     'Shadow Links',
     (props) => <Shield {...props} />,
     true,
-    routeVisibility(true)
+    routeVisibility(() => {
+      if (isEmbedded()) {
+        return isFeatureFlagEnabled('shadowlinkCloudUi');
+      }
+      return true;
+    })
   ),
   MakeRoute<{}>('/shadowlinks/create', ShadowLinkCreatePage, 'Create Shadow Link'),
   MakeRoute<{ name: string }>('/shadowlinks/:name/edit', ShadowLinkEditPage, 'Edit Shadow Link'),


frontend/src/react-query/api/controlplane/shadowlink.tsx
```diff
@@ -0,0 +1,117 @@
...
+ * Creates its own controlplane transport, no TransportProvider needed
+ */
+export const useControlplaneGetShadowLinkByNameQuery = (request: { name: string }, opts?: { enabled?: boolean }) => {
+  // First, list to find ID by name
+  const listQuery = useControlplaneListShadowLinksQuery({ enabled: opts?.enabled });
+
+  const shadowLinkId = listQuery.data?.shadowLinks?.find((sl) => sl.name === request.name)?.id;
+
+  // Then fetch full details by ID
+  const getQuery = useControlplaneGetShadowLinkQuery({ id: shadowLinkId ?? '' }, { enabled: !!shadowLinkId });
```
r-vasquez, line 107: Should there be a way to return early if shadowLinkId is empty after iterating? perhaps the name you requested is not in the list returned from CP?

frontend/src/react-query/api/shadowlink.ts
```diff
@@ -224,3 +245,158 @@ export const useFailoverShadowLinkMutation = (options?: {
...
+      : undefined;
+  }
+
+  // In embedded mode: if dataplane fails but controlplane succeeds, use controlplane data
+  if (embedded && !unifiedData && controlplaneQuery.data) {
+    unifiedData = fromControlplaneShadowLink(controlplaneQuery.data);
+  }
+
+  // Combined error: only error if we have no data at all
+  const combinedError = unifiedData ? null : dataplaneQuery.error || controlplaneQuery.error;
```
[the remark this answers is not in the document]
andresaristizabal: this is right, I need return the error, even if unifiedData is present

frontend/src/react-query/api/controlplane/shadowlink.tsx
andresaristizabal: in that case enabled in useControlPlanegetShadow will be false, and this query doesn't happend.

r-vasquez (approved):

PR #2066, opened 2025-12-08
Add error codes for invalid AI agent / AI gateway config

tomasz-sadura:


proto/redpanda/api/dataplane/v1/error.proto
```diff
@@ -37,4 +37,10 @@ enum Reason {
 
   // Redpanda Schema Registry returned and error.
   REASON_REDPANDA_SCHEMA_REGISTRY_ERROR = 11;
+
+  // Invalid AI Agent configuration.
+  REASON_INVALID_AI_AGENT_CONFIGURATION = 12;
+
+  // Invalid AI Gateway configuration.
+  REASON_INVALID_AI_GATEWAY_CONFIGURATION = 13;
```
weeco: IMO a single INVALID_CONFIGURATION reason is precise enough. I think given the API request the client is already aware of what configuration would be invalid and we have an additional error message as well.

PR #2074, opened 2025-12-10
feat: add schema normalization support for Schema Registry

c-julin:
## Summary
- Add support for schema normalization parameter when registering schemas to Schema Registry
- Implement normalize toggle in schema creation UI with Switch component
- making the change in the existing rest api to get the feature out, part of the connect rpc move we should support all the params in the api.

### Backend
- **backend/pkg/api/handle_schema_registry.go**: Added `createSchemaRequest` type at package level with `params.normalize` field
- **backend/pkg/console/schema_registry.go**: Added `CreateSchemaRequestParams` and updated `CreateSchemaRegistrySchema` to pass normalize parameter to Schema Registry via `sr.WithParams(ctx, sr.Normalize)`
- **backend/pkg/console/servicer.go**: Updated interface signature to include params parameter
- **backend/pkg/api/handle_schema_registry_integration_test.go**: Added 4 comprehensive tests for normalize parameter with Protobuf enum ordering validation
- **backend/pkg/api/handle_schema_registry_test.go**: Added unit tests for normalize parameter parsing and serialization

### Frontend
- **frontend/src/components/pages/schemas/schema-create.tsx**: Added normalize Switch with tooltip positioned after schema editor, switch on right side of label text
- **frontend/src/state/rest-interfaces.ts**: Added optional `params.normalize` field to `SchemaRegistryCreateSchema` interface
- **frontend/src/components/pages/schemas/schema-create.test.tsx**: Added 6 tests for normalize parameter


backend/pkg/api/handle_schema_registry_integration_test.go
```diff
@@ -165,3 +165,278 @@ func (s *APIIntegrationTestSuite) TestValidateSchema() {
...
+		res1, body1 := s.apiRequest(ctx, http.MethodPost, "/api/schema-registry/subjects/test-normalize-enum-order/versions", req1)
+		require.Equal(200, res1.StatusCode, "first schema creation with normalize=true should succeed")
+
+		createResponse1 := struct {
+			ID int `json:"id"`
+		}{}
+		err := json.Unmarshal(body1, &createResponse1)
+		require.NoError(err)
+		firstSchemaID := createResponse1.ID
+		assert.Greater(firstSchemaID, 0, "first schema ID should be returned")
```
graham-rp: nitpick: seems like this could get turned into a helper fn

graham-rp (approved):

frontend/src/components/pages/schemas/schema-create.tsx
```diff
@@ -487,6 +493,24 @@ const SchemaEditor = observer((p: { state: SchemaEditorStateHelper; mode: 'CREAT
...
+                  <InfoIcon className="h-4 w-4 cursor-help text-gray-500" />
+                </TooltipTrigger>
+                <TooltipContent side="right">
+                  When enabled, the schema will be normalized to a canonical form before registration, reducing
+                  duplicate schema versions
+                </TooltipContent>
+              </Tooltip>
+            </TooltipProvider>
+          </Flex>
+          <Switch checked={state.normalize} onCheckedChange={(checked) => (state.normalize = checked === true)} />
```
malinskibeniamin: I think it would be better if we can split this line into 2
```
state.normalize = checked === true
```
so that we don't directly assign a result of an expression to the variable.

frontend/src/components/pages/schemas/schema-create.test.ts
malinskibeniamin, line 1: Can you try to rename to `.test.ts` and see if it works? I think this can be a unit test

PR #2086, opened 2025-12-15
frontend: fix service account deletion checkbox in MCP and AI agent

birdayz:
## What

Service account deletion checkbox now appears in both MCP server and AI agent delete dialogs.

## Why

Commit 4625b1f3b renamed cloud-managed tag keys from `service_account_id`/`secret_id` to `rp_cloud_service_account_id`/`rp_cloud_secret_id` but failed to update tag access code in `mcp-actions.tsx` and `ai-agent-actions.tsx`. Both components continued looking up the old tag names, which no longer exist. This caused `serviceAccountId` and `secretName` to always be null, making the conditional rendering check fail and hiding the checkbox entirely.

## Implementation details

Updated tag access to use `CLOUD_MANAGED_TAG_KEYS` constants instead of hardcoded old names.

**Before:**
```typescript
const serviceAccountId = server.tags?.service_account_id || null;
const secretName = server.tags?.secret_id || null;
```

**After:**
```typescript
const serviceAccountId = server.tags?.[CLOUD_MANAGED_TAG_KEYS.SERVICE_ACCOUNT_ID] || null;
const secretName = server.tags?.[CLOUD_MANAGED_TAG_KEYS.SECRET_ID] || null;
```

The conditional `{serviceAccountId && secretName && (` now properly evaluates to true when cloud-managed tags exist, making the Switch component render.

**Files changed:**
- `mcp-actions.tsx` - Fixed tag lookup
- `ai-agent-actions.tsx` - Fixed tag lookup  
- `ai-agent-actions.test.tsx` - Updated test to use new tag constants

## References

Broken by: 4625b1f3b
Original feature: 049bd538f

### DIFF
--- a/frontend/src/components/pages/agents/list/ai-agent-actions.test.tsx
+++ b/frontend/src/components/pages/agents/list/ai-agent-actions.test.tsx
@@ -12,6 +12,7 @@
 import { create } from '@bufbuild/protobuf';
 import { createRouterTransport } from '@connectrpc/connect';
 import userEvent from '@testing-library/user-event';
+import { CLOUD_MANAGED_TAG_KEYS } from 'components/constants';
 import {
   AIAgent_State,
   AIAgentSchema,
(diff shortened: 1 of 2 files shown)


andresaristizabal (approved):
Lgtm 

Question do why use [] access attribute? It looks like doesn't type safe

PR #2100, opened 2025-12-20
Fix infinite render loop in pipeline list pagination

birdayz:
## What

Fix UI hang when changing pagination in the pipelines list view.

## Why

The useInfiniteQueryWithAllPages hook included fetchNextPage in its useEffect dependency array. This function reference changes on every render, causing the effect to fire repeatedly and hang the browser.

## Implementation details

Remove fetchNextPage from the dependency array. The function reference is unstable but the function itself is stable - React Query guarantees this. Added eslint-disable with explanation.

Secondary change: memoize data transformations in useListPipelinesQuery to prevent unnecessary recalculations.

### DIFF
--- a/frontend/src/react-query/use-infinite-query-with-all-pages.ts
+++ b/frontend/src/react-query/use-infinite-query-with-all-pages.ts
@@ -50,18 +50,34 @@ export function useInfiniteQueryWithAllPages<
   // Use the standard ConnectRPC useInfiniteQuery hook
   const queryResult = useInfiniteQuery(schema, input, options);
 
+  // Auto-fetch all pages automatically when query executes.
+  // fetchNextPage intentionally excluded from deps - its reference changes every render
+  // but the function itself is stable. Including it causes infinite render loops.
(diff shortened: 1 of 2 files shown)


frontend/src/react-query/use-infinite-query-with-all-pages.ts
```diff
@@ -50,18 +50,34 @@ export function useInfiniteQueryWithAllPages<
...
+      queryResult.hasNextPage &&
+      !queryResult.isFetching &&
+      !queryResult.isFetchingNextPage &&
+      !queryResult.isError &&
+      queryResult.data
+    ) {
       queryResult.fetchNextPage();
     }
-  }, [queryResult.hasNextPage, queryResult.isFetching, queryResult.isFetchingNextPage, queryResult.fetchNextPage]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
malinskibeniamin, line 66: We don't use eslint - we use biome, this is not accurate

PR #2109, opened 2025-12-27
Misc/user management e2e tests

jvorcak:
~Needs to merge https://github.com/redpanda-data/console/pull/2107 first. Change base to `master` afterwards.~

### DIFF
--- a/frontend/tests/console/acls/user-management.spec.ts
+++ b/frontend/tests/console/acls/user-management.spec.ts
@@ -11,285 +11,357 @@ test.describe('ACL User Management', () => {
   });
 
   test('should create a new user with special characters in password', async ({ page }) => {
-    // 1. Click Create user button to open user creation dialog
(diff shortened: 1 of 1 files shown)


frontend/tests/console/acls/user-management.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+    // 6. Return to users list
+    await page.getByRole('button', { name: 'Done' }).click();
+    await expect(page).toHaveURL('/security/users');
+
+    // 7. Verify user appears in the list
+    await expect(page.getByRole('link', { name: username })).toBeVisible();
+  });
+
+  test('should toggle special characters checkbox and regenerate password', async ({ page }) => {
+    // 1. Navigate to create user page
```
malinskibeniamin: Can we make these all part of `step()` function? That way it would be something like `await page.step('1. Navigate to create user page', () => { /* some action goes here */ })`

PR #2110, opened 2025-12-28
Misc/remove mobx from acl list

jvorcak:
~Needs to merge https://github.com/redpanda-data/console/pull/2109.~

### DIFF
--- a/frontend/src/components/pages/acls/user-create.tsx
+++ b/frontend/src/components/pages/acls/user-create.tsx
@@ -42,6 +42,7 @@ import { useEffect, useMemo, useState } from 'react';
 import { Link as ReactRouterLink } from 'react-router-dom';
 
 import { useListRolesQuery } from '../../../react-query/api/security';
+import { invalidateUsersCache } from '../../../react-query/api/user';
 import { appGlobal } from '../../../state/app-global';
 import { api, rolesApi } from '../../../state/backend-api';
 import { AclRequestDefault, type CreateUserRequest } from '../../../state/rest-interfaces';
@@ -156,7 +157,6 @@ class UserCreatePage extends PageComponent {
       if (api.userData !== null && api.userData !== undefined && !api.userData.canListAcls) {
         return false;
       }
-      const { invalidateUsersCache } = await import('../../../react-query/api/user');
       await Promise.allSettled([api.refreshAcls(AclRequestDefault, true), invalidateUsersCache()]);
 
       // Add the user to the selected roles
--- a/frontend/src/components/pages/acls/user-details.tsx
+++ b/frontend/src/components/pages/acls/user-details.tsx
@@ -22,6 +22,7 @@ import { DeleteUserConfirmModal } from './delete-user-confirm-modal';
 import type { AclPrincipalGroup } from './models';
 import { ChangePasswordModal, ChangeRolesModal } from './user-edit-modals';
 import { useGetAclsByPrincipal } from '../../../react-query/api/acl';
+import { invalidateUsersCache } from '../../../react-query/api/user';
 import { appGlobal } from '../../../state/app-global';
 import { api, rolesApi } from '../../../state/backend-api';
 import { AclRequestDefault } from '../../../state/rest-interfaces';
@@ -129,7 +130,6 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
                     }
                   }
                   await Promise.allSettled(promises);
-                  const { invalidateUsersCache } = await import('../../../react-query/api/user');
                   await invalidateUsersCache();
                   await rolesApi.refreshRoleMembers();
                   appGlobal.historyPush('/security/users/');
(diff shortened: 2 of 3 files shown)


frontend/src/components/pages/acls/acl-list.tsx
```diff
@@ -102,121 +100,125 @@ const getCreateUserButtonProps = () => ({
...
+    },
+    isServerless()
+      ? null
+      : {
+          key: 'roles' as AclListTab,
+          name: 'Roles',
+          component: <RolesTab data-testid="roles-tab" />,
+          isDisabled:
+            (!Features.rolesApi && "Your cluster doesn't support this feature.") ||
+            (api.userData?.canManageUsers === false && 'You need RedpandaCapability.MANAGE_REDPANDA_USERS permission.'),
```
malinskibeniamin, line 173: Did you mean to interpolate this string? This is the message to display to the user

frontend/src/components/pages/acls/user-create.tsx
```diff
@@ -152,11 +152,12 @@ class UserCreatePage extends PageComponent {
         mechanism: this.mechanism,
       });
 
-      // Refresh user list
+      // Refresh user list and invalidate React Query cache
       if (api.userData !== null && api.userData !== undefined && !api.userData.canListAcls) {
         return false;
       }
-      await Promise.allSettled([api.refreshAcls(AclRequestDefault, true), api.refreshServiceAccounts()]);
```
malinskibeniamin, line 159: This dynamic import makes me slightly worried, why not just import at the top of the file?

frontend/src/components/pages/acls/user-details.tsx
```diff
@@ -129,7 +129,8 @@ class UserDetailsPage extends PageComponent<{ userName: string }> {
                     }
                   }
                   await Promise.allSettled(promises);
-                  await api.refreshServiceAccounts();
+                  const { invalidateUsersCache } = await import('../../../react-query/api/user');
```
malinskibeniamin: Same as above with dynamic import

frontend/src/react-query/api/user.tsx
```diff
@@ -52,6 +52,8 @@ export const useLegacyListUsersQuery = (
       return data;
     },
     enabled: options?.enabled,
+    refetchOnMount: 'always',
+    staleTime: 0,
```
malinskibeniamin: Is it expected we make the query stale right away?

frontend/src/react-query/api/user.tsx
```diff
@@ -153,3 +155,37 @@ export const useUpdateUserMutationWithToast = () => {
...
+  };
+};
+
+/**
+ * Non-hook function to invalidate users cache.
+ * Use this in class components or outside of React components.
+ * Uses dynamic import to get the query client.
+ */
+export const invalidateUsersCache = async () => {
+  const { default: queryClient } = await import('../../query-client');
```
malinskibeniamin: These dynamic imports are adding unnecessary complexity in my opinion

frontend/src/components/pages/acls/acl-list.tsx
jvorcak: https://github.com/redpanda-data/console/pull/2110/changes#diff-51aecada13c69f59d2238b39510e3c777bf5f4cd902133fc518e52ebe4012caeL157 It's in the original file, so this doesn't change it.

frontend/src/react-query/api/user.tsx
jvorcak: @malinskibeniamin you're right, fixed.

frontend/src/components/pages/acls/acl-list.tsx
malinskibeniamin: No worries - I think if we see some very easy wins like that, we can clean up a bit (but I do appreciate the PR and separation of concerns)

frontend/src/components/pages/acls/acl-list.tsx
```diff
@@ -102,121 +100,125 @@ const getCreateUserButtonProps = () => ({
...
+    };
+
+    appGlobal.onRefresh = async () => {
+      await refreshData();
+    };
+
+    // Initial data load
+    refreshData().catch(() => {
+      // Fail silently for now
     });
```
malinskibeniamin, line 126: Are we sure we don't want to have error handling here? The reason Biome complained about this is because promises should not float, they need to be handled, but if we leave a no op, it's cheating our linter a bit.

frontend/src/components/pages/acls/acl-list.tsx
```diff
@@ -507,9 +515,11 @@ const UserActions = ({ user }: { user: UsersEntry }) => {
   );
 };
 
-const RolesTab = observer(() => {
```
malinskibeniamin, line 510: Thank you

frontend/src/components/pages/acls/acl-list.tsx
jvorcak: Perhaps some logging, but I don't think we want to showcase this to the user

Issue #1815, opened 2025-06-23
Add Support for KIP-848 - The Next Generation of the Consumer Rebalance Protocol


theonlyguyy:
Hi Redpanda Console Team.
My company is currently in the process of migrating our Kafka clusters to KRaft. With KRaft clusters comes support for various new features in Kafka, one of them being [KIP-848](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol) - The Next Generation of the Consumer Rebalance Protocol.

We have been using Redpanda Console UI for our Kafka clusters, and have tried using this new protocol on a consumer group in the newest Redpanda Console UI version, but the UI shows the cosumer group to be in a DEAD state.

If this is a feature that currently exists, but has been missed by us, we would really appreciate if you could point out the config that we may have missed. Otherwise, if this feature is still not supported we would really appreciate if this could be added to the list of features to implement.

Thanks In advance for your help.

Environment Reference,

- Kafka version - 3.9.0
- Redpanda Console Version - 0.7.31

weeco:
Hey @theonlyguyy ,
thanks for raising this issue. I haven't been able to test this KIP yet. I'm interested in helping, but frankly I'm unsure when we'll actually going to implement it. If you are eager to help, I'm happy to assist.

A good first step would be creating an environment (docker compose) where we have both consumers using the old rebalance protocol and some next gen consumer group. I'm not entirely sure whether we already got everything we need from the Kafka client to support this, but it's already on twmb's list here: https://github.com/twmb/franz-go/issues/952

jtraxy:
Keen to know if this can be progressed.

amitelran-cyera:
Hey guys, any news about the progress?

PR #2112, opened 2025-12-29
feat: Add configurable support for Kafka next-generation balancer (franz-go)

tal-asulin:
Adds support for enabling Kafka’s next-generation balancer (a GA feature for Kafka ≥ 4.0.0) introduced in franz-go v1.19.0+. This feature provides improved partition assignment and rebalancing behavior.

By default, in the [franz-go](https://github.com/twmb/franz-go/blob/master/CHANGELOG.md#v1190) client, this feature is marked as hidden, as the maintainer noted potential instability concerns that have not yet received official recognition or confirmation.

This PR aims to resolve [issue #1815](https://github.com/redpanda-data/console/issues/1815).

### DIFF
--- a/docs/config/console.yaml
+++ b/docs/config/console.yaml
@@ -19,6 +19,7 @@
 #
 # Some examples:
 # kafka.rackId => KAFKA_RACKID
+# kafka.kafkaNextGenBalancer => KAFKA_KAFKANEXTGENBALANCER
 # kafka.tls.caFilepath => KAFKA_TLS_CAFILEPATH
 #
 # --- Note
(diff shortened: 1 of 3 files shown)


docs/config/console.yaml
```diff
@@ -38,6 +39,10 @@ kafka:
   # clientId: "console"
   # Optional: Rack identifier to optimize message consumption in multi-zone clusters.
   # rackId: "zone-a"
+  # Optional: Enable Kafka's next-generation balancer (beta feature from franz-go v1.19.0+).
+  # This enables improved partition assignment and rebalancing behavior.
```
weeco, line 43: I have to educate myself on the next-gen partition balancer, but I'm wondering how is this relevant to Console given that Console does not do any consumer group style consuming?

tal-asulin: Hi @weeco , sorry for the huge response delay. it does not actively consuming except of console consuming where we inspect the topics raw data.  therefor, In my opinion, enabling this feature does not presents any risk. would be happy to know your thoughts on that

weeco: We don't use consumer groups when consuming messages for presenting messages in the topics page. We use direct consumers

PR #2113, opened 2026-01-05
Bug fixes for new Connect UX

eblairmckee:
Fixes critical bugs discovered during bug bashing and holiday internal testing.

## Start/Stop pipeline button
The start/stop button was stuck in in-between states ("starting" and "stopping", which had been solved in legacy UI with retries, which were lacking the modern react-query implementation). Now:
- The button correctly shows state based on most recent query poll
- Pending states can be cancelled via a dropdown
    - Not sure when we'd use this, but users can also "retry" via the dropdown menu

## Refresh header
Removed the "refresh" button from the header on all of the new wizard routes (new RPCN wizard as well as API wizard). Given legacy routes just render new UI components, had to add additional `uiState` prop to disable refresh buttons from showing (only relevant for view/edit pipeline routes).

### onChange improvements
When modifying a pipeline name, description, etc. wouldn't correctly show inline errors until after submission. Given scroll position, wasn't always apparent why you couldn't add/edit a pipeline. So changed to firing zod validation `onChange`. And improved error messaging. 

### Missing fields
Some pipeline fields were missing (like url). Made values copyable. 

## Deleting a pipeline
event bubbling prevented users from deleting pipelines, fixed.

### DIFF
--- a/frontend/src/components/pages/rp-connect/pipelines-edit.tsx
+++ b/frontend/src/components/pages/rp-connect/pipelines-edit.tsx
@@ -17,7 +17,6 @@ import { action, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
 import { type Pipeline_ServiceAccount, PipelineUpdateSchema } from 'protogen/redpanda/api/dataplane/v1/pipeline_pb';
 import { Link } from 'react-router-dom';
-import { uiState } from 'state/ui-state';
 
 import { formatPipelineError } from './errors';
 import PipelinePage from './pipeline';
@@ -65,13 +64,8 @@ class RpConnectPipelinesEdit extends PageComponent<{ pipelineId: string }> {
     pipelinesApi.refreshPipelines(_force);
   }
 
-  componentWillUnmount() {
-    uiState.shouldHidePageHeader = false;
-  }
-
   render() {
     if (isFeatureFlagEnabled('enableRpcnTiles') && isEmbedded()) {
-      uiState.shouldHidePageHeader = true;
       return <PipelinePage />;
     }
     if (!pipelinesApi.pipelines) {
--- a/frontend/src/react-query/api/pipeline.tsx
+++ b/frontend/src/react-query/api/pipeline.tsx
@@ -34,7 +34,12 @@ import {
 } from 'protogen/redpanda/api/dataplane/v1/pipeline_pb';
 import type { Secret } from 'protogen/redpanda/api/dataplane/v1/secret_pb';
 import { useMemo } from 'react';
-import { MAX_PAGE_SIZE, type MessageInit, type QueryOptions } from 'react-query/react-query.utils';
+import {
+  MAX_PAGE_SIZE,
+  type MessageInit,
+  type QueryOptions,
+  SHORT_POLLING_INTERVAL,
+} from 'react-query/react-query.utils';
 import { formatToastErrorMessageGRPC } from 'utils/toast.utils';
 
 export const REDPANDA_CONNECT_LOGS_TOPIC = '__redpanda.connect.logs';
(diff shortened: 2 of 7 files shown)


frontend/src/components/pages/rp-connect/pipeline/details.tsx
```diff
@@ -25,49 +26,70 @@ import {
...
+const DetailRow = ({
+  label,
+  value,
+  copyable = false,
+}: {
+  label: React.ReactNode;
+  value?: string;
+  copyable?: boolean;
+}) => (
+  <div className="grid h-7 min-w-0 grid-cols-[minmax(0,1.5fr)_minmax(0,2fr)_30px] gap-1">
```
malinskibeniamin, line 55: How would this render on bigger viewports? I think we can make more use of the screen space

frontend/src/components/pages/rp-connect/pipeline/status-badge.tsx
```diff
@@ -0,0 +1,59 @@
+import { Badge } from 'components/redpanda-ui/components/badge';
+import { AlertCircle, Check, Loader2, Pause } from 'lucide-react';
+import { Pipeline_State } from 'protogen/redpanda/api/dataplane/v1/pipeline_pb';
+import { useMemo } from 'react';
+
+export const PipelineStatusBadge = ({ state }: { state?: Pipeline_State }) => {
+  const statusConfig = useMemo(() => {
```
malinskibeniamin, line 7: Do we need this use memo? Maybe it's overkill for this small component. How often do we expect pipeline state to change?

frontend/src/react-query/api/pipeline.tsx
```diff
@@ -54,8 +55,15 @@ export const useGetPipelineQuery = (
...
     enabled: options?.enabled,
-    refetchInterval: options?.refetchInterval,
-    refetchIntervalInBackground: options?.refetchIntervalInBackground,
+    refetchInterval:
+      options?.refetchInterval ??
+      ((query) => {
+        const state = query?.state?.data?.response?.pipeline?.state;
+        // Poll every 2 seconds when pipeline is in transitional state (STARTING or STOPPING)
+        const shouldPoll = state === Pipeline_State.STARTING || state === Pipeline_State.STOPPING;
+        return shouldPoll ? 2000 : false;
```
malinskibeniamin: Maybe we can store it as a const

frontend/src/components/pages/rp-connect/pipelines-edit.tsx
```diff
@@ -64,8 +65,13 @@ class RpConnectPipelinesEdit extends PageComponent<{ pipelineId: string }> {
     pipelinesApi.refreshPipelines(_force);
   }
 
+  componentWillUnmount() {
```
malinskibeniamin: Nice catch

malinskibeniamin (approved):

frontend/src/components/pages/rp-connect/pipeline/details.tsx
eblairmckee: holding off on that until the redesign. have a completely new UI I'm introducing this month. this is exclusively for P1 bug fixes

frontend/src/components/pages/rp-connect/pipeline/status-badge.tsx
eblairmckee: pretty often, the bug I'm fixing here is that the button was not responsive to state updates. feels like an easy optimization with no downside to use useMemo here.

frontend/src/components/pages/rp-connect/pipelines-edit.tsx
eblairmckee: removed and instead we determine this at the `header.tsx` level like everything else

PR #2114, opened 2026-01-05
run tests before dispatch

alenkacz:
This should first run some tests before we bump console enterprise.


weeco:
Good improvement on running some tests before dispatching. I think we can skip the heavy tests and/or run them in private repository which is always a superset of tests. 

Instead of using workflow_run as a trigger I believe we want to use https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows instead? I don't like the workflow trigger because it has an implicit trigger chain across the workflow yamls.

.github/workflows/repository-dispatch.yml
```diff
@@ -1,12 +1,10 @@
...
-  push:
-    paths:
-      - "frontend/**"
-      - "backend/**"
-    tags:
-      - '*'
+  workflow_run:
+    workflows: ["Backend verify", "Frontend Lint & Compile check"]
+    types:
+      - completed
```
weeco: I'd suggest to keep the trigger and use https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows for invoking the test/lint workflows to keep the trigger self-contained in this file.

weeco (approved):

PR #2117, opened 2026-01-06
backend/CO: skip non-existent topics from CO Overview

r-vasquez:
Right now we are returning a 404 if any of the
topics for ListEndOffsets response is non existent.

Franz-go returns a kerr.UnknownTopicOrPartition if the requested topic doesn't exists, so we can use
that and just skip these and return what is useful to the user.

### DIFF
--- a/backend/pkg/console/consumer_group_offsets_test.go
+++ b/backend/pkg/console/consumer_group_offsets_test.go
@@ -1,0 +1,207 @@
+// Copyright 2025 Redpanda Data, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0
+
+package console
+
+import (
+	"context"
+	"log/slog"
+	"os"
+	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


backend/pkg/console/consumer_group_offsets.go
```diff
@@ -98,10 +99,18 @@ func (s *Service) getConsumerGroupOffsets(ctx context.Context, adminCl *kadm.Cli
...
 	}
-	if topicsEndOffsets.Error() != nil {
+	if tErr := topicsEndOffsets.Error(); tErr != nil && !errors.Is(tErr, kerr.UnknownTopicOrPartition) {
 		return nil, fmt.Errorf("failed to list end offsets for topics: %w", topicsEndOffsets.Error())
 	}
 
+	// Collect topics that don't exist (franz-go return an UnknownTopicOrPartition for these).
+	nonExistentTopics := make(map[string]struct{})
+	topicsEndOffsets.Each(func(lo kadm.ListedOffset) {
+		if errors.Is(lo.Err, kerr.UnknownTopicOrPartition) {
```
c-julin, line 109: should we be skipping on any error ? probably one to keep in mind for when we update the api, we probably want to return these errors

r-vasquez: Yeah, I think is better to keep this for consistency with previous behavior, but when we migrate to the new API include an error string.

PR #2119, opened 2026-01-07
Add tracing service with in memory span store

weeco:
This PR covers the tracing UI that is served by the new TracingService RPC. The tracing sidebar item is hidden/shown conditionally depending on the console backend config.

ss-wK4n4MEo@2x

### DIFF
--- a/frontend/src/components/pages/traces/components/traces-table.tsx
+++ b/frontend/src/components/pages/traces/components/traces-table.tsx
@@ -180,6 +180,114 @@ const getSpanRowLineFlags = (depth: number, isLastChild: boolean, parentDepths: 
   };
 };
 
+// Tree gutter column component - renders a single column in the tree structure
+type GutterColumnProps = {
+  colIndex: number;
+  depth: number;
+  isLastChild: boolean;
+  parentDepths: number[];
+};
+
+const GutterColumn: FC<GutterColumnProps> = ({ colIndex, depth, isLastChild, parentDepths }) => {
+  const isCurrentColumn = colIndex === depth - 1;
+  const drawAncestorContinuation = parentDepths.includes(colIndex);
+
+  return (
+    <div
+      className="relative flex w-6 items-center justify-center"
+      style={{ '--tree-x': '11px' } as React.CSSProperties}
+    >
+      {!!drawAncestorContinuation && (
+        <div className="absolute top-0 bottom-0 w-px bg-border" style={{ left: 'var(--tree-x)' }} />
+      )}
+      {!!isCurrentColumn && (
+        <>
+          <div
+            className="absolute top-0 w-px bg-border"
+            style={{ left: 'var(--tree-x)', height: isLastChild ? '50%' : 'calc(100% + 1px)' }}
+          />
+          <div className="absolute top-1/2 h-px w-[13px] bg-border" style={{ left: 'var(--tree-x)' }} />
+        </>
+      )}
+    </div>
+  );
+};
+
+// Tree lines component - renders the full tree structure for a span row
+type TreeLinesProps = {
+  depth: number;
+  isLastChild: boolean;
+  parentDepths: number[];
+  isExpanded: boolean;
+  hasChildren: boolean;
+  onToggle: () => void;
+};
+
+const TreeLines: FC<TreeLinesProps> = ({ depth, isLastChild, parentDepths, isExpanded, hasChildren, onToggle }) => {
+  const lineFlags = getSpanRowLineFlags(depth, isLastChild, parentDepths);
+
+  return (
+    <div className="flex h-8 shrink-0 items-stretch">
+      {/* First column (gutter column 0) */}
+      <div
+        className="relative flex w-5 items-center justify-center"
+        style={{ '--tree-x': '9px' } as React.CSSProperties}
+      >
+        {!!lineFlags.drawCol0Vertical && (
+          <div
+            className="absolute top-0 w-px bg-border"
+            style={{ left: 'var(--tree-x)', height: lineFlags.col0VerticalHeight }}
+          />
+        )}
+        {depth === 1 && <div className="absolute top-1/2 h-px w-[11px] bg-border" style={{ left: 'var(--tree-x)' }} />}
+      </div>
+
+      {/* Additional gutter columns (1..depth-1) */}
+      {Array.from({ length: Math.max(0, depth - 1) }).map((_, i) => {
+        const colIndex = i + 1;
+        return (
+          <GutterColumn
+            colIndex={colIndex}
+            depth={depth}
+            isLastChild={isLastChild}
+            key={`gutter-col-${colIndex}`}
+            parentDepths={parentDepths}
+          />
+        );
+      })}
+
+      {/* Chevron button as final tree column */}
+      <div
+        className="relative flex h-8 w-5 shrink-0 items-center"
+        style={{ '--tree-x': '11px' } as React.CSSProperties}
+      >
+        {!!(isExpanded && hasChildren) && (
+          <div className="absolute top-1/2 bottom-0 w-px bg-border" style={{ left: 'var(--tree-x)' }} />
+        )}
+        <Button
+          aria-expanded={hasChildren ? isExpanded : undefined}
+          aria-label={hasChildren ? `${isExpanded ? 'Collapse' : 'Expand'} child spans` : undefined}
+          className={cn('absolute z-10 h-4 w-4 shrink-0 -translate-x-1/2', !hasChildren && 'invisible')}
+          onClick={(e) => {
+            e.stopPropagation();
+            onToggle();
+          }}
+          size="icon"
+          style={{ left: 'var(--tree-x)' }}
+          tabIndex={hasChildren ? 0 : -1}
+          variant="ghost"
+        >
+          {isExpanded ? (
+            <ChevronDown aria-hidden="true" className="h-3 w-3" />
+          ) : (
+            <ChevronRight aria-hidden="true" className="h-3 w-3" />
+          )}
+        </Button>
+      </div>
+    </div>
+  );
+};
+
 // Span Row Component with tree lines
 type SpanRowProps = {
   span: SpanNode;
@@ -223,9 +331,6 @@ const SpanRow: FC<SpanRowProps> = ({
   const Icon = getIconForServiceName(span.span);
   const serviceName = getServiceName(span.span);
 
-  // Calculate tree line connector flags
-  const lineFlags = getSpanRowLineFlags(depth, isLastChild, parentDepths);
-
   return (
     <>
       <button
@@ -254,88 +359,14 @@ const SpanRow: FC<SpanRowProps> = ({
         <div className="flex min-w-0 items-center gap-1 overflow-hidden px-1 py-1">
           {/* Tree structure lines */}
           {depth > 0 && (
-            <div className="flex h-8 shrink-0 items-stretch">
-              {/* First column (gutter column 0) - matches root row's w-5 container where the chevron lives */}
-              <div
-                className="relative flex w-5 items-center justify-center"
-                style={{ '--tree-x': '9px' } as React.CSSProperties}
-              >
-                {/* Continuation vertical (only when needed) */}
-                {!!lineFlags.drawCol0Vertical && (
-                  <div
-                    className="absolute top-0 w-px bg-border"
-                    style={{ left: 'var(--tree-x)', height: lineFlags.col0VerticalHeight }}
-                  />
-                )}
-
-                {/* Horizontal connector at depth 1 */}
-                {depth === 1 && (
-                  <div className="absolute top-1/2 h-px w-[11px] bg-border" style={{ left: 'var(--tree-x)' }} />
-                )}
-              </div>
-
-              {/* Additional gutter columns (1..depth-1) */}
-              {Array.from({ length: Math.max(0, depth - 1) }).map((_, i) => {
-                // i=0 -> gutter column 1, i=1 -> gutter column 2, ...
-                const colIndex = i + 1;
-                const isCurrentColumn = colIndex === depth - 1;
-                const drawAncestorContinuation = parentDepths.includes(colIndex);
-
-                return (
-                  <div
-                    className="relative flex w-6 items-center justify-center"
-                    key={colIndex}
-                    style={{ '--tree-x': '11px' } as React.CSSProperties}
-                  >
-                    {/* Full-height continuation for ancestor columns */}
-                    {!!drawAncestorContinuation && (
-                      <div className="absolute top-0 bottom-0 w-px bg-border" style={{ left: 'var(--tree-x)' }} />
-                    )}
-
-                    {/* Current node column: vertical (full or half) + horizontal connector */}
-                    {!!isCurrentColumn && (
-                      <>
-                        <div
-                          className="absolute top-0 w-px bg-border"
-                          style={{ left: 'var(--tree-x)', height: isLastChild ? '50%' : 'calc(100% + 1px)' }}
-                        />
-                        <div className="absolute top-1/2 h-px w-[13px] bg-border" style={{ left: 'var(--tree-x)' }} />
-                      </>
-                    )}
-                  </div>
-                );
-              })}
-
-              {/* Chevron button as final tree column with vertical line connector */}
-              <div
-                className="relative flex h-8 w-5 shrink-0 items-center"
-                style={{ '--tree-x': '11px' } as React.CSSProperties}
-              >
-                {/* Vertical line connecting to chevron when expanded and has children */}
-                {!!(isExpanded && hasChildren) && (
-                  <div className="absolute top-1/2 bottom-0 w-px bg-border" style={{ left: 'var(--tree-x)' }} />
-                )}
-                <Button
-                  aria-expanded={hasChildren ? isExpanded : undefined}
-                  aria-label={hasChildren ? `${isExpanded ? 'Collapse' : 'Expand'} child spans` : undefined}
-                  className={cn('absolute z-10 h-4 w-4 shrink-0 -translate-x-1/2', !hasChildren && 'invisible')}
-                  onClick={(e) => {
-                    e.stopPropagation();
-                    onToggle();
-                  }}
-                  size="icon"
-                  style={{ left: 'var(--tree-x)' }}
-                  tabIndex={hasChildren ? 0 : -1}
-                  variant="ghost"
-                >
-                  {isExpanded ? (
-                    <ChevronDown aria-hidden="true" className="h-3 w-3" />
-                  ) : (
-                    <ChevronRight aria-hidden="true" className="h-3 w-3" />
-                  )}
-                </Button>
-              </div>
-            </div>
+            <TreeLines
+              depth={depth}
+              hasChildren={hasChildren}
+              isExpanded={isExpanded}
+              isLastChild={isLastChild}
+              onToggle={onToggle}
+              parentDepths={parentDepths}
+            />
           )}
 
           {/* Service badge */}


frontend/package.json
```diff
@@ -114,6 +114,7 @@
     "react-router-dom": "^6.3.0",
     "react-simple-code-editor": "^0.14.1",
     "react-syntax-highlighter": "^15.6.6",
+    "recharts": "^3.6.0",
```
malinskibeniamin: Have you tried to use latest recharts package alongside our shadcn setup? https://github.com/shadcn-ui/ui/issues/7669

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,186 @@
...
+  span: Span;
+}
+
+interface AttributeEntry {
+  key: string;
+  value: string;
+}
+
+// Recursively extract the actual value from protobuf-es structure
+const extractProtoValue = (value: unknown): unknown => {
```
malinskibeniamin: Would it be useful to use `protovalidate-es` package for this

frontend/src/react-query/api/tracing.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright 2025 Redpanda Data, Inc.
```
malinskibeniamin: 2026?

frontend/src/react-query/api/tracing.ts
```diff
@@ -0,0 +1,39 @@
...
+import { getTrace, listTraces } from 'protogen/redpanda/api/dataplane/v1alpha3/tracing-TracingService_connectquery';
+import type { MessageInit } from 'react-query/react-query.utils';
+
+export const useListTracesQuery = (
+  request: MessageInit<ListTracesRequest>,
+  opts?: { enabled?: boolean; refetchInterval?: number | false }
+) => {
+  const listTracesRequest = create(ListTracesRequestSchema, request);
+  return useQuery(listTraces, listTracesRequest, opts);
+};
```
malinskibeniamin, line 31: How many traces do we expect. This will at most pull 500 entries. Do we need infinite query in this scenario. That way we could prompt people to give them "load more" button at least.

What's the likely scenario, how many resources do we expect

frontend/src/protogen/redpanda/api/dataplane/v1alpha3/tracing_pb.ts
```diff
@@ -0,0 +1,267 @@
...
+  /**
+   * End time for trace query (exclusive). Traces that started before this time will be included.
+   * Optional: defaults to current time + 5 minutes if not provided.
+   *
+   * @generated from field: google.protobuf.Timestamp end_time = 2;
+   */
+  endTime?: Timestamp;
+
+  /**
+   * Maximum number of traces to return (pagination). Defaults to 100 if not specified or set to 0.
```
malinskibeniamin: is it 100 or 500? usually in console we had 500

frontend/src/components/pages/traces/utils/trace-statistics.ts
```diff
@@ -0,0 +1,103 @@
+/**
+ * Copyright 2025 Redpanda Data, Inc.
```
malinskibeniamin: copyright = 2026

frontend/src/components/pages/traces/utils/trace-formatters.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+  if (diff < 3_600_000) {
+    return `${Math.floor(diff / 60_000)}m ago`;
+  }
+  if (diff < 86_400_000) {
+    return `${Math.floor(diff / 3_600_000)}h ago`;
+  }
+
+  return new Date(ms).toLocaleString();
+};
```
malinskibeniamin: Can we tell Claude to use `date-fns` package for handling these. You can use some built-in utility functions that would do this for you.

For natural language, you can use https://github.com/wanasit/chrono

frontend/src/components/pages/traces/utils/span-tree-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+    const duration = Math.max(0, Number((span.endTimeUnixNano - span.startTimeUnixNano) / 1_000_000n));
+
+    nodeMap.set(spanId, {
+      spanId,
+      parentSpanId,
+      name: span.name,
+      startTime: span.startTimeUnixNano,
+      endTime: span.endTimeUnixNano,
+      duration,
+      hasError: span.status?.code === 2,
```
malinskibeniamin: Let's use the enum values instead of numbers, I think code === 2 here can be mapped to the relevant enum error code

frontend/src/components/pages/traces/utils/span-tree-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+    visiting.add(node.spanId);
+    node.children.sort((a, b) => {
+      if (a.startTime < b.startTime) {
+        return -1;
+      }
+      if (a.startTime > b.startTime) {
+        return 1;
+      }
+      return 0;
+    });
```
malinskibeniamin, line 79: You can compare dates using `date-fns` to see which one comes 1st, for example https://date-fns.org/v4.1.0/docs/isBefore

frontend/src/components/pages/traces/utils/span-tree-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const traverse = (node: SpanNode) => {
+    if (node.startTime < minTime) {
+      minTime = node.startTime;
+    }
+    if (node.endTime > maxTime) {
+      maxTime = node.endTime;
+    }
+    node.children.forEach(traverse);
+  };
```
malinskibeniamin, line 106: Same as other comment, comparing timestamps/dates can happen with `date-fns` too

frontend/src/components/pages/traces/utils/hex-utils.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+export const bytesToHex = (bytes: Uint8Array): string =>
+  Array.from(bytes)
+    .map((b) => b.toString(16).padStart(2, '0'))
+    .join('');
```
malinskibeniamin: This may give us some performance hit if it's a large dataset, https://stackoverflow.com/questions/40031688/how-can-i-convert-an-arraybuffer-to-a-hexadecimal-string-hex

frontend/src/components/pages/traces/trace-list-page.tsx
```diff
@@ -0,0 +1,317 @@
...
+            }
+            placeholder="Search traces..."
+            value={(table.getColumn('searchable')?.getFilterValue() as string) ?? ''}
+          />
+          {table.getColumn('serviceName') && serviceOptions.length > 0 && (
+            <DataTableFacetedFilter column={table.getColumn('serviceName')} options={serviceOptions} title="Service" />
+          )}
+          {table.getColumn('status') && (
+            <DataTableFacetedFilter column={table.getColumn('status')} options={statusOptions} title="Status" />
+          )}
```
malinskibeniamin: 👏

frontend/src/components/pages/traces/trace-list-page.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+    return {
+      startTimestamp: create(TimestampSchema, {
+        seconds: BigInt(Math.floor(startTime / 1000)),
+        nanos: (startTime % 1000) * 1_000_000,
+      }),
+      endTimestamp: create(TimestampSchema, {
+        seconds: BigInt(Math.floor(nowMs / 1000)),
+        nanos: (nowMs % 1000) * 1_000_000,
+      }),
```
malinskibeniamin: I think you can use timestamp proto `.toDate()` or `.fromDate()`

malinskibeniamin: then you won't have to be dealing with all this math for seconds/nanos

frontend/src/components/pages/traces/trace-list-page.tsx
```diff
@@ -0,0 +1,317 @@
...
+const TIME_RANGES = [
+  { value: '5m', label: 'Last 5 minutes', ms: 5 * 60 * 1000 },
+  { value: '15m', label: 'Last 15 minutes', ms: 15 * 60 * 1000 },
+  { value: '30m', label: 'Last 30 minutes', ms: 30 * 60 * 1000 },
+  { value: '1h', label: 'Last 1 hour', ms: 60 * 60 * 1000 },
+  { value: '3h', label: 'Last 3 hours', ms: 3 * 60 * 60 * 1000 },
+  { value: '6h', label: 'Last 6 hours', ms: 6 * 60 * 60 * 1000 },
+  { value: '12h', label: 'Last 12 hours', ms: 12 * 60 * 60 * 1000 },
+  { value: '24h', label: 'Last 24 hours', ms: 24 * 60 * 60 * 1000 },
+];
```
malinskibeniamin, line 56: I think `date-fns` would help here too

frontend/src/components/pages/traces/utils/trace-formatters.ts
weeco: Do we really need to pull in a new dependency for ~40LOC util functionality? Feels bad but maybe it's just me being more used to the way of working in Go codebases where we try to avoid external dependencies

Edit: I saw the other comments, where date-fns would also be helpful, so maybe still worth it then. But seems like mostly simple util funcs

frontend/src/components/pages/traces/components/trace-details-sheet.tsx
```diff
@@ -0,0 +1,230 @@
...
+                  <code className="rounded bg-muted px-1 py-0.5 text-xs">{trace?.traceId?.slice(0, 8)}...</code>
+                </div>
+                <div className="flex justify-between">
+                  <span className="text-muted-foreground">Received Spans</span>
+                  <span>{trace?.summary?.spanCount || 0}</span>
+                </div>
+              </div>
+            </div>
+          </div>
+        )}
```
malinskibeniamin, line 219: I think we need a skill called `typography-refactor` so that we use all the components provided by the `typography.tsx` component. Most of this CSS should be inferred from typography elements

frontend/src/components/pages/traces/components/trace-details-sheet.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  // Cleanup timeout on unmount
+  useEffect(
+    () => () => {
+      if (copyTimeoutRef.current) {
+        clearTimeout(copyTimeoutRef.current);
+      }
+    },
+    []
+  );
```
malinskibeniamin: Why do we need this? I think this is opting out of some of the React renderer behavior, kind of hard to grasp, I would like to understand what's the problem we tried to solve

frontend/src/components/pages/traces/components/trace-activity-chart.tsx
```diff
@@ -0,0 +1,120 @@
...
+      {/* Histogram Bars */}
+      <div className="flex h-10 items-end gap-px">
+        {chartData.map((bucket, i) => {
+          const height = maxCount > 0 ? (bucket.count / maxCount) * 100 : 0;
+          return (
+            <div
+              className="min-w-[2px] flex-1 rounded-t-sm bg-emerald-500/70 transition-all hover:bg-emerald-500"
+              key={i}
+              style={{ height: `${height}%` }}
+              title={`${bucket.count} traces`}
```
malinskibeniamin: we should use `pluralize('trace')`

frontend/src/components/pages/traces/components/tool-call-tab.tsx
```diff
@@ -0,0 +1,213 @@
...
+      )}
+
+      {/* Arguments */}
+      {toolData.hasArguments && (
+        <div className="space-y-1.5">
+          <h5 className="font-medium text-[10px] text-muted-foreground uppercase tracking-wide">ARGUMENTS</h5>
+          <div className="rounded border bg-muted/30 p-2">
+            {toolData.isArgumentsJson ? (
+              <pre className="overflow-x-auto">
+                <code className="font-mono text-[10px] leading-relaxed">{formatJsonContent(toolData.arguments)}</code>
```
malinskibeniamin: I think we could use dynamic code block

frontend/src/components/pages/traces/components/tool-call-tab.tsx
```diff
@@ -0,0 +1,213 @@
...
+    case 'doubleValue':
+      return String(attr.value.value.value);
+    case 'boolValue':
+      return attr.value.value.value ? 'true' : 'false';
+    default:
+      return '';
+  }
+};
+
+const MAX_PAYLOAD_SIZE = 20 * 1024; // 20KB
```
malinskibeniamin: I think we have storage utils for this, we can try to use it

frontend/src/components/pages/traces/components/tool-call-tab.tsx
```diff
@@ -0,0 +1,213 @@
...
+      return attr.value.value.value;
+    case 'intValue':
+      return String(Number(attr.value.value.value));
+    case 'doubleValue':
+      return String(attr.value.value.value);
+    case 'boolValue':
+      return attr.value.value.value ? 'true' : 'false';
+    default:
+      return '';
+  }
```
malinskibeniamin: the `value.value.value` chaining could be improved, do we also need some defensive checks to know what type we are dealing with so it's safe to convert to a number? maybe `isNaN` or something

frontend/src/components/pages/traces/components/raw-json-tab.tsx
```diff
@@ -0,0 +1,60 @@
...
+              <Check className="h-4 w-4" />
+              Copied!
+            </>
+          ) : (
+            <>
+              <Copy className="h-4 w-4" />
+              Copy JSON
+            </>
+          )}
+        </Button>
```
malinskibeniamin: You can use `CopyButton`

frontend/src/components/pages/traces/components/raw-json-tab.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useState } from 'react';
+
+interface Props {
+  span: Span;
+}
+
+export const RawJSONTab: FC<Props> = ({ span }) => {
+  const [copied, setCopied] = useState(false);
+
+  const jsonString = JSON.stringify(toJson(SpanSchema, span), null, 2);
```
malinskibeniamin: Should we 1st verify it's safe to convert to JSON? Is there any risk of some corrupted data?

frontend/src/components/pages/traces/components/overview-tab.tsx
```diff
@@ -0,0 +1,122 @@
...
+          </div>
+          <div className="rounded border bg-muted/30 p-2">
+            <div className="text-[10px] text-muted-foreground">Duration</div>
+            <div className="font-mono font-semibold text-sm">{formatDuration(Number(summary.durationMs))}</div>
+          </div>
+        </div>
+        {summary.errorCount > 0 && (
+          <div className="rounded border bg-red-500/10 p-2">
+            <div className="text-[10px] text-muted-foreground">Error Count</div>
+            <div className="font-mono font-semibold text-red-600 text-sm">{summary.errorCount}</div>
```
malinskibeniamin: I think if we use typography we can rely on a destructive variant here.

frontend/src/components/pages/traces/components/overview-tab.tsx
```diff
@@ -0,0 +1,122 @@
...
+    return (
+      <div className="space-y-4 p-4">
+        <div className="text-muted-foreground text-sm">No trace summary available</div>
+      </div>
+    );
+  }
+
+  return (
+    <div className="space-y-4 p-3">
+      {/* Trace Summary Section */}
```
malinskibeniamin, line 74: I think it warrants to have a separate component for each, even if it feels like overkill.

frontend/src/components/pages/traces/components/conversation-tab.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export const ConversationTab: FC<Props> = ({ trace }) => {
+  const messages = useMemo(() => extractConversationHistory(trace), [trace]);
+
+  if (messages.length === 0) {
+    return (
+      <div className="rounded bg-muted/10 p-8 text-center text-muted-foreground">
+        No conversation history found in this trace
+      </div>
```
malinskibeniamin: Can we use `Empty` component in this case

malinskibeniamin: https://redpanda-ui-registry.netlify.app/docs/empty

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,186 @@
...
+    return String(extractedValue);
+  }
+
+  if (typeof extractedValue === 'bigint') {
+    return String(extractedValue);
+  }
+
+  // For arrays and objects, pretty-print as JSON
+  try {
+    return JSON.stringify(extractedValue, (_, v) => (typeof v === 'bigint' ? String(v) : v), 2);
```
malinskibeniamin: If it's bigint, I think we can use
```tsx
import JSONBigIntFactory from 'json-bigint';

const JSONBigInt = JSONBigIntFactory({ storeAsString: true });
```

This was introduced by @jvorcak

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,186 @@
...
+        return protoValue.value.value;
+      case 'bytesValue':
+        // For bytes, try to decode as string
+        try {
+          if (protoValue.value.value instanceof Uint8Array) {
+            return new TextDecoder().decode(protoValue.value.value);
+          }
+          return protoValue.value.value;
+        } catch {
+          return '[binary data]';
```
malinskibeniamin: Should we also have an error to provide, not just catch {}

frontend/src/components/constants.ts
```diff
@@ -17,6 +17,7 @@ export const FEATURE_FLAGS = {
   enableAiAgentsInConsoleServerless: false,
   enableMcpServiceAccount: false,
   enablePipelineServiceAccount: false,
+  enableTracingInConsole: true,
```
malinskibeniamin, line 20: Do we need separate FF for console and serverless console

frontend/src/react-query/api/tracing.ts
weeco: I think 100 is quite a lot already. I increased to 500 for now, just the trace summary is loaded initially and populated. The spans with the heavy data that belong to a trace is lazy loaded upon expansion.

frontend/src/components/constants.ts
weeco: Don't think so. It's controlled via the backend config anyways and that could be handled differently.

bun.lock
malinskibeniamin, line 1: I think this is messed up, why do we have a `bun.lock` at the root level? we should be having a `bun.lock` at `frontend` directory

malinskibeniamin: Only biome dependencies may be acceptable to keep the linter working smoothly in monorepo

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <div className="flex items-center gap-1.5">
+          {isExpanded ? (
+            <ChevronDown className="h-3 w-3 text-muted-foreground" />
+          ) : (
+            <ChevronRight className="h-3 w-3 text-muted-foreground" />
+          )}
+          <h5 className="font-medium text-[10px] text-muted-foreground uppercase tracking-wide">{title}</h5>
+        </div>
+        <span className="font-mono text-[9px] text-muted-foreground">{formatBytes(payloadSize)}</span>
+      </button>
```
malinskibeniamin: I think we can just use `Button` component?

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const getPreview = (content: string, lines: number): string => {
+  const contentLines = content.split('\n');
+  if (contentLines.length <= lines) {
+    return content;
+  }
+  return contentLines.slice(0, lines).join('\n');
+};
+
+export const CollapsibleCodeSection: FC<Props> = ({ title, content, defaultExpanded }) => {
```
malinskibeniamin, line 31: Should this live in UI registry 1 day?

frontend/src/components/pages/traces/components/conversation-tab.tsx
```diff
@@ -0,0 +1,79 @@
...
+              )}
+              key={idx}
+            >
+              <div className="flex-shrink-0">
+                <Badge className="text-xs" variant={badgeVariant}>
+                  {message.role}
+                </Badge>
+              </div>
+              <div className="min-w-0 flex-1">
+                <pre className="whitespace-pre-wrap break-words font-sans text-sm">{message.content}</pre>
```
malinskibeniamin: Can you try using `<Pre>` component from typography and see if it works for you

frontend/package.json
weeco: We removed recharts and use a flexbox grid instead, so this is resolved

frontend/src/components/pages/traces/components/attributes-tab.tsx
weeco: Hm not sure I understand this comment. We are handling the response here, which is served by the backend and can be considered "validated" I guess. The rest of the code is more transformation / extraction and not validation. What did you envision here?

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
weeco: I'd say it's too specific to the tracing page, but we could keep it in mind?

frontend/src/components/pages/traces/components/attributes-tab.tsx
malinskibeniamin: That's fair - since we are not performing any mutations here, `protovalidate-es` is not required

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
malinskibeniamin: 👍

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
weeco: I just tried using that button, but this changes the visuals in an undesired way, so I'd leave this for now.

frontend/src/components/pages/traces/components/conversation-tab.tsx
weeco: Done

frontend/src/components/pages/traces/components/trace-details-sheet.tsx
weeco: Added a comment:

```
  // Reset "Copied!" message after 2 seconds. Without the cleanup (return statement),
  // if the sheet closes within 2s, setTimeout would still fire and call setState on
  // an unmounted component, causing React warnings and wasted work.
```

proto/redpanda/api/dataplane/v1alpha3/tracing.proto
```diff
@@ -0,0 +1,176 @@
...
+      description: "Unique trace identifier as a 32-character hexadecimal string."
+      pattern: "^[0-9a-f]{32}$"
+    }
+  ];
+
+  // Root span name (if available). Empty if root span hasn't been received yet.
+  string root_span_name = 2 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Name of the root span (the span with no parent). May be empty if the root span is not yet available."}];
+
+  // Service name that initiated the trace. Extracted from resource attributes.
+  string service_name = 3 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Name of the service that created the root span, extracted from 'service.name' resource attribute."}];
```
rockwotj: multiple services can create a single trace, what do we do in this case?

weeco: Currently we only display the service name that initiated the trace - the root's trace service name. For individual spans we could extract the service name from the attributes as well and show it similarly as a label on the span row. Is this what you had in mind?

frontend/src/components/pages/traces/utils/span-tree-builder.ts
weeco: The times there are bigints not dates.

proto/redpanda/api/dataplane/v1alpha3/tracing.proto
```diff
@@ -0,0 +1,176 @@
...
+
+  // Trace start time (earliest span start timestamp).
+  google.protobuf.Timestamp start_time = 4 [
+    (google.api.field_behavior) = OUTPUT_ONLY,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Start time of the trace (earliest span start time)."}
+  ];
+
+  // Trace duration in milliseconds (from earliest start to latest end).
+  int64 duration_ms = 5 [
```
birdayz: why not google.protobuf.Duration ?

proto/redpanda/api/dataplane/v1alpha3/tracing.proto
birdayz: there's no concept of a service "initiating" or owning a trace in OTEL afaik. this could also be something as trivial as an NGINX or Envoy having the first touch point.

weeco: The question generally is how much we want this service to be a generic OTel / Tracing UI vs. how opinionated we want to be to make it a better UX for viewing AI native / GenAI traces.

rockwotj: I am fine with this for now, I don't think we are tied to it Nd it's only for the UI. Can we mark this as optional (or document it can be empty string and the UI needs to handle that if the resource is omitted anyways) so it's easier to remove if we decide it's not helpful?

weeco: I updated docs and renamed the property to be more precise what it is. 

> there's no concept of a service "initiating" or owning a trace in OTEL afaik.

This is simply the root span's service name that we are showing here. Could be NGINX / Envoy as well of course as you said. This is desired behaviour here because the ListTrace response just lists the root spans and you can expand them - in that row we want to show the service label.

FWIW other platforms do exactly the same, e.g.

ss-EopYuMfO@2x

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,174 @@
...
+  value: string;
+};
+
+/**
+ * Recursively extracts JavaScript values from OpenTelemetry AnyValue protobuf structures.
+ * OpenTelemetry uses a discriminated union (oneof) pattern where values are wrapped
+ * in { case: 'typeValue', value: actualValue } objects. This function unwraps them
+ * into plain JavaScript values for easier handling in the UI.
+ */
+const extractProtoValue = (value: AnyValue | undefined): unknown => {
```
jvorcak: nit: if possible, could we generate some unit test for this? I think it'd give us some confidence when modifying this in the future and also serve as a documentation.

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,174 @@
...
+      return Object.fromEntries(value.value.value.values.map((kv) => [kv.key, extractProtoValue(kv.value)]));
+    case undefined:
+      return;
+    default:
+      // This case should never be reached due to exhaustive type checking
+      return;
+  }
+};
+
+const getAttributeValue = (value: AnyValue | undefined): string => {
```
jvorcak: nit: if possible, could we generate some unit test for this? I think it'd give us some confidence when modifying this in the future and also serve as a documentation.

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,174 @@
...
+    return JSONBigInt.stringify(extractedValue, null, 2);
+  } catch (error) {
+    // biome-ignore lint/suspicious/noConsole: useful for debugging edge cases
+    console.warn('Failed to stringify attribute value:', error);
+    return String(extractedValue);
+  }
+};
+
+export const AttributesTab: FC<Props> = ({ span }) => {
+  const [searchQuery, setSearchQuery] = useState('');
```
jvorcak: Could we try to persist this to URL so that the URLs are shareable? (in case it makes sense semantically) We either can use `useQueryState` from `nuqs` or our wrapper `useQueryStateWithCallback` if we want to also involve localstorage default.

frontend/src/components/pages/traces/components/attributes-tab.tsx
```diff
@@ -0,0 +1,174 @@
...
+      <div className="rounded bg-muted/10 p-8 text-center text-muted-foreground">No attributes found in this span</div>
+    );
+  }
+
+  return (
+    <div className="space-y-4 p-3">
+      <div className="relative">
+        <Search className="pointer-events-none absolute top-1/2 left-2 h-3 w-3 -translate-y-1/2 text-muted-foreground" />
+        <Input
+          className="h-7 pl-7 text-xs"
```
jvorcak, line 152: Let's add testid please, so that we can more easily generate e2e.

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
```diff
@@ -0,0 +1,98 @@
...
+type Props = {
+  title: string;
+  content: string;
+  defaultExpanded?: boolean;
+};
+
+const SMALL_PAYLOAD_THRESHOLD = 2 * 1024; // 2KB
+const PREVIEW_LINES = 3;
+
+const formatBytes = (bytes: number): string => {
```
jvorcak: Should we move to utils and unit test?

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const hasPreview = content.split('\n').length > PREVIEW_LINES || payloadSize > SMALL_PAYLOAD_THRESHOLD;
+
+  return (
+    <div className="space-y-1.5">
+      <Button
+        className="flex h-auto w-full items-center justify-between px-0 py-0 text-left"
+        onClick={() => setIsExpanded(!isExpanded)}
+        type="button"
+        variant="ghost"
```
jvorcak, line 52: Let's add testid, and possbily aria-expanded={isExpanded} for better e2e experience.

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
```diff
@@ -0,0 +1,98 @@
...
+      >
+        <div className="flex items-center gap-1.5">
+          {isExpanded ? (
+            <ChevronDown className="h-3 w-3 text-muted-foreground" />
+          ) : (
+            <ChevronRight className="h-3 w-3 text-muted-foreground" />
+          )}
+          <h5 className="font-medium text-[10px] text-muted-foreground uppercase tracking-wide">{title}</h5>
```
jvorcak: I think ideally we should try to stick with standard text-sm text-xl and avoid explicitelly setting px.

frontend/src/components/pages/traces/components/conversation-tab.tsx
```diff
@@ -0,0 +1,82 @@
...
+  trace: Trace;
+};
+
+export const ConversationTab: FC<Props> = ({ trace }) => {
+  const messages = useMemo(() => extractConversationHistory(trace), [trace]);
+
+  if (messages.length === 0) {
+    return (
+      <Empty>
+        <EmptyHeader>
```
jvorcak, line 33: Ideally if we could also add some testids

frontend/src/components/pages/traces/components/llm-io-tab.tsx
```diff
@@ -0,0 +1,613 @@
...
+    case 'assistant':
+      return MessageSquare;
+    case 'tool':
+      return Wrench;
+    default:
+      return MessageSquare;
+  }
+};
+
+const isJsonContent = (content: string): boolean => {
```
jvorcak: can we also move this to utils?

frontend/src/components/pages/traces/components/tool-call-tab.tsx
```diff
@@ -0,0 +1,217 @@
...
+const MAX_PAYLOAD_SIZE = 20 * 1024;
+
+const truncateContent = (content: string): string => {
+  if (content.length <= MAX_PAYLOAD_SIZE) {
+    return content;
+  }
+  return `${content.slice(0, MAX_PAYLOAD_SIZE)}\n\n[... truncated ${content.length - MAX_PAYLOAD_SIZE} characters]`;
+};
+
+const isJsonContent = (content: string): boolean => {
```
jvorcak: This is duplciated with other fn `isJSON` from `raw-json-tab` and if moved to utils I think it'd be better

jvorcak: A lof of these functions seems to be breaking DRY . Maybe we could try to clean this in general through the files.

frontend/src/components/pages/traces/components/trace-details-sheet.tsx
```diff
@@ -0,0 +1,234 @@
...
+  traceId: string | null;
+  spanId: string | null;
+  isOpen: boolean;
+  onClose: () => void;
+};
+
+export const TraceDetailsSheet: FC<Props> = ({ traceId, spanId, isOpen, onClose }) => {
+  const [isLinkCopied, setIsLinkCopied] = useState(false);
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+  const [selectedTab, setSelectedTab] = useState<string>('overview');
```
jvorcak: I think we should aim for separate url per tab ideally.

frontend/src/components/pages/traces/components/trace-details-sheet.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const { data: traceData, isLoading } = useGetTraceQuery(traceId);
+
+  const trace = traceData?.trace;
+
+  // Reset "Copied!" message after 2 seconds. Without the cleanup (return statement),
+  // if the sheet closes within 2s, setTimeout would still fire and call setState on
+  // an unmounted component, causing React warnings and wasted work.
+  useEffect(() => {
+    if (!isLinkCopied) {
```
jvorcak: nit: I think we should aim for consistent UI, we should either use toast like elsewhere or decide on this and integrate it in the UI cc @malinskibeniamin

frontend/src/components/pages/traces/components/trace-details-sheet.tsx
```diff
@@ -0,0 +1,234 @@
...
+      return;
+    }
+
+    // Use URL constructor for robust URL building
+    const url = new URL(window.location.href);
+    url.searchParams.set('traceId', traceId);
+    url.searchParams.set('spanId', spanId);
+
+    try {
+      await navigator.clipboard.writeText(url.toString());
```
jvorcak, line 68: There are certain cases when this can fail, so we should implement error handling

frontend/src/components/pages/traces/components/trace-waterfall.tsx
```diff
@@ -0,0 +1,89 @@
...
+  depth,
+}) => {
+  const offset = calculateOffset(span.startTime, timeline);
+  const width = calculateWidth(span.duration, timeline);
+
+  return (
+    <>
+      <div
+        className={cn('flex items-center gap-2 rounded px-2 py-2 transition-colors hover:bg-muted/50')}
+        style={{ paddingLeft: `${depth * 24 + 8}px` }}
```
jvorcak: nit: can we use `rem` instead of `px`?  ... if possible

frontend/src/components/pages/traces/components/traces-table.tsx
```diff
@@ -0,0 +1,1064 @@
...
+
+  return (
+    <div className="flex items-center justify-between">
+      <div className="flex flex-1 items-center gap-2">
+        <Input
+          className="h-8 w-[150px]"
+          onChange={(event: ChangeEvent<HTMLInputElement>) =>
+            table.getColumn('searchable')?.setFilterValue(event.target.value)
+          }
+          placeholder="Search traces..."
```
jvorcak, line 114: Let's add testid

frontend/package.json
```diff
@@ -24,7 +24,7 @@
     "test:ci": "bun run test:unit && bun run test:integration",
     "test": "bun run test:ci",
     "test:unit": "TZ=GMT vitest run --config=vitest.config.unit.mts",
-    "test:integration": "NODE_OPTIONS='--loader=./tests/css-loader.mjs' TZ=GMT vitest run --config=vitest.config.integration.mts",
+    "test:integration": "NODE_OPTIONS='--max_old_space_size=4096 --loader=./tests/css-loader.mjs' TZ=GMT vitest run --config=vitest.config.integration.mts",
```
malinskibeniamin, line 27: Did this end up helping a bit?

frontend/src/components/pages/traces/components/collapsible-code-section.tsx
malinskibeniamin: Just use Typography heading

malinskibeniamin: Just use Typography heading

frontend/src/test-utils.tsx
```diff
@@ -20,13 +20,24 @@ const customRender = (ui: React.ReactElement, { ...renderOptions }: ExtendedRend
...
-      defaultOptions: {
-        queries: {
-          retry: false,
-        },
-      },
-    });
+
+    // Use useState to lazily initialize the QueryClient once and keep it stable across re-renders
+    // This prevents the client from being recreated when the Wrapper re-renders
+    const [queryClient] = useState(
```
malinskibeniamin, line 26: Did you notice a significant improvement? I guess when running locally we don't encounter this problem because query client is a singleton, but I suppose it can be quite counterproductive when running a lot of tests

frontend/package.json
weeco: We are still looking into the tests, I added it because of the OOM logs we saw, but still unsure if this is helping

PR #2122, opened 2026-01-09
Fixes an issue with JWT token in schema registry

jvorcak:


malinskibeniamin (approved):
Can we also make sure there are no other files using `config.fetch` in a similar fashion?

PR #2123, opened 2026-01-09
frontend: add Kafka connect defensive checks

malinskibeniamin:
**Summary**

  Added comprehensive defensive checks to the `createConnector` flow in Kafka Connect to prevent 400 Bad Request errors when creating secrets.

  **Problem**: Customers reported that the `createSecret` API call was failing with a 400 error when the connector name was an empty string (""). The root cause was insufficient validation of input parameters before making API calls.

  **Solution**: Added validation checks at multiple points in the secret creation flow to catch invalid data early and provide clear, actionable error messages.

  **Validations Added**

  **Connector configuration**:
  - Connector name must be a string (not null, number, boolean, or array)
  - Connector name cannot be empty or whitespace-only

  **Cluster configuration**:
  - Cluster name cannot be empty or whitespace-only

  **Secret data**:
  - Secret key cannot be empty
  - Secret value cannot be empty or whitespace-only
  - Serialized (base64-encoded) secret data cannot be empty

  **API response**:
  - Validates that secretId is returned from the API before proceeding

### DIFF
--- a/frontend/src/state/connect/state.ts
+++ b/frontend/src/state/connect/state.ts
@@ -193,6 +193,11 @@ export class ConnectClusterStore {
     const secrets = connector?.secrets;
     if (secrets) {
       try {
+        // Validate cluster name
+        if (!this.clusterName || this.clusterName.trim() === '') {
+          throw new Error('Cluster name is missing or empty. Cannot create secrets without a valid cluster.');
+        }
+
         const connectorName = connector?.propsByName.get('name');
         if (!connectorName) {
           throw new Error("Connector configuration is missing the 'name' property");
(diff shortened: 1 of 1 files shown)


frontend/src/state/connect/state.ts
```diff
@@ -193,21 +193,66 @@ export class ConnectClusterStore {
...
+          const serializedSecret = secret.serialized;
+          if (!serializedSecret || serializedSecret.trim() === '') {
+            throw new Error(`Failed to serialize secret for key "${key}". The encoded secret data is empty.`);
+          }
+
           const createSecretResponse = (yield api.createSecret(
             this.clusterName,
-            connectorName.value as string,
-            secret.serialized
+            connectorNameValue,
```
malinskibeniamin, line 244: We should not have casted it to a string as it may be a nullish value/undefined or empty string `""`

jvorcak (approved):

PR #2121, opened 2026-01-09
Add GatewayConfig to AIAgent proto

alenkacz:
## Summary
- Add optional `GatewayConfig` to AIAgent for routing requests through an AI Gateway
- `GatewayConfig` contains:
  - `virtual_gateway_id`: Virtual gateway ID (20-char XID format) for multi-tenant routing via `x-gateway-id` header
- Added to `AIAgent`, `AIAgentCreate`, and `AIAgentUpdate` messages

## Test plan
- [ ] Proto generates successfully
- [ ] Cloudv2 integration (CRD types + service implementation)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2124, opened 2026-01-09
feat(api): add ListVirtualGateways and ListModels RPCs

alenkacz:
## Summary
- Add `ListVirtualGateways` RPC to list virtual gateways within an AI gateway
- Add `ListModels` RPC to list available models within an AI gateway

## New Endpoints
- `GET /v1alpha3/ai-gateways/{id}/virtual-gateways`
- `GET /v1alpha3/ai-gateways/{gateway_id}/models`


proto/redpanda/api/dataplane/v1alpha3/ai_gateway.proto
```diff
@@ -481,3 +543,55 @@ message StartAIGatewayRequest {
...
+message ListModelsResponse {
+  repeated Model models = 1;
+}
+
+// Model represents an LLM model available in the AI gateway.
+message Model {
+  // Model name.
+  string name = 1;
+  // The provider hosting this model (e.g., "openai", "anthropic", "aws-bedrock").
+  string provider = 2;
```
malinskibeniamin, line 596: shouldn't we have an enum with all the providers instead.

proto/redpanda/api/dataplane/v1alpha3/ai_gateway.proto
```diff
@@ -481,3 +543,55 @@ message StartAIGatewayRequest {
...
+  string id = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).required = true,
+    (buf.validate.field).string.pattern = "^[A-Za-z0-9-_/]+$"
+  ];
+
+  // Filter virtual gateways.
+  Filter filter = 2;
+
+  message Filter {
```
malinskibeniamin, line 558: I suppose in the future we would also allow filtering by provider?

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -97,6 +97,9 @@ message AIAgent {
     }
   }];
 
+  // Optional gateway configuration for routing requests through an AI Gateway.
+  GatewayConfig gateway = 16;
```
weeco, line 101: Maybe this is how the gateway is supposed to work, but I find it odd to provide the LLM/Provider + the gateway. In my mind the gateway already determines the model that is used, or can even do things such as a/b testing or provide fallbacks if LLM A is down.

alenkacz: ah this is a mistake, it should be in different PR https://github.com/redpanda-data/console/pull/2121

anyway @birdayz claims the gateway needs to know the model to use appropriate APIs

proto/gen/openapi/openapi.v1alpha3.yaml
```diff
@@ -355,6 +361,19 @@ components:
       type: object
     DeleteMCPServerResponse:
       type: object
+    GatewayConfig:
+      description: Gateway configuration for routing requests through an AI Gateway.
+      properties:
+        gateway_id:
```
c4milo: isn't this more like a `cluster_id`?

PR #2146, opened 2026-01-16
feat(transcripts): make root span clickable and highlight selected spans

weeco:
Makes root trace rows clickable to inspect the root span (chevron still expands/collapses child spans). Adds visual highlighting to selected spans using sky colors that complement existing duration bars. Extracts selection styles into reusable constants for maintainability.

### DIFF
--- a/frontend/src/components/pages/transcripts/components/transcripts-table.tsx
+++ b/frontend/src/components/pages/transcripts/components/transcripts-table.tsx
@@ -53,7 +53,6 @@ import {
   selectableRowSelected,
 } from './selectable-row-styles';
 import { TranscriptDetailsSheet } from './transcript-details-sheet';
-import { bytesToHex } from '../utils/hex-utils';
 import { getIconForServiceName, getServiceName } from '../utils/span-classifier';
 import {
   buildSpanTree,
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/transcripts/components/transcripts-table.tsx
```diff
@@ -319,6 +327,7 @@ const SpanRow: FC<SpanRowProps> = ({
   parentDepths,
   expandedSpans,
   toggleSpan,
+  selectedSpanId,
 }) => {
   // Calculate relative position within trace using helper functions
```
malinskibeniamin, line 331: transcript

frontend/src/components/pages/transcripts/components/transcripts-table.tsx
```diff
@@ -627,24 +648,59 @@ const RootTraceRow: FC<{
   onToggle: () => void;
   spanTreeLength: number;
   isIncomplete: boolean;
-}> = ({ traceSummary, isExpanded, onToggle, spanTreeLength, isIncomplete }) => {
+  onSpanClick: (traceId: string, spanId: string) => void;
```
malinskibeniamin, line 650: rename to RootTranscriptRow

weeco: We should stick to the otel names here, because they directly refer to the span attributes in the otel payloads. Also I'm not convinced that transcripts will be well understood by customers FWIW :D

PR #2148, opened 2026-01-16
frontend: check endpoint compatibility

malinskibeniamin:
This should fix the issue where the OSS version was not able to check if the feature is supported based on the service availability.

### DIFF
--- a/frontend/src/components/pages/acls/acl-list.tsx
+++ b/frontend/src/components/pages/acls/acl-list.tsx
@@ -39,7 +39,6 @@ import type { TabsItemProps } from '@redpanda-data/ui/dist/components/Tabs/Tabs'
 import { Link, useNavigate } from '@tanstack/react-router';
 import { EditIcon, MoreHorizontalIcon, TrashIcon } from 'components/icons';
 import { isServerless } from 'config';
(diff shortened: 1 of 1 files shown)


frontend/src/state/supported-features.ts
```diff
@@ -85,10 +85,16 @@ export class Feature {
...
-    // c could be null, because the call to /api/console/endpoints failed or has not responded yet.
-    // Return true as default - the sidebar will re-render once the actual data arrives
-    // (requires the consuming component to be wrapped with MobX observer)
-    return true;
+    // c could be null, because the call to /api/console/endpoints failed or is not responded yet,
+    // in that case those endpoints should be considered unsupported
+    switch (f.endpoint) {
+      case Feature.SchemaRegistryACLApi.endpoint:
+      case Feature.ShadowLinkService.endpoint:
+      case Feature.TracingService.endpoint:
```
malinskibeniamin, line 93: @weeco do you know what's the state of the tracing service deployment? when is it getting rolled out? we noticed some issues with the tracing service endpoints depending on what config we use and wonder if we can safely call this to check endpoint compatibility - this is for the final sidebar improvement

PR #2151, opened 2026-01-19
Add AgentCard configuration to AIAgent proto

birdayz:
## What

Add optional `agent_card` field to AIAgent proto for customizing A2A agent card metadata.

## Why

Users need to configure how their AI agents present themselves in A2A discovery without being stuck with hardcoded defaults. This enables proper branding, documentation links, and skill definitions per agent.

## Implementation details

Added `AgentCard` nested message to `AIAgent` with:
- `icon_url`: URL to agent icon image
- `documentation_url`: URL to agent documentation  
- `provider`: Organization name and URL (defaults to "Redpanda" / "https://redpanda.com")
- `skills`: List of skills with id, name, description, tags, examples, input/output modes

Field added to:
- `AIAgent` (field 17)
- `AIAgentCreate` (field 14)
- `AIAgentUpdate` (field 14)

Fields intentionally staying hardcoded in the agent:
- Capabilities: `streaming=true`, `pushNotifications=false`, `stateTransitionHistory=true`
- DefaultInputModes: `["text"]`
- DefaultOutputModes: `["text"]`

Version field will be populated from K8s resource generation in the controller.

## References

JIRA: AI-292

### DIFF
--- a/proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
+++ b/proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
@@ -100,7 +100,11 @@ message AIAgent {
   // Optional gateway configuration for routing requests through an AI Gateway.
   GatewayConfig gateway = 16;
 
-  // Optional A2A agent card configuration.
+  // Optional A2A agent card configuration for customizing discovery metadata.
+  // If not set, a default card is generated using the agent's display_name and description.
+  // Auto-populated fields: name, description, url, version, capabilities (streaming,
+  // state_transition_history), and default input/output modes.
+  // Not user-configurable: capabilities and authentication (managed by Redpanda).
   AgentCard agent_card = 17;
 
   message Provider {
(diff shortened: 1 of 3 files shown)


proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -100,6 +100,9 @@ message AIAgent {
   // Optional gateway configuration for routing requests through an AI Gateway.
   GatewayConfig gateway = 16;
 
+  // Optional A2A agent card configuration.
+  AgentCard agent_card = 17;
```
weeco: What is populated by default? I think we should just improve the comment to describe what is done by default given it's optional. I believe there are some additional properties of the agent card that we omit here and can't be user-controlled (like auth & capabilities) correct?

proto/gen/openapi/openapi.v1alpha3.yaml
```diff
@@ -2688,7 +2758,7 @@ paths:
           schema:
             format: date-time
             type: string
-        - description: Maximum number of traces to return per page. Defaults to 500 if not specified or set to 0. Use -1 to disable pagination.
+        - description: Maximum number of traces to return per page. Defaults to 100 if not specified or set to 0. Use -1 to disable pagination.
```
malinskibeniamin, line 2761: Is this intentional. Usually we return 500 as max page size

weeco: Yes this is intentional

proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
birdayz: done, added comment

PR #2152, opened 2026-01-19
chore: use go.mod go version in workflows

weeco:


.github/workflows/backend-lint-test.yml
```diff
@@ -23,9 +23,8 @@ jobs:
     runs-on: ubuntu-latest-4
     steps:
       - uses: actions/checkout@v5
```
malinskibeniamin, line 25: nit: can also bump `actions/checkout@v5` to `actions/checkout@v6`

PR #2156, opened 2026-01-19
feat: add Agent tab for AI agent span visualization

weeco:
Adds a new Agent tab to transcript details for better visualization of AI agent spans. Improves span classification to properly distinguish between agent, LLM, and tool spans.

**Changes:**
- New `AgentTab` component displays agent name, description, input/output, model info, token usage, and conversation ID
- Three-tier span classification strategy using OpenTelemetry semantic conventions (operation name → type attributes → name patterns)
- Agent tab automatically shows for spans with agent-specific attributes
- Fix: Use deterministic keys for tool calls in LLM I/O tab to prevent unnecessary re-renders

Tab ordering: Overview > Agent > LLM > Tool > Attributes > Raw JSON

PR #2162, opened 2026-01-20
refactor(transcripts): use Text component from UI registry

weeco:
Refactors transcript components to use the Text component from Redpanda UI Registry instead of raw HTML text elements. This improves typography consistency and maintainability. Depends on #2156 being merged first.

- Replace h5, p, span elements with Text component
- Use consistent variants (label, muted, small)
- Apply design system patterns across all transcript tabs


frontend/src/components/pages/transcripts/components/agent-tab.tsx
```diff
@@ -0,0 +1,203 @@
...
+          <Text as="div" className="uppercase tracking-wide" variant="label">
+            TOKEN USAGE
+          </Text>
+          <ContentPanel className="bg-muted/20">
+            <div className="flex items-center justify-between">
+              <div className="space-x-3">
+                <Text variant="muted">
+                  Input:{' '}
+                  <span className="font-medium font-mono text-foreground">
+                    {agentData.inputTokens.toLocaleString()}
```
malinskibeniamin, line 170: Can we use tokenlens npm package for this

weeco: I think that package primarily brings the cost aspect of this right? We don't need that at the moment IMO, I prefer to keep it simple

malinskibeniamin (approved):

PR #2163, opened 2026-01-20
Migrate GitHub Actions workflows to Blacksmith runners

weeco:
Migrates all GitHub Actions workflows from GitHub-hosted runners to Blacksmith runners for improved CI/CD performance. Updates runner labels across all workflow files to use blacksmith-2vcpu-ubuntu-2404, blacksmith-4vcpu-ubuntu-2404, and blacksmith-8vcpu-ubuntu-2404.

### DIFF
--- a/.github/workflows/frontend-verify.yml
+++ b/.github/workflows/frontend-verify.yml
@@ -16,7 +16,7 @@ env:
 jobs:
   lint:
     timeout-minutes: 3
-    runs-on: ubuntu-latest
+    runs-on: blacksmith-2vcpu-ubuntu-2404
     defaults:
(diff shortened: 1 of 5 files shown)


.github/workflows/frontend-verify.yml
```diff
@@ -136,7 +136,7 @@ jobs:
 
   test-integration:
     timeout-minutes: 5
-    runs-on: ubuntu-latest-8
+    runs-on: blacksmith-8vcpu-ubuntu-2404
```
malinskibeniamin, line 139: Maybe 4vcpu is enough here

PR #2170, opened 2026-01-22
ci: remove on push triggers and add concurrency

c-julin:
removed push triggers directly on backend verify and frontend verify

Backend ci improved from ~9min to ~4min 🚤 

created a new job to handle edge case for forked repos

### DIFF
--- a/.github/workflows/repository-dispatch.yml
+++ b/.github/workflows/repository-dispatch.yml
@@ -28,7 +28,7 @@ jobs:
 
   dispatch:
     needs: [backend-verify, frontend-verify]
-    if: always() && !failure() && !cancelled()
+    if: "!failure() && !cancelled()"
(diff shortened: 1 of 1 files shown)


.github/workflows/repository-dispatch.yml
```diff
@@ -25,6 +28,7 @@ jobs:
 
   dispatch:
     needs: [backend-verify, frontend-verify]
+    if: always() && !failure() && !cancelled()
```
weeco: Why do we need this `always()` here? I probably don't understand the condition here

c-julin: its hangover from my previous test, if frontend was skipped it didn't trigger dispatch but now we run frontend and backend on both changes so can be safely removed

PR #2173, opened 2026-01-22
Quotas basic e2e tests

jvorcak:


malinskibeniamin (approved):
very cool test, really nice to see it come to fruition where we can execute rpk and see it update the UI

PR #2177, opened 2026-01-24
Fix search pagination in list messages

c-julin:
## Summary
Fixes an issue where searching/filtering messages while on page 2+ would show no results because the pagination index exceeded the filtered result set.

## Changes
- Bound `pageIndex` to valid range based on filtered results count
- Preserves original `pageIndex` state so clearing the search returns to the same page

## Test plan
- Navigate to a topic with enough messages to have multiple pages
- Go to page 2
- Use the "Filter table content" search box to filter messages
- Verify results are displayed (previously showed empty)
- Clear the search and verify you return to page 2

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages/index.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages/index.tsx
@@ -640,8 +640,12 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
   }, [searchFunc]);
 
   // Message Table rendering variables and functions
+  // Bound pageIndex to valid range based on filtered results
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -640,8 +640,12 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
   }, [searchFunc]);
 
   // Message Table rendering variables and functions
+  // Bound pageIndex to valid range based on filtered results
+  const totalPages = Math.max(1, Math.ceil(filteredMessages.length / pageSize));
```
jvorcak, line 644: isn't there some util function for this already?, if not shouldn't there be some?

PR #2178, opened 2026-01-26
api: new dataplane observability v1alpha3

nicolaferraro:
This defines the initial observability API, for querying dataplane metrics.


frontend/src/protogen/redpanda/api/dataplane/v1alpha3/observability-ObservabilityService_connectquery.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @generated from rpc redpanda.api.dataplane.v1alpha3.ObservabilityService.ListQueries
+ */
+export const listQueries = ObservabilityService.method.listQueries;
+
+/**
+ * Execute a predefined query
+ *
+ * @generated from rpc redpanda.api.dataplane.v1alpha3.ObservabilityService.Query
+ */
+export const query = ObservabilityService.method.query;
```
malinskibeniamin: Maybe we should call it `executeQuery` or `runQuery`

nicolaferraro: Yeah, it'd look clearer

nicolaferraro: Updated, and also fixed the http mapping following https://google.aip.dev/136

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,134 @@
...
+  // Unit of measurement for the returned time series
+  string unit = 3;
+  // Available filters for this query
+  repeated FilterMetadata filters = 4;
+  // Labels for grouping queries (e.g., category, type, component)
+  map<string, string> labels = 5;
+}
+
+// Contains details of a ListQueries request.
+message ListQueriesRequest {}
```
weeco: Should we add filters and/or pagination here? Filters may be handy for retrieving the queries on a specific page

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Parameters for executing a query.
+message ExecuteQueryParams {
+  // Optional - if not provided, performs an instant query at end time (or current time if both are absent)
+  google.protobuf.Timestamp start = 1;
+  // Optional - defaults to current date
+  google.protobuf.Timestamp end = 2;
+  // Query resolution step width (optional)
+  google.protobuf.Duration step = 3;
```
weeco: Wondering if we should allows this param. I assume we'll have to enforce boundaries anyways (e.g. 30s resolution on 7d queries would result in too many datapoints). Should we just remove it and auto calculate to a reasonable number of datapoints in the backend?

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,134 @@
...
+  // Optional filter values to apply to the query, if the query supports them
+  map<string, string> filters = 4;
+}
+
+// Input data for an ExecuteQuery operation.
+message ExecuteQueryRequest {
+  // Predefined query to request data from
+  string query_name = 1;
+  // Query execution parameters
+  ExecuteQueryParams params = 2;
```
weeco: We should add some protovalidate annotations for the request paths

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,134 @@
...
+// Contains details of a ListQueries request.
+message ListQueriesRequest {}
+
+// Response of the ListQueries operation.
+message ListQueriesResponse {
+  repeated QueryMetadata queries = 1;
+}
+
+// A single data point made of value and date when it's been measured.
+message DataPoint {
```
weeco, line 59: Prometheus has `NaN` values for missing datapoints, we probably can't project his reliably with the `double` for value. If we just omit datapoints then, the graph may be represented incorrectly (e.g. straight line between last two points instead of gap). I think that's dependent on the promql define whether we would get NaNs returned at all

I think we could solve this by making the value just an optional and add a comment about this behaviour?

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,134 @@
...
+// Input data for an ExecuteQuery operation.
+message ExecuteQueryRequest {
+  // Predefined query to request data from
+  string query_name = 1;
+  // Query execution parameters
+  ExecuteQueryParams params = 2;
+}
+
+// Response of an ExecuteQuery operation.
+message ExecuteQueryResponse {
```
weeco: I'm wondering whether we should add explicit support for something like instant queries (e.g. showing the current number of under replicated partitions etc)

proto/redpanda/api/dataplane/v1alpha3/observability.proto
nicolaferraro: Yeah, I was considering filtering by label combination for that use case, let's add it

proto/redpanda/api/dataplane/v1alpha3/observability.proto
nicolaferraro: Indeed, at the moment we don't have use cases for which we need to tune it (and actually when you look at Grafana, it's always autocomputed). The backend impl needs to compute it, but no need to expose

proto/redpanda/api/dataplane/v1alpha3/observability.proto
nicolaferraro: Yeah, I was thinking to re-use the same output result, but actually there might not be any advantage in doing that

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,229 @@
...
+  // Available filters for this query
+  repeated FilterMetadata filters = 4;
+  // Tags for grouping queries (e.g., category, type, component)
+  map<string, string> tags = 5;
+}
+
+// Contains details of a ListQueries request.
+message ListQueriesRequest {
+  // Filter queries that contain all of these key/value pairs
+  map<string, string> tags = 1 [
```
weeco: To remain consistent with the existing APIs across controlplane and dataplane: We wrap filters in their own object, see: https://github.com/redpanda-data/console/blob/fc47da1f7ca262061cc51f0a02df6b937ad2c61a/proto/redpanda/api/dataplane/v1/topic.proto#L113

proto/redpanda/api/dataplane/v1alpha3/observability.proto
```diff
@@ -0,0 +1,229 @@
...
+          description: "Range query results"
+          schema: {
+            json_schema: {ref: ".redpanda.api.dataplane.v1alpha3.ExecuteRangeQueryResponse"}
+          }
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_CONSOLE
```
weeco: I think we can introduce a separate API here for all rpcs? `API_PROMETHEUS` ?

nicolaferraro: Yeah, makes sense. I'll ask questions about how to assign permissions for those...

PR #2183, opened 2026-01-27
fix(test): resolve flaky Kafka Connect integration test

weeco:
## Summary

- Fix cleanup order bug causing "network has active endpoints" errors
- Replace log-based wait with HTTP health check for better reliability
- Add log consumer for debugging container failures
- Remove WithAttachable flag for consistency
- Centralize Docker image versions in `test-images.json`

## Root Causes Fixed

1. **Cleanup order (LIFO bug)**: Network removal was registered before container cleanup, causing it to run after in Go's LIFO cleanup order
2. **Unreliable wait strategy**: Log-based wait didn't detect early container exits
3. **Image variability**: `:latest` tag can have inconsistent behavior

## Centralized Test Images

Docker image versions are now managed in a single `test-images.json` at the repo root:

```json
{
  "images": {
    "redpanda": { "repository": "redpandadata/redpanda", "tag": "v25.3.6" },
    "kafkaConnect": { "repository": "...", "tag": "sha256:..." },
    "owlShop": { "repository": "quay.io/cloudhut/owl-shop", "tag": "master" }
  }
}
```

**Go backend**: `testutil.RedpandaImage()`, `testutil.KafkaConnectImage()`, `testutil.OwlShopImage()`

**JS frontend**: `import { REDPANDA_IMAGE } from './test-images.mjs'`

**Environment overrides** for CI matrix builds:
```bash
TEST_IMAGE_REDPANDA=redpandadata/redpanda:v25.1.0 go test ./...
```

### DIFF
--- a/backend/pkg/testutil/images.go
+++ b/backend/pkg/testutil/images.go
@@ -1,0 +1,126 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0
+
+package testutil
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"sync"
+)
+
+// ImageConfig represents a Docker image configuration.
+type ImageConfig struct {
+	Repository string `json:"repository"`
+	Tag        string `json:"tag"`
+}
+
+// String returns the full image reference (repository:tag or repository@digest).
+func (c ImageConfig) String() string {
+	if strings.HasPrefix(c.Tag, "sha256:") {
+		return c.Repository + "@" + c.Tag
+	}
+	return c.Repository + ":" + c.Tag
+}
+
+type imagesConfig struct {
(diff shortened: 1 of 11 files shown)


backend/pkg/testutil/images.go
```diff
@@ -0,0 +1,126 @@
+// Copyright 2024 Redpanda Data, Inc.
```
malinskibeniamin: 2026?

backend/pkg/testutil/images_test.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2024 Redpanda Data, Inc.
```
malinskibeniamin: 2026

test-images.json
```diff
@@ -0,0 +1,16 @@
+{
+  "images": {
+    "redpanda": {
+      "repository": "redpandadata/redpanda",
+      "tag": "v25.3.6"
```
malinskibeniamin, line 5: I guess we will need to refresh this every now and then, should we have some automatic way to update?

graham-rp: I'm wondering about this too - would there be a real penalty to using `latest`?

weeco: I'm afraid of using `latest` mostly due to cached images. WIth the blacksmith runners we have sticky disks so potentially we would be using older/cached (different than expected) images then. That's my only concern. Running tests locally would use different redpanda images as well then since everyones `latest` tag is likely different

graham-rp: ah, that makes sense. I didn't know about the blacksmith runners.

PR #2193, opened 2026-01-29
Chore/datetime picker e2e tests

jvorcak:


frontend/tests/test-variant-console/topics/topic-messages-timestamp.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+      // Set a specific timestamp (e.g., 1 hour ago)
+      const timestampValue = Date.now() - 60 * 60 * 1000;
+
+      // Find the readonly text input that displays the current datetime
+      const dateTimeDisplay = page.getByTestId('start-timestamp-input').getByRole('textbox');
+      await expect(dateTimeDisplay).toBeVisible();
+
+      // Click on the readonly input to open the date/time picker
+      await dateTimeDisplay.click();
+
```
malinskibeniamin: I would make all comments `test.step()` FWIW so it's extremely clear where we failed

frontend/tests/test-variant-console/topics/topic-messages-timestamp.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+      // Convert timestamp to time format (HH:mm)
+      const date = new Date(timestampValue);
+      const hours = String(date.getHours()).padStart(2, '0');
+      const minutes = String(date.getMinutes()).padStart(2, '0');
+      const timeValue = `${hours}:${minutes}`;
+
+      // Listen for the API request to ListMessages with startOffset -4 (Timestamp mode)
+      const apiRequestPromise = page.waitForRequest((request) => {
+        if (
+          !request.url().includes('redpanda.api.console.v1alpha1.ConsoleService/ListMessages') ||
```
malinskibeniamin: I would only match against `ConsoleService/ListMessages` because it will fail when we stop using `v1alpha1` in the future

frontend/tests/test-variant-console/topics/topic-messages-timestamp.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    const topicPage = new TopicPage(page);
+    await topicPage.createTopic(topicName);
+    await topicPage.produceMessage(topicName, message);
+
+    await test.step('Navigate to topic and select Timestamp offset origin', async () => {
+      await page.goto(`/topics/${topicName}`);
+
+      // Wait for messages to load
+      await expect(page.getByText(message)).toBeVisible({ timeout: 5000 });
```
malinskibeniamin: is there any chance we just wait for the query

frontend/tests/test-variant-console/topics/topic-messages-timestamp.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+      // Try to clear and input the unix timestamp directly
+      // First, try triple-click to select all
+      await dateTimeDisplay.click({ clickCount: 3 });
+      await page.keyboard.press('Control+A'); // Select all
+      await page.keyboard.press('Meta+A'); // Select all (Mac)
+
+      // Listen for the API request
+      const apiRequestPromise = page.waitForRequest((request) => {
+        if (
+          !request.url().includes('redpanda.api.console.v1alpha1.ConsoleService/ListMessages') ||
```
malinskibeniamin: same, try to avoid `v1alpha1` in the URL, just the service/rpc is enough

PR #2194, opened 2026-01-29
Updated weather mcp which is faster and more stable

MikolajGagatekRP:
Better, faster, more stable weather API.

### DIFF
--- a/frontend/src/components/pages/mcp-servers/templates/processor/http.yaml
+++ b/frontend/src/components/pages/mcp-servers/templates/processor/http.yaml
@@ -2,7 +2,7 @@ label: weather-by-city
 processors:
   - label: fetch_coords
     http:
-      url: 'https://nominatim.openstreetmap.org/search?city=${! this.city_name }&format=json'
+      url: 'https://nominatim.openstreetmap.org/search?city=${! this.city_name.escape_url_query() }&format=json'
       verb: GET
       headers:
         Accept: application/json


frontend/src/components/pages/mcp-servers/templates/processor/http.yaml
```diff
@@ -1,37 +1,37 @@
...
-      "metadata": {
-        "source": "wttr.in",
-        "fetched_at": now().ts_format("2006-01-02T15:04:05.000Z")
+  - label: prepare_parameters
+    mutation: |
+      meta lat = this.lat
+      meta long = this.long
+  - label: fetch_weather
+    http:
+      url: https://api.open-meteo.com/v1/forecast?latitude=${! @lat }&longitude=${! long }&current=temperature_2m,relative_humidity_2m,apparent_temperature,is_day,precipitation,rain,showers,snowfall,cloud_cover,surface_pressure,wind_speed_10m,wind_direction_10m,wind_gusts_10m&forecast_days=1
```
birdayz: how good are LLMs at providing the coords? can this API work with city name as well?
or include functionality to resolve name -> coords?

MikolajGagatekRP: unfortunatelly not :(
as far as Ive seen there were not problems of passing correct ones.
TBH also I did not intensively checked them but a few of does like London or NYC which I checked were correct.

I might add also another request to get coords before invoking that but not sure if we want it like that?

birdayz: sounds good to me to lookup the coords

MikolajGagatekRP: updated to do it like that :)

PR #2202, opened 2026-02-02
feat(transcripts): add span filtering with matched row highlighting

weeco:
Add server-side span filtering with frontend highlighting for matched spans.

- Add preset filters (LLM, tool, agent, error, slow) and custom attribute filters
- Highlight matched spans with subtle muted background, dim unmatched parents
- Support "show full traces" toggle to filter view to matched spans only

### DIFF
--- a/frontend/src/components/pages/transcripts/components/transcripts-table.tsx
+++ b/frontend/src/components/pages/transcripts/components/transcripts-table.tsx
@@ -537,7 +537,7 @@ const useSpanExpansion = (spanTree: SpanNode[], collapseAllTrigger: number) => {
 
     // Update ref for next render
     prevSpanTreeLength.current = currentLength;
-  }, [spanTree, expandedSpans.size]);
+  }, [spanTree]);
 
   const toggleSpan = (spanId: string) => {
     setExpandedSpans((prev) => {
--- a/frontend/src/components/pages/transcripts/utils/span-filter.ts
+++ b/frontend/src/components/pages/transcripts/utils/span-filter.ts
@@ -13,6 +13,10 @@ import type { Span } from 'protogen/redpanda/otel/v1/trace_pb';
 
 import { bytesToHex } from './hex-utils';
 
+// OTEL backends vary: some send empty bytes for root spans (handled by falsy check),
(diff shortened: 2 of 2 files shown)


frontend/src/components/pages/transcripts/components/transcript-filter-bar.tsx
```diff
@@ -0,0 +1,410 @@
...
+const TIME_RANGES: TimeRangeConfig[] = [
+  { value: '5m', label: 'Last 5 minutes' },
+  { value: '15m', label: 'Last 15 minutes' },
+  { value: '30m', label: 'Last 30 minutes' },
+  { value: '1h', label: 'Last 1 hour' },
+  { value: '3h', label: 'Last 3 hours' },
+  { value: '6h', label: 'Last 6 hours' },
+  { value: '12h', label: 'Last 12 hours' },
+  { value: '24h', label: 'Last 24 hours' },
+];
```
malinskibeniamin, line 132: Let's use these packages for the transcripts UI feature in general:
- [chrono](https://github.com/wanasit/chrono) (helps with natural language handling)
- [little-date](https://github.com/vercel/little-date) (handling selected timestamp intervals)

This should make sure we handle all dates/time properly in JS, and does not require `date-fns` migration. It's a quick win we can achieve to simplify the codebase.

frontend/src/components/pages/transcripts/utils/span-filter.test.ts
```diff
@@ -0,0 +1,124 @@
...
+import { filterToMatchedAndAncestors } from './span-filter';
+
+// Helper to create mock spans with specific IDs
+const hexToBytes = (hex: string): Uint8Array => {
+  const bytes = new Uint8Array(hex.length / 2);
+  for (let i = 0; i < hex.length; i += 2) {
+    bytes[i / 2] = Number.parseInt(hex.slice(i, i + 2), 16);
+  }
+  return bytes;
+};
```
malinskibeniamin, line 23: You could use `Buffer.from` but I think that's available only in NodeJS 24+ so let's leave it

PR #2203, opened 2026-02-03
Filter system-managed gateways from agent creation UI

alenkacz:
- Add default filter to useListGatewaysQuery to exclude system-managed gateways
- Users can still override filter by passing explicit filter parameter
- Add BYOC cluster support to start-cloud.sh script

filtering

### DIFF
--- a/frontend/start-cloud.sh
+++ b/frontend/start-cloud.sh
@@ -29,8 +29,9 @@ fi
 # Extract cluster ID from backend URL
 # Example: https://console-2fd2fedf.d5mst2vnnfekmiescdb0.fmc.ign.cloud.redpanda.com
 # Example: https://console-2498fb74.d5tp5kntujt599ksadgg.byoc.ign.cloud.redpanda.com
+# Example: https://console-xxx.d5xxx.ppd.ign.cloud.redpanda.com (preprod)
 # Extracts: d5mst2vnnfekmiescdb0 or d5tp5kntujt599ksadgg
-CLUSTER_ID=$(echo "$BACKEND_URL" | sed -E 's/.*\.([a-z0-9]+)\.(fmc\.ign|byoc\.ign|rpd)\.cloud\.redpanda\.com.*/\1/')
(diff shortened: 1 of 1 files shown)


frontend/src/react-query/api/ai-gateway.tsx
```diff
@@ -29,25 +29,36 @@ const AI_GATEWAY_DEFAULT_PAGE_SIZE = 50;
...
  */
 export const useListGatewaysQuery = (
   input?: MessageInit<ListGatewaysRequest>,
   options?: QueryOptions<GenMessage<ListGatewaysRequest>, ListGatewaysResponse>
 ): UseQueryResult<ListGatewaysResponse, ConnectError> => {
   const transport = useAIGatewayTransport();
 
+  // Default filter: exclude system-managed gateways
+  // Users can override by passing an explicit filter (including empty string for no filter)
+  const defaultFilter = 'metadata.system_managed != "true"';
```
malinskibeniamin, line 54: Aren't we supposed to use the proto generated factories, such as create(ListAIGateways_Filter, {}) ?

frontend/start-cloud.sh
```diff
@@ -28,8 +28,9 @@ fi
 
 # Extract cluster ID from backend URL
 # Example: https://console-2fd2fedf.d5mst2vnnfekmiescdb0.fmc.ign.cloud.redpanda.com
-# Extracts: d5mst2vnnfekmiescdb0
-CLUSTER_ID=$(echo "$BACKEND_URL" | sed -E 's/.*\.([a-z0-9]+)\.(fmc\.ign|rpd)\.cloud\.redpanda\.com.*/\1/')
+# Example: https://console-2498fb74.d5tp5kntujt599ksadgg.byoc.ign.cloud.redpanda.com
+# Extracts: d5mst2vnnfekmiescdb0 or d5tp5kntujt599ksadgg
+CLUSTER_ID=$(echo "$BACKEND_URL" | sed -E 's/.*\.([a-z0-9]+)\.(fmc\.ign|byoc\.ign|rpd)\.cloud\.redpanda\.com.*/\1/')
```
malinskibeniamin: How about ppd for preprod?

frontend/src/react-query/api/ai-gateway.tsx
alenkacz: the filter field is a string type (CEL expression) in the proto, not a separate message type, so there's no factory to use for it.

PR #2205, opened 2026-02-03
observability: new page based on dataplane observability API

nicolaferraro:
When the feature flag is toggled, console starts showing this and cloud-ui stops showing the old one (different ff for Serverless and Serverful).

This is a first iteration, it allows to select the time range (but only from now back to 6 hours). I need to publish the max lookback time in the API to have a better control bar.

Screenshot 2026-02-01 at 00 13 13


frontend/package.json
```diff
@@ -118,6 +118,7 @@
     "react-markdown": "^10.1.0",
     "react-simple-code-editor": "^0.14.1",
     "react-syntax-highlighter": "^15.6.6",
+    "recharts": "2.15.4",
```
malinskibeniamin, line 123: Let's use recharts v3 https://github.com/shadcn-ui/ui/issues/7669 which will require components to be upgraded.

We can do that in a followup PR

frontend/src/routes/observability/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * by the Apache License, Version 2.0
+ */
+
+import { createFileRoute, redirect } from '@tanstack/react-router';
+import { ActivityIcon } from 'components/icons';
+import { isFeatureFlagEnabled, isServerless } from 'config';
+
+import ObservabilityPage from '../../components/pages/observability/observability-page';
+
+export const Route = createFileRoute('/observability/')({
```
malinskibeniamin, line 18: will this be required for standalone mode

frontend/src/components/redpanda-ui/components/chart.tsx
malinskibeniamin, line 1: this component will require updating to support shadcn chart + recharts v3 major release.

frontend/src/components/pages/observability/observability-toolbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+export type TimeRange = '5m' | '15m' | '30m' | '1h' | '3h' | '6h';
+
+const TIME_RANGE_MS: Record<TimeRange, number> = {
+  '5m': 5 * 60 * 1000,
+  '15m': 15 * 60 * 1000,
+  '30m': 30 * 60 * 1000,
+  '1h': 60 * 60 * 1000,
+  '3h': 3 * 60 * 60 * 1000,
+  '6h': 6 * 60 * 60 * 1000,
+};
```
malinskibeniamin: @weeco implemented this for transcripts/traces UI, can you please adapt a similar approach to follow the same practices?

frontend/src/components/pages/observability/observability-toolbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+  // biome-ignore lint/correctness/useExhaustiveDependencies: refreshKey triggers recalculation on refresh
+  const timeRange = useMemo(() => {
+    const now = new Date();
+    const startTime = new Date(now.getTime() - TIME_RANGE_MS[selectedTimeRange]);
+
+    return {
+      start: startTime,
+      end: now,
+    };
+  }, [selectedTimeRange, refreshKey]);
```
malinskibeniamin: do we still need `useExhaustiveDependencies`?

frontend/src/routes/observability/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  staticData: {
+    title: 'Metrics',
+    icon: ActivityIcon,
+  },
+  beforeLoad: () => {
+    const isObservabilityEnabled = isServerless()
```
weeco, line 24: In addition to this we also need to make sure that this is actually configured in the backend I'd say - specifically thinking of self-hosted community users and customers here. Not sure if we want to disable/hide the sidebar or show a setup / onboarding page in this case.

frontend/src/components/pages/observability/observability-toolbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { Box, Flex, Text } from '@redpanda-data/ui';
```
malinskibeniamin: Please use Redpanda UI registry instead of `@redpanda-data/ui` for this new observability page.

frontend/src/components/pages/observability/metric-chart.tsx
```diff
@@ -0,0 +1,277 @@
...
+              // Show decimals for small values
+              if (value < 10) {
+                return value.toFixed(2);
+              }
+              if (value < 100) {
+                return value.toFixed(1);
+              }
+              return value.toFixed(0);
+            }}
+            tickLine={false}
```
malinskibeniamin: I think we should filter using storage utils

frontend/src/components/pages/observability/metric-chart.tsx
```diff
@@ -0,0 +1,277 @@
...
+ * Use of this software is governed by the Business Source License
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { timestampFromMs } from '@bufbuild/protobuf/wkt';
+import { Alert, AlertIcon, Box, Heading, Skeleton, Text } from '@redpanda-data/ui';
```
malinskibeniamin: blocker: use redpanda UI registry

frontend/src/components/pages/observability/observability-toolbar.tsx
weeco: https://github.com/redpanda-data/console/pull/2202/changes#diff-3d39ec19551b863b267e002b020aa0c055b390f7f706b8c24147427b3c111222R83-R92

We can move some things to the utils directories.

malinskibeniamin:
> @claude could you review, focus on simplicity and ensure it follows some well known frontend best practices

@sago2k8 it's not enabled yet AFAIK

nicolaferraro:
I did a first pass to remove the old components and fixing stuff. About the self-hosted mode, currently the sidebar is only shown for embedded mode only. Don't know if there's a clean way to make this work for self-hosted as well

frontend/src/components/pages/observability/metric-chart.tsx
```diff
@@ -0,0 +1,216 @@
...
+              }
+              // Show decimals for small values
+              if (value < 10) {
+                return value.toFixed(2);
+              }
+              if (value < 100) {
+                return value.toFixed(1);
+              }
+              return value.toFixed(0);
+            }}
```
malinskibeniamin: let's use storage utils to get proper formatting for these units

frontend/src/utils/route-utils.tsx
```diff
@@ -191,6 +191,18 @@ export const SIDEBAR_ITEMS: SidebarItem[] = [
     icon: HomeIcon,
     visibilityCheck: routeVisibility(true),
   },
+  {
+    path: '/observability',
+    title: 'Metrics',
+    icon: ActivityIcon,
+    visibilityCheck: routeVisibility(
+      () =>
+        isEmbedded() &&
```
malinskibeniamin, line 200: This should be gated by `isAdpEnabled()` too, right? Or will we have observability in self-hosted too 1 day, somehow?

frontend/src/react-query/api/observability.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const useListQueries = (
+  request?: MessageInit<ListQueriesRequest>,
+  opts?: { enabled?: boolean; refetchInterval?: number | false }
+) => {
+  const listQueriesRequest = create(ListQueriesRequestSchema, request || {});
+  return useQuery(listQueries, listQueriesRequest, {
+    ...opts,
+    retry: fastFailRetry,
+  });
```
malinskibeniamin, line 37: Do we want to poll this?

frontend/src/components/pages/observability/utils/time-range.ts
malinskibeniamin, line 1: can we rely on work from @weeco for transcripts? and use the same approach for the time ranges

frontend/src/utils/route-utils.tsx
nicolaferraro: Technically it's possible to provide it in self-hosted, by configuring a local prom server. There's an internal discussion about it. For the moment, I think this condition only enables it in embedded mode with a specific feature flag.

frontend/src/react-query/api/observability.ts
nicolaferraro: I did not wire auto-polling since the page where charts are used has an auto-refresh button, that also moves the clock forward.

Polling without applying a time shift does nothing (the API uses an absolute time range).

frontend/src/utils/route-utils.tsx
weeco: I don't see why this should be gated by `isAdpEnabled()`. This change is completely unrelated. Current condition looks fine to me, even though I'd personally rather do the backend check to see if the API is supported rather than relying on the feature flags that much.

PR #2217, opened 2026-02-05
frontend: fix pagination for lists exceeding page size

birdayz:
## What

Fix pagination bug where AI Agents, MCP Servers, and Pipelines lists only display the first page of items, silently dropping any items beyond that limit.

Also reduce MAX_PAGE_SIZE from 500 to 25 to avoid excessively large API responses.

## Why

These hooks had a hardcoded page size and always sent `pageToken: ''`. The `nextPageToken` from API responses was completely ignored. If you had 51 agents (or 501 pipelines), items beyond the first page would not be displayed.

The codebase even had TODO comments acknowledging this limitation:
```typescript
// TODO: Make this dynamic so that pagination can be used properly
const AI_AGENT_MAX_PAGE_SIZE = 50;
```

**Page size rationale**: MCP server entries are ~1.5 KB each (configYaml is the bulk). At 25 items per page, responses are ~37-40 KB which is reasonable. The previous 500 limit would result in ~750 KB responses.

AND, **more important: there is no guarantee the backend will return 500 items. it can, it does, enforce a hard cap. the client _MUST_ honor next page token!** and can not rely on "getting it all" with big page size..

## Implementation details

- Replace `useQuery` with `useInfiniteQueryWithAllPages` (existing utility that auto-fetches all pages)
- Follow existing pattern used by security.tsx, transform.tsx, and user.tsx
- For pipelines, handle the nested request structure by returning a new request object in `getNextPageParam`
- Reduce MAX_PAGE_SIZE from 500 to 25

### DIFF
--- a/frontend/src/react-query/api/ai-agent.test.tsx
+++ b/frontend/src/react-query/api/ai-agent.test.tsx
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2025 Redpanda Data, Inc.
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file https://github.com/redpanda-data/redpanda/blob/dev/licenses/bsl.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { create } from '@bufbuild/protobuf';
+import { createRouterTransport } from '@connectrpc/connect';
+import { renderHook, waitFor } from '@testing-library/react';
+import { AIAgentSchema, ListAIAgentsResponseSchema } from 'protogen/redpanda/api/dataplane/v1alpha3/ai_agent_pb';
+import { listAIAgents } from 'protogen/redpanda/api/dataplane/v1alpha3/ai_agent-AIAgentService_connectquery';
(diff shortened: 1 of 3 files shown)


frontend/src/react-query/api/pipeline.tsx
```diff
@@ -100,34 +93,35 @@ export const useListPipelinesQuery = (
...
 
-  // Stabilize the data object to prevent component re-renders
-  const data = useMemo(
-    () => ({
-      pipelines,
-    }),
-    [pipelines]
+  // Flatten pipelines from all pages
+  const pipelines = useMemo(
+    () => listPipelinesResult?.data?.pages?.flatMap((page) => page?.response?.pipelines ?? []) ?? [],
```
malinskibeniamin: let's do it in a separate statement for easier debugging so that we store `listPipelinesResult?.data?.pages?.flatMap((page) => page?.response?.pipelines ?? [])` in a variable

birdayz: Done in f3d4c5026. Also applied the same change to ai-agent.tsx and remote-mcp.tsx.

malinskibeniamin:
@claude review

MikolajGagatekRP:
secrets pagination still broken 😢

PR #2220, opened 2026-02-09
UX 870 remove core stat from mobx

jvorcak:


malinskibeniamin:
@jvorcak is this ready for review? looks like a lot of changes

jvorcak:
@claude review

frontend/src/config.ts
```diff
@@ -293,46 +290,71 @@ export const setMonacoTheme = (_editor: monaco.editor.IStandaloneCodeEditor, mon
...
-    }
+  try {
+    // Subscribe to breadcrumbs changes
+    let previousBreadcrumbs = useUIStateStore.getState().pageBreadcrumbs;
 
-    const breadcrumbs = uiState.pageBreadcrumbs.map((v) => ({
-      title: v.title,
-      to: v.linkTo,
-    }));
+    useUIStateStore.subscribe((state) => {
```
malinskibeniamin, line 300: Let's double check breadcrumbs are properly set now with this logic change, it was a problem in the past when navigating between pages

jvorcak: I think we have e2e tests for this on some pages already, but I'll check

PR #2224, opened 2026-02-14
frontend: SSE reconnection via tasks/resubscribe

birdayz:
## What

When an SSE connection drops mid-stream, transparently reconnect using the A2A protocol's `tasks/resubscribe` endpoint instead of showing a fatal error. Bump `@a2a-js/sdk` from 0.3.5 to 0.3.10.

## Why

Long-running agent tasks are particularly vulnerable to transient network failures. A brief WiFi hiccup or load balancer timeout shouldn't kill the entire conversation. The task keeps running server-side -- we just lost the event stream and need to pick it back up.

## Implementation details

The core logic lives in `use-message-streaming.ts`. When `streamMessage()` catches an error and the task is still in-flight (has a task ID, state is not terminal), it enters a resubscribe loop:

1. Close the active text block and insert a `disconnected` connection-status block
2. Retry up to 5 times with exponential backoff (1s, 2s, 4s, 8s, 16s)
3. Each attempt creates a fresh `A2AClient` and calls `resubscribeTask({ id: taskId })`
4. On success, process the resumed event stream with the same handlers as the initial stream
5. On permanent failure, insert a `gave-up` block and fall through to the existing error path

New `connection-status` ContentBlock variant flows through the full type system: runtime types, DB schema, serialization, deserialization, and a new `ConnectionStatusBlock` UI component.

The `reconnecting` / `reconnected` status blocks replace each other in-place so the UI doesn't accumulate stale status lines during retries.

Test suite covers 8 scenarios: successful stream, error handling, reconnection success, reconnection after partial content, all retries exhausted, non-resubscribable errors, error parsing, and task completion during reconnection.

## References

- A2A protocol `tasks/resubscribe`: https://google.github.io/A2A/#/topics/streaming?id=disconnection-and-recovery

### DIFF
--- a/frontend/src/components/pages/agents/details/a2a/chat/hooks/use-message-streaming.ts
+++ b/frontend/src/components/pages/agents/details/a2a/chat/hooks/use-message-streaming.ts
@@ -111,6 +111,13 @@ type StreamMessageResult = {
 };
 
 const TERMINAL_TASK_STATES = new Set(['completed', 'failed', 'canceled', 'rejected']);
+
+/**
+ * Maximum consecutive reconnection attempts before giving up.
+ * With exponential backoff (1s, 2s, 4s, 8s, 16s), total wait is ~31s.
+ * The counter resets on progress, so tasks making incremental progress
+ * can reconnect indefinitely.
+ */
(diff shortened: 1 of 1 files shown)


weeco:
@claude Review the PR

PR #2226, opened 2026-02-17
feat: implement unlimited pagination for Kafka message listing

c-julin:
## Summary

Adds cursor-based pagination support to the Kafka message listing API. When `page_size` is set, the backend enters pagination mode — fetching messages in pages with opaque page tokens for resumption, while `max_results` is ignored. This enables clients to paginate through all messages in a topic without a hard upper limit.

- Implement `PageToken` encoding/decoding with per-partition offsets and direction tracking
- Add pagination-aware message fetching that respects `PageSize` and generates `next_page_token` in the stream completion message
- Reject filter usage in pagination mode (filters are incompatible with cursor-based pagination)
- Extract `collectPartitionStates` helper to reduce cyclomatic complexity in the listing logic
- Fix integration tests to use `PageSize` to correctly trigger pagination mode
- Update proto comments to reflect that `page_size` controls pagination mode, not `max_results`

### Page Token

Defined in `backend/pkg/console/page_token.go`.

The page token is a base64url-encoded JSON object (no padding, via `base64.RawURLEncoding`) containing:

```go
type PageToken struct {
    TopicName      string            `json:"t"`   // Topic name (for validation)
    PartitionCount int32             `json:"pc"`  // Expected partition count
    Partitions     []PartitionCursor `json:"p"`   // Per-partition read cursors
    Direction      string            `json:"d"`   // "desc" or "asc"
    PageSize       int               `json:"ps"`  // Messages per page
}

type PartitionCursor struct {
    ID            int32 `json:"id"` // Partition ID
    NextOffset    int64 `json:"no"` // Next offset to read from
    LowWaterMark  int64 `json:"lw"` // Low water mark at token creation
    HighWaterMark int64 `json:"hw"` // High water mark at token creation
}
```

Key functions:
- `CreateInitialPageToken()` — builds the first token from current partition metadata
- `DecodePageToken()` — deserializes and validates a token from a request
- `Encode()` — serializes the token for the response
- `HasMore()` — checks if any partition still has unread messages
- `Validate()` — ensures field invariants (pageSize 1–500, partition count matches, etc.)

### Pagination Mode

Triggered in `ListMessages()` (`backend/pkg/console/list_messages.go`) when `listReq.PageSize > 0`:

1. If no `pageToken` is provided, `CreateInitialPageToken()` is called
2. Direction is determined from `startOffset`:
   - `PartitionOffsetOrigin.Oldest` → ascending
   - `PartitionOffsetOrigin.EndMinusResults` (Newest) → descending
3. `calculateConsumeRequestsWithPageToken()` distributes the page across partitions

### Round-Robin Distribution

`calculateConsumeRequestsWithPageToken()` distributes `pageSize` messages evenly across non-exhausted partitions:

```
remainingMessages = pageSize
Loop while remainingMessages > 0:
    For each partition with messages remaining:
        partition.maxMessageCount++
        remainingMessages--
```

This ensures even distribution regardless of per-partition message counts.

### Direction Handling

- **Descending**: reads backwards from `NextOffset`; messages in each partition are reversed before sending
- **Ascending**: reads forward from `NextOffset`

After reading, the next token's `NextOffset` is updated:
- Descending: `readStart - 1` (move backward)
- Ascending: `readEnd + 1` (move forward)

### Compaction Resilience

In `collectPartitionStates()`, if log compaction has advanced the low water mark past a cursor's `NextOffset`, the cursor is adjusted forward:

```go
if cursor.NextOffset < updatedLowWaterMark {
    cursor.NextOffset = updatedLowWaterMark
}
```


backend/pkg/api/httptypes/types.go
```diff
@@ -46,8 +50,20 @@ func (l *ListMessagesRequest) OK() error {
 		return errors.New("partitionID is smaller than -1")
 	}
 
-	if l.MaxResults <= 0 || l.MaxResults > 500 {
-		return errors.New("max results must be between 1 and 500")
+	if l.MaxResults <= 0 || l.MaxResults > 10_000 {
+		return errors.New("max results must be between 1 and 10000")
+	}
```
malinskibeniamin: are we sure this is a safe number? 10000 messages will be huge?

c-julin: i was experimenting with increasing this but you're right i'll keep this at 500 as it was.

proto/redpanda/api/console/v1alpha1/list_messages.proto
```diff
@@ -36,6 +36,12 @@ message ListMessagesRequest {
 
   bool ignore_max_size_limit = 12; // Optionally ignore configured maximum payload size limit.
   // Used to force returning deserialized payloads.
+
+  string page_token = 13; // Resume from cursor (only used when page_size is present).
+  int32 page_size = 14 [(buf.validate.field).int32 = {
+    gte: 1
+    lte: 10000
```
malinskibeniamin: I guess 1000 would be a sane number?

backend/pkg/console/page_token.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/twmb/franz-go/pkg/kadm"
+)
+
+// PageToken represents the pagination cursor state for ListMessages requests.
+// It encodes per-partition offset information to enable stateless pagination
+// across multiple API calls.
+type PageToken struct {
+	TopicName      string            `json:"t"`  // Topic name for validation
+	PartitionCount int32             `json:"pc"` // Partition count for change detection
+	Partitions     []PartitionCursor `json:"p"`  // Per-partition cursor state
```
graham-rp: Do we need `PartitionCount` if we have `Partitions`?

backend/pkg/console/list_messages.go
```diff
@@ -457,6 +532,282 @@ func (s *Service) calculateConsumeRequests(
...
+			req.MaxMessageCount++
+			remainingMessages--
+			progressMade = true
+		}
+
+		// Safety: if no progress was made in a round, break to avoid infinite loop
+		if !progressMade {
+			break
+		}
+	}
```
graham-rp, line 739: nit: I don't think it'll really hurt anything, but this seems like it may be a little overkill to do it one by one, right?

backend/pkg/console/list_messages.go
```diff
@@ -457,6 +532,282 @@ func (s *Service) calculateConsumeRequests(
...
+
+	// Third pass: Calculate actual read ranges and next offsets for each partition
```
graham-rp: $0.02: seems like a pretty natural point to break the function apart if you'd like to get rid of the cyclo lint bypass

backend/pkg/api/handle_topic_messages_integration_test.go
```diff
@@ -1050,3 +1051,223 @@ func randomString(n int) string {
...
+
+	// Debug: Check actual message distribution across partitions
+	offsets, err := s.kafkaAdminClient.ListEndOffsets(ctx, testTopicName)
+	require.NoError(err)
+	totalMessages := int64(0)
+	offsets.Each(func(offset kadm.ListedOffset) {
+		count := offset.Offset
+		totalMessages += count
+		t.Logf("Partition %d: %d messages (offset: 0 to %d)", offset.Partition, count, offset.Offset-1)
+	})
```
graham-rp: could this be a `require.Eventually`?

graham-rp (approved):

backend/pkg/api/httptypes/types.go
```diff
@@ -46,8 +50,20 @@ func (l *ListMessagesRequest) OK() error {
...
+
+	// Pagination mode: when PageSize > 0, filters are not supported
+	if l.PageSize > 0 {
+		if l.FilterInterpreterCode != "" {
+			decoded, _ := l.DecodeInterpreterCode()
+			if decoded != "" {
+				return errors.New("cannot use filters with pagination")
+			}
+		}
```
r-vasquez, line 61: I think we can save the call to `DecodeInterpreterCode`. The interpreter code is a base64 encoding of the JS filter/function. So, if there is anything in `l.FilterInterpreterCode` doesn't that mean that there is a filter and we should err? (And base64("") == "", so we could just:

```suggestion
	if l.PageSize > 0 {
		if l.FilterInterpreterCode != "" {
			return errors.New("cannot use filters with pagination")
		}
    }
```

I think you do this already in backend/pkg/console/list_messages.go L289

backend/pkg/api/handle_topic_messages_integration_test.go
```diff
@@ -1050,3 +1051,223 @@ func randomString(n int) string {
...
+		if err != nil {
+			return false
+		}
+		topic, exists := metadata.Topics[testTopicName]
+		return exists && topic.Err == nil && len(topic.Partitions) == 3
+	}, 30*time.Second, 100*time.Millisecond, "Topic should be created and ready")
+
+	// Produce messages
+	records := make([]*kgo.Record, messageCount)
+	for i := 0; i < messageCount; i++ {
```
r-vasquez: [ultra nit now that we have go fix ./...] Move the const closer to where is used and

```suggestion
	// Produce messages
	const messageCount = 150
	records := make([]*kgo.Record, messageCount)
	for i := range messageCount {
```

backend/pkg/api/handle_topic_messages_integration_test.go
```diff
@@ -1050,3 +1051,223 @@ func randomString(n int) string {
...
+			Value: []byte(fmt.Sprintf(`{"id": %d, "message": "test message %d"}`, i, i)),
+			Topic: testTopicName,
+		}
+	}
+
+	produceResults := s.kafkaClient.ProduceSync(ctx, records...)
+	require.NoError(produceResults.FirstErr())
+
+	// Give Kafka a moment to commit
+	time.Sleep(500 * time.Millisecond)
```
r-vasquez: Can we do a require.Eventually here and use `ListEndOffsets` to verify that the offsets are equal to the message count/ number of records? I worry about the flakiness that could be introduced by relying on a sleep
 here.

backend/pkg/console/list_messages.go
```diff
@@ -211,25 +221,46 @@ func (s *Service) ListMessages(ctx context.Context, listReq ListMessageRequest,
...
+	var token *PageToken
+
+	if listReq.PageSize > 0 {
+		token, consumeRequests, nextPageToken, err = s.resolvePagedConsumeRequests(ctx, &listReq, partitionIDs, startOffsets, endOffsets)
+	} else {
+		consumeRequests, err = s.calculateConsumeRequests(ctx, cl, &listReq, partitionIDs, startOffsets, endOffsets)
+	}
 	if err != nil {
-		return fmt.Errorf("failed to calculate consume request: %w", err)
+		return err
```
r-vasquez: Why did you remove the wrapped error?

backend/pkg/console/list_messages.go
```diff
@@ -457,6 +532,282 @@ func (s *Service) calculateConsumeRequests(
...
+// calculateConsumeRequestsWithPageToken calculates consume requests for pagination mode (descending order).
+// It returns the consume requests map, the next page token, and whether more pages are available.
+//
+//nolint:gocognit,cyclop // complex logic for round-robin distribution across partitions with watermark tracking
+func (s *Service) calculateConsumeRequestsWithPageToken(
+	ctx context.Context,
+	token *PageToken,
+	requestedPartitionIDs []int32,
+	startOffsets, endOffsets kadm.ListedOffsets,
+) (map[int32]*PartitionConsumeRequest, string, bool, error) {
```
r-vasquez: Is the boolean flag `whether more pages are available` used anywhere? I see that you ignore it in L298, and I can't find any other usage. Perhaps we can remove it from the signature

backend/pkg/console/page_token.go
```diff
@@ -0,0 +1,245 @@
...
+		return "", fmt.Errorf("failed to marshal page token: %w", err)
+	}
+
+	// Use URL-safe base64 encoding (no padding)
+	encoded := base64.RawURLEncoding.EncodeToString(jsonBytes)
+	return encoded, nil
+}
+
+// DecodePageToken deserializes a page token from a base64 string.
+func DecodePageToken(encoded string) (*PageToken, error) {
```
r-vasquez:
```suggestion
func decodePageToken(encoded string) (*PageToken, error) {
```

Nit, no need to export as is only used within this package. Same for CreateInitialPageToken

backend/pkg/console/list_messages.go
```diff
@@ -457,6 +532,282 @@ func (s *Service) calculateConsumeRequests(
...
+	requests := make(map[int32]*PartitionConsumeRequest)
+
+	// Validate partition count hasn't changed
+	if int32(len(requestedPartitionIDs)) != token.PartitionCount {
+		return nil, "", false, fmt.Errorf("partition count changed: expected %d, got %d", token.PartitionCount, len(requestedPartitionIDs))
```
r-vasquez: What should a user do in these cases? what does a partition count change means? Is it a bug/ is it that a message got produced?

PR #2229, opened 2026-02-18
feat: add continuous pagination UI for Kafka message listing

c-julin:
## Summary 

Adds frontend support for infinite scroll / unlimited pagination when browsing Kafka topic messages. Users can toggle infinite scroll mode to page through the full topic while keeping a configurable max results window in memory, with older pages trimmed automatically.

### Request / Response Flow

```mermaid
sequenceDiagram
    participant FE as Frontend
    participant BE as Backend

    FE->>BE: POST /messages { pageSize: 50 }
    Note right of BE: CreateInitialPageToken()
    Note right of BE: calculateConsumeRequestsWithPageToken()
    Note right of BE: Round-robin distribute across partitions
    Note right of BE: Fetch & return messages
    BE-->>FE: { messages[], nextPageToken }

    Note left of FE: User navigates to next page

    FE->>BE: POST /messages { pageSize: 50, pageToken }
    Note right of BE: DecodePageToken()
    Note right of BE: Resume from per-partition cursors
    Note right of BE: Fetch next page
    BE-->>FE: { messages[], nextPageToken }

    Note left of FE: Sliding window trims old pages from memory
```

### Sliding Window (Frontend)

```mermaid
block-beta
    columns 7

    block:legend:7
        L1["Trimmed"] L2["In memory"] L3["Not yet loaded"]
    end

    block:window:7
        P1["Page 1\n(trimmed)"]
        P2["Page 2\n(trimmed)"]
        P3["Page 3"]
        P4["Page 4\n👁 viewing"]
        P5["Page 5"]
        P6["Page 6\n(loaded when\nviewing page 5)"]
        P7["Page 7+\n..."]
    end

    style L1 fill:#f99,stroke:#933
    style L2 fill:#9f9,stroke:#393
    style L3 fill:#ddf,stroke:#99b

    style P1 fill:#f99,stroke:#933
    style P2 fill:#f99,stroke:#933
    style P3 fill:#9f9,stroke:#393
    style P4 fill:#9f9,stroke:#393,stroke-width:3px
    style P5 fill:#9f9,stroke:#393
    style P6 fill:#ddf,stroke:#99b
    style P7 fill:#ddf,stroke:#99b
```

## Changes

- **Infinite scroll toggle** with info tooltip explaining behavior and supported offsets (Newest, Beginning only)
- **Eager pre-fetching** of the second page when max results >= page size for smoother UX
- **Buffer indicator** showing the full message buffer size in the bottom pagination area
- **Table page size capping** when max results < page size to prevent empty pages
- **Descending timestamp sort** for "Newest" offset in infinite scroll mode so newest messages appear first
- **Auto-disable** infinite scroll when switching to unsupported offsets (Custom, Timestamp, Latest/Live)

## Files changed

- `frontend/src/components/pages/topics/Tab.Messages/index.tsx` — main UI logic for infinite scroll, pagination controls, and tooltip
- `frontend/src/state/backend-api.ts` — message search load-more support
- `frontend/src/stores/topic-settings-store.ts` — persisted infinite scroll toggle state

## Example
infinite-scroll

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages/index.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages/index.tsx
@@ -69,7 +69,6 @@ import {
   SkipBackIcon,
   TabIcon,
   TimerIcon,
-  WarningIcon,
 } from 'components/icons';
 import { Button as RegistryButton } from 'components/redpanda-ui/components/button';
 import { DataTablePagination } from 'components/redpanda-ui/components/data-table';
@@ -512,13 +511,13 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
     sortingParser.withDefault([])
   );
 
-  // Infinite scroll toggle state
-  const [infiniteScrollEnabled, setInfiniteScrollEnabled] = useQueryStateWithCallback<boolean>(
+  // Continuous pagination toggle state
+  const [continuousPaginationEnabled, setContinuousPaginationEnabled] = useQueryStateWithCallback<boolean>(
     {
       onUpdate: (val) => {
-        setSearchParams(props.topic.topicName, { infiniteScrollEnabled: val });
+        setSearchParams(props.topic.topicName, { continuousPaginationEnabled: val });
       },
-      getDefaultValue: () => getSearchParams(props.topic.topicName)?.infiniteScrollEnabled ?? false,
+      getDefaultValue: () => getSearchParams(props.topic.topicName)?.continuousPaginationEnabled ?? false,
     },
     'inf',
     parseAsBoolean.withDefault(false)
@@ -590,10 +589,10 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
       })
     : messages;
 
-  // For infinite scroll, just use the filtered messages directly
+  // For continuous pagination, just use the filtered messages directly
   // We don't use placeholders as they cause page content to shift
   const filteredMessages =
-    infiniteScrollEnabled && startOffset === PartitionOffsetOrigin.EndMinusResults
+    continuousPaginationEnabled && startOffset === PartitionOffsetOrigin.EndMinusResults
       ? [...baseFilteredMessages].sort((a, b) => b.timestamp - a.timestamp)
       : baseFilteredMessages;
 
@@ -638,12 +637,19 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
     };
   }, []);
 
-  // Clear sorting when enabling infinite scroll mode
+  // Clear sorting when enabling continuous pagination mode
   useEffect(() => {
-    if (infiniteScrollEnabled && sorting.length > 0) {
+    if (continuousPaginationEnabled && sorting.length > 0) {
       setSortingState([]);
     }
-  }, [infiniteScrollEnabled, sorting.length, setSortingState]);
+  }, [continuousPaginationEnabled, sorting.length, setSortingState]);
+
+  // Auto-cap page size to maxResults when continuous pagination is enabled
+  useEffect(() => {
+    if (continuousPaginationEnabled && maxResults < pageSize) {
+      setPageSize(maxResults);
+    }
+  }, [continuousPaginationEnabled, maxResults, pageSize, setPageSize]);
 
   // Reset to page 1 when start offset changes (e.g., switching from Newest to Beginning)
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


frontend/src/state/backend-api.ts
```diff
@@ -2831,6 +2842,8 @@ export function createMessageSearch() {
       req.startTimestamp = BigInt(searchRequest.startTimestamp);
       req.partitionId = searchRequest.partitionId;
       req.maxResults = searchRequest.maxResults;
+      req.pageToken = searchRequest.pageToken ?? '';
+      req.pageSize = searchRequest.pageSize ?? 0; // Backend will default to 50 if 0
```
malinskibeniamin: shouldn't we just use the PAGE_SIZE const instead of 0 then

malinskibeniamin: having page size of 0 referenced here just makes it a bit confusing IMO

frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -848,6 +1192,50 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
     },
   ];
 
+  const expanderColumn: ColumnDef<TopicMessage> = {
+    id: 'expander',
+    size: 40,
+    enableSorting: false,
+    cell: ({ row }) =>
+      row.getCanExpand() ? (
+        <button aria-label={row.getIsExpanded() ? 'Collapse row' : 'Expand row'} className="cursor-pointer p-1" onClick={row.getToggleExpandedHandler()} type="button">
```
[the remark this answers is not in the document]
malinskibeniamin: Should we use a new redpanda UI registry <Button> component (or Chakra UI button if it's a legacy page)

frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -1206,48 +1641,217 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
...
+              </div>
+            </div>
+          )}
+
+          {/* Virtual page indicator for infinite scroll mode */}
+          {infiniteScrollEnabled && messages.length > 0 && (
+            <Flex align="center" className="text-muted-foreground text-sm" gap={2} justify="center" mt={2}>
+              <span>
+                Loaded messages {virtualStartIndex + 1}-{virtualStartIndex + messages.length}
+                {` (pages ${windowStartPage + 1}–${windowStartPage + loadedPages} in memory)`}
```
graham-rp, line 1836: Do we want to expose this to users?

frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -1206,48 +1641,217 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
...
+          {/* Warning when maxResults is less than page size */}
+          {infiniteScrollEnabled && maxResults < pageSize && (
+            <Alert mt={4} status="warning">
+              <WarningIcon />
+              <AlertDescription ml={2}>
+                Max Results ({maxResults}) is less than the page size ({pageSize}). Increase Max Results or decrease the
+                page size to allow infinite scrolling to fill each page.
+              </AlertDescription>
+            </Alert>
+          )}
```
graham-rp: This seems a little clunky. Can we be smarter about presentation here (maybe automatically increase max results if it's under page size)?

frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -970,11 +1373,39 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
...
               }}
-              options={[1, 3, 5, 10, 20, 50, 100, 200, 500].map((i) => ({ value: i }))}
+              options={[...[1, 3, 5, 10, 20, 50, 100, 200, 500, 1000, 10_000].map((i) => ({ value: i }))]}
               value={maxResults}
             />
           </Label>
 
+          <Label
+            style={{}}
+            text="Infinite Scroll"
```
graham-rp: $0.02: is it a tiny bit weird to call it 'Infinite Scroll'? I'm not sure I've got a better naming idea, but it seems like it's not quite what we're doing here

graham-rp (approved):
Couple of minor points, but looks good overall!

frontend/src/components/pages/topics/Tab.Messages/index.tsx
c-julin: i agree, ive renamed it again to continuous pagination which is more inline with what it does, but not quite got the same pizzaz :D

frontend/src/components/pages/topics/Tab.Messages/index.tsx
c-julin: great idea, yes i'll get that updated

frontend/src/components/pages/topics/Tab.Messages/index.tsx
c-julin: Only because i dont want users confused when they do the fuzzy search and something doesn't appear

PR #2233, opened 2026-02-19
[UX-902] fix: create user client-side validation

eblairmckee:
[UX-902](https://redpandadata.atlassian.net/browse/UX-902)

Problem: Users got an obtuse server error when clicking "Create" with a username exceeding 128 characters — no client-side validation told them what was wrong, so they had to guess the limit by trial and  error.                              
                                                                                                                                                                                                               
Fix: Added client-side validation with clear error messages for username/password length limits, derived directly from protobuf field constraints so they can't drift from the server.                       
                                                                                                                                                                                                               
  Changes:                                                                                                                                                                                                     
  - proto-constraints.ts (new) — utility that extracts buf.validate annotations from proto schemas and builds Zod validators automatically                                                                     
  - user-create-form-schema.ts (new) — Zod schema using proto-derived constraints for username (max 128) and password (min 3, max 128) with real-time error messages                                           
  - user-create.tsx — migrated from legacy patterns to react-hook-form + Zod + Registry components; validation errors now show inline before submission                                                        
  - user.ts — constants like USERNAME_MAX_LENGTH now derived from proto at runtime instead of hardcoded                                                                                                        
  - user-create.test.tsx (new) — 9 integration tests covering character limit validation, whitespace, duplicate users, and form submission                                                                     
  - proto-constraints.test.ts (new) — unit tests for the proto-to-Zod utility                                                                                                                                  
  - Minor fixes in copy-button.tsx, input.tsx, user-edit-modals.tsx, add-user-step.tsx

## After
<img width="1804" height="743" alt="Screenshot 2026-02-19 at 2 00 30 PM" src="https://github.com/user-attachments/assets/66ad9c20-e3c3-4db6-a2fb-ca5c74ea1d82" />


## Before
<img width="1805" height="634" alt="Screenshot 2026-02-19 at 1 51 14 PM" src="https://github.com/user-attachments/assets/02d94cfd-0684-4b1c-9be8-664e2a43efd6" />
<img width="1815" height="651" alt="Screenshot 2026-02-19 at 1 51 25 PM" src="https://github.com/user-attachments/assets/9e9c4722-9589-4961-9d7c-a6e4973e406a" />


[UX-902]: https://redpandadata.atlassian.net/browse/UX-902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/components/pages/acls/user-create.tsx
+++ b/frontend/src/components/pages/acls/user-create.tsx
@@ -13,7 +13,7 @@ import { create } from '@bufbuild/protobuf';
 import { ConnectError } from '@connectrpc/connect';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { Link, useNavigate } from '@tanstack/react-router';
-import { Loader2, RotateCwIcon } from 'lucide-react';
+import { Loader2, RefreshCcw } from 'lucide-react';
 import { UpdateRoleMembershipRequestSchema } from 'protogen/redpanda/api/dataplane/v1/security_pb';
 import { CreateUserRequestSchema } from 'protogen/redpanda/api/dataplane/v1/user_pb';
 import { useEffect, useMemo, useState } from 'react';
@@ -136,7 +136,7 @@ const UserCreatePage = () => {
 
   return (
     <PageContent>
-      <form className="flex max-w-[460px] flex-col gap-8" onSubmit={form.handleSubmit(onSubmit)}>
+      <form className="mt-4 flex max-w-[460px] flex-col gap-8" onSubmit={form.handleSubmit(onSubmit)}>
         <Controller
           control={form.control}
           name="username"
@@ -182,14 +182,14 @@ const UserCreatePage = () => {
                         type="button"
                         variant="ghost"
                       >
-                        <RotateCwIcon className="size-4" />
+                        <RefreshCcw className="size-3" />
                       </Button>
                     </TooltipTrigger>
                     <TooltipContent>Generate new random password</TooltipContent>
                   </Tooltip>
                   <Tooltip>
                     <TooltipTrigger asChild>
-                      <CopyButton content={field.value} size="icon" variant="ghost" />
+                      <CopyButton content={field.value} variant="ghost" />
                     </TooltipTrigger>
                     <TooltipContent>Copy password</TooltipContent>
                   </Tooltip>
--- a/frontend/src/components/redpanda-ui/components/copy-button.tsx
+++ b/frontend/src/components/redpanda-ui/components/copy-button.tsx
@@ -18,7 +18,7 @@ const buttonVariants = cva(
         outline:
           'border bg-background shadow-xs hover:bg-accent hover:text-accent-foreground dark:border-input dark:bg-input/30 dark:hover:bg-input/50',
         secondary: 'bg-primary text-primary-foreground shadow-xs hover:bg-primary/90',
-        ghost: 'hover:bg-accent hover:text-accent-foreground dark:hover:bg-accent/50',
+        ghost: 'hover:bg-accent dark:hover:bg-accent/50 text-action-primary',
       },
       size: {
         md: 'h-9 px-4 py-2 has-[>svg]:px-3',
--- a/frontend/src/components/redpanda-ui/components/input.tsx
+++ b/frontend/src/components/redpanda-ui/components/input.tsx
@@ -216,7 +216,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
   }
 );
 
-const inputEndClassNames = 'absolute top-1/2 -translate-y-1/2 z-10 pointer-events-none right-2';
+const inputEndClassNames = 'absolute top-1/2 -translate-y-1/2 z-10 pointer-events-none right-2 flex items-center justify-center';
 
 const InputContext = createContext<{
   setStartWidth: (width: number) => void;


frontend/src/utils/user.ts
```diff
@@ -14,20 +14,57 @@
...
+  let result = '';
+  for (const n of randomValues) {
+    const index = n % alphabet.length;
+    const sym = alphabet[index];
+
+    result += sym;
+  }
+
+  return result;
+}
```
malinskibeniamin, line 70: I think we should be using bcrypt or similar to generate the hash, crypto is getting better in native nodejs but nonetheless I would feel safer if we use bcrypt

frontend/src/components/pages/acls/user-create-form-schema.ts
```diff
@@ -0,0 +1,46 @@
...
+    username: (protoSchema.shape.name as z.ZodString)
+      .regex(USERNAME_REGEX, USERNAME_ERROR_MESSAGE)
+      .refine((val) => !existingUsers.includes(val), 'User already exists'),
+    // Proto provides min(3) + max(128)
+    password: protoSchema.shape.password as z.ZodString,
+    // Keep as string enum for form UX (proto uses numeric enum)
+    mechanism: z.enum(SASL_MECHANISMS),
+    // Not in proto — frontend-only field for role assignment
+    roles: z.array(z.string()).default([]),
+  });
```
malinskibeniamin, line 37: Is there any way we can get it working without casting such as `as z.ZodString`?

Ideally I would like to rely as much as possible on `zod` and `protovaildate` internals.

That being said, we need to discuss whether we go with:
a) merging the zod and protovalidate validation step and present the user with errors accordingly
b) 1st rely on the `z.infer()` handler where we get all the zod powered schema validation on every form change and then do a 2nd pass on protovalidate on submission? Ideally the 2 schemas should not be very far away from each other
c) have a dedicated package which would have the power to convert all protovalidate schemas to zod, but then we would lose on the protovalidate descriptive errors, so perhaps we may want to consolidate them. Just talking out loud

Personally I would be leaning towards the approach in Gateway where you have `protovalidateResolver`, but with the additional zod resolver.

frontend/src/components/pages/acls/user-create.tsx
```diff
@@ -9,455 +9,364 @@
...
+                {...field}
+                autoComplete="off"
+                autoFocus
+                placeholder="Username"
+                spellCheck={false}
+                testId="create-user-name"
               />
+              <FieldDescription>
+                Must not contain any whitespace. Dots, hyphens and underscores may be used. Maximum{' '}
+                {USERNAME_MAX_LENGTH} characters.
```
malinskibeniamin, line 156: did you mean to interpolate the string there

frontend/src/utils/proto-constraints.ts
malinskibeniamin, line 1: before we continue with this approach, let's consider creating protovalidate resolver like disccused in this thread https://github.com/redpanda-data/console/pull/2233/changes#r2831228170

frontend/src/components/pages/acls/user-create-form-schema.ts
malinskibeniamin: Discussed with @eblairmckee, we can do this in a separate PR.

frontend/src/components/pages/acls/user-create.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { useCreateUserMutation, useLegacyListUsersQuery } from 'react-query/api/user';
+
+import UserCreatePage from './user-create';
+
+// jsdom polyfills
+global.ResizeObserver = class ResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+};
```
malinskibeniamin, line 63: If this is required, I would put it in the setup utils because it probably should be reused across tests.

frontend/src/components/pages/acls/user-create.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+    } as any);
+
+    vi.mocked(useUpdateRoleMembershipMutation).mockReturnValue({
+      mutateAsync: vi.fn().mockResolvedValue({}),
+      isPending: false,
+    } as any);
+
+    vi.mocked(useListRolesQuery).mockReturnValue({
+      data: { roles: [] },
+    } as any);
```
malinskibeniamin, line 102: Let's use an approach such as
```tsx
    const getAIAgentMock = vi.fn().mockReturnValue(getAIAgentResponse);

    const transport = createRouterTransport(({ rpc }) => {
      rpc(getAIAgent, getAIAgentMock);
    });
```

malinskibeniamin: I think if there is a REST endpoint it makes sense to mock the function for now (if we can't use connectRPC, connect RPC migration can be a separate PR too). For connect native RPC calls, let's use `createRouterTransport` pattern.

frontend/src/components/pages/acls/user-create.tsx
```diff
@@ -9,455 +9,364 @@
  * by the Apache License, Version 2.0
  */
 
+import { create } from '@bufbuild/protobuf';
+import { ConnectError } from '@connectrpc/connect';
+import { zodResolver } from '@hookform/resolvers/zod';
+import { Link, useNavigate } from '@tanstack/react-router';
+import { Loader2, RefreshCcw } from 'lucide-react';
```
malinskibeniamin, line 16: I think you can use `components/icons` file

frontend/src/components/pages/acls/user-create.tsx
```diff
@@ -9,455 +9,364 @@
...
-  const [selectedRoles, setSelectedRoles] = useState<string[]>([]);
 
   const { data: usersData } = useLegacyListUsersQuery();
-  const users = usersData?.users?.map((u) => u.name) ?? [];
+  const existingUsers = useMemo(() => usersData?.users?.map((u) => u.name) ?? [], [usersData]);
 
-  const isValidUsername = validateUsername(username);
-  const isValidPassword = validatePassword(password);
+  const { mutateAsync: createUser, isPending: isCreating } = useCreateUserMutation();
+  const { mutateAsync: updateRoleMembership } = useUpdateRoleMembershipMutation();
```
malinskibeniamin, line 72: Should we also check on `isPending` for the other mutation?

frontend/src/components/pages/acls/user-create.tsx
```diff
@@ -9,455 +9,364 @@
...
+      setSubmittedData({
+        username: values.username,
+        password: values.password,
+        mechanism: values.mechanism,
       });
-    } finally {
-      setIsCreating(false);
+    } catch (error) {
+      const connectError = ConnectError.from(error);
+      toast.error(formatToastErrorMessageGRPC({ error: connectError, action: 'create', entity: 'user' }));
```
malinskibeniamin, line 118: 👏

PR #2236, opened 2026-02-20
frontend: fix message viewer subComponent instability causing Monaco remounts

weeco:
## Summary

- Wrap the inline `subComponent` prop in `useCallback` to stabilize its identity across renders, preventing unnecessary unmount/remount cycles of expanded-row Monaco editors during rapid filter input
- Memoize `JSON.stringify` in `KowlJsonView` with `useMemo` to skip redundant serialization when `srcObj` hasn't changed

### DIFF
--- a/frontend/src/components/pages/topics/Tab.Messages/index.tsx
+++ b/frontend/src/components/pages/topics/Tab.Messages/index.tsx
@@ -905,6 +905,30 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
     startOffset >= 0 ? PartitionOffsetOrigin.Custom : startOffset
   ) as PartitionOffsetOriginType;
 
+  const subComponent = useCallback(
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -1227,25 +1227,28 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
...
-                  })
-                }
-                msg={original}
-                onCopyKey={(msg) => onCopyKey(msg, toast)}
-                onCopyValue={(msg) => onCopyValue(msg, toast)}
-                onDownloadRecord={() => {
-                  setDownloadMessages([original]);
-                }}
-              />
+            subComponent={useCallback(
```
malinskibeniamin: Let's avoid calling useCallback inside the prop. This may break some rules of hooks in react. Try to move it into a variable

malinskibeniamin (approved):

PR #2242, opened 2026-02-25
backend: add SR context to endpoints/ compat URL

r-vasquez:
This will detect whether the Schema Registry
Context feature is enabled in the cluster. This
will require users that which to use this feature
to also have configured the Admin API as there is
no way of telling this within the SR API directly.

### DIFF
--- a/backend/pkg/console/redpanda_feature.go
+++ b/backend/pkg/console/redpanda_feature.go
@@ -34,7 +34,14 @@ const (
 
 	// redpandaFeatureSchemaRegistryACL represents Schema Registry ACL feature.
 	redpandaFeatureSchemaRegistryACL redpandaFeature = "redpanda_feature_schema_registry_acl"
+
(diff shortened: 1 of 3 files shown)


backend/pkg/console/redpanda_feature.go
```diff
@@ -64,6 +67,17 @@ func (*Service) checkRedpandaFeature(ctx context.Context, redpandaCl redpandafac
 			}
 		}
 		return true
+	case redpandaFeatureSchemaRegistryContexts:
+		cfg, err := redpandaCl.SingleKeyConfig(ctx, "schema_registry_enable_qualified_subjects")
```
c-julin: supernit : can be a const "schema_registry_enable_qualified_subjects"

r-vasquez: Done, fixed

PR #2250, opened 2026-03-02
refactor: remove MobX 

jvorcak:
Removes observer() from components that don't access MobX observables directly:
- error-display.tsx, error401-page.tsx, kowl-json-view.tsx
- operation.tsx, require-auth.tsx
- message-headers.tsx, message-meta-data.tsx, payload-component.tsx

### DIFF
--- a/frontend/src/components/pages/admin/admin-debug-bundle-progress.tsx
+++ b/frontend/src/components/pages/admin/admin-debug-bundle-progress.tsx
@@ -14,7 +14,7 @@ import { Link } from '@tanstack/react-router';
 import { api } from '../../../state/backend-api';
 import '../../../utils/array-extensions';
 import { Box, Button, Flex, Text } from '@redpanda-data/ui';
-import type { FC } from 'react';
+import { type FC, useEffect } from 'react';
 
 import DebugBundleOverview from './debug-bundle-overview';
 import { appGlobal } from '../../../state/app-global';
@@ -42,6 +42,16 @@ export default class AdminPageDebugBundleProgress extends PageComponent {
 }
 
 const AdminPageDebugBundleProgressContent: FC = () => {
+  // Poll for status updates while bundle is in progress
+  useEffect(() => {
+    const interval = setInterval(() => {
+      api.refreshDebugBundleStatuses().catch(() => {
+        // Error handling managed by API layer
+      });
+    }, 2000);
+    return () => clearInterval(interval);
+  }, []);
+
   return (
     <Box>
       <Text data-testid="debug-bundle-description">
--- a/frontend/src/components/pages/overview/overview.tsx
+++ b/frontend/src/components/pages/overview/overview.tsx
@@ -72,6 +72,9 @@ class Overview extends PageComponent {
     api.refreshDebugBundleStatuses().catch(() => {
       // Error handling managed by API layer
     });
+    api.listLicenses().catch(() => {
+      // Error handling managed by API layer
+    });
   }
 
   render() {
--- a/frontend/src/components/pages/page.ts
+++ b/frontend/src/components/pages/page.ts
@@ -60,6 +60,9 @@ export abstract class PageComponent<TRouteParams = NoRouteParams> extends React.
       useRpcnSecretManagerStore.subscribe(update),
       useTransformsStore.subscribe(update),
     ];
+    // Force a re-render to catch any store updates that arrived before this subscription was set up
+    // (e.g. fast API responses from initPage's constructor-time calls)
+    this.forceUpdate();
   }
 
   componentWillUnmount() {


frontend/src/components/layout/sidebar.tsx
```diff
@@ -220,7 +219,7 @@ function SidebarNavItem({ item, isActive, onNavClick }: NavItemProps) {
   );
 }
 
-const SidebarNavigation = observer(() => {
```
malinskibeniamin, line 223: Let's double check if that still works when in embedded mode with cloud UI, sidebar items need to refresh

frontend/src/components/pages/acls/operation.tsx
```diff
@@ -34,47 +33,45 @@ const OptionContent: FC<{
...
+  style?: CSSProperties;
+}) => {
+  const disabled = p.disabled ?? false;
 
-    const operationName =
-      typeof p.operation === 'string'
-        ? p.operation
-        : Object.keys(AclOperation).find((key) => AclOperation[key as keyof typeof AclOperation] === p.operation) ||
-          'Unknown';
+  const operationName =
```
malinskibeniamin, line 45: I think this deserves its own utility function

frontend/src/components/pages/acls/operation.tsx
```diff
@@ -34,47 +33,45 @@ const OptionContent: FC<{
...
+      <SingleSelect<AclStrPermission>
+        components={{
+          DropdownIndicator: null,
+        }}
+        isDisabled={disabled}
+        onChange={p.onChange}
+        options={[
+          {
+            value: 'Any',
+            label: <OptionContent icon={icons.minus}>Not set</OptionContent>,
```
malinskibeniamin, line 62: Not set or Any?

frontend/src/components/pages/acls/role-details.tsx
```diff
@@ -22,154 +21,151 @@ import { api, type RolePrincipal, rolesApi } from '../../../state/backend-api';
...
+async function refreshRoleData(force: boolean) {
+  if (api.userData !== null && api.userData !== undefined && !api.userData.canListAcls) {
+    return;
   }
+  await Promise.allSettled([api.refreshAcls(AclRequestDefault, force), api.refreshServiceAccounts()]);
+  await rolesApi.refreshRoles();
+  await rolesApi.refreshRoleMembers();
+}
 
+class RoleDetailsPage extends PageComponent<{ roleName: string }> {
```
malinskibeniamin, line 35: I guess functional components can be used later on?

frontend/src/components/pages/connect/helper.tsx
```diff
@@ -678,20 +672,20 @@ export const ConfirmModal = observer(<T,>(props: ConfirmModalProps<T>) => {
...
     </AlertDialog>
   );
-});
+};
 
 // Takes an observable object that is either a single connector (runningTasks and totalTasks properties)
 // or an array of connectors (in which case it will show the sum)
 
-export const TasksColumn = observer((props: { observable: ClusterConnectors | ClusterConnectorInfo }) => {
+export const TasksColumn = (props: { observable: ClusterConnectors | ClusterConnectorInfo }) => {
```
malinskibeniamin, line 690: Do we still need to keep the observable if we removed the observer pattern?

frontend/src/components/pages/connect/helper.tsx
```diff
@@ -754,90 +748,90 @@ export const ConnectorsColumn = observer((props: { observable: ConnectorInfo | C
...
-    }
-
-    let stateContent = (
-      <span className="capitalize" style={{ display: 'flex', alignItems: 'center', gap: '4px', height: 'auto' }}>
-        {icon}
-        {String(state).toLowerCase()}
-      </span>
-    );
+  let icon: JSX.Element = <></>;
+  if (state === ConnectorState.Running) {
```
malinskibeniamin, line 766: I think it should be its own component or utility method

frontend/src/components/pages/connect/helper.tsx
jvorcak: it's just a variable name, I see it's not optimal, but I think this can be addressed later on, I want to make as few changes as possible.

frontend/src/components/misc/router-sync.tsx
```diff
@@ -26,7 +25,7 @@ import { api } from '../../state/backend-api';
  * In embedded mode, it also notifies the shell (Cloud UI) when Console
  * navigates internally, allowing the shell to sync its router state.
  */
-export const RouterSync = observer(() => {
+export const RouterSync = () => {
```
malinskibeniamin, line 28: Let's double check embedded mode functionality with cloud UI route syncing

frontend/package.json
```diff
@@ -95,8 +95,6 @@
     "jwt-decode": "^4.0.0",
     "lottie-react": "^2.4.1",
     "lucide-react": "^0.563.0",
-    "mobx": "^6.10.0",
-    "mobx-react": "^7.6.0",
```
malinskibeniamin, line 99: big moment if we can finally get there. thank you

frontend/bun.lock
malinskibeniamin, line 1: also regenerate `yarn.lock`

frontend/src/config.ts
```diff
@@ -345,11 +344,14 @@ setTimeout(() => {
...
-    // Disposer intentionally not captured — this reaction lives for the app's lifetime.
-    autorun(updateSidebarItems);
+    // Call once on initialization; also re-call whenever endpointCompatibility
+    // becomes available (it starts null and is populated after the first API fetch).
+    updateSidebarItems();
+    useApiStore.subscribe((state, prev) => {
+      if (state.endpointCompatibility !== prev.endpointCompatibility) {
+        updateSidebarItems();
+      }
+    });
```
malinskibeniamin, line 354: would this be sufficient? this would mean sidebar items get updated twice, I just want to make sure no race conditions appear.

frontend/src/hooks/use-list-quotas.ts
```diff
@@ -14,9 +14,8 @@ import { useQuery } from '@connectrpc/connect-query';
 
 import { ListQuotasRequestSchema } from '../protogen/redpanda/api/dataplane/v1/quota_pb';
 import { listQuotas } from '../protogen/redpanda/api/dataplane/v1/quota-QuotaService_connectquery';
-import { MAX_PAGE_SIZE } from '../react-query/react-query.utils';
 
-const request = create(ListQuotasRequestSchema, { pageSize: MAX_PAGE_SIZE });
+const request = create(ListQuotasRequestSchema, { pageSize: 1000 });
```
malinskibeniamin: why not keep using `MAX_PAGE_SIZE`?

frontend/src/components/pages/rp-connect/pipelines-edit.tsx
```diff
@@ -71,182 +56,170 @@ class RpConnectPipelinesEdit extends PageComponent<{ pipelineId: string }> {
...
-            value={this.tasks}
-          />
-        </FormField>
+const RpConnectPipelinesEditContent = ({ pipeline, pipelineId }: { pipeline: Pipeline; pipelineId: string }) => {
+  const [displayName, setDisplayName] = useState(pipeline.displayName);
+  const [description, setDescription] = useState(pipeline.description);
+  const [tasks, setTasks] = useState(cpuToTasks(pipeline?.resources?.cpuShares) || MIN_TASKS);
+  const [editorContent, setEditorContent] = useState(pipeline.configYaml);
+  const [isUpdating, setIsUpdating] = useState(false);
+  const [tags] = useState(pipeline.tags);
```
malinskibeniamin: Do we need `useState` hook if we are never changing the state for the tags? I think they will always remain the same based on this code?

frontend/src/components/pages/rp-connect/pipelines-edit.tsx
```diff
@@ -71,182 +56,170 @@ class RpConnectPipelinesEdit extends PageComponent<{ pipelineId: string }> {
...
+          isClosable: false,
+          title: 'Pipeline updated',
+        });
+
+        const retUnits = cpuToTasks(r.response?.pipeline?.resources?.cpuShares);
+        if (retUnits && tasks !== retUnits) {
           toast({
-            status: 'success',
-            duration: 4000,
+            status: 'warning',
```
malinskibeniamin: should it be info instead? resizing pipeline should not be a warning, it's just a message

malinskibeniamin: nevermind, that was the functionality before, fine with either

frontend/src/config.ts
jvorcak: Good question — no race condition here. The two calls are intentional: the first fires immediately so the sidebar is populated on startup (in case `endpointCompatibility` is already set); the second fires when it changes (e.g., after the initial API fetch populates it from null). Since `updateSidebarItems()` reads the current store snapshot and calls `setSidebarItems()` synchronously, calling it twice is harmless — the second just overwrites with the same or more accurate result.

frontend/bun.lock
jvorcak: `yarn` is not installed in this environment — can you regenerate `yarn.lock` locally with `yarn install` after merging `bun.lock`?

frontend/src/components/pages/topics/topic-details.tsx
```diff
@@ -141,28 +139,46 @@ const warnIcon = (
...
-  constructor(props: Readonly<PageProps<{ topicName: string }>>) {
-    super(props);
+  // consumers are lazy loaded because they're (relatively) expensive
+  if (uiSettings.topicDetailsActiveTabKey === 'consumers') {
+    api.refreshTopicConsumers(topicName, force);
+  }
 
-    if (isServerless()) {
-      this.topicTabs.removeAll((x) => x.id === 'documentation');
-    }
```
malinskibeniamin, line 155: Does this logic still exist anywhere? Do we still need it?

frontend/src/components/pages/topics/topic-details.tsx
```diff
@@ -172,311 +188,278 @@ class TopicDetails extends PageComponent<{ topicName: string }> {
...
-    if (config === null || config.error !== null) {
-      return null;
-    }
-    return config.configEntries;
+  render() {
+    // Touch observables so PageComponent's Reaction tracks them and triggers re-renders
+    // when async data arrives (TopicDetailsContent is a functional component child
+    // that reads api.* outside the MobX tracking context).
+    void api.topics;
+    void api.topicConfig.get(this.props.topicName);
```
malinskibeniamin: not sure if still relevant given mobx is fully cleaned up

frontend/src/components/pages/topics/topic-produce.tsx
```diff
@@ -225,7 +223,7 @@ const PublishTopicForm: FC<{ topicName: string }> = observer(({ topicName }) =>
     }
 
     return partitions;
-  });
+  })();
```
malinskibeniamin, line 226: nit: having anonymous arrow function called makes it hard to read but it's fine for now

frontend/src/components/pages/topics/topic-details.tsx
jvorcak: Yes, it's still needed — `api.topicPartitions.get(topic.topicName)` is read in `tab-partitions.tsx`. Removed the stale comment about 'statistics bar' and updated it to reflect the actual usage.

frontend/src/components/pages/topics/topic-details.tsx
jvorcak: You're right — removed. `PageComponent` now re-renders via Zustand store subscriptions (`useApiStore.subscribe(forceUpdate)`), so the MobX-style observable touches are no longer needed.

frontend/bun.lock
malinskibeniamin: You can do "bun install --yarn"

frontend/src/components/pages/admin/admin-debug-bundle-progress.tsx
```diff
@@ -43,76 +36,85 @@ export default class AdminPageDebugBundleProgress extends PageComponent {
...
-          )}
-        </Box>
+const AdminPageDebugBundleProgressContent: FC = () => {
+  // Poll for status updates while bundle is in progress
+  useEffect(() => {
+    const interval = setInterval(() => {
+      api.refreshDebugBundleStatuses().catch(() => {
+        // Error handling managed by API layer
+      });
+    }, 2000);
```
malinskibeniamin, line 51: In the future we could replace it by using tanstack query polling mode

PR #2253, opened 2026-03-03
feat: add react-doctor

malinskibeniamin:
## Summary

Integrate [react-doctor](https://www.react.doctor/) as a CI quality gate and fix all detected React performance anti-patterns across the frontend codebase to achieve a **100/100 score**.

- Add react-doctor CI workflow and configuration
- Fix React Compiler compatibility issues (immutability, hooks, refs, memoization)
- Eliminate `useEffect`-based state sync in favor of render-time derivation patterns
- Replace array-index-as-key anti-patterns with stable keys
- Add `use no memo` directives and React Compiler source filters where needed
- Fix accessibility, async-parallel, and error-boundary warnings
- Wrap Zustand store with typed hook (`useApiStoreHook`) for React Compiler compatibility

### CI & Tooling
- `.github/workflows/frontend-react-doctor.yml` — runs react-doctor on PRs touching `frontend/`, fails on warnings
- `frontend/react-doctor.config.json` — config with all rules enabled, ignores vendored UI components (`redpanda-ui/`, `ui/`, `ai-elements/`), `diff: false` for full-repo analysis
- `frontend/package.json` — add `react-doctor` dependency and `bun run doctor` script

### React Compiler Fixes
- Replace `watch()` with `useWatch({ control, name })` for memoization compatibility in `topic-produce.tsx`
- Extract external store mutations into standalone functions to avoid immutability violations
- Add `use no memo` directives for components incompatible with React Compiler
- Add React Compiler source filter in `rsbuild.config.ts`
- Wrap `useApiStore` with `useApiStoreHook` to provide proper typed selectors

### Rule-by-Rule Fixes (145 files, +3774 / -1824)

| Rule | Fix applied |
|------|-------------|
| `no-derived-useState` | Replace `useEffect` state sync with render-time derivation (`prevX` pattern) |
| `no-cascading-set-state` | Move state updates out of effects into event handlers or render |
| `no-effect-event-handler` | Move event subscriptions/handlers out of `useEffect` |
| `no-array-index-as-key` | Replace array indices with stable unique keys |
| `no-render-in-render` | Extract inline component definitions to module scope |
| `no-prevent-default` | Use proper form/event handling patterns |
| `prefer-useReducer` | Replace complex multi-setState with `useReducer` where flagged |
| `prefer-dynamic-import` | Add dynamic imports where appropriate |
| `static-components` | Extract static JSX outside render |
| `purity` | Remove side effects from render path |
| `hooks` | Fix hook ordering and conditional hook calls |
| `immutability` | Replace direct mutations with immutable updates |
| `refs` | Fix ref access patterns (no writes in render) |
| `set-state-in-effect` | Remove setState calls from effects |
| `error-boundaries` | Add proper error boundary patterns |
| `rules-of-hooks` | Fix ESLint react-hooks/rules-of-hooks violations |

### Notable Refactors
- **Knowledge base details**: Remove `formHasChanges` useState + useEffect sync, use `isDirty` from `form.formState` directly; replace tag sync effect with render-time derivation
- **Admin roles page**: Extract `refreshData` callbacks with `useCallback`
- **Payload component**: Refactor ref write patterns and fix JSON render data types
- **Backend API store**: Add `useApiStoreHook` wrapper for type-safe Zustand selectors

### Other
- `frontend/globals.css` — add `prefers-reduced-motion` CSS for accessibility
- Various type fixes across components touched by react-doctor changes

## Test Plan
- [ ] CI passes: type-check, lint, unit tests, integration tests, e2e tests
- [ ] React Doctor CI check passes with 0 errors and 0 warnings
- [ ] Knowledge base edit flow works correctly (edit, save, cancel)
- [ ] Topic produce page works correctly (publish messages with different encodings)
- [ ] Shadow links CRUD flow works correctly
- [ ] Pipeline editor and onboarding wizard function correctly
- [ ] ACL/roles management pages work correctly

### DIFF
--- a/frontend/src/components/pages/rp-connect/pipeline/index.tsx
+++ b/frontend/src/components/pages/rp-connect/pipeline/index.tsx
@@ -530,33 +530,6 @@ export default function PipelinePage() {
     </>
   );
 
-  const renderContent = () => {
-    if (mode === 'create') {
-      return content;
-    }
-
-    if (mode === 'view' && pipeline) {
-      return (
-        <Card size="full">
-          <Tabs defaultValue="configuration">
-            <TabsList>
-              <TabsTrigger value="configuration">Configuration</TabsTrigger>
-              <TabsTrigger value="logs">Logs</TabsTrigger>
-            </TabsList>
-            <TabsContents>
-              <TabsContent value="configuration">{content}</TabsContent>
-              <TabsContent value="logs">
-                <LogsTab pipeline={pipeline} />
-              </TabsContent>
-            </TabsContents>
-          </Tabs>
-        </Card>
-      );
-    }
-
-    return <Card size="full">{content}</Card>;
-  };
-
   return (
     <div>
       {mode === 'edit' && (
@@ -570,7 +543,24 @@ export default function PipelinePage() {
             <Toolbar pipelineId={pipelineId} pipelineName={form.getValues('name')} pipelineState={pipeline?.state} />
           )}
 
-          {renderContent()}
+          {mode === 'create' && content}
+          {mode === 'view' && pipeline && (
+            <Card size="full">
+              <Tabs defaultValue="configuration">
+                <TabsList>
+                  <TabsTrigger value="configuration">Configuration</TabsTrigger>
+                  <TabsTrigger value="logs">Logs</TabsTrigger>
+                </TabsList>
+                <TabsContents>
+                  <TabsContent value="configuration">{content}</TabsContent>
+                  <TabsContent value="logs">
+                    <LogsTab pipeline={pipeline} />
+                  </TabsContent>
+                </TabsContents>
+              </Tabs>
+            </Card>
+          )}
+          {mode === 'edit' && <Card size="full">{content}</Card>}
         </div>
         {(mode === 'create' || mode === 'edit') && (
           <CreatePipelineSidebar
--- a/frontend/src/components/pages/rp-connect/pipeline/pipeline-flow-nodes.tsx
+++ b/frontend/src/components/pages/rp-connect/pipeline/pipeline-flow-nodes.tsx
@@ -46,10 +46,10 @@ export function PipelineFlowSkeleton({ error }: PipelineFlowSkeletonProps) {
         {SKELETON_SECTIONS.map((section) => (
           <div className="flex flex-col gap-2" key={section.label}>
             <Skeleton className={error ? 'animate-none! opacity-40' : ''} variant="text" width="xs" />
-            {Array.from({ length: section.leaves }, (_, i) => (
+            {Array.from({ length: section.leaves }, (_, leafIndex) => (
               <Skeleton
                 className={cn('ml-10', error ? 'animate-none! opacity-40' : '')}
-                key={`${section.label}-${i}`}
+                key={`${section.label}-leaf-${leafIndex}`}
                 size="lg"
                 variant="rounded"
                 width="md"


frontend/package.json
```diff
@@ -37,6 +37,8 @@
     "format:check": "biome format src/*",
     "format": "biome format src/* --write",
     "load-git-hooks": "cd .. && git config core.hooksPath \"./.git-hooks\"",
+    "doctor": "bunx react-doctor . -y --fail-on error",
+    "quality:gate": "bun run lint && bun run doctor && bun run type:check && bun run build && bun run test:unit && bun run test:integration",
```
graham-rp, line 41: 👍 maybe even `check` or something we'd definitely remember before committing?

frontend/src/components/pages/knowledgebase/details/knowledge-base-details-page.tsx
```diff
@@ -282,9 +277,12 @@ export const KnowledgeBaseDetailsPage = () => {
   // Subscribe to form changes by accessing isDirty during render
   const { isDirty } = form.formState;
 
-  useEffect(() => {
+  // Sync isDirty to formHasChanges (adjusting state during render instead of useEffect)
+  const [prevIsDirty, setPrevIsDirty] = useState(isDirty);
+  if (isDirty !== prevIsDirty) {
+    setPrevIsDirty(isDirty);
```
graham-rp: $0.02: it feels like there's probably a cleaner way to do this, right?

frontend/src/components/pages/topics/topic-details.tsx
```diff
@@ -294,11 +294,12 @@ class TopicDetails extends PageComponent<{ topicName: string }> {
         () => topic,
         'partitions',
         'viewPartitions',
-        <Flex gap={1}>
-          Partitions
+        <Flex gap={1} key="partitions-title">
```
graham-rp: I'm not sure I'm following the need for a key here

graham-rp (approved):

frontend/package.json
malinskibeniamin: We already have "check" used across various repos for other purposes.

I will make it deeply ingrained into LLM use so you won't have to remember ideally. Every UI will have this command called `quality:gate` to run so if you switch between projects, it should be seamless

frontend/src/components/pages/knowledgebase/details/basic-info-section.tsx
```diff
@@ -102,7 +104,7 @@ export const BasicInfoSection = ({
           <div className="flex gap-2">
             {isEditMode ? (
               <>
-                <Button disabled={!formHasChanges || isUpdating} onClick={onSave} variant="primary">
```
graham-rp, line 105: out of curiosity, why the name change?

frontend/src/components/pages/knowledgebase/details/knowledge-base-details-page.tsx
malinskibeniamin: Good call — the previous code had `updatePageTitle` inside a `useEffect` which was unnecessary since it is synchronous and idempotent. Calling it during render is the React-recommended pattern for derived values (no effect needed). That said, I agree it reads oddly. I have refactored it to move the title update into the parent route loader/component where it belongs, keeping the details component pure.

frontend/src/components/pages/topics/topic-details.tsx
malinskibeniamin: This was resolved during the rebase — the topic-details.tsx changes were already incorporated into master via the MobX removal refactor, so the key prop change is no longer in the diff.

frontend/src/components/pages/knowledgebase/details/basic-info-section.tsx
malinskibeniamin: Renamed `formHasChanges` → `isDirty` to use react-hook-form's built-in `isDirty` from `formState` directly, eliminating the redundant `useState` + `useEffect` sync that was tracking the same value. Less code, same behavior.

PR #2257, opened 2026-03-04
fix: fixed pipeline component names on pipelines list page

eblairmckee:
Connect pipeline list was showing `label` for some component names because the config parsing logic was picking up the first key in a component config, but template output `label` as the first key, so needed to add an exclusion when parsing. 

Before
Screenshot 2026-03-04 at 8 46 14 AM

After
Screenshot 2026-03-04 at 8 45 56 AM

### DIFF
--- a/frontend/src/components/pages/rp-connect/utils/yaml.ts
+++ b/frontend/src/components/pages/rp-connect/utils/yaml.ts
@@ -519,9 +519,35 @@ export const getConnectTemplate = ({
 // Config Component Parsing (used by pipeline list)
 // ============================================================================
 
-/** Extract the first key from an object, or undefined if not a valid object. */
-const firstKey = (obj: unknown): string | undefined =>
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/rp-connect/utils/yaml.ts
```diff
@@ -519,9 +519,35 @@ export const getConnectTemplate = ({
...
+  'while',
+  'workflow',
+] as const;
+
+/** Extract the component name from an object, skipping reserved metadata keys like `label`. */
+const componentName = (obj: unknown): string | undefined => {
+  if (!obj || typeof obj !== 'object' || Array.isArray(obj)) {
+    return;
+  }
+  return Object.keys(obj).find((k) => !RESERVED_COMPONENT_KEYS.has(k));
```
graham-rp, line 549: are we expecting a bunch more down the line?

eblairmckee: not sure.. but down the line this will actually be coming from the connect backend. so temporary solution, thankfully.

PR #2260, opened 2026-03-05
fix: resolve Kafka Connect connector creation hang after MobX removal

c-julin:
## Summary

- After MobX was removed from `ConnectClusterStore` and `ConnectorPropertiesStore`, several components still relied on `observer` to reactively track plain boolean properties (`isInitialized`, `initPending`), causing the UI to hang on a skeleton loader indefinitely
- Removed `observer` wrapper from 5 dynamic-ui child components (`ConnectorStepComponent`, `PropertyGroupComponent`, `PropertyComponent`, `ErrorWrapper`, `TopicInput`) that received `Property`/`PropertyGroup` objects with circular references (`property.propertyGroup.properties[n]`), which caused MobX's `deepEnhancer` to stack overflow
- Converted `TopicInput` from `useLocalObservable` to `useState`/`useMemo` to avoid storing `Property` objects in MobX observables

### DIFF
--- a/frontend/src/state/connect/state.test.ts
+++ b/frontend/src/state/connect/state.test.ts
@@ -21,7 +21,7 @@ vi.mock('../backend-api', () => ({
   },
 }));
 
-import { ConnectorPropertiesStore } from './state';
+import { ConnectorPropertiesStore, SecretsStore } from './state';
 import { api } from '../backend-api';
 
 function createMockProperty(overrides: {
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/connect/dynamic-ui/components.tsx
```diff
@@ -28,7 +27,14 @@ export type ConfigPageProps = {
   context: 'CREATE' | 'EDIT';
 };
 
-export const ConfigPage: React.FC<ConfigPageProps> = observer(({ connectorStore, context }: ConfigPageProps) => {
+export const ConfigPage: React.FC<ConfigPageProps> = ({ connectorStore, context }: ConfigPageProps) => {
+  const [, forceUpdate] = useReducer((x: number) => x + 1, 0);
```
malinskibeniamin: not too sure about this, feels like a hack

PR #2320, opened 2026-03-17
Adds line-height for InlineCode component

jvorcak:
Screenshot 2026-03-17 at 19 08 57

### DIFF
--- a/frontend/src/components/redpanda-ui/components/typography.tsx
+++ b/frontend/src/components/redpanda-ui/components/typography.tsx
@@ -213,7 +213,7 @@ interface InlineCodeProps extends React.HTMLAttributes<HTMLElement>, SharedProps
 export function InlineCode({ className, children, testId, ...props }: InlineCodeProps) {
   return (
     <code
-      className={cn('relative rounded bg-muted px-[0.3rem] py-[0.2rem] font-mono font-semibold text-sm', className)}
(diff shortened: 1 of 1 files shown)


malinskibeniamin (approved):

frontend/src/components/redpanda-ui/components/typography.tsx
```diff
@@ -213,7 +213,7 @@ interface InlineCodeProps extends React.HTMLAttributes<HTMLElement>, SharedProps
 export function InlineCode({ className, children, testId, ...props }: InlineCodeProps) {
   return (
     <code
-      className={cn('relative rounded bg-muted px-[0.3rem] py-[0.2rem] font-mono font-semibold text-sm', className)}
+      className={cn('relative rounded bg-muted px-[0.3rem] py-[0.2rem] font-mono font-semibold text-sm leading-8', className)}
```
eblairmckee, line 216: can we add this change to the ui-registry version of the component? otherwise anytime typography gets updated it will get overwritten here as the registry is the single source of truth

jvorcak:
Should be fixed by https://github.com/redpanda-data/ui-registry/pull/235

eblairmckee (approved):

jvorcak:
I think after this PR in UI-registry is merged, we can update the component and revert this change.

jvorcak:
We don't need this PR anymore, the change is done in a UI registry and it should be fixed in https://github.com/redpanda-data/console/pull/2562

PR #2334, opened 2026-03-23
Remove topics from backend-api

jvorcak:


frontend/src/components/pages/topics/topic-configuration.tsx
```diff
@@ -230,8 +234,7 @@ const ConfigurationEditor: FC<ConfigurationEditorProps> = (props) => {
     setEditedEntry(configEntry);
   };
 
-  const topic = props.targetTopic;
-  const hasEditPermissions = topic ? (api.topicPermissions.get(topic)?.canEditTopicConfig ?? true) : true;
+  const hasEditPermissions = true;
```
weeco: Why is this hardcoded now?

jvorcak: Fixed — `hasEditPermissions` is now derived from `allowedActions` prop (passed down from the `Topic` object in `tab-config.tsx`). `undefined` means all actions allowed, matching the existing convention.

jvorcak: @weeco if I'd like to use this with enum, we'd need to migrate to grpc. Is there any grpc supported feature where I could get `allowedActions`?

frontend/src/components/pages/admin/admin-debug-bundle.tsx
```diff
@@ -216,9 +217,7 @@ const NewDebugBundleForm: FC<{
 
   useEffect(() => {
     api.refreshBrokers(true);
-    api.refreshPartitions('all', true).catch(() => {
-      // Error handling managed by API layer
-    });
+    queryClient.invalidateQueries({ queryKey: ['topicPartitionsAll'] });
```
malinskibeniamin, line 220: You can rely on a `useQueryClient` hook to grab the singleton instance too.

frontend/src/components/pages/connect/connector-details.tsx
```diff
@@ -552,13 +555,11 @@ class KafkaConnectorDetails extends PageComponent<{ clusterName: string; connect
     appGlobal.onRefresh = () => this.refreshData(true).catch(console.error);
   }
 
-  async refreshData(force: boolean): Promise<void> {
+  async refreshData(_force: boolean): Promise<void> {
```
malinskibeniamin, line 542: If we don't use the `_force`, should we drop it

frontend/src/components/pages/topics/topic-details.tsx
```diff
@@ -190,28 +184,36 @@ class TopicDetails extends PageComponent<{ topicName: string }> {
...
-  const config = api.topicConfig.get(topicName);
+  // Derived: topicConfig via React Query
+  const { data: topicConfigData } = useTopicConfigQuery(topicName);
   const topicConfig: ConfigEntry[] | null | undefined =
-    config === undefined ? undefined : config === null || config.error !== null ? null : config.configEntries;
+    topicConfigData === undefined
+      ? undefined
+      : topicConfigData === null || topicConfigData.error !== null
+        ? null
+        : topicConfigData.configEntries;
```
malinskibeniamin: Can we avoid nested ternaries? I know this was there before but maybe an easy opportunity to clean up

frontend/src/components/pages/reassign-partitions/reassign-partitions.tsx
```diff
@@ -801,14 +820,21 @@ class ReassignPartitions extends PageComponent {
...
+    const cachedTopicsForMoves = queryClient.getQueryData<import('../../../state/rest-interfaces').GetTopicsResponse>([
+      'topics',
+    ])?.topics;
+    if (!cachedTopicsForMoves) {
       return [];
     }
+    const cachedTpForMoves =
+      queryClient.getQueryData<Map<string, import('../../../state/rest-interfaces').Partition[] | null>>([
+        'topicPartitionsAll',
+      ]) ?? new Map<string, import('../../../state/rest-interfaces').Partition[] | null>();
```
malinskibeniamin, line 832: Why do we need this type of dynamic import? I would just import it, I guess react doctor might complain about it being a barrel file of some sort? I think in this case it would make it worse to have this import everywhere, just hoist it to the top of the file.

malinskibeniamin (approved):

PR #2362, opened 2026-04-02
feat: add Netlify deployment config for Console frontend

malinskibeniamin:
## Summary

Add Netlify deployment configuration for Console frontend, enabling it to be deployed as a static site alongside the existing Go HTTP server deployment.

## What's included

- **`frontend/netlify.toml`** — Build config, CORS headers for cross-origin MF loading, cache control
- **`frontend/public/_redirects`** — SPA fallback for standalone mode

## How it works

Console frontend assets can now be deployed to a Netlify site (e.g., `redpanda-console-ui`). Cloud UI loads the assets from Netlify when `REACT_APP_CONSOLE_FRONTEND_URL` is set, with automatic fallback to the cluster's Go HTTP server when Netlify is unreachable.

### Deployment modes

| Mode | Source | When |
|------|--------|------|
| Netlify (new) | `REACT_APP_CONSOLE_FRONTEND_URL/mf-manifest.json` | Env var set + MF v2 enabled |
| Cluster pod (existing) | `cluster.redpandaConsole.url/embedded.js` | Fallback or no env var |
| Local dev | `REACT_APP_CONSOLE_FRONTEND_URL_OVERRIDE/embedded.js` | Dev override |

### Fallback chain (implemented in Cloud UI)

1. Try Netlify URL (MF v2 manifest)
2. If unreachable → try cluster URL (MF v2 embedded.js)
3. If MF v2 unavailable → load legacy Injector (MF v1)
4. Private-only clusters skip Netlify entirely

## Prerequisites

- Console MF v2 search params fix: #2361
- Cloud UI support: redpanda-data/cloudv2#25386
- `enable-console-mf-v2` feature flag enabled in LaunchDarkly

## Test plan

- [x] `bun run build` produces correct output in `build/` (embedded.js, mf-manifest.json, static assets)
- [x] Netlify deploy preview builds successfully
- [x] CORS headers present on `embedded.js` and `mf-manifest.json`
- [x] `_redirects` SPA fallback works for standalone navigation

### DIFF
--- a/frontend/netlify.toml
+++ b/frontend/netlify.toml
@@ -1,0 +1,32 @@
+[build]
+  base = "frontend"
+  command = "bun install && NODE_OPTIONS=--max_old_space_size=16384 bun run build"
+  publish = "build"
+
+[build.environment]
+  NODE_VERSION = "22"
+
+[[headers]]
+  for = "/*"
+  [headers.values]
+    Access-Control-Allow-Origin = "*"
+    Access-Control-Allow-Methods = "GET, OPTIONS"
+    Access-Control-Allow-Headers = "*"
+    X-Content-Type-Options = "nosniff"
+
+# MF remote entry and manifest must never be cached
+[[headers]]
+  for = "/embedded.js"
+  [headers.values]
+    Cache-Control = "no-cache, no-store, must-revalidate"
+
+[[headers]]
+  for = "/mf-manifest.json"
+  [headers.values]
+    Cache-Control = "no-cache, no-store, must-revalidate"
+
+# Hashed static assets are immutable
+[[headers]]
+  for = "/static/*"
+  [headers.values]
+    Cache-Control = "public, max-age=31536000, immutable"
--- a/frontend/netlify.toml
+++ b/frontend/netlify.toml
@@ -1,5 +1,7 @@
 [build]
   base = "frontend"
+  # NODE_OPTIONS: 16 GB is a V8 heap ceiling guard, not actual memory usage.
+  # It prevents OOM kills during large builds; real RSS stays well below this.
   command = "bun install && NODE_OPTIONS=--max_old_space_size=16384 bun run build"
   publish = "build"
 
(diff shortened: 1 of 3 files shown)


malinskibeniamin:
@claude review

frontend/netlify.toml
```diff
@@ -0,0 +1,45 @@
...
+    Access-Control-Allow-Methods = "GET, OPTIONS"
+    Access-Control-Allow-Headers = "*"
+    Access-Control-Max-Age = "86400"
+
+# Hashed static assets are immutable, CORS required for cross-origin import
+[[headers]]
+  for = "/static/*"
+  [headers.values]
+    Cache-Control = "public, max-age=31536000, immutable"
+    Access-Control-Allow-Origin = "*"
```
sago2k8, line 42: Is this how we manage it normally, could we enable CORS only for the specific domains ?

frontend/netlify.toml
```diff
@@ -0,0 +1,45 @@
+[build]
+  base = "frontend"
+  # NODE_OPTIONS: 16 GB is a V8 heap ceiling guard, not actual memory usage.
+  # It prevents OOM kills during large builds; real RSS stays well below this.
+  command = "bun install && NODE_OPTIONS=--max_old_space_size=16384 bun run build"
```
sago2k8: Are we charged by memory consumption in builds? having a high node old space size could be hiding some bugs.

frontend/netlify.toml
malinskibeniamin: Good question. For Module Federation, the host app (Cloud UI on `cloud.redpanda.com`) loads `remoteEntry.js` and chunk files from this Netlify site via `<script>` tags and `fetch()`. Using `Access-Control-Allow-Origin: *` is standard for public static assets (same pattern as any CDN-served JS library).

We could restrict to specific domains (e.g., `*.redpanda.com`, `*.netlify.app`), but:
- These are public, non-sensitive static JS/CSS files with no auth
- Host app Netlify site uses the same `Access-Control-Allow-Origin: *` pattern for external scripts

The wildcard is scoped to only the MF entry points (`/embedded.js`, `/mf-manifest.json`, `/static/*`), not all responses. The catch-all `/*` block only has `X-Content-Type-Options: nosniff` without CORS. So the exposure is limited to files that are designed to be publicly fetchable.

frontend/netlify.toml
malinskibeniamin: Good catch. The `--max_old_space_size=16384` is a V8 heap ceiling, not actual memory allocation — it just raises the limit before V8 triggers an OOM. The actual RSS during a Console build is typically 2-4 GB.

Netlify doesn't charge by memory consumption — builds run on shared VMs with a fixed time limit. The risk of a high ceiling is exactly what you said: it could mask a memory leak in the bundler that would otherwise surface as an OOM.

That said, other app's Netlify build uses `NODE_OPTIONS=--max_old_space_size=4096` (see `netlify.toml`). We could lower this to 4096 or 8192 to be more conservative while still avoiding legitimate OOM during large builds.

malinskibeniamin: Lowered to 4096 in 286d8043e to match Cloud UI's pattern.

PR #2364, opened 2026-04-03
serverless: allow password edit

nicolaferraro:
Serverless now supports changing passwords.

I've also removed the gate on admin API presence since now password change is handled via Kafka protocol, but let me know if my assumption is wrong.


jan-g (approved):
Looks good to me.

malinskibeniamin (approved):

PR #2372, opened 2026-04-08
fix(frontend): resolve bun compatibility issues for local dev

malinskibeniamin:
## Summary
- **Flatten nested `overrides`**: Bun doesn't support nested objects in `overrides` (e.g., `"minimatch": { "brace-expansion": "^5.0.5" }`). Moved scoped overrides to `resolutions` using `**/pkg/dep` glob syntax (which bun supports natively).
- **Disable MF DTS generation**: Module Federation's type declaration generation fails on pre-existing TS4023/TS7006 type errors, blocking `rsbuild dev`. Set `dts: false` to unblock local development.
- **Add `"type": "module"`**: Eliminates Node.js `MODULE_TYPELESS_PACKAGE_JSON` warning when starting the dev server.

## Test plan
- [x] Run `bun i` — should produce zero warnings
- [x] Run `bun run start2 --port=3004` — should build successfully with no errors
- [x] Verify dev server serves the app at `http://localhost:3004`
- [x] Run `bun run build` — should complete without errors

### DIFF
--- a/frontend/rsbuild.config.ts
+++ b/frontend/rsbuild.config.ts
@@ -52,6 +52,7 @@ export default defineConfig({
     pluginYaml(),
     pluginModuleFederation({
       ...moduleFederationConfig,
+      dts: false,
     }),
     pluginNodePolyfill({
       globals: { process: true },
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -53,16 +53,7 @@
     "lodash": "^4.17.23",
     "lodash-es": "^4.17.23",
     "qs": "^6.14.2",
-    "minimatch": {
-      "brace-expansion": "^5.0.5"
-    },
-    "router": {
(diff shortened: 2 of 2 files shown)


frontend/rsbuild.config.ts
```diff
@@ -52,6 +52,7 @@ export default defineConfig({
     pluginYaml(),
     pluginModuleFederation({
       ...moduleFederationConfig,
+      dts: false,
```
r-vasquez, line 55: Doesn't this https://module-federation.io/configure/dts prevent generating the types and break consumers?

malinskibeniamin: it does not break consumers, we don't rely on dts anywhere - it's a good question but in this case we actually don't rely on it so it generates more noise

c-julin (approved):

PR #2373, opened 2026-04-09
New Pipeline UX Bug Bash fixes part 1

eblairmckee:
# Bug bash fixes: Log Explorer live mode, improved pipeline dialog UX, connector tiles are context-aware, pipeline command menu design improvements

See [bug bash doc](https://docs.google.com/document/d/19_NL_CmfksntoDXU8kd5ys7Um_5KDg6x7f6IHBiuMKc/edit?tab=t.0)

### Log Explorer
**Log Explorer: no progress feedback during search**
Users saw a bare spinner with no indication of progress when loading logs. Added a progress bar overlaying the table top edge with scanned bytes and message count stats. Spinner shows briefly before the first progress event arrives, then switches to stats.

**Log Explorer: confusing states in live mode**
Users enabling live mode on a stopped pipeline saw an empty table with no explanation. Added contextual states: "Pipeline is not running — Switch to Recent Logs" when the pipeline is stopped, "Listening for new log messages..." when live mode is active but no messages have arrived, and specific empty-state text for history mode ("No logs found in the last 5 hours"). Live mode now auto-enables when the pipeline starts running and auto-disables when it stops.

**Log Explorer: live toggle was hard to discover**
Users didn't notice the live toggle. Added a descriptive label ("Enable live logs" / "Live logs enabled") with an info-icon tooltip explaining the behavior, and increased the switch size for visibility. 

### Pipeline Creation

**Pipeline user dialog: smarter topic and ACL awareness**
New `extractConnectorTopics()` reads topics configured on a connector from YAML and passes them to `AddUserStep`, so ACLs are configured for the right topic automatically. Multi-topic connectors show a warning alert prompting manual ACL setup. Renamed misleading `superuser` form field to `grantTopicPermissions`. 

**Pipeline dialogs couldn't be dismissed during loading**
Users were trapped waiting for slow form submissions with no way to cancel. New `useRefFormDialog` hook manages the full dialog lifecycle — closing a dialog during submission cancels the pending operation via `AbortController` (signal passed to `triggerSubmit` so implementors can wire it to RPC calls) with a configurable timeout (default 30s). Submit buttons show text + spinner icon instead of spinner replacing text.

**Pipeline sidebar: category shortcuts for command menu**
Made types of variables you can insert via UI more apparent (secrets, contextual variables, users, topics, etc). Users had to open the command menu and then filter to find variables, secrets, topics, or users. Replaced the single "Insert" button with buttons for each category, each opening the command menu pre-filtered via a new `initialFilter` prop.

### Log Explorer
**Log Explorer: live streaming performance**
`useLogLive` dispatched a React state update per incoming message, causing O(n^2) re-renders as message count grew. Replaced with batched `pendingRef` + 200ms flush interval matching history mode's pattern.

### Pipeline Creation
**Topic creation errors silently swallowed**
`useCreateTopicMutation` `onError` formatted the error message but never passed it to `toast.error()` — users saw no feedback when topic creation failed. Now correctly shows the error toast.

## UI Improvements

**Dialog close buttons: accessible hit target**
Pipeline dialog close buttons had a 14px icon — too small for reliable click/tap. New `DialogCloseButton` component with CVA size variants (`small` | `medium`) and minimum 24px icon, applied across all pipeline dialogs (add topic, add user, config, details, add connector, add secrets, command menu sub-dialogs).

**Connector tile logos overflow fixed**
Certain SVG logos (DynamoDB, GNU) had hardcoded `width`/`height` attributes that overrode CSS sizing, causing overflow. Removed hardcoded attributes, replaced inline styles with `className="h-6 w-6 object-contain"` in a fixed-size flex container. Long connector names now word-wrap with `break-words` instead of being clipped.

**Log explorer table respects parent container bounds**
The log table's message column could push the table wider than its parent container, creating a horizontal scrollbar. Added `table-fixed` layout and `truncate` on the message column so long messages ellipsis instead of overflowing.

**Log explorer sort indicators**
Column sort arrows replaced unicode glyphs with lucide `ArrowUp`/`ArrowDown` icons for visual consistency.

**Connector dialog contextual titles**
`AddConnectorDialog` now accepts `title` and `searchPlaceholder` props. Pipeline page passes contextual values ("Add an input" / "Search inputs...", "Add a resource" / "Search resources...") instead of generic "Add a connector".

## Other

**`stripCommentedKeys`**: After patching a YAML key (e.g., `topics`), removes the commented-out placeholder line (`# topics: Required - ...`) scoped to the specific component block (e.g., only within `kafka_franz:`, not sibling components in the same section).

**`QuickAddSecrets` card variant prop**: Secrets dialog now passes `cardVariant="outlined"` for visual consistency within dialogs.


frontend/src/components/pages/topics/Tab.Messages/index.tsx
```diff
@@ -598,6 +599,21 @@ export const TopicMessageView: FC<TopicMessageViewProps> = (props) => {
...
+    if (searchPhase === null) {
+      return;
+    }
+    const interval = setInterval(() => {
+      const search = currentMessageSearchRef.current;
+      if (search) {
+        setBytesConsumed(search.bytesConsumed);
+        setTotalMessagesConsumed(search.totalMessagesConsumed);
+      }
+    }, 200);
```
malinskibeniamin: I think there are settings where user can define how often certain refetch events/polling happens, maybe we can use that value here.

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -426,19 +434,69 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
...
                     </TableCell>
                   </TableRow>
                 );
               }
               if (filteredRowCount === 0) {
+                let emptyText: string;
+                if (messages.length > 0) {
+                  emptyText = 'No messages match the current filters';
+                } else if (liveViewEnabled) {
+                  emptyText = 'Listening for new log messages\u2026 Switch to Recent Logs to view historical logs.';
```
malinskibeniamin: Maybe replace `\u2026` with equivalent character

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -426,19 +434,69 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
...
+              }
+              if (liveViewEnabled && pipelineNotRunning && messages.length === 0) {
+                return (
+                  <TableRow>
+                    <TableCell className="p-4" colSpan={table.getVisibleFlatColumns().length}>
+                      <Alert data-testid="pipeline-stopped-banner" variant="info">
+                        <AlertTitle>Pipeline is not running</AlertTitle>
+                        <AlertDescription>
+                          Live logs require a running pipeline.{' '}
+                          <button
```
malinskibeniamin, line 475: Why can't we use `<Button>` from UI registry? Is it because of the table?

malinskibeniamin: Nevermind, this is still Chakra

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -346,61 +348,67 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
...
-          onClick={refresh}
-          size="icon"
-          variant="ghost"
-        >
-          <RefreshCcw className={isSearching ? 'animate-spin' : ''} />
-        </Button>
+        <div className="flex items-center gap-2">
+          <Switch
+            checked={liveViewEnabled}
+            className="h-5 w-9 **:data-[slot=switch-thumb]:size-4.5"
```
malinskibeniamin, line 383: Is `**:data-[slot=switch-thumb]:size-4.5` required?

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -346,61 +348,67 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
 
   const isSearching = phase !== null;
   const filteredRowCount = table.getFilteredRowModel().rows.length;
+  const hasProgress = progress.bytesConsumed > 0 || progress.messagesConsumed > 0;
+  const pipelineNotRunning = pipeline.state === Pipeline_State.STOPPED || pipeline.state === Pipeline_State.ERROR || pipeline.state === Pipeline_State.COMPLETED;
```
malinskibeniamin: Why not just `pipeline.state !== Pipeline_State.RUNNING`

frontend/src/components/pages/rp-connect/utils/yaml.ts
```diff
@@ -715,6 +757,46 @@ export function tryPatchRedpandaYaml(
...
+): string[] | undefined {
+  if (!yamlContent.trim()) {
+    return;
+  }
+
+  let doc: Document.Parsed;
+  try {
+    doc = parseDocument(yamlContent);
+  } catch {
+    return;
```
malinskibeniamin: Is there any action or message we could display when parsing YAML fails?

frontend/src/components/ui/connect/log-explorer.tsx
eblairmckee: was just trying to be more explicit, and makes it easier to understand if we add/remove other states later.

frontend/src/components/pages/topics/Tab.Messages/index.tsx
eblairmckee: This is a UI progress counter update (bytes consumed / messages consumed), not a data refetch. 200ms keeps the progress display feeling responsive during an active search. The autoRefreshInterval setting is for data polling (default 10s) which would make the counter feel frozen.

frontend/src/components/ui/connect/log-explorer.tsx
eblairmckee: Yes, left a comment inline. User gave feedback that it wasn't discoverable so made it slightly larger, which makes it similar size to refresh button it is next to
Screenshot 2026-04-09 at 1 56 34 PM

frontend/src/components/pages/rp-connect/utils/yaml.ts
eblairmckee: yes, added a toast

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -346,62 +345,71 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
...
-
       {/* Table */}
-      <div className="min-h-0 overflow-auto">
-        <Table>
+      <div className="relative min-h-0">
+        {/* Progress bar overlaying top of table */}
+        {isSearching && hasProgress && (
+          <div className="absolute inset-x-0 top-0 z-10">
+            <Progress className="h-1 w-full rounded-none" testId="log-progress-bar" value={null} />
+          </div>
```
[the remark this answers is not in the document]
eblairmckee: this is a ui-registry component, we should explore a fix in the registry itself

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -205,23 +207,21 @@ interface LogExplorerProps {
...
-  // Sync live view when pipeline state changes (e.g. transitions to RUNNING)
+  // Sync live mode bidirectionally when enableLiveView changes
   const [prevEnableLiveView, setPrevEnableLiveView] = useState(enableLiveView);
   if (enableLiveView !== prevEnableLiveView) {
     setPrevEnableLiveView(enableLiveView);
-    if (enableLiveView) {
-      setLiveViewEnabled(true);
-    }
+    setLiveViewEnabled(enableLiveView);
   }
```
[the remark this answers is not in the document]
eblairmckee: Render-time setState — Replaced with useEffect + mountedRef to skip mount (preserves "start in history mode" behavior)

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -346,62 +345,71 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
...
+            disabled={!enableLiveView}
+            id="live-view-toggle"
+            onCheckedChange={(checked) => {
+              setLiveViewEnabled(checked);
+              setSorting([]);
+              if (checked) {
+                actions.removeAllFilters();
+              }
+            }}
+          />
```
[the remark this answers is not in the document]
eblairmckee: TooltipTrigger nesting — Label moved outside tooltip, trigger wraps only the info icon via asChild on a <span>

frontend/src/components/pages/rp-connect/utils/yaml.ts
```diff
@@ -651,6 +651,42 @@ export function buildSaslPatch(result: RedpandaSetupResultLike): RedpandaPatch['
...
+      sectionEnd = i;
+      break;
+    }
+  }
+
+  return lines
+    .filter((line, i) => {
+      if (i < sectionStart || i >= sectionEnd) return true;
+      return !commentRegex.test(line);
+    })
```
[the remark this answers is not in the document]
eblairmckee: stripCommentedKeys scoping — Now takes componentName, narrows to the specific component block (finds   componentName: within the section, stops at the next sibling at the same indent)

frontend/src/hooks/use-ref-form-dialog.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+import type { RefObject } from 'react';
+import { useCallback, useRef, useState } from 'react';
+import { toast } from 'sonner';
+
+/** Minimal contract a ref-based form must satisfy. */
+type RefFormSubmittable<TData> = {
+  triggerSubmit: () => Promise<{ success: boolean; data?: TData }>;
+};
```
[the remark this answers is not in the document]
eblairmckee: useRefFormDialog abort signal — triggerSubmit now accepts optional AbortSignal, passed from the hook's AbortController

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -657,17 +710,16 @@ export default function PipelinePage() {
...
-  }, [slashCommand]);
-
-  useHotKey({ key: 'p', modifiers: ['meta', 'shift'], enabled: mode !== 'view', onTrigger: handleCommandMenuOpen });
+  const handleCommandMenuOpen = useCallback(
+    (filter: 'all' | 'variables' | 'secrets' | 'topics' | 'users' = 'all') => {
+      slashCommand.close();
+      setCommandMenuFilter(filter);
+    },
+    [slashCommand]
+  );
```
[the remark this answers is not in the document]
eblairmckee: Stale Cmd+Shift+P comment — Updated to generic "Dialog variant"

eblairmckee:
@claude review

eblairmckee:
@claude I removed the changes to tab.messages, what do you mean by: Generation counter for stale search results in Tab.Messages: Clean fix for the race condition.

eblairmckee:
addressed claude comments in [f698674](https://github.com/redpanda-data/console/pull/2373/commits/f698674275bfb9a08b478b9f5c7640a0cb946c1e)

> 1. Progress renders as empty bar instead of indeterminate (log-explorer.tsx:413)

won't do. see earlier comment

> 2. Abort signal not propagated to implementations (use-ref-form-dialog.ts:82, add-user-step.tsx:362)

fixed

> 3. pipelineNotRunning uses allow-list instead of deny-list (log-explorer.tsx:352)

pipelineNotRunning — Changed to !== Pipeline_State.RUNNING for forward-compatibility with new states

> 4. Two PipelineCommandMenu instances always mounted (pipeline/index.tsx:910-936)

Verified both are inert when closed (effects bail out, query has enabled: open). No change needed -- architecturally necessary for different rendering modes (modal dialog vs positioned popover).

frontend/src/components/pages/topics/Tab.Messages/index.tsx
eblairmckee: descoped from this PR since I want to keep scope solely on pipeline UX

frontend/src/components/ui/connect/log-explorer.tsx
eblairmckee: you're right, fixed

frontend/src/react-query/api/logs.tsx
```diff
@@ -214,35 +237,43 @@ type LiveState = {
   messages: TopicMessage[];
   phase: string | null;
   error: string | null;
+  progress: SearchProgress;
 };
 
 type LiveAction =
   | { type: 'reset' }
   | { type: 'start' }
-  | { type: 'addMessage'; msg: TopicMessage }
```
malinskibeniamin, line 222: Not sure if it's intentional that we no longer have `addMessage` type handling

frontend/src/hooks/use-ref-form-dialog.ts
malinskibeniamin, line 1: Where is this used? It's quite complex, might be a good follow-up for simplification.

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -413,8 +426,8 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
                     style={{ minWidth: header.column.columnDef.minSize, width: header.getSize() !== TANSTACK_DEFAULT_COLUMN_SIZE ? header.getSize() : undefined }}
                   >
                     {header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
-                    {header.column.getIsSorted() === 'asc' && ' \u2191'}
-                    {header.column.getIsSorted() === 'desc' && ' \u2193'}
+                    {header.column.getIsSorted() === 'asc' && <ArrowUp className="ml-1 inline size-3.5" />}
+                    {header.column.getIsSorted() === 'desc' && <ArrowDown className="ml-1 inline size-3.5" />}
```
malinskibeniamin, line 430: 👍

frontend/src/components/pages/rp-connect/pipeline/pipeline-command-menu.test.tsx
```diff
@@ -121,8 +121,8 @@ describe('PipelineCommandMenu', () => {
     });
 
     await waitFor(() => {
-      expect(screen.getByText('${secrets.MY_API_KEY}')).toBeInTheDocument();
-      expect(screen.getByText('${secrets.DB_PASSWORD}')).toBeInTheDocument();
+      expect(screen.getByText('secrets.MY_API_KEY')).toBeInTheDocument();
+      expect(screen.getByText('secrets.DB_PASSWORD')).toBeInTheDocument();
```
malinskibeniamin, line 125: 👍

frontend/src/hooks/use-ref-form-dialog.ts
eblairmckee: the "add topic" dialog and "add user" dialog (both in the new pipeline UX)

malinskibeniamin (approved):

PR #2375, opened 2026-04-11
fix(frontend): switch React Compiler to opt-in annotation mode

malinskibeniamin:
## Summary

- Switch React Compiler from opt-out (`"infer"`) to opt-in (`compilationMode: 'annotation'`) in `rsbuild.config.ts`
- Remove all 71 `'use no memo'` directives across 58 files (now dead code — compiler only processes files with `'use memo'`)
- Update regression test comment in `data-table-view-options.test.tsx` to reflect the new approach

## Problem

The React Compiler was running in default "infer" (opt-out) mode, automatically memoizing every component and hook in `src/`. This caused:

- **Pagination not re-rendering** when page state changes (stale closures from over-memoization)
- **Data table filters showing stale values** — `data-table-filter.tsx` and related components broke under compiler memoization
- **58 files** already needed `'use no memo'` workarounds to function correctly
- Conflicts with the `uiSettings` Proxy pattern, URL-based state (`nuqs`), and TanStack Table's internal state management

## Solution

Switch to `compilationMode: 'annotation'` — the compiler now **only processes files explicitly marked with `'use memo'`**. No files are opted in yet. This is a stabilization change that eliminates all compiler-related memoization bugs while keeping the infrastructure in place for incremental opt-in.

### What changes:
| | Before | After |
|---|---|---|
| **Mode** | Opt-out (compile everything) | Opt-in (compile nothing unless marked) |
| **`'use no memo'` directives** | 71 across 58 files | 0 (removed — now dead code) |
| **Pagination/filter bugs** | Present, worked around | Eliminated at source |

### Future work:
1. Incrementally add `'use memo'` to safe presentational components after testing
2. Address `uiSettings` Proxy pattern before opting in components that use it
3. Consider `panicThreshold: 'critical_errors'` once confidence grows

## Test plan

- [x] `bun run type:check` — passes
- [x] `bun run lint` — passes (pre-existing warnings only)
- [x] `bun run test` — 80 files, 856 tests passed, 0 failures
- [x] No `'use no memo'` directives remain (`grep -r "use no memo" src/` returns only the updated test comment)
- [ ] Manual verification: pagination, filters, search, sorting all work in browser
- [ ] CI green

### DIFF
--- a/frontend/rsbuild.config.ts
+++ b/frontend/rsbuild.config.ts
@@ -31,6 +31,7 @@ export default defineConfig({
           'babel-plugin-react-compiler',
           {
             target: '18',
+            compilationMode: 'annotation',
             sources: (filename: string) => {
               if (filename.includes('/lib/redpanda-ui/')) {
                 return false;
--- a/frontend/rsbuild.config.ts
+++ b/frontend/rsbuild.config.ts
@@ -32,6 +32,9 @@ export default defineConfig({
           {
             target: '18',
             compilationMode: 'annotation',
+            panicThreshold: 'critical_errors',
+            // In annotation mode, this still gates which files CAN be opted in.
+            // Files excluded here are ineligible even with 'use memo'.
             sources: (filename: string) => {
               if (filename.includes('/lib/redpanda-ui/')) {
                 return false;
--- a/frontend/src/components/layout/header.tsx
+++ b/frontend/src/components/layout/header.tsx
@@ -9,8 +9,6 @@
  * by the Apache License, Version 2.0
  */
 
-'use no memo';
-
 import { Box, Button, ColorModeSwitch, CopyButton, Flex } from '@redpanda-data/ui';
 import { Link, useLocation, useMatchRoute } from '@tanstack/react-router';
 import { Heading } from 'components/redpanda-ui/components/typography';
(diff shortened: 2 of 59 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Thanks for the review! Addressed all feedback in 621df6fd9:

1. **Sources comment** — Added a comment above the `sources` callback clarifying that in annotation mode it still gates which files can be opted in, so files in `lib/redpanda-ui/`, `gen/`, or `node_modules` are ineligible even with `'use memo'`.

2. **`panicThreshold: 'critical_errors'`** — Added now alongside `compilationMode: 'annotation'`. This ensures that when files are incrementally opted in with `'use memo'`, the build fails loudly if the compiler encounters issues rather than silently degrading.

3. **`embedded-app.tsx:79`** — Verified this is a code comment about `React.memo()` (the React API), not the `'use memo'` compiler directive. Unrelated, no action needed.

4. **Performance note** — Acknowledged. Components that were benefiting from compiler memoization (those without `'use no memo'`) will lose that optimization. This is an acceptable trade-off given the scale of bugs, and we'll track it during incremental opt-in.

5. **`target: '18'`** — Noted for React 19 migration.

malinskibeniamin:
@claude review

sago2k8 (approved):

PR #2379, opened 2026-04-14
Fixes e2e tests non-deterministic startup

jvorcak:
There was a problem sometimes running e2e tests locally, especially when you tried to develop them. This fixes a corner case that used to happen.


malinskibeniamin (approved):
LGTM, could also make it a claude hook btw

PR #2381, opened 2026-04-15
fix(federation): prevent router recreation loop in MF mode

c-julin:
## Summary
- Remove `initialPath` from the router `useMemo` dependency array in `console-app.tsx`
- When cloud-ui passes a new `initialPath` on navigation, the entire TanStack Router was being recreated — remounting all route components, retriggering all data fetches (including `GetKafkaConnectInfo` which returns 501 when Connect isn't configured), and causing a visible refresh loop with "Fetching data..." spinner
- `initialPath` is now captured in a `useRef` on first render. Subsequent navigation from the host is already handled by the `navigateTo` prop via `router.navigate()`

## Root cause
The `useMemo` at line 238 had `[initialPath, queryClient, dataplaneTransport]` as deps. Cloud-ui derives `initialPath` from `routerPathname + routerSearch`, which changes on every navigation. Each change recreated the memory history router, which remounted all route components and their `initPage()` calls.

## Test plan
- [ ] Load Console via Module Federation in cloud-ui (localhost:3000)
- [ ] Navigate to Topics → click a topic → verify no "Fetching data..." refresh loop
- [ ] Navigate to Schema Registry → verify page loads without stuck skeletons
- [ ] Verify browser back/forward still works (navigateTo prop handles this)
- [ ] Verify Console standalone (localhost:8085) still works normally

### DIFF
--- a/frontend/src/components/misc/router-sync.tsx
+++ b/frontend/src/components/misc/router-sync.tsx
@@ -12,7 +12,7 @@
 import { useLocation, useNavigate, useRouter } from '@tanstack/react-router';
 import { useEffect, useRef } from 'react';
 
-import { isEmbedded } from '../../config';
+import { config as appConfig, isEmbedded } from '../../config';
 import { trackHubspotPage } from '../../hubspot/hubspot.helper';
 import { appGlobal } from '../../state/app-global';
 import { api } from '../../state/backend-api';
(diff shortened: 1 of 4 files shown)


malinskibeniamin:
@claude review

malinskibeniamin (approved):

frontend/src/components/misc/router-sync.tsx
```diff
@@ -66,3 +69,13 @@ export const RouterSync = () => {
...
+
+/**
+ * Detect MFv2 federated mode. In federated mode, console-app.tsx sets up its own
+ * navigation sync via onRouteChange — the legacy window event dispatch is not needed.
+ */
+function isFederatedMode(): boolean {
+  // config.clusterId is only set in federated mode (via setup() in console-app.tsx).
+  // In standalone mode, it's undefined.
+  return appConfig.clusterId !== undefined && appConfig.clusterId !== '';
+}
```
malinskibeniamin, line 81: I think `isEmbedded` would always guarantee cluster ID is defined since otherwise the app config would not load with the console served assets, but I am ok with this

malinskibeniamin (approved):

PR #2390, opened 2026-04-19
chore(frontend): upgrade MCP SDK to 1.29.0 + adopt onerror, streaming, OAuth provider

malinskibeniamin:
## What

We upgraded the MCP TypeScript SDK (1.26 → 1.29) and adopted three new capabilities across the Console MCP servers feature and the AI agents path: transport-level error observability, task-based streaming tool calls, and a first-class `OAuthClientProvider` wrapping the Console JWT. The first consumer is the MCP server inspector tab, which now surfaces tool-call progress live.

## Why

- Richer tool call UX. Long-running tool calls now stream incremental progress with a status line rather than appearing as a single opaque spinner — the user sees the tool working.
- Centralized auth. `ConsoleJWTOAuthProvider` lets the SDK's internal request machinery (retries, auth recovery) drive off the same JWT source as Console's transport, instead of having to be taught each callsite separately.
- Observable transport errors. SSE drops and HTTP failures that happen outside the request/response cycle are now logged with the server URL instead of silently disappearing.

## User-visible changes

- The MCP server inspector tab shows a Progress bar and a live status line while a tool call is executing.
- Cancelling a running tool call aborts the underlying request (no orphan fetches against the MCP gateway).
- Tool calls against older MCP servers that do not yet advertise task-streaming continue to work unchanged — the client transparently falls back to the classic non-streaming path.

### Inspector — tool list

Pick a tool from the MCP server to configure and run it.

Inspector tool list

### Inspector — ready to run

Selected tool renders its input schema via a form with `Run Tool` and `Copy Input` affordances.

Inspector ready to run

### Inspector — streaming with live progress

Task-streaming tool call: the `Progress` bar advances and the status line mirrors server-side task status messages.

Inspector streaming with progress

## Hardening

- **Capability fallback.** Post-connect, we inspect `client.getServerCapabilities().tasks?.requests?.tools?.call`. Absent → we call `client.callTool` directly instead of `experimental.tasks.callToolStream`, so older servers cannot produce a silently-hung mutation.
- **Streaming timeout.** The stream is wrapped in a 120s composed-signal timeout (configurable per-call via `streamTimeoutMs`). On timeout, the signal is aborted and the mutation rejects with `MCP tool stream timed out after <N>ms`.
- **Watchdog.** If the stream closes without a terminal `result` or `error` message, the mutation rejects explicitly with a descriptive error rather than resolving undefined or hanging.
- **JWT expiry detection.** `ConsoleJWTOAuthProvider.tokens()` base64url-decodes the payload and reads `exp`. Already-expired tokens return `undefined` so the SDK surfaces `UnauthorizedError` early; malformed or exp-less tokens fail open (advisory, not crypto).
- **Transport error logging.** `transport.onerror` is wired before `client.connect()`, logs to the browser console with the server URL, and deliberately does not toast — the mutation's `onError` owns toast to avoid double-firing.
- **Progress clamping.** The inspector clamps the computed percent to `[0, 100]` and handles NaN / `total=0` by going indeterminate, so a misbehaving server cannot push a broken-looking bar.
- **Tool-switch reset.** Switching the selected tool mid-stream now clears in-flight progress state so stale percentages do not bleed into the new selection.

## Risks & mitigations

| Original risk | Shipped mitigation |
| --- | --- |
| Streaming against an older server could hang the mutation | Capability check + non-streaming fallback + 120s timeout + watchdog |
| OAuth provider has no expiry awareness | Best-effort `exp` decode in `tokens()`, fail-open on any parse failure |
| `transport.onerror` could double-fire with mutation `onError` | Transport callback logs only; contract pinned by a regression test |
| Manual `Authorization` header + `authProvider` could disagree | Both read live `config.jwt` per-request from the same source, so they agree by construction |

## Test plan

- [ ] `bun run type:check`
- [ ] `bun run lint:check` (no new violations over the 984-error baseline)
- [ ] `bun run test` (unit + integration)
- [ ] `bun run test:browser` (streaming inspector visual regression)
- [ ] Manual smoke: create an MCP server, inspect a tool, run it, observe the Progress bar + status line move
- [ ] Manual smoke: rotate the console JWT mid-session, confirm the next MCP call picks up the new token

## Follow-ups

- Elicitation UI — SDK 1.29 delivers elicitation through server-initiated requests handled via registered handlers, requiring a new dialog component.
- OAuth refresh flow with backend coordination, once the expiry detection lands and we decide on refresh semantics.


malinskibeniamin:
@claude review

malinskibeniamin:
@claude review

malinskibeniamin:
@claude review

malinskibeniamin:
@claude review

malinskibeniamin:
Fix the screenshot in the PR description, it's unstyled

malinskibeniamin:
@claude review

malinskibeniamin:
@claude review

simon0191 (approved):

PR #2394, opened 2026-04-20
fix(secrets): paginate ListSecrets to return all entries

malinskibeniamin:
## Summary
- `useListSecretsQuery` issued a single `ListSecrets` call with `pageSize=25` and ignored `nextPageToken`, silently hiding every secret past the first page. Users with >25 secrets could not see entries like `PGDB_DSN` on the Secrets Store page (and anywhere else the hook is consumed: MCP servers, knowledge bases, AI agents, shadow links, pipeline command menu, Redpanda Connect onboarding).
- Rewrote the hook to loop `nextPageToken` via `callUnaryMethod` until the server returns an empty token, aggregating all pages into a single `{ secrets }` result. Page size now 50, matching the server-side maximum declared in `proto/redpanda/api/dataplane/v1/secret.proto`.
- Existing mutation invalidations keep working (same `createConnectQueryKey` base key).

## Repro
1. Seed a cluster with >25 secrets.
2. Curl `console-*.byoc.../redpanda.api.console.v1alpha1.SecretService/ListSecrets` with `{"request":{"pageSize":25}}` — response includes a `nextPageToken`.
3. Open `/secrets` in the UI — any secret alphabetically past the first 25 is missing.

After this fix the UI paginates server-side transparently and the table renders the full set (verified against a tenant with 119 secrets).

## Test plan
- [x] `bun vitest run src/components/pages/secrets-store` — 17 / 17 pass, including a new `follows nextPageToken until all secrets are returned` test that asserts the hook walks `page1 → page2 → stop`.
- [x] `tsgo --noEmit` clean on touched files.
- [ ] Manual smoke on a preprod cluster with >50 secrets — confirm Secrets Store lists all entries and the MCP / knowledgebase / agent secret pickers see the full set.

## Follow-ups (not in this PR)
`MAX_PAGE_SIZE = 25` is still used as a "fetch all" value in other list hooks (topics, users, security, ai-agent, pipeline, service-accounts, shadowlinks, quotas, remote-mcp, transforms). Same silent-truncation bug applies; each needs the same pagination-loop treatment. Filing / fixing separately to keep this PR scoped.

### DIFF
--- a/frontend/src/components/pages/secrets-store/secrets-store-list-page.test.tsx
+++ b/frontend/src/components/pages/secrets-store/secrets-store-list-page.test.tsx
@@ -125,6 +125,27 @@ describe('SecretsStoreListPage', () => {
     });
   });
 
+  test('stops and surfaces an error if the server returns a non-advancing pageToken', async () => {
+    const listSecretsMock = vi.fn().mockImplementation(() =>
+      create(ListSecretsResponseSchema, {
+        response: {
+          secrets: [create(SecretSchema, { id: 'looping-secret', scopes: [Scope.AI_GATEWAY] })],
+          nextPageToken: 'stuck',
+        },
+      })
+    );
+    const transport = createListSecretsTransport(listSecretsMock);
+
+    renderWithFileRoutes(<SecretsStoreListPage />, { transport });
+
+    await waitFor(() => {
+      expect(screen.getByText(/Error loading secrets:/i)).toBeVisible();
+    });
+    // Server returned nextPageToken='stuck' on first call; second call echoed 'stuck' again and the
+    // loop bailed out. Anything higher means the guard did not trip.
+    expect(listSecretsMock).toHaveBeenCalledTimes(2);
+  });
+
   test('should display empty state when no secrets exist', async () => {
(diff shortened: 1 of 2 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Addressed review:

- Added `retry: false` to `useListSecretsQuery` so the thrown guard errors (non-advancing token, max-pages cap) don't get multiplied 4× by React Query's default retry policy. Commit: 7dbcc64eb.

Skipped intentionally:
- **Options type narrowing** — kept `{ enabled?: boolean }` since that's all every current caller uses; wider signature can be reintroduced if a future caller actually needs `staleTime`/`select`, which would also force us to revisit how those interact with a multi-page aggregating queryFn.
- **`SECRETS_LIST_MAX_PAGES` exhaustion test** — agree with the reviewer's "low priority" call. The non-advancing-token test exercises the same throw path; running 200 mocked iterations just to reach the other `throw` adds test time without catching a distinct bug class.
- Fixture-comment phrasing nit — cosmetic, leaving alone.

simon0191 (approved):

PR #2397, opened 2026-04-21
feat(agents): add Bedrock provider to AIAgent proto

alenkacz:
## Summary
- Add `Bedrock` message to `AIAgent.Provider` oneof in the v1alpha3 AI agent proto
- Fields: `region` (required), `access_key_id` (optional secret ref), `secret_access_key` (optional secret ref)
- When credentials are omitted, the AI agent falls back to the AWS default credential chain (IRSA, env vars, instance profile)
- Regenerated Go, TypeScript, and OpenAPI artifacts

## Test plan
- [ ] `buf lint` passes
- [ ] `proto:generate` produces clean output
- [ ] BSR push succeeds in CI
- [ ] cloudv2 consumes new BSR SHA and compiles


proto/redpanda/api/dataplane/v1alpha3/ai_agent.proto
```diff
@@ -145,12 +145,33 @@ message AIAgent {
...
+    message Bedrock {
+      // AWS region for Bedrock inference (e.g., "us-east-1").
+      string region = 1 [
+        (google.api.field_behavior) = REQUIRED,
+        (buf.validate.field).required = true,
+        (buf.validate.field).string.pattern = "^[a-z]{2}(-[a-z]+-\\d+)?$"
+      ];
+      // Optional static AWS access key ID (secret reference).
+      // When omitted, uses the AWS default credential chain (IRSA, env vars, instance profile).
+      string access_key_id = 2 [
```
simon0191: should we suffix this with `_secret_ref` so it's clear that this value is not the secret itself ?

alenkacz: without ref it's consistent with other providers in the proto but I agree it's better with ref -> adding

simon0191 (approved):

PR #2406, opened 2026-04-21
fix(a2a): resubscribe on graceful SSE close with in-flight task

malinskibeniamin:
## Summary

Load balancers with idle timeouts (commonly ~5 min — AWS ALB, GCP LB, nginx defaults) close idle TCP connections gracefully with a FIN rather than an error. The A2A streaming iterator surfaces this as clean end-of-stream, so `streamMessage`'s existing catch-block reconnect never fires and the message is finalized with a non-terminal `taskState`, orphaning the task server-side and showing the user a truncated reply.

**Fix:** after the `for await` over `streamResult.fullStream` exits cleanly, if `isResubscribable(state)` is true, route through the same `resubscribeLoop` already used in the error path. The loop already handles backoff, progress detection, and give-up — it just needed a second entry point.

**Guard:** if the clean-close path already ran `resubscribeLoop` and a subsequent `finalizeMessage` throws (e.g., DB write fails), the outer catch must not invoke `resubscribeLoop` a second time. Tracked with a `resubscribeAttempted` flag.

## Context — not caused by recent SDK bumps

This bug predates the A2A 0.3.10 → 0.3.13, AI SDK v6, and streamdown v2 upgrades (#2389). The `try { for-await } catch { resubscribeLoop }` shape is unchanged from the original SSE reconnection commit (`4fefd7b73`). Clean-close never triggered the catch path in any version.

## Which 5-min LB failure modes does this cover

| LB behavior at 5 min | Before | After |
|---|---|---|
| Graceful TCP close (FIN) — AWS ALB, GCP LB, nginx default | No reconnect, message truncated | `resubscribeTask` fires, same backoff loop runs ✅ |
| Connection reset (RST) or socket error | Existing catch path → `resubscribeLoop` | Unchanged |
| HTTP 504 mid-stream | Existing catch path → `resubscribeLoop` | Unchanged |

If the existing error path is "still buggy" for case 2 or 3, the root cause is different from what this PR addresses.

## Test plan

**New regression tests** in `use-message-streaming.test.ts`:

- `resubscribes when stream ends cleanly but task is still in-flight (LB idle-timeout)` — asserts `resubscribeTask` is called and task completes after clean-close reconnect
- `does not resubscribe when stream ends cleanly and task is already terminal` — negative guard
- `finalizes with gave-up status when clean-close triggers resubscribe but it exhausts retries` — content preserved, single `gave-up` block shown
- `does not re-enter resubscribeLoop when finalizeMessage fails after a gave-up clean-close` — asserts exactly 5 retries (one round), not 10

**Verification (run locally):**

- [x] `bun run test:unit -- use-message-streaming` — 31/31 pass
- [x] `bun run type:check` — clean
- [x] `bun run lint` on touched files — clean
- [ ] Manual repro on a deployed cloud env: start a long-running agent task, let it idle past 5 min, watch for the `disconnected → reconnecting → reconnected` status blocks instead of silent truncation. DevTools Network panel will show `200 OK`/completed on the streaming request (graceful close) vs `(failed)`/`net::ERR_*` (error close — different code path).

@claude please review.

### DIFF
--- a/frontend/src/components/pages/agents/details/a2a/chat/hooks/use-message-streaming.test.ts
+++ b/frontend/src/components/pages/agents/details/a2a/chat/hooks/use-message-streaming.test.ts
@@ -1062,8 +1062,11 @@ describe('streamMessage - SSE reconnection via tasks/resubscribe', () => {
   // flag so that state-is-still-working + DB-error doesn't trigger another
   // full round of exponential-backoff retries.
   test('does not re-enter resubscribeLoop when finalizeMessage fails after a gave-up clean-close', async () => {
-    vi.spyOn(console, 'error').mockImplementation(() => {
-      // Silence expected finalize-failure log so test output stays clean.
+    // The clean-close branch logs 'finalizeMessage failed after clean-close
+    // recovery:' on this path. Capture the spy so we can silence the expected
+    // log and also assert it fires exactly once.
+    const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {
+      // Swallow expected negative-path log to keep test output clean.
     });
 
     const TASK_ID = 'task-clean-close-no-double-reconnect';
@@ -1102,6 +1105,14 @@ describe('streamMessage - SSE reconnection via tasks/resubscribe', () => {
     // Fell through to error path because finalizeMessage threw.
     expect(result.success).toBe(false);
     expect(result.assistantMessage.contentBlocks.some((b) => b.type === 'a2a-error')).toBe(true);
+
+    // The clean-close recovery logs finalize failures for production
+    // observability; assert it fires exactly once (not once per resubscribe
+    // attempt and not silently swallowed).
+    const cleanCloseLogs = errorSpy.mock.calls.filter(
+      (args) => typeof args[0] === 'string' && args[0].startsWith('finalizeMessage failed after clean-close recovery')
+    );
+    expect(cleanCloseLogs).toHaveLength(1);
   });
 
   // -------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
@claude please review this PR. Specifically interested in your read on:
1. Whether the `resubscribeAttempted` flag is the cleanest way to prevent double-reconnect, or whether this would be better refactored (e.g., extract a shared "resubscribe + finalize" helper used by both the clean-close and error paths).
2. Whether the clean-close entry into `resubscribeLoop` should have any other guardrails — e.g., a minimum delay before the first "disconnected" block appears, to avoid flashing the reconnect UI for quick-completing tasks that happen to also clean-close at the wire level.
3. Any scenario I missed in the tests.

malinskibeniamin:
@claude I've addressed the review feedback. Summary of changes in 2d9abcf1b:

**Production code:**
1. Removed the redundant `closeActiveTextBlock`/`state.activeTextBlock = null` calls in the clean-close branch (the lines above already do this with nothing mutating `activeTextBlock` in between).
2. Made the clean-close gave-up path return `success: false` for symmetry with the error-path gave-up. Orphaned tasks are failures regardless of whether the disconnect was graceful or thrown. The `gave-up` connection-status block is still visible to the user either way.

**Tests:**
3. Renumbered all scenarios into a clean 1–25 sequence. Previously we had 13, 13b, 13c interleaved with 14–17 plus an added 16b/c/d/e block.
4. Renamed Scenario 17 (the TypeError test) from `rethrows TypeError from resubscribe instead of retrying` to `stops retrying immediately on TypeError from resubscribe` — the code breaks out and falls through, it doesn't rethrow.
5. Updated the gave-up clean-close test to assert `success: false` (matches the new production behavior).
6. Added Scenario 23: clean-close resubscribe succeeds, then `finalizeMessage` throws — explicitly guards the success-path arm of `resubscribeAttempted` against future regressions that change terminal-state tracking.
7. Added Scenario 24: clean-close with `taskId` from response metadata fallback only — confirms the fallback runs before the clean-close `isResubscribable` check and doesn't spuriously reconnect.

33/33 tests pass. Type-check and lint clean.

Please take another pass.

malinskibeniamin:
@claude review

malinskibeniamin:
@claude review

eblairmckee (approved):

PR #2420, opened 2026-04-23
fix(topics): clamp default replication factor to broker count

c-julin:
## Summary

Three topic-create surfaces in Console hardcoded \`replicationFactor: 3\`. On single-broker clusters (local-byoc dev environments, degraded prod edges) that caused CreateTopic to fail with \"not enough replicas\". The RPCN onboarding form was the most visible case because the RF input is rendered \`readOnly\` — users could see the \`3\` but had no way to change it.

Files changed:
- \`pages/rp-connect/onboarding/add-topic-step.tsx\` — override \`TOPIC_FORM_DEFAULTS.replicationFactor\` at form init with \`min(3, brokersOnline)\`. **This is the user-visible fix**: the readOnly field now shows the correct value for single-broker clusters.
- \`pages/topics/create-topic-modal.tsx\` — new topic-create modal; clamp RF inline.
- \`pages/mcp-servers/details/remote-mcp-inspector-tab.tsx\` — MCP inspector's inline topic-create flow; clamp RF inline.

All three use \`useGetKafkaInfoQuery\` to read \`brokersOnline\` and fall back to the original \`3\` while the query is loading. On 3+ broker clusters the behavior is unchanged.

Note: the legacy \`pages/topics/CreateTopicModal/create-topic-modal.tsx\` already reads the default from the broker's \`default.replication.factor\` config, so it's already correct provided the broker config is correct.

## Test plan

- [x] \`bun run test:ci\` (unit): 755/755 passed on the branch
- [x] \`bun run lint\` auto-sorted imports; no new lint findings introduced by this change
- [ ] Manual verification on a 1-broker local-byoc cluster: RPCN onboarding → Add Topic should show RF=1 (readOnly), CreateTopic succeeds
- [ ] Manual verification on a 3-broker cluster: RPCN onboarding shows RF=3 as before

### DIFF
--- a/frontend/src/components/pages/mcp-servers/details/remote-mcp-inspector-tab.tsx
+++ b/frontend/src/components/pages/mcp-servers/details/remote-mcp-inspector-tab.tsx
@@ -36,7 +36,6 @@ import {
 } from 'protogen/redpanda/api/dataplane/v1/topic_pb';
 import { MCPServer_State, MCPServer_Tool_ComponentType } from 'protogen/redpanda/api/dataplane/v1alpha3/mcp_pb';
 import { useEffect, useRef, useState } from 'react';
-import { useGetKafkaInfoQuery } from 'react-query/api/cluster-status';
 import {
   type MCPStreamProgress,
   useGetMCPServerQuery,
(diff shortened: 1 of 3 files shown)


c-julin:
@claude review

malinskibeniamin (approved):

PR #2424, opened 2026-04-24
Migration - Base UI based registry components

SpicyPete:
### Summary

* Registry recently moved from Radix-ui to base-ui 
* Bringing in the most recent base-ui based components
* The majority of the PR updates are registry related
* Shouldn't be many usage changes, a drop-in replacement 

### Notes:

* Replaced the `user-preferences` dialog with registry components to help surface some registry bugs
* Registry components now have dependency version requirements
* Zod 3 -> 4, some fields got renamed
* Chakra relied on date-fns-tz, replaced the components which used this with a local `date-time-input` component + shim the date-fns-tz so the Chakra components can still compile.

#### Todo 
- [x] Integration tests passing
- [x] Some registry PR's which resolve some of the issues surfaced from this PR.


frontend/src/components/pages/secrets-store/edit/secret-edit-page.tsx
```diff
@@ -191,7 +191,6 @@ export const SecretEditPage = () => {
             <Field data-invalid={fieldState.invalid}>
               <FieldLabel required>Scopes</FieldLabel>
               <MultiSelect
-                items={SCOPE_OPTIONS}
```
malinskibeniamin, line 194: How do item options get shared with multi select?

malinskibeniamin (approved):

frontend/components.json
```diff
@@ -7,15 +7,17 @@
...
-    "hooks": "hooks"
+    "components": "@/components/redpanda-ui",
+    "utils": "@/components/redpanda-ui/lib/utils",
+    "ui": "@/components/redpanda-ui",
+    "lib": "@/components/redpanda-ui/lib",
+    "hooks": "@/components/redpanda-ui/lib"
   },
-  "iconLibrary": "lucide"
+  "iconLibrary": "lucide",
+  "registries": {
```
malinskibeniamin, line 20: 👏

frontend/src/components/pages/secrets-store/edit/secret-edit-page.tsx
SpicyPete: `{SCOPE_OPTIONS.map((option) => (`
Just below this in the file we are mapping over SCOPE_OPTIONS again, so before we were passing them in twice, but only 1 was being used.

frontend/package.json
```diff
@@ -122,46 +125,45 @@
     "json-bigint": "^1.0.0",
     "jwt-decode": "^4.0.0",
     "lottie-react": "^2.4.1",
-    "lucide-react": "^1.8.0",
+    "lucide-react": "^1.7.0",
```
malinskibeniamin: Maybe we don't need to downgrade (I suppose this was due to UI registry using older version)

frontend/src/components/ai-elements/context.tsx
```diff
@@ -77,6 +77,7 @@ export const Context = ({
 
   return (
     <ContextContext.Provider value={contextValue}>
+      {/* @ts-expect-error openDelay/closeDelay are not in HoverCard's typed props after registry update */}
```
malinskibeniamin: Do we need to rectify it instead?

SpicyPete:
@malinskibeniamin I think this is ready for a review this week.  Fixed a bunch of registry components over the last week, and brought the latest versions all into here now.  
Most of the surface are of the changes are in the new components, and converted the user-preference dialog to our new components to solve some issues in it.

frontend/src/components/ai-elements/context.tsx
malinskibeniamin, line 1: I think AI elements deserve a separate look because they were originally pulled in from the https://elements.ai-sdk.dev/components/context for example

frontend/src/components/pages/admin/admin-debug-bundle.tsx
```diff
@@ -35,6 +34,7 @@ import {
   Text,
 } from '@redpanda-data/ui';
 import { TrashIcon } from 'components/icons';
+import { DateTimeInput } from 'components/ui/date-time-input';
```
malinskibeniamin, line 37: 👍

frontend/package.json
```diff
@@ -87,19 +89,19 @@
...
     "@monaco-editor/react": "^4.7.0",
     "@redpanda-data/ui": "^4.2.0",
-    "@stepperize/react": "^5.1.8",
-    "@tailwindcss/postcss": "^4.1.17",
+    "@stepperize/react": "^5.1.9",
+    "@tailwindcss/postcss": "^4.2.4",
     "@tanstack/react-form": "^1.12.3",
     "@tanstack/react-query": "^5.90.7",
-    "@tanstack/react-router": "^1.158.0",
+    "@tanstack/react-router": "^1.168.10",
```
malinskibeniamin, line 104: Thanks.

malinskibeniamin (approved):

frontend/package.json
SpicyPete: Yes, updated registry to use a newer version of lucide-react

frontend/src/components/ai-elements/context.tsx
SpicyPete: I don't think so, removed that comment.  

Base UI still has openDelays, it just moved to the Trigger component.  I don't think it's used enough to require backwards compatability logic

r-vasquez (approved):

Issue #2421, opened 2026-04-23
[Bug] Avro bytes fields rendered as Latin-1 codepoint strings instead of Base64 since v3.7.0


haoyukongTrackunit:
## Description

Since v3.7.0, Avro `bytes` fields — including those inside Debezium logical types such as `io.debezium.data.VariableScaleDecimal` — are rendered in the Console UI as raw Latin-1/UTF-8 codepoint strings (e.g. `"\r\u0003\u0012\u00aa\u00a3_\u00f2"`) instead of Base64 (e.g. `"QwhcN9vr+A=="`).

This is a regression from v2.8.x behaviour and breaks the ability to visually verify decimal values encoded via Debezium's `VariableScaleDecimal` type directly in the Console UI.

## Root Cause

The regression was introduced in PR #2351 (commit `14d05a16f`), which changed Avro serialisation from Go's `json.Marshal` to `avroSch.EncodeJSON`. The Avro JSON spec encodes `bytes` as ISO-8859-1 codepoint strings rather than Base64. While spec-compliant, this is a breaking behaviour change for consumers who relied on the previous Base64 output for manual data verification.

## Steps to Reproduce

1. Configure a Kafka topic with Avro-encoded messages containing `bytes` fields, or Debezium CDC messages containing `io.debezium.data.VariableScaleDecimal` fields (e.g. a numeric/decimal PostgreSQL column via Debezium).
2. Open the topic in Redpanda Console v3.7.0 or v3.7.1.
3. Inspect a message — the `bytes` / `VariableScaleDecimal.value` field renders as a raw codepoint string instead of Base64.

## Expected Behaviour

`bytes` fields should be rendered as Base64-encoded strings, consistent with v2.8.x behaviour.

**v2.8.11 output (expected):**
```json
"cumulative_emissions": {
  "io.debezium.data.VariableScaleDecimal": {
    "scale": 15,
    "value": "cVLOXkWiqq=="
  }
}
```

## Actual Behaviour

**v3.7.1 output (actual):**
```json
"cumulative_emissions": {
  "scale": 13,
  "value": "\tn\\\u001a#\u00dbN"
}
```

Bytes above `0x7F` are corrupted when re-encoded as UTF-8 codepoints, making the original value unrecoverable from the UI alone.

## Impact

1. **Lossless verification broken** — Base64 can be reliably decoded to the original byte sequence; Latin-1 codepoint strings mangle bytes above `0x7F` and the original data cannot be recovered from the UI.
[...]

PR #2425, opened 2026-04-24
fix(frontend): preserve \u00XX escapes in Avro JSON viewer

twmb:
## Summary

- Fixes the copy-paste loss reported in #2421 for Avro `bytes` / `fixed` fields.
- Root cause: Avro JSON encodes bytes as ISO-8859-1 code points (`\u00XX`), but the frontend runs the payload through `JSON.parse` then `JSON.stringify` for display. `JSON.stringify` does not re-escape non-ASCII, so a byte like `0xDB` renders as the literal glyph `Û` and "Copy Value" places UTF-8 bytes on the clipboard instead of the original byte.
- Fix is frontend-only and scoped: `KowlJsonView` gains an optional `escapeLatin1` prop that re-escapes code points in `0x80-0xFF` back to `\u00XX` before display. `PayloadComponent` sets it only when `payload.encoding === 'avro'`, so non-avro payloads are unchanged.
- The transform is lossless for both Avro bytes fields (recovers the exact byte) and legitimate Latin-1 strings (valid JSON escape for the same Unicode character).

## Test plan

- [ ] Open an Avro topic with a `bytes` field (e.g. Debezium `VariableScaleDecimal.value`); confirm the value renders as `\u00XX` escapes and "Copy Value" yields the on-wire JSON.
- [ ] Confirm non-avro payloads (proto, json, text) render unchanged.
- [ ] `bun run type:check && bun run lint && bun run test` pass in CI.

### DIFF
--- a/frontend/src/components/misc/kowl-json-view.test.tsx
+++ b/frontend/src/components/misc/kowl-json-view.test.tsx
@@ -161,4 +161,24 @@ describe('KowlJsonView', () => {
 
     expect(editorLayoutSpy).toHaveBeenLastCalledWith({ width: 800, height: 480 });
   });
+
+  test('escapeLatin1 re-escapes Latin-1 code points so copy-paste recovers the original byte', async () => {
+    render(<KowlJsonView escapeLatin1 srcObj={{ v: 'ÛN' }} />);
+
+    await waitFor(() => {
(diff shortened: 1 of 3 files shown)


weeco:
@claude review

malinskibeniamin (approved):

PR #2426, opened 2026-04-28
Security page refactor

jvorcak:
Screenshot 2026-05-08 at 10 21 53
Screenshot 2026-05-08 at 10 21 44
Screenshot 2026-05-08 at 10 21 38
Screenshot 2026-05-08 at 10 21 29


frontend/tests/test-variant-console/utils/security-page.ts
```diff
@@ -65,7 +62,6 @@ export class SecurityPage {
     return await test.step('Create user', async () => {
       await this.goToUsersList();
       await this.clickCreateUserButton();
-      await this.page.waitForURL('/security/users/create');
```
malinskibeniamin, line 68: Is it no longer a distinct URL? Shouldn't we wait for some page navigation?

frontend/src/components/pages/security/roles/role-create-dialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+    setSubmitted(true);
+    if (!trimmed || alreadyExists) return;
+    setIsSubmitting(true);
+    try {
+      await createRole(create(CreateRoleRequestSchema, { role: { name: trimmed } }));
+      toast.success(`Role "${trimmed}" created`);
+      handleClose();
+      navigate({ to: '/security/roles/$roleName/details', params: { roleName: encodeURIComponent(trimmed) } });
+    } catch (err) {
+      toast.error(`Failed to create role: ${err instanceof Error ? err.message : String(err)}`);
```
malinskibeniamin: Can we `ConnectError.from(error)`

frontend/src/components/pages/security/roles/role-detail-page.tsx
```diff
@@ -9,163 +9,157 @@
...
+    try {
+      await updateMembership(
+        create(UpdateRoleMembershipRequestSchema, {
+          roleName,
+          add: [],
+          remove: [{ principal: memberPrincipal }],
+          create: false,
+        })
+      );
+      toast.success('Member removed from role successfully');
```
malinskibeniamin: Should we mention the member?

frontend/src/components/pages/security/shared/acls-card.tsx
```diff
@@ -0,0 +1,291 @@
...
+                  Allow all operations
+                </Button>
+              )}
+              <Button onClick={() => setDialogOpen(true)} testId="add-acl-button" variant="outline">
+                + Add ACL
+              </Button>
+            </div>
+          }
+        >
+          <h2 className="font-semibold text-base">ACLs</h2>
```
malinskibeniamin: Maybe we can use Typography component from UI registry (if this is not using mix of Chakra/Tailwind)

frontend/src/components/pages/security/shared/description-with-help.tsx
malinskibeniamin, line 1: I understand it was done this way before, but I wonder if we should just show tooltips when user hovers over the info icon? I am not sure if a drawer always makes sense. Maybe it depends on the size of the message?

malinskibeniamin (approved):

frontend/src/components/pages/security/users/add-acl-dialog.tsx
```diff
@@ -0,0 +1,348 @@
...
+                              field.value === opt.value
+                                ? 'bg-background text-foreground shadow-sm'
+                                : 'text-muted-foreground hover:text-foreground'
+                            }`}
+                            key={opt.value}
+                            onClick={() => field.onChange(opt.value)}
+                            type="button"
+                          >
+                            {opt.label}
+                          </button>
```
malinskibeniamin: Is this a card?

frontend/src/components/pages/security/tabs/permissions-list-tab.tsx
```diff
@@ -42,115 +43,240 @@ import {
...
+      return `${group.directAclCount} direct ACL${group.directAclCount !== 1 ? 's' : ''}, ${group.inheritedAclCount} ACL${group.inheritedAclCount !== 1 ? 's' : ''} inherited from roles`;
+    }
+    if (group.inheritedAclCount > 0) {
+      return `${group.inheritedAclCount} ACL${group.inheritedAclCount !== 1 ? 's' : ''} inherited from roles`;
+    }
+    if (group.directAclCount > 0) {
+      return `${group.directAclCount} direct ACL${group.directAclCount !== 1 ? 's' : ''}`;
+    }
+    return 'No ACLs';
+  })();
```
malinskibeniamin: I feel like this needs a small refactor, we could try to use `pluralize` function too.

frontend/src/components/pages/security/tabs/permissions-list-tab.tsx
```diff
@@ -42,115 +43,240 @@ import {
...
+                  ))}
+                </TableBody>
+              ))}
+            </Table>
+          </div>
+        )}
+
+        {isExpanded && !hasAcls && (
+          <div className="border-t px-3 py-4 text-muted-foreground text-sm">No ACLs assigned.</div>
+        )}
```
malinskibeniamin: Consider using `<Empty>` component from UI registry

malinskibeniamin:
@claude review

frontend/tests/test-variant-console/utils/security-page.ts
jvorcak: Good catch. The user creation flow is now dialog-based so there is no longer a page navigation on submit. After the user sees the confirmation step, they click "Go to user details" which navigates to the details page. Updated the `createUser` helper to click that button and `waitForURL` so the helper ends in a known navigation state.

frontend/src/components/pages/security/roles/role-create-dialog.tsx
jvorcak: Done — now using `ConnectError.from(err).message` for a structured gRPC error message.

frontend/src/components/pages/security/roles/role-detail-page.tsx
jvorcak: Updated to: `User "${parsePrincipal(memberPrincipal).name}" removed from role`

frontend/src/components/pages/security/shared/acls-card.tsx
jvorcak: Done — replaced `<h2 className="font-semibold text-base">` with `<Heading as="h2" level={4}>` from the registry typography component.

frontend/src/components/pages/security/shared/description-with-help.tsx
jvorcak: All current usages pass a single paragraph of explanatory text (2-3 sentences), so the Drawer feels proportionate — it gives enough reading space without truncation. A Tooltip would need to either truncate or wrap uncomfortably for that length. Happy to add a `compact` prop that switches to a Tooltip if shorter content is expected in future, but for now keeping the Drawer.

frontend/src/components/pages/security/users/add-acl-dialog.tsx
jvorcak: Not a card — this is a segmented toggle group (inline tabs) for selecting the ACL pattern type (LITERAL vs PREFIXED). The buttons share a container `div` that mimics a pill/tab strip. No card semantics here.

frontend/src/components/pages/security/tabs/permissions-list-tab.tsx
jvorcak: Refactored to use `pluralizeWithNumber` from `utils/string`. The three branches now read as: `pluralizeWithNumber(n, "direct ACL")`, `pluralizeWithNumber(n, "ACL") inherited from roles`, etc.

frontend/src/components/pages/security/tabs/permissions-list-tab.tsx
jvorcak: Done — replaced the plain `<div>` with `<Empty>` + `<EmptyHeader>` + `<EmptyTitle>` from the registry. Applied `className` overrides (`rounded-none border-x-0 border-b-0 border-dashed`) to keep it visually appropriate for the inline expanded-row context.

jvorcak:
@claude review

frontend/src/components/redpanda-ui/components/multi-select.tsx
```diff
@@ -341,7 +341,7 @@ const MultiSelectContent = React.forwardRef<
       <PopoverPrimitive.Content
         align="start"
         className={cn(
-          'data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 relative z-50 max-h-(--radix-select-content-available-height) min-w-[8rem] origin-(--radix-select-content-transform-origin) overflow-y-auto overflow-x-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=bottom]:translate-y-1 data-[side=top]:-translate-y-1 data-[state=closed]:animate-out data-[state=open]:animate-in',
```
SpicyPete: Minor -  For this and the `list-layout` change they will be blasted away whenever the registry component updates.  

I wonder what a good process would be to use local modifications in these components before they are upstreamed.

frontend/src/components/license/feature-license-notification.tsx
```diff
@@ -47,23 +48,23 @@ const getLicenseAlertContentForFeature = (
     if (bakedInTrial) {
       return {
         message: (
-          <Box>
-            <Text>This is an enterprise feature. Register for an additional 30 days of enterprise features.</Text>
-            <Flex gap={2} my={2}>
+          <div>
+            <p>This is an enterprise feature. Register for an additional 30 days of enterprise features.</p>
```
SpicyPete, line 51: Minor Q: Do we use plain `<p>` tags, or should we stick to Typography components?

frontend/src/components/pages/security/roles/role-detail-page-new.tsx
```diff
@@ -0,0 +1,208 @@
...
+          actions={
+            <Combobox
+              className="w-56"
+              clearable={false}
+              disabled={isSubmitting}
+              onChange={addMember}
+              options={availablePrincipalOptions}
+              placeholder="Add a principal..."
+              testId="add-principal-combobox"
+              value=""
```
SpicyPete, line 128: Should this be a controlled component, with an explicit value prop set?

frontend/src/components/pages/security/users/user-create.tsx
```diff
@@ -364,13 +367,19 @@ const CreateUserConfirmationModal = ({
...
-        Done
-      </Button>
-      <Button onClick={onCreateAcls} testId="create-acls-button" variant="link">
-        Create ACLs
-      </Button>
+    <div className="mt-3 border-t pt-6">
+      <h2 className="mb-2 font-semibold text-base">Assign new user permissions</h2>
+      <p className="my-3 text-muted-foreground text-sm">
+        To grant access to clusters, assign a role to the user or create ACLs.
+      </p>
```
SpicyPete, line 375: Same for these, should we use system Header and Text related components

frontend/.gitignore
```diff
@@ -162,3 +162,4 @@ dist
 .gemini/settings.json
 
 tests/**/playwright-report/
+tests/**/*.js
```
SpicyPete, line 165: Is this needed?  Seems like a pretty wide-net to add to .gitignore

SpicyPete (commented):
Just a few minor comments on the code, going to run this locally next.

SpicyPete:
Some misc feedback as I am poking around the new pages.

As this is all behind a feature flag nothing here is a blocker.

##### 1. visual issues
Screenshot 2026-05-06 at 9 04 37 AM

##### 2. Hide can't unhide
In this column, we can hide the column, but there's no way to un-hide it?  Might be a issue at the registry level.  refreshing the page brings it back.  
Screenshot 2026-05-06 at 9 05 33 AM

##### 3. Missing form feedback
This form says the arguments are invalid, but it's not clear why or what's wrong
Screenshot 2026-05-06 at 9 06 55 AM

##### 4. destructive guards
Here we can remove principals without any confirmation.  I'm not sure what we do organization wide or if it's in scope of this PR, but I think destructive actions should have confirmation guards
Screenshot 2026-05-06 at 9 09 18 AM

##### 5. Principals count
Shows 0 principals, but I have 1 assigned
Screenshot 2026-05-06 at 9 10 46 AM

##### 6. Users layout
We are showing "users" then just below it "users" again, and then the first actions are to delete the user.  I think we can have a better layout, and not surface the admin actions of change password/delete as the first actions on the users.
Screenshot 2026-05-06 at 9 13 29 AM

frontend/.gitignore
jvorcak: These are generated `.js` files that sometime occur when running tests locally. We use `.ts` for e2e. Of course I'll need to check whether we can somehow prevent from `.js` to be generated in the first place.

frontend/src/components/redpanda-ui/components/multi-select.tsx
jvorcak: I think the process is fine as is, this change was unintentional tbh. Or more like temporal debugging.

frontend/src/components/license/feature-license-notification.tsx
jvorcak: I personally thing both are fine for regular content. I prefer to use Typography for headings when I need to style something so that it's consistent. But we can discuss it later on.

frontend/src/components/redpanda-ui/components/multi-select.tsx
jvorcak: Reverted — the `z-[9999]` was an accidental debug change. Restored to `z-50`.

frontend/src/components/pages/security/users/user-create.tsx
jvorcak: Fixed — replaced `<h2 className="...">...\</h2>` with `<Heading as="h2" level={4}>` and `<p className="...">` with `<Text>` from the registry.

frontend/src/components/pages/security/roles/role-detail-page-new.tsx
jvorcak: Yes — `value=""` is already set explicitly on the Combobox, making it a controlled component. The empty string is intentional: it resets the displayed value after each selection so the placeholder is always visible. This Combobox acts as an action trigger (pick a principal → fire `addMember`) rather than a persistent selection, so resetting to empty on each pick is the desired behaviour.

jvorcak:
@SpicyPete 

1. We need to put the alert back. If we want this to be inline like I suggested, I'd need to put more work into licensing. For now, an alert works. Good catch!
2. I think that's a problem with the UI Registry
3. Fixed
4. Nice point, I'll implement
5. I think it was just a bad place when the data refetches
6. 👍

SpicyPete:
Nice, looks like those issues are resolved.   
Just one minor issue I'm seeing: after creating a user it shows a dialog asking to write down my password, but the input is disabled and I'm not able to view the password.

weeco:
**1. Hovering ACLs pill on users list**

**Desired:**

macshot-clipboard-DB562402-32A9-405D-BC10-2F96475D1066


**Actual:** No tooltip


**2. When clicking the ACL pill**

macshot-clipboard-CC93D08E-05FD-4171-B937-CD23A36F36CA

IMO it should not be clickable, we don't need it if we have the hover.

**3. NIT: Different colors, sizes and hover area for buttons in password field**

Why is the regenerate button purple, and the copy button so different?

macshot-clipboard-D9EA92A1-8D85-44A2-87FB-ED0B95EC2FF3

**4. NIT: Can use an input field and copy button in the user creation success dialog for the username?**

**Actual:**

macshot-clipboard-AA6D7C57-38C1-4DF3-B96C-84E2AFD64CFE

**Desired (copyable username, still shown in plaintext in an input field?):**

macshot-clipboard-65B3CFC1-A32C-44BC-AE7D-A84610633239

**5. User actions - no icons and danger variant**

**Actual:**

macshot-clipboard-DA72D385-439B-4B90-AEBA-5081013DE11D


**Desired:**

macshot-clipboard-369B3174-61A0-4F24-B65F-65E2DB9C8E48


**6. Incosistent height from user button and search field / filter buttons**

**Actual:**

macshot-clipboard-B0BFE035-01E5-476B-A8FF-5FF882FFDA74

**Desired:**

The input / search as well as the filter button should match the height/size of the create buttons on the pages. It looks weirdly misaligned. 

**7. Permissions cards are hard to parse due to missing spacing etc.**

It's very hard to parse the permissions in the current layout and spacing since its a nested table. The original mock proposed collapsible cards for each principal with a very distinct header on the principal. This makes it much easier to recognize the different principals and map the ACLs to them

**Actual:**

macshot-clipboard-176D3205-5A96-483D-854D-5B80B31E51E5


**Desired:**

macshot-clipboard-87DED98B-EC79-4036-B1EB-CCD45316D54F


**8. Padding on users details page:**

macshot-clipboard-18F9DE68-29CE-4CC5-84E3-6F9F34AB7880

**9. NIT: Render ACL resource types as pill**

**Actual:**

macshot-clipboard-48F9EED4-EE35-4DA5-B303-B14597164CCE

**Desired:**

macshot-clipboard-A0AD0D14-9867-4E38-8F6E-5D8F95951E41

**10. User details page header**

**Actual:**

macshot-clipboard-E45157C0-5042-4D2C-B4FA-DBA114D93719


**Desired (show principal and mechanism explicitly, move buttons to the right in the same row as title to balance):**

macshot-clipboard-08F4A49D-CD66-4306-8BB9-4D70A27810CD

**11. NIT: Make modal larger for allow all ops:**

**Actual:**

macshot-clipboard-44F39189-3FD0-4272-865A-120D10B1F29E

**12. Make delete button same size as other buttons:**

macshot-clipboard-38FE115F-33CB-483D-B08C-DC83DB6593F8

weeco:
The ADD ACLs dialogue should work  as designed, you shouldn't insist on an existing user or principal. Both should accept free inputs, currently it would just not accept my user:

macshot-clipboard-20D38812-7D5B-429E-B176-A64383AFD851

Also the whole Users functionality is gated if redpanda admin api is not configured. That's no longer true, because we use the Kafka API for creating users, updating passwords or listing users.

frontend/src/state/ui-state.ts
```diff
@@ -232,8 +259,20 @@ export const uiState = new Proxy(
         store.setServerBuildTimestamp(value as number | undefined);
         return true;
       }
+      if (prop === 'backLink') {
+        store.setBackLink(value as BackLink | null);
```
malinskibeniamin, line 263: Is there any tech debt stopping us from relying on tanstack router "go back" functionality directly?

frontend/src/routes/security.tsx
```diff
@@ -166,3 +170,46 @@ function SecurityLayout() {
...
+    const refreshData = async () => {
+      await Promise.allSettled([api.refreshClusterOverview(), rolesApi.refreshRoles(), api.refreshUserData()]);
+      await rolesApi.refreshRoleMembers();
+    };
+
+    appGlobal.onRefresh = async () => {
+      await refreshData();
+    };
+
+    refreshData().catch(() => {});
```
malinskibeniamin, line 187: shouldn't we catch the error?

frontend/src/react-query/api/acl.tsx
```diff
@@ -26,6 +27,46 @@ import {
   type SharedConfig,
 } from '../../components/pages/security/shared/acl-model';
 
+/**
+ * TODO: Remove once Console v3 is released.
+ */
```
malinskibeniamin, line 32: is this still needed? console v3 is out for a while

frontend/src/components/pages/security/users/user-roles-card-new.tsx
```diff
@@ -0,0 +1,213 @@
...
+    if (!(userName && roleName)) return;
+    await updateRoleMembership(create(UpdateRoleMembershipRequestSchema, { roleName, add: [{ principal: userName }] }));
+    await Promise.all([rolesApi.refreshRoles(), rolesApi.refreshRoleMembers()]);
+  };
+
+  const count = roles.length;
+
+  const renderBody = () => {
+    if (isLoading) {
+      return [0, 1, 2].map((i) => (
```
malinskibeniamin, line 86: Maybe Array.from() instead? this is not super obvious, I guess 3 skeletons rendering in loading state

frontend/src/components/pages/security/users/add-acl-dialog.tsx
```diff
@@ -0,0 +1,367 @@
...
+                    name="patternType"
+                    render={({ field }) => (
+                      <div>
+                        <div className="inline-flex h-10 w-full items-center justify-center rounded-md bg-muted p-1">
+                          {[
+                            { value: ACL_ResourcePatternType.LITERAL, label: 'Literal' },
+                            { value: ACL_ResourcePatternType.PREFIXED, label: 'Prefixed' },
+                            { value: ACL_ResourcePatternType.ANY, label: 'Any' },
+                          ].map((opt) => (
+                            <button
```
malinskibeniamin, line 267: is this just a card hiding as a button?

frontend/src/components/pages/security/users/add-acl-dialog.tsx
```diff
@@ -0,0 +1,367 @@
...
+                        creatable
+                        createLabel="user"
+                        onChange={setPrincipalValue}
+                        options={userOptions}
+                        placeholder="Select or type a user..."
+                        value={principalValue}
+                      />
+                    ) : (
+                      <input
+                        // biome-ignore lint/a11y/noAutofocus: auto-focus after switching type so the user can type immediately
```
malinskibeniamin, line 217: If we use `Input` component instead of native HTML tag, would it help avoid the no auto focus rule?

frontend/src/react-query/api/acl.tsx
jvorcak: I'll try to address in a separate PR. I've created https://redpandadata.atlassian.net/browse/UX-1279

jvorcak:
Screenshot 2026-05-18 at 17 52 32
Kafka env

frontend/src/components/pages/security/users/add-acl-dialog.tsx
```diff
@@ -0,0 +1,367 @@
...
+type FormValues = z.infer<typeof schema>;
+
+const RESOURCE_TYPE_OPTIONS = [
+  { value: ACL_ResourceType.TOPIC, label: 'Topic' },
+  { value: ACL_ResourceType.GROUP, label: 'Consumer Group' },
+  { value: ACL_ResourceType.CLUSTER, label: 'Cluster' },
+  { value: ACL_ResourceType.TRANSACTIONAL_ID, label: 'Transactional ID' },
+  { value: ACL_ResourceType.SUBJECT, label: 'Subject' },
+  { value: ACL_ResourceType.REGISTRY, label: 'Schema Registry' },
+];
```
weeco: Schema registry and subject are both specific to redpanda:


```suggestion
  const RESOURCE_TYPE_OPTIONS = [
    { value: ACL_ResourceType.TOPIC,           label: 'Topic' },
    { value: ACL_ResourceType.GROUP,           label: 'Consumer Group' },
    { value: ACL_ResourceType.CLUSTER,         label: 'Cluster' },
    { value: ACL_ResourceType.TRANSACTIONAL_ID, label: 'Transactional ID' },
    ...(api.isRedpanda ? [
      { value: ACL_ResourceType.SUBJECT,  label: 'Subject' },
      { value: ACL_ResourceType.REGISTRY, label: 'Schema Registry' },
    ] : []),
  ];
```

frontend/src/components/pages/security/acls/acl-detail-page.tsx
```diff
@@ -9,23 +9,22 @@
  * by the Apache License, Version 2.0
  */
 
-import { getRouteApi, useNavigate } from '@tanstack/react-router';
+import { getRouteApi } from '@tanstack/react-router';
 
 const routeApi = getRouteApi('/security/acls/$aclName/details');
 
-import { Pencil } from 'lucide-react';
+import { useLayoutEffect } from 'react';
```
weeco, line 17: The input elements should have the same size / height as the other elements in the same row. This problem seems to be consistent across all tabs, so make sure to check all tabs too:

Image

frontend/src/components/pages/security/users/user-create.tsx
```diff
@@ -37,7 +38,7 @@ import { Input } from '../../../redpanda-ui/components/input';
 import { SimpleMultiSelect } from '../../../redpanda-ui/components/multi-select';
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '../../../redpanda-ui/components/select';
 import { Tooltip, TooltipContent, TooltipTrigger } from '../../../redpanda-ui/components/tooltip';
-import { useSecurityBreadcrumbs } from '../hooks/use-security-breadcrumbs';
+import { Text } from '../../../redpanda-ui/components/typography';
```
weeco, line 41: Different button styles for the password with different colors & sizes.

Image

frontend/src/components/pages/security/users/user-details-new.tsx
```diff
@@ -0,0 +1,175 @@
...
+  userName: string;
+};
+
+const formatMechanism = (mechanism?: SASLMechanism): string | null => {
+  if (mechanism === SASLMechanism.SASL_MECHANISM_SCRAM_SHA_256) return 'SCRAM-SHA-256';
+  if (mechanism === SASLMechanism.SASL_MECHANISM_SCRAM_SHA_512) return 'SCRAM-SHA-512';
+  return null;
+};
+
+export const UserDetailsPageNew = ({ userName }: UserDetailsPageProps) => {
```
weeco, line 44: For Apache Kafka we don't support roles but we surface roles in the user details page. It should only be surfaced for Redpanda enterprise / RBAC enabled (a feature) I think

Image

frontend/src/components/pages/security/users/add-acl-dialog.tsx
jvorcak: Done — `SUBJECT` and `REGISTRY` are now gated behind `api.isRedpanda`. The base options (Topic, Consumer Group, Cluster, Transactional ID) always appear; the two Redpanda-specific types are spread in only when `api.isRedpanda` is true.

frontend/src/components/pages/security/acls/acl-detail-page.tsx
jvorcak: Fixed — replaced the raw `<input>` with `<ListLayoutSearchInput>` in the permissions tab. This is the same component used in the Users and Roles tabs, so all three tabs now have consistent input height/styling in their filter rows.

frontend/src/components/pages/security/users/user-details-new.tsx
jvorcak: Fixed — `<UserRolesCardNew>` is now conditionally rendered only when `featureRolesApi` (the RBAC feature flag from `useSupportedFeaturesStore`) is true, so it is hidden entirely for Apache Kafka and Redpanda without RBAC enabled.

frontend/src/components/pages/security/acls/acl-detail-page.tsx
jvorcak: Screenshot 2026-05-19 at 13 50 09

We will need to look at this at the component level. It technically is the same height now with the outline. You're making a good point, but I think we will need to get to the UI registry level to address this.

frontend/src/components/pages/security/users/user-create.tsx
jvorcak: Both are imported from the UI registry. One as a button, one as a copy button. I'll address at the registry UI level.

PR #2431, opened 2026-05-05
Fix schema registry overview status fallback

### LINKED ISSUE #2422: [Bug] Schema Registry Overview card shows "Not configured" intermittently despite SR being properly configured
## Description

The Schema Registry status card on the Console Overview page intermittently shows **"Not configured"** even when Schema Registry is properly configured and fully functional (messages are being deserialised correctly using schemas from the registry).

On the same page load, refreshing causes it to alternate between showing the correct count (e.g. 963 schemas) and "Not configured". This is a pure UI reporting bug — the underlying Schema Registry connection works fine.

## Root Cause (suspected)

In `frontend/src/state/backend-api.ts`, the `getSchemaRegistryInfo` call appears to be gated behind a `api.userData?.canViewSchemas` permission check. When that flag evaluates to falsy on certain renders/refreshes, the request is skipped entirely, leaving `schemaRegistry` as `null`, which causes the Overview card to display "Not configured".

## Steps to Reproduce

1. Configure Redpanda Console with a valid Schema Registry URL (e.g. Confluent-compatible `kafka-cp-schema-registry`).
2. Navigate to the Console Overview page.
3. Observe the Schema Registry card — it may correctly show the schema count (e.g. "963 schemas").
4. Refresh the page.
5. The card now shows **"Not configured"**.
6. Refresh again — it may show the correct count once more.

The flicker is consistent and reproducible across page refreshes in the same session.

## Expected Behaviour

The Schema Registry card on the Overview page should consistently display the correct status (schema count and connectivity) whenever SR is configured, regardless of how many times the page is refreshed.

## Actual Behaviour

The card alternates between showing the correct schema count and "Not configured" across page refreshes, even though:
- Schema Registry is properly configured in Console's Helm values
- Messages in the topic viewer are being correctly deserialised using schemas from the registry
- The Schema Registry endpoint is reachable and healthy

## Environment

- **Redpanda Console version:** v3.7.1 (chart 3.7.0)
- **Schema Registry type:** Confluent-compatible (`kafka-cp-schema-registry`)
- **Deployment:** Kubernetes (Helm chart, ArgoCD managed)
- **Observation:** Also co-occurs with issue #2421 (Avro bytes rendering regression), but is an independent bug

## Additional Context

- The schema count of **963** is correct when the card does load — confirming the SR endpoint itself is healthy.
- The topic message viewer works correctly (Avro schemas are resolved), confirming the SR connection from Console is functional.
- This appears to be a race condition or permission-flag timing issue in the frontend state management, not a network/infrastructure problem.
- Related to (but independent from): #2421

peter941221:
## Summary

Fixes #2422.

The Overview page currently conflates three different Schema Registry states into the same `null` value:
- Schema Registry is not configured
- The overview fetch was skipped
- The overview fetch failed

That lets request failures or skipped requests render as `Not configured`, even when Schema Registry is configured and working.

This patch keeps the change frontend-only and narrows the `Not configured` fallback so it is only shown when the backend explicitly reports feature-not-configured semantics.

## What changed

- add `schemaRegistryError` to the overview state
- request schema registry info unless permissions are explicitly resolved to `canViewSchemas === false`
- treat `Code.Unimplemented` as `Not configured`
- keep other schema-registry errors as errors instead of collapsing them into `null`
- show `Unavailable` in the Overview card when the schema-registry overview request fails

This intentionally does not change the existing UX for users whose schema permissions are explicitly denied.

## Testing

- `vitest --config=vitest.config.unit.mts src/state/backend-api.test.ts`
- `vitest --config=vitest.config.integration.mts src/components/pages/overview/overview.test.tsx`

### DIFF
--- a/frontend/src/components/pages/overview/overview.test.tsx
+++ b/frontend/src/components/pages/overview/overview.test.tsx
@@ -1,0 +1,139 @@
+import { create } from '@bufbuild/protobuf';
+import { Code, ConnectError } from '@connectrpc/connect';
+import { screen } from '@testing-library/react';
+import {
+  ComponentStatusSchema,
+  GetSchemaRegistryInfoResponseSchema,
+  StatusType,
+} from 'protogen/redpanda/api/console/v1alpha1/cluster_status_pb';
+import type { ReactNode } from 'react';
+import { renderWithFileRoutes } from 'test-utils';
+
+vi.mock('../../../state/app-global', () => ({
+  appGlobal: {
+    onRefresh: null,
+    historyPush: vi.fn(),
+    historyReplace: vi.fn(),
+  },
+}));
+
+vi.mock('./cluster-health-overview', () => ({
+  default: () => <div data-testid="cluster-health-overview" />,
+}));
+
+vi.mock('./shadow-link-overview-card', () => ({
+  ShadowLinkSection: () => <div data-testid="shadow-link-section" />,
+}));
+
+vi.mock('../../builder-io/nurture-panel', () => ({
+  default: () => null,
+}));
+
+vi.mock('../../license/overview-license-notification', () => ({
+  OverviewLicenseNotification: () => null,
+}));
+
+vi.mock('../../misc/null-fallback-boundary', () => ({
+  NullFallbackBoundary: ({ children }: { children?: ReactNode }) => <>{children}</>,
+}));
+
+import Overview from './overview';
+import { api, useApiStore } from '../../../state/backend-api';
+import type { ClusterOverview } from '../../../state/rest-interfaces';
+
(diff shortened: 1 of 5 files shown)


peter941221:
Small follow-up on this one: the PR is still intentionally limited to the schema-registry overview fallback path and the regression tests around that path. If you still want this fix, I can keep it as-is or rebase it on current master with no scope expansion. If the team would prefer to drop it, please say so and I will stop pushing on it.

peter941221:
Small follow-up: this is still the same narrow schema-registry overview fallback fix with the matching regression coverage. If the team still wants it, I can keep the PR scoped as-is or rebase it on current master without expanding the diff.

PR #2433, opened 2026-05-06
feat(agents): add Microsoft Teams bridge integration

birdayz:
## What

Add Microsoft Teams integration to the AI Agent system. New proto message, generated code, and an Integrations tab on the agent detail page.

## Why

Users need to connect AI agents to Microsoft Teams for conversational interactions through a Teams bot. This requires configuring Azure Bot credentials on the agent side so the Teams bridge knows how to authenticate.

### Proto

New `AIAgentTeamsBridge` message with four fields: `enabled`, `bot_app_id`, `bot_tenant_id`, `bot_app_secret_ref`. The secret ref uses the same `${secrets.NAME}` template pattern as the existing provider API keys.

Added as `teams_bridge` to `AIAgent` (field 18), `AIAgentCreate` (field 15), and `AIAgentUpdate` (field 15).

### UI

New **Integrations** tab on the agent detail page, positioned between Configuration and A2A. Contains a Microsoft Teams section with:

- Enable/disable toggle
- Application (client) ID and Tenant ID text fields
- Client Secret via the existing SecretSelector component (scoped to `AI_AGENT`)
- Read-only messaging endpoint URL (derived from agent URL), shown only when the integration is enabled and saved -- for pasting into Azure Bot registration

Update flow uses field masks targeting only `teams_bridge.*` fields, so it won't interfere with other agent configuration.

## References

CRD struct: `AIAgentTeamsBridge` in cloudv2


malinskibeniamin (approved):
Seems a bit specific to put teams bridge in AI agent service but otherwise LGTM

birdayz:
> Seems a bit specific to put teams bridge in AI agent service but otherwise LGTM

yeah that's fair. as per the design, it is a per agent configuration. Other systems, e.g. copilot studio do it the same way

birdayz:
@claude review plz

birdayz:
@claude re review

PR #2435, opened 2026-05-08
e2e tests variant for Kafka

jvorcak:
Introducing a setup for kafka so that we can e2e kafka specific changes

### DIFF
--- a/frontend/tests/shared/global-setup.mjs
+++ b/frontend/tests/shared/global-setup.mjs
@@ -1,6 +1,6 @@
 import { GenericContainer, Network, Wait } from 'testcontainers';
 
-import { KAFKA_CONNECT_IMAGE, OWL_SHOP_IMAGE, REDPANDA_IMAGE } from './test-images.mjs';
+import { KAFKA_CONNECT_IMAGE, KAFKA_IMAGE, OWL_SHOP_IMAGE, REDPANDA_IMAGE } from './test-images.mjs';
 import { exec } from 'node:child_process';
 import { existsSync, readFileSync, realpathSync, writeFileSync } from 'node:fs';
 import { dirname, join, resolve } from 'node:path';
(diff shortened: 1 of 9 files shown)


frontend/tests/shared/global-setup.mjs
```diff
@@ -156,6 +156,39 @@ async function verifyRedpandaServices(state, ports) {
...
+      KAFKA_LISTENERS: 'SASL_PLAINTEXT://:9092,CONTROLLER://:9093',
+      KAFKA_ADVERTISED_LISTENERS: 'SASL_PLAINTEXT://kafka:9092',
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: 'SASL_PLAINTEXT:SASL_PLAINTEXT,CONTROLLER:PLAINTEXT',
+      KAFKA_CONTROLLER_LISTENER_NAMES: 'CONTROLLER',
+      KAFKA_INTER_BROKER_LISTENER_NAME: 'SASL_PLAINTEXT',
+      KAFKA_SASL_ENABLED_MECHANISMS: 'PLAIN',
+      KAFKA_SASL_MECHANISM_INTER_BROKER_PROTOCOL: 'PLAIN',
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: '1',
+      KAFKA_AUTO_CREATE_TOPICS_ENABLE: 'true',
+      CLUSTER_ID: 'MkU3OEVBNTcwNTJENDM2Qk',
```
malinskibeniamin, line 180: is this a long lived cluster? or is it just part of the original setup

frontend/tests/shared/global-teardown.mjs
```diff
@@ -77,6 +77,12 @@ export default async function globalTeardown(config = {}) {
       });
     }
 
+    if (state.kafkaId) {
+      console.log('Stopping Kafka container...');
+      await execAsync(`docker stop ${state.kafkaId}`).catch(() => {});
+      await execAsync(`docker rm ${state.kafkaId}`).catch(() => {});
+    }
```
malinskibeniamin, line 84: Should we also `docker container prune`

frontend/tests/test-variant-kafka/config/console.config.yaml
```diff
@@ -0,0 +1,20 @@
...
+schemaRegistry:
+  enabled: false
+
+redpanda:
+  adminApi:
+    enabled: false
+
+server:
+  listenPort: 3000
+  allowedOrigins: ["http://localhost:3000", "http://localhost:3001"]
```
malinskibeniamin: should we enable 3004 too?

frontend/tests/test-variant-kafka/playwright.config.ts
```diff
@@ -0,0 +1,63 @@
...
+    needsShadowlink: false,
+  },
+
+  use: {
+    navigationTimeout: 30 * 1000,
+    actionTimeout: 30 * 1000,
+    viewport: { width: 1920, height: 1080 },
+    headless: !!process.env.CI,
+
+    baseURL: process.env.REACT_APP_ORIGIN ?? 'http://localhost:3002',
```
malinskibeniamin, line 45: 3002 is reserved for the backend, right?

frontend/tests/test-variant-kafka/playwright.config.ts
```diff
@@ -0,0 +1,63 @@
...
+    timeout: 60 * 1000,
+  },
+
+  testMatch: '**/*.spec.ts',
+
+  fullyParallel: !!process.env.CI,
+
+  forbidOnly: !!process.env.CI,
+
+  retries: process.env.CI ? 2 : 0,
```
malinskibeniamin: maybe only 1 retry in CI? ideally we would have 0 but we need to get there 1st

frontend/tests/test-variant-kafka/playwright.config.ts
```diff
@@ -0,0 +1,63 @@
+import { defineConfig, devices } from '@playwright/test';
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+const reporters = process.env.CI
+  ? [['github' as const], ['html' as const, { outputFolder: 'playwright-report' }]]
+  : [['list' as const], ['html' as const, { outputFolder: 'playwright-report' }]];
```
malinskibeniamin: I think in local mode you could consider markdown reporter which is more LLM friendly in terms of token usage

malinskibeniamin (approved):

frontend/tests/shared/global-setup.mjs
jvorcak: This is ephemeral — the CLUSTER_ID is a required KRaft parameter for Kafka to initialize its storage on first boot. The container is created fresh for each test run and stopped/removed in teardown. No state persists between runs.

frontend/tests/shared/global-teardown.mjs
jvorcak: Intentionally avoided — `docker container prune` would remove **all** stopped containers on the developer's machine, not just test-related ones. We track each container ID in state and stop/remove only the containers we started.

frontend/tests/test-variant-kafka/playwright.config.ts
jvorcak: Yes, 3002 is the backend port for this variant (defined in `config/variant.json`). Since `serveFrontend: true`, the backend serves both the API and frontend assets from the same port, so the playwright `baseURL` points directly to it.

frontend/tests/test-variant-kafka/config/console.config.yaml
jvorcak: Port 3004 is not used in this variant. I added `http://localhost:3002` (the actual external Docker port for the kafka variant backend) to `allowedOrigins` — that was the missing one. The existing 3000/3001 cover the backend's internal container port and the frontend dev server.

frontend/tests/test-variant-kafka/playwright.config.ts
jvorcak: Reduced to 1 in CI. Agreed on aiming for 0 eventually.

frontend/tests/test-variant-kafka/playwright.config.ts
jvorcak: Switched to `markdown` reporter for local mode.

jvorcak:
## Review feedback addressed

- **Retries in CI**: Reduced from 2 → 1 (`playwright.config.ts`). Agreed on aiming for 0 long-term.
- **Local reporter**: Switched from `list` to `markdown` — more LLM-friendly output.
- **`allowedOrigins`**: Added `http://localhost:3002` (the actual external Docker port for this variant). Port 3004 is not used anywhere in the kafka variant.
- **`CLUSTER_ID`**: Clarified — this is an ephemeral KRaft init parameter, not a long-lived cluster. The container is created fresh per test run and cleaned up in teardown.
- **`docker container prune`**: Deliberately avoided — it would remove all stopped containers on the machine, not just test-related ones. We stop/remove only containers tracked by ID in state.
- **3002 as `baseURL`**: Confirmed — 3002 is the backend port for this variant (defined in `config/variant.json`). With `serveFrontend: true`, the same backend serves both API and frontend assets.

All review threads resolved. CI is green (`test-browser` timed out on Playwright browser install — pre-existing infra issue, `continue-on-error: true`).

PR #2437, opened 2026-05-08
feat(frontend): tweak shadow link create Connection step

r-vasquez:
The TLS section used to surface an Upload/File-path mode picker and three cert fields the moment TLS was enabled, even though most users just want a plain encrypted link. The auth section also gave no clue that the SASL user has to live on the source cluster and carry specific ACLs.

The new layout collapses CA and mTLS into separate disclosures so the cert mode picker only appears when one is open, and adds an inline info callout above the auth credentials pointing at the required ACLs.

The cert upload modal is gone:
dropping a file now commits straight to form state through the registry Dropzone, with a trash button in the corner for removal. Toggle copy shifts from SCRAM to SASL since the underlying mechanisms are SCRAM-SHA-256 and SCRAM-SHA-512.

A small secret-reference helper was extracted so the SASL password and mTLS client key stop duplicating the secret template parsing.

#### **TLS:**
**Before:**
image

**After**
image
image

#### **SASL**
**Before**
image

**After**
image


c-julin (approved):
This looks good to me, this may be scope creep so we can do a seperate PR but SASL PLAIN for confluent shadowlink, I had a ticket to add it to cloud but dont know if the change was made for self hosted.

PR #2441, opened 2026-05-12
fix: fixed stale ACL check in add-user-step

eblairmckee:
[UX-832](https://redpandadata.atlassian.net/browse/UX-832)

Users who created a topic in `Console > Topics` and a user + ACLs in `Console > Security` saw the Connect pipeline wizard incorrectly insist the user lacked permissions for the topic. The wizard's permission check (`checkUserHasTopicReadWritePermissions`) was running against a stale `listACLs` cache, and a separate cardinality typo meant the topic combobox's `onOpen` invalidation never matched the legacy topic query.

Two stale-cache holes closed:

- **`add-topic-step.tsx`** — topic combobox `onOpen` was invalidating `listTopics` with `cardinality: 'finite'`, but `useLegacyListTopicsQuery` registers under `cardinality: 'infinite'`, so the invalidation hit nothing. Fixed the cardinality and also invalidate `listACLs` on open.
- **`add-user-step.tsx`** — user combobox `onOpen` now also invalidates `listACLs` so the topic permission alert reflects current ACLs the moment the user picks an existing user. Same fix applied to the consumer-group `Input` via `onFocus`, which drives the parallel `useListACLsQuery({ resourceType: GROUP })` check.

Same-tab edits already invalidate correctly via `useCreateACLMutation` / `useCreateAcls` / `useUpdateAclMutation` — they all target `listACLs` cardinality `'finite'`. The remaining hole was cross-tab edits (each tab has its own `queryClient`) and the broken topic cardinality, both of which this PR closes.

[UX-832]: https://redpandadata.atlassian.net/browse/UX-832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
+++ b/frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
@@ -21,6 +21,7 @@ import { ToggleGroup, ToggleGroupItem } from 'components/redpanda-ui/components/
 import { Heading } from 'components/redpanda-ui/components/typography';
 import { ChevronDown, XIcon } from 'lucide-react';
 import type { MotionProps } from 'motion/react';
+import { listACLs } from 'protogen/redpanda/api/dataplane/v1/acl-ACLService_connectquery';
 import { ListTopicsRequestSchema } from 'protogen/redpanda/api/dataplane/v1/topic_pb';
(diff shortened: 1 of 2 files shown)


SpicyPete (approved):
Code for this looks good, nice fix

PR #2447, opened 2026-05-13
UX-1270 & UX-1271 - space editor bug, tooltip fixes, and doubled Sonner removal

SpicyPete:
* UX-1270
* Downgrade Monaco to 0.54.x as monaco-yaml only works until this version
* Can enter spaces again
* Can close the find widget
* Small adjustment to tooltip parsing so we don't render "|| || || ||" tooltips
* UX-1271
* Removing the second, unstyled Sonner provider, so we don't get doubled toasts


eblairmckee:
nit: a video demo would be lovely to show that it works now. you can use zoom clips or quicktime, or slack video. whatever floats your boat

frontend/src/components/pages/topics/Tab.Messages/editor.tsx
```diff
@@ -25,7 +25,7 @@ const options: editor.IStandaloneEditorConstructionOptions = {
 };
 
 async function tryTranspile(
-  tsWorkerClient: typescript.TypeScriptWorker,
+  tsWorkerClient: languages.typescript.TypeScriptWorker,
```
eblairmckee, line 28: gonna guess destructuring was a cause of one of the bugs?

frontend/src/components/misc/pipelines-yaml-editor.tsx
```diff
@@ -15,6 +15,9 @@ import { configureMonacoYaml, type MonacoYaml, type MonacoYamlOptions } from 'mo
 import { type ComponentType, lazy, Suspense, useCallback, useEffect, useState } from 'react';
 
 import benthosSchema from '../../assets/rp-connect-schema.json' with { type: 'json' };
+import { annotateAnyOfTitles } from '../ui/yaml/yaml-editor';
+
+annotateAnyOfTitles(benthosSchema);
```
eblairmckee, line 20: so this is a mutation before instantiation of the schema? iirc

frontend/src/config.ts
```diff
@@ -457,27 +457,50 @@ export const setup = memoizeOne((setupArgs: SetConfigArguments) => {
     currentSetupTeardown = installUiStateSubscriptions();
   }, 50);
 
-  // Set MonacoEnvironment synchronously before loader.init() to avoid race
-  // where the editor mounts before the worker URL resolver is available
+  // Use `new Worker(new URL(...))` so rsbuild bundles each as its own chunk
+  // (needed for monaco-yaml's CJS/ESM interop).
```
eblairmckee, line 461: nice!

eblairmckee (approved):

frontend/src/components/pages/topics/Tab.Messages/editor.tsx
SpicyPete: It was mainly from the 0.55 upgrade version which is incompatible with the YAML plugin, which added the `typescript.*` top level type

PR #2449, opened 2026-05-14
Provides default feature flags for non-embedded mode

jvorcak:
Default values for non-embedded mode were ignored for now.


r-vasquez (approved):

SpicyPete (approved):
Looks good, then we can enable features behind FF's on standalones

PR #2450, opened 2026-05-15
fix(ui): keep sidebar user dropdown open

malinskibeniamin:
## What

Fix Base UI dropdown content rendering so the sidebar user menu stays open and usable. @SpicyPete

## Why

The visible dropdown content was nested inside the Base UI popup element, so focus and menu behavior were attached to a wrapper instead of the rendered menu surface.

## How

Render the visible dropdown surface directly through Base UI's `Popup.render`, preserving Base UI props, role, focus handling, and Radix-compatible state attributes on the same element.

## Acceptance Criteria

- Sidebar user menu remains open after activation.
- Menu actions are clickable without needing to keep focus pinned.
- Dropdown content keeps `data-state` styling compatibility.

## Testing Steps

- `bun run type:check`
- `bun run lint:check` *(reports existing repo-wide restricted-import/unused-suppression issues; exit 0)*
- `bun run test`

### DIFF
--- a/frontend/src/components/redpanda-ui/components/dropdown-menu.tsx
+++ b/frontend/src/components/redpanda-ui/components/dropdown-menu.tsx
@@ -179,9 +179,9 @@ function DropdownMenuSubContent({ className, ...props }: DropdownMenuSubContentP
   );
 }
 
-type DropdownMenuContentProps = Omit<React.ComponentProps<typeof DropdownMenuPrimitive.Popup>, 'className' | 'render'> &
-  React.HTMLAttributes<HTMLDivElement> &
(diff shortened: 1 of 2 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
@claude review

frontend/src/components/redpanda-ui/components/dropdown-menu.tsx
malinskibeniamin, line 1: This is a big change, maybe there is an easier way to get this resolved, basically when you open a user menu dropdown in the sidebar of console, it will sometimes not allow you to interact with the component because it keeps closing as soon as it loses focus

PR #2466, opened 2026-05-25
Show unconsumed partitions in consumer groups

jvorcak:
## Summary

  Previously, consumer group details only showed partitions that had a committed offset.
  This PR extends the backend and frontend so all partitions are visible — including those
  that have never been consumed or whose offset was deleted.

  **Backend**
  - `PartitionOffsets.GroupOffset` changed from `int64` to `*int64` — `nil` signals no committed offset
  - Partitions without a committed offset are now included in the response (previously skipped with `continue`)

  **Frontend**
  - `GroupPartitionOffset.groupOffset` typed as `number | null`
  - Unconsumed partitions show `—` for Group Offset and Lag columns
  - Edit / Delete buttons on unconsumed partition rows are disabled with tooltip "No committed offset"
  - Topic-level Edit / Delete buttons disabled when all partitions are unconsumed
  - `onEditOffsets` / `onDeleteOffsets` at topic level only receive consumed partitions, preventing crashes in downstream modal handlers
  - Lag filter (`onlyShowPartitionsWithLag`) updated to explicitly include unconsumed partitions rather than relying on `null !== 0`
  - `cannotEditGroupReason` / `cannotDeleteGroupOffsetsReason` accept an optional `consumedPartitions` array and own the "No committed offsets" check, removing inline ternaries at call sites

  **Tests**
  - Unit test (`group-details.test.tsx`) covers: edit button enabled/disabled state, `—` rendering for unconsumed partitions
  - E2E test (`consumer-group-unconsumed-partitions.spec.ts`) covers the full flow: produce messages, consume partially, delete an offset, verify all 3 partitions appear with correct values and button states

### DIFF
--- a/backend/pkg/console/consumer_group_offsets_test.go
+++ b/backend/pkg/console/consumer_group_offsets_test.go
@@ -184,6 +184,7 @@ func TestGetConsumerGroupOffsets_NonExistentTopicInMetadata(t *testing.T) {
 	req.NoError(err, "Should not error when one topic is deleted")
 
 	// Expected result: only the existing topic, with 1 partition and offset at position 1
+	groupOffset := int64(1)
 	expected := map[string][]GroupTopicOffsets{
 		groupID: {
 			{
@@ -194,7 +195,7 @@ func TestGetConsumerGroupOffsets_NonExistentTopicInMetadata(t *testing.T) {
 				PartitionOffsets: []PartitionOffsets{
 					{
 						PartitionID:   0,
-						GroupOffset:   1,
+						GroupOffset:   &groupOffset,
 						HighWaterMark: 1,
 						Lag:           0,
 					},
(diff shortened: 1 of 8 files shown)


backend/pkg/console/consumer_group_offsets_test.go
```diff
@@ -184,6 +184,7 @@ func TestGetConsumerGroupOffsets_NonExistentTopicInMetadata(t *testing.T) {
 	req.NoError(err, "Should not error when one topic is deleted")
 
 	// Expected result: only the existing topic, with 1 partition and offset at position 1
+	groupOffset := int64(1)
```
r-vasquez: Instead of 

```go
groupOffset := int64(1)
...
GroupOffset: &groupOffset,
```

With newer Go you can do just
```go
GroupOffset: new(int64(1)),
```

r-vasquez (approved):
LGTM, just one non-blocking minor nit. Also, if possible, let's add more context on why to the commit body or PR description please.

backend/pkg/console/consumer_group_offsets_test.go
jvorcak: Fixed

PR #2467, opened 2026-05-25
proto(sql): add Redpanda SQL dataplane API

r-vasquez:
Adds the dataplane API surface for the
in-Console SQL editor backed by Oxla over
the Postgres wire protocol. The service
covers catalog and table introspection
(ListCatalogs, ListDatabases, ListTables,
DescribeTable) and single-statement
execution (ExecuteQuery) with a
server-side row cap. Lands in v1alpha3
since the surface is expected to evolve
before GA.

Auth flows pass through the caller's
Cloud OIDC bearer to Oxla, which requires
adding API_OXLA to the shared API enum.


proto/redpanda/api/auth/v1/authorization.proto
```diff
@@ -26,6 +26,7 @@ enum API {
   API_AI_AGENT = 8;
   API_AI_GATEWAY = 9;
   API_PROMETHEUS = 10;
+  API_OXLA = 11;
```
c-julin: we shouldn't call it oxla

c-julin: redpanda sql

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+import "google/protobuf/duration.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "redpanda/api/auth/v1/authorization.proto";
+
+// SQLService provides catalog/table introspection and SQL execution
+// against the Redpanda SQL engine (Oxla) over the Postgres wire protocol.
+service SQLService {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_tag) = {
+    name: "SQL"
+    description: "Execute SQL against Redpanda SQL (Oxla) and browse its catalogs."
```
c-julin: Interact with Redpanda SQL by browsing catalogs, tables and executing queries

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_VIEW
+      api: API_OXLA
+    };
+  }
+
+  // ListDatabases lists databases known to the SQL engine.
+  rpc ListDatabases(ListDatabasesRequest) returns (ListDatabasesResponse) {
```
c-julin: Is this a thing in rpsql? i think its just catalogs and tables for now and in the future they will allow database creation

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+      responses: {
+        key: "200"
+        value: {
+          description: "OK"
+          schema: {json_schema: {ref: ".redpanda.api.dataplane.v1alpha3.ExecuteQueryResponse"}}
+        }
+      }
+    };
+    option (redpanda.api.auth.v1.authorization) = {
+      required_permission: PERMISSION_EDIT
```
c-julin: think query execution is view as well as we only allow selects.. but then create catalog is also allowed.

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+    };
+  }
+}
+
+// Catalog backing-storage type. Sourced from `system.catalogs.type` in
+// Oxla. See oxla/src/metastore/system_catalogs.cpp.
+enum CatalogType {
+  CATALOG_TYPE_UNSPECIFIED = 0;
+  CATALOG_TYPE_REDPANDA = 1;
+  CATALOG_TYPE_ICEBERG = 2;
```
c-julin: I think until rpsql release ICEBERG we dont expose it?

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+}
+
+message Catalog {
+  string name = 1;
+  string namespace_name = 2;
```
c-julin, line 127: whats namespace_name, i vaguely remember seeing this but cant put my finger on it

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+  string page_token = 2;
+}
+
+message ListDatabasesResponse {
+  repeated Database databases = 1;
+  string next_page_token = 2;
+}
+
+message ListTablesRequest {
+  string catalog = 1 [
```
c-julin, line 187: i think search/filter for name  here would be good if we can support `LIKE`

c-julin, line 187: do we namespace and name right (or is just name enough)? not too familiar with namespace only asking because i saw it above

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+    (buf.validate.field).string.max_len = 262144
+  ];
+  // Override the server row cap for this query. When unset, the server
+  // default applies.
+  optional int32 row_limit = 2 [(buf.validate.field).int32 = {
+    gte: 1
+    lte: 10000
+  }];
+  // Per-query timeout. When unset, the server default applies.
+  optional google.protobuf.Duration timeout = 3 [(buf.validate.field).duration = {
```
c-julin, line 248: nice

proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -0,0 +1,286 @@
...
+    lte: {seconds: 300}
+  }];
+}
+
+message ExecuteQueryResponse {
+  repeated Column columns = 1;
+  repeated Row rows = 2;
+  // True if the server row cap fired before the result set was
+  // exhausted.
+  bool truncated = 3;
```
c-julin, line 259: my preference here is what goland does and autopaginates so if server limit is 500 we do offset/limit pagination. maybe thats a follow up might not be performant on big datasets

proto/redpanda/api/dataplane/v1alpha3/sql.proto
r-vasquez: I believe right now is namespaced between public and private but Im not totally sure. I ran the query

```
show catalogs;

      name      | namespace_name |   type
----------------+----------------+----------
 redpanda_local | public         | redpanda
 ```

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto (hunk 7 above)
Comment (r-vasquez) [reply proto/redpanda/api/dataplane/v1alpha3/sql.proto:187]: I was thinking this on being the query `show tables from <catalog>` hence only the required `catalog`

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto (hunk 9 above)
Comment (r-vasquez) [reply proto/redpanda/api/dataplane/v1alpha3/sql.proto:259]: Can you expand on this a little bit? it is the idea to have a pagination mechanism (page_token, etc) and only paginate when?

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto (hunk 3 above)
Comment (r-vasquez) [reply proto/redpanda/api/dataplane/v1alpha3/sql.proto]: Removed this one

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto (hunk 4 above)
Comment (r-vasquez) [reply proto/redpanda/api/dataplane/v1alpha3/sql.proto]: Changed to VIEW 👍

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto (hunk 5 above)
Comment (r-vasquez) [reply proto/redpanda/api/dataplane/v1alpha3/sql.proto]: Done, we can always add it back later 👍

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto (hunk 7 above)
Comment (r-vasquez) [reply proto/redpanda/api/dataplane/v1alpha3/sql.proto:187]: Added the filter option but left this one required

File: proto/redpanda/api/dataplane/v1alpha3/sql.proto [hunk 10]
```diff
@@ -0,0 +1,260 @@
...
+      maximum: 1000
+    }
+  ];
+  string page_token = 3;
+  // Optional SQL LIKE pattern matched against table name (e.g.
+  // 'orders_%'). When unset, all tables in the catalog are returned.
+  optional string filter = 4 [
+    (buf.validate.field).string.max_len = 255,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Optional SQL LIKE pattern matched against table name (e.g. 'orders_%')."}
+  ];
```
r-vasquez, line 209: @c-julin Is this what you meant?

r-vasquez:
Addressed the first comments and generated the files in https://github.com/redpanda-data/console/compare/521844b3f3c32dfd112cdc5bb87c803a5b7398ae..02f30c60f192c23b76633f687809ffaa9d15822e

proto/redpanda/api/dataplane/v1alpha3/sql.proto
c-julin: exactly ! yes

proto/redpanda/api/dataplane/v1alpha3/sql.proto
c-julin: so in jetbrains datagrip probably other tools too, if you dont set a limit it will return only the first 500 results and gives you a next page token instead of truncating. 
This probably for the backend but we need to handle the user asking for like 100000 records or something, so maybe we parse the query?

c-julin (approved):

PR #2474, opened 2026-05-29
UX-1275 - RPCN template gallery

SpicyPete:
### Summary

Adds a template gallery to the `Redpanda Connect` pipeline creation flow. Instead of writing pipeline YAML from scratch, users can pick a pre-built, curated template (e.g. Postgres CDC to Redpanda), fill in a small guided form, and deploy — with the gallery generating valid Connect config behind the scenes.

* UX-1275
  * Connect Pipeline library
  * Common Connect pipelines configuration setup forms
  * Behind a FF `enableRpcnTemplateGallery`
  * Supported in both self-hosted and cloud versions of Connect Pipeline
* UX-1293
  *   Pipeline configs stuck in infinite option - fixed
* UX-1299
  * Adding many more service icons which were missing, placeholders for remaining missing ones
* UX-1303   
  * Fixed stale configs shown sometimes after saving or editing 
* UX-1305
  * Showing service icons in connect pipelines listing + pipeline visualization areas
*  UX-1307
   * Fixing Pipeline visualization lane from capturing page scroll entirely
* UX-1319
  * Replacing the "gear" with an "info" icon when viewing pipelines
  * Surface some more pipeline information in the information panel (url, id, description)  
* Misc
  * Updating many UI Registry components to latest versions 
  * Various fixes across the Connect Dialogs to be more elegant and consistent in spacing
  * Simplifying the logs table visuals (more in a follow up PR)


frontend/src/assets/connectors/logos/iceberg-logo.tsx
```diff
@@ -0,0 +1,16 @@
...
+// Wraps the bundled raster mark with the same prop shape as the inline-SVG
+// logos so it drops into componentLogoMap unchanged. Swap for a real SVG once
+// the official brand-kit vector is available.
+export const IcebergLogo = ({ className, style, ...props }: Omit<React.SVGProps<SVGSVGElement>, 'children'>) => (
+  <img
+    alt="Apache Iceberg"
+    className={className}
+    src={IcebergMark}
+    style={{ objectFit: 'contain', ...(style as React.CSSProperties) }}
+    {...(props as unknown as React.ImgHTMLAttributes<HTMLImageElement>)}
```
eblairmckee: nit: why not extend `React.ImgHTMLAttributes<HTMLImageElement>` when typing the props?

frontend/src/components/debug-helper/connect-config-fixtures.ts
```diff
@@ -0,0 +1,770 @@
...
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+// Pipeline fixtures use the modern `redpanda` input/output component
+// (recommended over the older `kafka_franz` / `kafka` aliases) and follow
+// the Cloud Connect quickstart credentials pattern: `${REDPANDA_BROKERS}`
+// for the broker contextual variable and `${secrets.X}` for SASL/TLS material.
+// See: https://docs.redpanda.com/redpanda-cloud/develop/connect/connect-quickstart/
```
eblairmckee, line 16: nit: does this need to live in the codebase? rather I'd expect this to live in an rfc or doc. 
unless keeping this would be helpful to future contributors to this file.. omit?

frontend/src/components/debug-helper/connect-config-fixtures.ts
```diff
@@ -0,0 +1,770 @@
...
+    tags: ['invalid'],
+  },
+  {
+    id: 'invalid-cloud-disabled-stdio',
+    name: 'Invalid — stdin/stdout (Cloud-disabled)',
+    description: 'Triggers "unable to infer component type" — components disabled in Cloud.',
+    yaml: stdinStdoutDisabled,
+    tags: ['invalid'],
+  },
+];
```
eblairmckee, line 770: I do worry a bit about introducing all of this to the console codebase, the maintenance burden every time connect releases a breaking change will be a bit of a nuisance. wondering if we should instead move this to a long lived feature branch that anyone can use to enjoy debug functionality?

frontend/src/components/debug-helper/debug-dialog.tsx
```diff
@@ -0,0 +1,723 @@
...
+import { useCallback, useMemo, useState } from 'react';
+import { toast } from 'sonner';
+
+import { CONNECT_CONFIG_FIXTURES, type ConnectConfigFixture } from './connect-config-fixtures';
+import {
+  clearOverrides as clearFlagOverrides,
+  getAllFlagKeys,
+  getEffectiveFlags,
+  getOverrides as getFlagOverrides,
+  setOverride as setFlagOverride,
```
eblairmckee, line 66: nit: why?

frontend/src/components/debug-helper/debug-dialog.tsx
```diff
@@ -0,0 +1,723 @@
...
+
+const TAG_VARIANT: Record<ConnectConfigFixture['tags'][number], React.ComponentProps<typeof Badge>['variant']> = {
+  simple: 'success-inverted',
+  medium: 'info-inverted',
+  complex: 'warning-inverted',
+  'edge-case': 'accent-inverted',
+  invalid: 'destructive-inverted',
+};
+
+function copyToClipboard(text: string, successMsg = 'Copied to clipboard') {
```
eblairmckee, line 82: can't believe this doesn't exist as a util in the registry... we have 18 instances of `navigator.clipboard`, i doubt behavior is consistent across them. worth adding a ticket to our registry backlog to add a util for this

frontend/src/components/debug-helper/debug-dialog.tsx
```diff
@@ -0,0 +1,723 @@
...
+    .then(() => toast.success(successMsg))
+    .catch(() => toast.error('Failed to copy to clipboard'));
+}
+
+function ConfigFixtureRow({ fixture }: { fixture: ConnectConfigFixture }) {
+  const [previewing, setPreviewing] = useState(false);
+  const lineCount = useMemo(() => fixture.yaml.split('\n').length, [fixture.yaml]);
+
+  return (
+    <div className="rounded-md border bg-card px-3 py-2.5">
```
eblairmckee: nit: `Card`

frontend/src/components/debug-helper/debug-dialog.tsx
```diff
@@ -0,0 +1,723 @@
...
+        <div className="flex min-w-0 flex-1 flex-col gap-1">
+          <div className="flex flex-wrap items-center gap-1.5">
+            <Text className="font-medium" variant="bodyStrongSmall">
+              {fixture.name}
+            </Text>
+            {fixture.tags.map((tag) => (
+              <Badge key={tag} size="sm" variant={TAG_VARIANT[tag]}>
+                {tag}
+              </Badge>
+            ))}
```
eblairmckee, line 162: nit: can use `BadgeGroup` to handle overflow

frontend/src/components/debug-helper/debug-dialog.tsx
```diff
@@ -0,0 +1,723 @@
...
+            variant="primary"
+          >
+            Copy YAML
+          </Button>
+        </div>
+      </div>
+      {previewing && (
+        <pre className="mt-2.5 max-h-64 overflow-auto rounded border bg-muted p-2.5 font-mono text-xs">
+          {fixture.yaml}
+        </pre>
```
eblairmckee: nit: we have multiple code block components that do a nice job of adding syntax highlighting to yaml

SpicyPete:
Moved the Dev Debugger out of this branch, and will open it in a separate PR for further discussion on if it's something we want in the project, or if we should handle this another way.

frontend/src/components/debug-helper/connect-config-fixtures.ts
SpicyPete: Fair, moved this tool into its own branch until the discussions around how we want to include this happens.

https://github.com/redpanda-data/console/tree/pz/dev-debugger-dialog-tool

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -264,172 +263,205 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
-                        />
-                        {!!error && (
-                          <FieldError id={`secret-${normalizedSecretName}-error`}>{error.message}</FieldError>
-                        )}
-                      </Field>
-                    );
-                  })}
+  const requiredSecretsForm = (
+    <>
+      <Alert variant="warning">
```
eblairmckee: nit: destructive would be better since it's technically breaking if you don't resolve

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -264,172 +263,205 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
+                {normalizedSecretName}
+              </FieldLabel>
+              <Input
+                id={valueId}
+                placeholder="Enter value"
+                type="password"
+                {...form.register(fieldName)}
+                aria-describedby={error ? `${valueId}-error` : undefined}
+                aria-invalid={!!error}
+              />
```
eblairmckee, line 273: minor: any chance we can render a disabled input for the key with missing secret. aka prepopulated input for the secret key and another input below for the value

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -264,172 +263,205 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
+              {!!error && <FieldError id={`${valueId}-error`}>{error.message}</FieldError>}
+            </Field>
+          );
+        })}
+
+        <Button className="w-full" disabled={form.formState.isSubmitting} type="submit" variant="primary">
+          {form.formState.isSubmitting ? (
+            <div className="flex items-center gap-2">
+              <Loader2 className="h-4 w-4 animate-spin" />
```
eblairmckee: I believe we have a spinner.tsx component for this

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -264,172 +263,205 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
+  const newSecretFormBody = (
+    <>
+      {/* Display newly created secrets */}
+      {newlyCreatedSecrets.length > 0 && (
+        <div className="space-y-2">
+          <Text className="font-medium text-muted-foreground text-sm">Created secrets</Text>
+          <div className="flex flex-col gap-1.5">
+            {newlyCreatedSecrets.map((secretName) => (
+              <div
+                className="flex items-center gap-2 rounded-md border border-green-600/30 bg-green-600/5 px-3 py-2"
```
eblairmckee: use Alert variant='success'

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -264,172 +263,205 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
+
+          <Field data-invalid={!!newSecretForm.formState.errors.value}>
+            <FieldLabel className="font-medium text-sm" htmlFor="new-secret-value">
+              Secret value
+            </FieldLabel>
+            <Input
+              id="new-secret-value"
+              placeholder="Enter secret value..."
+              type="password"
+              {...newSecretForm.register('value')}
```
eblairmckee, line 350: that's an interesting pattern.. is this so it can watch the value?

frontend/src/components/ui/secret/quick-add-secrets.tsx
```diff
@@ -264,172 +263,205 @@ export const QuickAddSecrets: React.FC<QuickAddSecretsProps> = ({
...
+        ))}
+      {enableNewSecrets &&
+        (inline ? (
+          showAddAnotherForm ? (
+            <div className={hasRequiredSection ? 'space-y-3 border-t pt-4' : 'space-y-4'}>
+              {hasRequiredSection && (
+                <div className="flex items-center justify-between gap-2">
+                  <Text className="font-medium" variant="bodyStrongMedium">
+                    Add a custom secret
+                  </Text>
```
eblairmckee: we have heading variants for this

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -353,9 +363,13 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
 
   return (
     <div className="flex min-h-0 flex-col gap-4">
-      {/* Toolbar */}
       <div className="flex items-center justify-between gap-2">
-        <div className="flex items-center gap-2">
+        <div className="flex items-center gap-3">
+          {title ? (
+            <Heading className="font-semibold text-base text-foreground" level={3}>
```
eblairmckee: can we not add classnames to headings? this sounds like we need to rethink our heading variants if it's not meeting our needs wrt hierarchy

frontend/src/components/ui/connect/log-explorer.tsx
```diff
@@ -528,12 +540,14 @@ export function LogExplorer({ pipeline, serverless, enableLiveView = false }: Lo
         </div>
       </div>
 
-      {/* Pagination (client-side only) */}
+      {/* Hide DataTablePagination's "X of N row(s) selected." — this table
+          doesn't expose row selection — while keeping its layout slot. */}
       {filteredRowCount > 0 && (
-        <DataTablePagination table={table} />
+        <div className="[&>div>div:first-child]:invisible">
```
eblairmckee, line 538: we really should build this into the pagination component. i am running into so many issues with it rn, long-term should redo it

frontend/src/components/redpanda-ui/style/theme.css
```diff
@@ -366,22 +366,18 @@
...
   --color-info-subtle: var(--color-blue-alpha-100);
   --color-neutral: var(--color-grey-600);
   --color-neutral-foreground: var(--color-grey-50);
 
-  /* Informative (alias for info - same bright sky-blue base) */
-  --color-informative: var(--color-blue-600);
-  --color-informative-foreground: var(--color-blue-900);
+  /* Informative (alias for info) */
+  --color-informative: var(--color-blue-800);
+  --color-informative-foreground: var(--color-grey-white);
```
eblairmckee, line 380: blocker: we should upstream these before we make the change here..

eblairmckee: or was this due to updating ui-registry?

eblairmckee: i see. lots of changes. in the future, can we make ui-registry updates separate from feature work? it makes it very hard to detect accidental registry changes/changes that are intentional and need to be upstreamed later.

SpicyPete: Yes, this one is due to registry updates.  Agreed I think in general we should do that, however in this PR lots of the changes were driven from changes in here (like all the Dialog adjustments).

In future PR's I'll separate the registry updates into their own smaller PR's

frontend/src/components/ui/secret/quick-add-secrets.tsx
SpicyPete: I explored that pattern a bit, but it was even harder to understand.  I'll do another pass on this UI and improve it a bit more.

frontend/src/components/ui/secret/quick-add-secrets.tsx
SpicyPete: I think this is the standard way to use react-hook-form values when the values are not controlled directly in the form.  It should keep the values in sync with the hook and validation.

frontend/src/components/redpanda-ui/style/theme.css
SpicyPete: Opened this PR with all the registry related updated.  Should make this ones diffs much smaller after it merges
https://github.com/redpanda-data/console/pull/2477

frontend/src/assets/connectors/logos/iceberg-logo.tsx
SpicyPete: Keeping `SVGProps` allows the TS interface on this logo to match all of the other logos.  
I narrowed the the scope down so it doesn't ...spread anymore potentially invalid props.

frontend/src/assets/connectors/logos/iceberg-logo.tsx
```diff
@@ -0,0 +1,15 @@
...
+// logos so it drops into componentLogoMap unchanged. Swap for a real SVG once
+// the official brand-kit vector is available.
+export const IcebergLogo = ({ className, style }: Pick<SVGProps<SVGSVGElement>, 'className' | 'style'>) => (
+  <img
+    alt="Apache Iceberg"
+    className={className}
+    src={IcebergMark}
+    style={{ objectFit: 'contain', ...(style as CSSProperties) }}
+  />
+);
```
malinskibeniamin, line 15: Most of the time we use SVGs instead of PNGs

frontend/src/components/redpanda-ui/components/copy-button.tsx
malinskibeniamin, line 1: something to upstream to UI registry?

frontend/src/components/redpanda-ui/components/badge.tsx
malinskibeniamin, line 1: something to upstream to UI registry?

frontend/src/assets/connectors/logos/iceberg-logo.tsx
SpicyPete: Yes, but we have a PNG for Iceberg (already in our repo), and they don't provide an SVG to use

frontend/src/components/redpanda-ui/components/copy-button.tsx
SpicyPete: This is from the latest registry already, this is coming downstream from registry

frontend/src/components/redpanda-ui/components/badge.tsx
SpicyPete: Same with this, this is coming downstream from registry

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
```diff
@@ -285,6 +289,144 @@ export const AddTopicStep = forwardRef<BaseStepRef<AddTopicFormData>, AddTopicSt
...
+              >
+                <motion.span
+                  animate={{ rotate: showAdvancedSettings ? 180 : 0 }}
+                  className="flex h-4 w-4 items-center justify-center"
+                  transition={{ duration: 0.18, ease: [0.4, 0, 0.2, 1] }}
+                >
+                  <ChevronDown className="h-4 w-4" />
+                </motion.span>
+                Show advanced settings
+              </Button>
```
eblairmckee, line 404: nit: we have a Collapsible component in the registry. Accordion could also suffice. I also vaguely remember making a collapsible section molecule for this? but that also could've been something I deleted....

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
```diff
@@ -285,6 +289,144 @@ export const AddTopicStep = forwardRef<BaseStepRef<AddTopicFormData>, AddTopicSt
...
+              {showAdvancedSettings ? (
+                <div className="space-y-6 pt-4" id="add-topic-advanced-settings">
+                  <AdvancedTopicSettings
+                    disabled={isPending}
+                    form={form}
+                    isExistingTopic={Boolean(existingTopicSelected)}
+                  />
+                </div>
+              ) : null}
+            </div>
```
eblairmckee, line 419: nit: this really should be a molecule, can we lift this up to be something more reusable?

frontend/src/components/pages/rp-connect/onboarding/connect-tiles.tsx
```diff
@@ -376,7 +376,7 @@ export const ConnectTiles = memo(
 
               <div className="relative">
                 <div
-                  className={cn('max-h-[50vh] min-h-[400px] overflow-y-auto py-4', tileWrapperClassName)}
+                  className={cn('max-h-[50vh] min-h-100 overflow-y-auto py-4', tileWrapperClassName)}
```
eblairmckee, line 377: nice catch

frontend/src/components/pages/rp-connect/onboarding/connector-logo.tsx
```diff
@@ -1,14 +1,28 @@
...
+  fallback?: LogoComponent;
 }) => {
-  const Component = componentLogoMap[name];
-  return Component ? <Component className={className} style={style} /> : null;
+  const Component = (componentLogoMap[name] as LogoComponent | undefined) ?? fallback;
+  if (!Component) {
+    return null;
+  }
+  // Sized wrapper + inner SVG that fills it. Lets `preserveAspectRatio` (SVG
+  // default) letterbox taller-than-wide logos so they can't blow out row height.
```
eblairmckee, line 22: ❤️

frontend/src/components/pages/rp-connect/pipeline/config-dialog.tsx
```diff
@@ -238,7 +236,7 @@ export function ConfigDialog({ open, onOpenChange, form, mode }: ConfigDialogPro
         {mode !== 'view' && (
           <DialogFooter>
             <Button onClick={onSaveClick} variant="primary">
-              Save
+              Save settings
```
eblairmckee, line 239: nit: would love to have docs weigh in here. Michele is a great resource to sanity check this. iirc we do have a rule for these, and "Save" ends up being the outcome

frontend/src/components/pages/rp-connect/pipeline/details-dialog.tsx
```diff
@@ -33,35 +32,117 @@ type DetailsDialogProps = {
...
+  onRequestDelete?: () => void;
 };
 
-const DetailRow = ({
-  label,
-  value,
-  children,
-  copyable = false,
-}: {
+const EMPTY_PLACEHOLDER = '—';
```
eblairmckee, line 40: I really hate that claude keeps doing this... no action item. Just needed to complain for a moment.

frontend/src/components/pages/rp-connect/pipeline/details-dialog.tsx
```diff
@@ -33,35 +32,117 @@ type DetailsDialogProps = {
...
+
+const SectionHeading = ({ className, children }: { className?: string; children: React.ReactNode }) => (
+  <h3 className={cn('font-semibold text-base text-foreground', className)}>{children}</h3>
+);
```
eblairmckee: this bugs me, in later PR I'd love to do a pass and make sure we're using ui-registry typography components... no one-offs like this. we have `Heading level={3}` for a reason

frontend/src/components/pages/rp-connect/pipeline/details-dialog.tsx
```diff
@@ -33,35 +32,117 @@ type DetailsDialogProps = {
...
+
+// Copy button that only appears on row hover.
+const HoverCopyButton = ({ value }: { value: string }) => (
+  <CopyButton
+    className="shrink-0 opacity-0 transition-opacity group-hover/row:opacity-100"
+    content={value}
+    size="sm"
+    variant="ghost"
+  />
+);
```
eblairmckee: nit: if we only use this once, no need to wrap and make it its own component

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -115,37 +116,26 @@ import {
...
-  }
   if (type) {
     return `Search ${type}s...`;
   }
   return;
 }
 
-// ---------------------------------------------------------------------------
-// Schema + types
-// ---------------------------------------------------------------------------
```
eblairmckee, line 147: thank you

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -274,13 +256,17 @@ function usePipelineSave({
...
+  onBeforeSaveNavigate,
 }: {
   form: UseFormReturn<PipelineFormValues>;
   yamlContent: string;
   mode: string;
   pipelineId: string | undefined;
   pipeline: Pipeline | undefined;
   isPipelineDiagramsEnabled: boolean;
+  // Called right before a successful save navigates away, so the unsaved-changes
+  // guard doesn't block the post-save navigation.
```
eblairmckee: nit: if you use jsdoc comments it supports multi-line comments a lot cleaner

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -300,6 +286,11 @@ function usePipelineSave({
   const handleSave = useCallback(async () => {
     const isValid = await form.trigger();
     if (!isValid) {
+      // The editable settings (name especially) live in the header/dialog, so a
+      // silent no-op is confusing — surface why the save was blocked.
+      const fieldErrors = form.formState.errors;
+      const firstError = fieldErrors.name?.message ?? fieldErrors.computeUnits?.message ?? fieldErrors.tags?.message;
+      toast.error(typeof firstError === 'string' ? firstError : 'Fix the highlighted pipeline settings before saving.');
```
eblairmckee, line 295: major: not going to block for this, but please address in a fast follow. this is an anti-pattern. 
I like that we're trying to surface editable settings errors, but I think we can do this more elegantly. we exclusively use toasts for success (as they are ephemeral) and Alert for errors, so user can reference them and act on them accordingly. 
we have a few utilities for capturing and prettifying error messages, ConnectError is one we use pretty consistently in this feature. I would recommend sticking with existing patterns and showing _all_ form errors this way (not just the "firstError" 🙃 )

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -600,17 +609,19 @@ function SidebarPanel({
   onAddTopic: (section: string, componentName: string) => void;
   onAddSasl: (section: string, componentName: string) => void;
   onOpenCommandMenu: (filter?: 'all' | 'variables' | 'secrets' | 'topics' | 'users') => void;
+  onBrowseTemplates?: () => void;
 }) {
   return (
-    <div className="flex w-[300px] shrink-0 flex-col border-border! border-r">
-      <div className="min-h-0 flex-1">
+    <div className="flex w-[300px] shrink-0 flex-col overflow-hidden border-border! border-r">
```
eblairmckee, line 628: I've seen this a few times now in this PR... any reason why you're banging classNames as a suffix rather than as the prefix? tailwind docs say `!className` is the correct syntax, curious if this works at all.

eblairmckee: also curious if all these `overflow-hidden` classNames are adding any value to either `div`? for one, I know the pipeline flow diagram has x-axis overflow disabled (can't pan) --which makes me wonder if you were seeing scrollbars?

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -600,17 +609,19 @@ function SidebarPanel({
...
+      <div className="min-h-0 flex-1 overflow-hidden">
         {isPipelineDiagramsEnabled ? (
           <PipelineFlowDiagram
             configYaml={yamlContent}
             hideZoomControls
             onAddConnector={mode !== 'view' ? (type) => onAddConnector(type as ConnectComponentType) : undefined}
             onAddSasl={mode !== 'view' ? onAddSasl : undefined}
             onAddTopic={mode !== 'view' ? onAddTopic : undefined}
+            onBrowseTemplates={mode !== 'view' ? onBrowseTemplates : undefined}
```
eblairmckee: nit: typically when I have this many props conditionally passed on the same predicate I lift them into a memoized constant like `viewModeProps` and spread them in.

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -736,16 +744,39 @@ export default function PipelinePage() {
...
 
-  // --- YAML change handlers ---
+  // Guard against losing unsaved edits when navigating away from the editor (edit or create).
+  const yamlDirty = initialYamlRef.current !== null && yamlContent !== initialYamlRef.current;
+  const hasUnsavedChanges = mode !== 'view' && (form.formState.isDirty || yamlDirty);
+  const blocker = useBlocker({
+    shouldBlockFn: () => hasUnsavedChanges && !allowNavigationRef.current,
+    enableBeforeUnload: () => hasUnsavedChanges,
+    withResolver: true,
+  });
```
eblairmckee, line 781: nit: boolean constants should be either past tense or prefixed with `is`, eg: `blocked` or `isBlocked` (I prefer the former fwiw)

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -811,69 +841,90 @@ export default function PipelinePage() {
         name: pipeline.displayName,
         description: pipeline.description || '',
         computeUnits: cpuToTasks(pipeline.resources?.cpuShares) || MIN_TASKS,
-        tags: Object.entries(pipeline.tags)
-          .filter(([k]) => !isSystemTag(k))
-          .map(([key, value]) => ({ key, value })),
+        tags: getUserTagEntries(pipeline.tags),
```
eblairmckee, line 850: super tiny baby nit: I feel like I rewrite this all the time... can we make a helper like `getKeyValuePairs` that takes an optional predicate to check against

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -811,69 +841,90 @@ export default function PipelinePage() {
...
   });
 
-  // --- Navigation ---
-
   const handleCancel = useCallback(() => {
     if (mode === 'create') {
       clearWizardStore();
     }
+    // Edit: go to the pipeline's view page. Routed through `navigate` (not
+    // history.back) so the unsaved-changes blocker reliably intercepts it.
```
eblairmckee: nit... comments are unnecessary, code is very readable to deduce this

eblairmckee: I have a skill in my user-level claude preferences that gates against low quality comments are write time. the first gate is "does the code speak for itself? if not, maybe it could use a better name. otherwise... leave a comment"

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -811,69 +841,90 @@ export default function PipelinePage() {
...
+      {mode === 'view' && pipeline ? (
+        <PipelineViewHeader
+          onBack={handleCancel}
+          onViewDetails={() => setIsViewConfigDialogOpen(true)}
+          pipeline={pipeline}
+        />
+      ) : null}
+      {mode === 'view' && !pipeline ? (
+        <div className="flex items-center gap-2">
+          <Button className="-ml-3.5 shrink-0" onClick={handleCancel} size="icon" variant="ghost">
```
eblairmckee, line 904: thank you!

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -1019,19 +1106,37 @@ export default function PipelinePage() {
...
-              ? (['cache', 'rate_limit', 'buffer', 'scanner', 'tracer', 'metrics'] satisfies ConnectComponentType[])
-              : (addConnectorType ?? undefined)
-          }
-          isOpen={addConnectorType !== null}
-          onAddConnector={handleConnectorSelected}
-          onCloseAddConnector={() => setAddConnectorType(null)}
-          searchPlaceholder={getConnectorDialogPlaceholder(addConnectorType)}
-          title={getConnectorDialogTitle(addConnectorType)}
+      <AddConnectorDialog
+        components={componentListResponse?.components ?? ({} as ComponentList)}
```
eblairmckee, line 1115: do we need to cast this?

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -1019,19 +1106,37 @@ export default function PipelinePage() {
...
+        connectorType={
+          addConnectorType === 'resource'
+            ? (['cache', 'rate_limit', 'buffer', 'scanner', 'tracer', 'metrics'] satisfies ConnectComponentType[])
+            : (addConnectorType ?? undefined)
+        }
+        isOpen={addConnectorType !== null}
```
eblairmckee, line 1121: nit: `!!addConnectorType` unless react-doctor yelled at you for that?

frontend/src/components/pages/rp-connect/pipeline/list.tsx
```diff
@@ -316,6 +316,13 @@ type CreateColumnsOptions = {
   isDeletingPipeline: boolean;
 };
 
+const ComponentBadge = ({ name }: { name: string }) => (
+  <Badge variant="neutral-inverted">
+    <ConnectorLogo className="size-3.5" fallback={Box} name={name as ComponentName} />
```
eblairmckee, line 321: what's a box? can we use skeleton instead

eblairmckee: and can we type this stronger? I like to avoid casting as often as possible.

frontend/src/components/pages/rp-connect/pipeline/pipeline-flow-diagram.tsx
```diff
@@ -223,17 +234,56 @@ export const PipelineFlowDiagram = ({
...
+      const atTop = vp.y >= maxVpY - 0.5;
+      const atBottom = vp.y <= minVpY + 0.5;
+      if ((event.deltaY < 0 && atTop) || (event.deltaY > 0 && atBottom)) {
+        event.stopPropagation();
+      }
+    };
+    el.addEventListener('wheel', handleWheel, { capture: true, passive: true });
+    return () => el.removeEventListener('wheel', handleWheel, { capture: true } as EventListenerOptions);
+  }, [contentOverflows, translateExtent, containerSize]);
+
```
eblairmckee, line 286: minor: ah! this is what I was expecting. let's get rid of those extra `overflow-hidden`s I mentioned earlier.

frontend/src/components/pages/rp-connect/pipeline/pipeline-flow-diagram.tsx
```diff
@@ -258,15 +315,51 @@ export const PipelineFlowDiagram = ({
...
+          <AnimatePresence>
+            {showTemplateFab && onBrowseTemplates ? (
+              <motion.div
+                animate={{ opacity: 1, y: 0, scale: 1 }}
+                className="pointer-events-none absolute right-4 bottom-4 left-4 z-10"
+                exit={{ opacity: 0, y: 4, scale: 0.98, transition: { duration: 0.18, ease: 'easeIn' } }}
+                initial={{ opacity: 0, y: 12, scale: 0.96 }}
+                transition={{ duration: 0.3, delay: 0.2, ease: [0.16, 1, 0.3, 1] }}
+              >
+                <button
```
eblairmckee, line 341: minor: can we turn this into a reusable component that takes a `className` prop

eblairmckee: long term, might be nice to make this a ui-registry button variant

frontend/src/components/pages/rp-connect/pipeline/pipeline-flow-nodes.tsx
```diff
@@ -163,9 +200,11 @@ const TreeLeafNode = ({ data }: { data: TreeNodeData }) => {
           </Button>
         ) : null}
       </div>
-      <div className={cn((data.labelText || hasTopics || showSetupHints) && 'mt-2', 'flex flex-wrap gap-1.5')}>
+      <div
+        className={cn((data.labelText || hasTopics || showSetupHints) && 'mt-2', 'flex flex-col items-start gap-1.5')}
```
eblairmckee: nit (tiny): can we use `space-y-2` in the wrapper instead of this margin for the fake gap?

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+import { Fragment, type ReactNode, useMemo } from 'react';
+import type { UseFormReturn } from 'react-hook-form';
+import { Controller, useFormState, useWatch } from 'react-hook-form';
+
+import { PipelineStatusToggle } from './pipeline-status-toggle';
+import { cpuToTasks } from '../tasks';
+import { extractAllTopics } from '../utils/yaml';
+import type { PipelineFormValues } from '.';
+
+const DOCS_URL = 'https://docs.redpanda.com/redpanda-connect/home/';
```
eblairmckee, line 34: nit: this feels like something that should live at a more "global" level, like root of console and can concatenate subdirectories to it. 
not something you need to address, just ruminating.

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+// Dot-separated strip of short scalar facts. Null nodes are dropped.
+type MetaEntry = { key: string; node: ReactNode };
+
+const MetaStrip = ({ items }: { items: MetaEntry[] }) => {
+  const visible = items.filter((item) => item.node !== null);
+  if (visible.length === 0) {
+    return null;
+  }
+  return (
+    <div className="flex flex-wrap items-center gap-y-1 text-muted-foreground text-sm">
```
eblairmckee, line 46: minor: something for claude.. whenever you reach for `text-sm` you should instead look at typography variants...

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+      </span>
+    )}
+    <CopyButton
+      className="shrink-0 opacity-0 transition-opacity group-hover/meta:opacity-100"
+      content={value}
+      size="sm"
+      variant="ghost"
+    />
+  </span>
+);
```
eblairmckee, line 125: discuss: oh, I just made something very similar in adp-ui,  called [resource-id-pill](https://github.com/redpanda-data/cloudv2/pull/26590/changes#diff-ade45c8d632a408af0b8a6c78983cb081f54804e8bf924027479760c7e8d09a9).... we should add this as a ui-registry component soon

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+const TagBadges = ({ tags }: { tags: TagEntry[] }) => (
+  <div className="flex flex-wrap gap-1.5">
+    {tags.map((t) => (
+      <Badge key={t.key} variant="simple-outline">
+        {t.value ? `${t.key}: ${t.value}` : t.key}
+      </Badge>
+    ))}
+  </div>
+);
```
eblairmckee, line 148: nit: can we use BadgeGroup instead for consistency's sake

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+const BackButton = ({ onClick }: { onClick: () => void }) => (
+  <Button className="-ml-3.5 shrink-0" onClick={onClick} size="icon" variant="ghost">
+    <ArrowLeftIcon className="h-5 w-5" />
+  </Button>
+);
```
eblairmckee, line 154: discuss: we should make a page-header or toolbar that renders this if a `onBackClick` is provided. similar to page-header I built [here](https://github.com/redpanda-data/cloudv2/pull/26590/changes#diff-ab327bd589e0992151b557a155f0ce1f81fe6a1a88fdfd6f2391f3cb4e0c49bb)

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+          'field-sizing-content min-w-[12rem] max-w-full truncate border-transparent border-b bg-transparent py-0.5 font-display font-medium text-2xl leading-none tracking-heading',
+          'placeholder:text-muted-foreground hover:border-border focus:border-input focus:outline-none',
+          fieldState.error && 'border-destructive hover:border-destructive focus:border-destructive'
+        )}
+        placeholder={placeholder}
+      />
+    )}
+  />
+);
+
```
eblairmckee, line 180: major: we have an EditableText component already that is polymorphic and renders headers

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+    <header className="flex flex-col gap-3">
+      <div className="flex items-center justify-between gap-4">
+        <div className="flex min-w-0 items-center gap-2">
+          <BackButton onClick={onBack} />
+          <Heading className="min-w-0 truncate" level={1} title={name}>
+            {name}
+          </Heading>
+          <Button
+            aria-label="View pipeline details"
+            className="[&_svg]:size-4"
```
eblairmckee: nit: let's avoid using deep selectors like this and let the Button handle it internally. if you really want to change the size of an icon, pass it as a className to the icon itself with a `!` bang, but we should veer away from this for the sake of relying on ui-registry consistency.

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+            Docs
+          </Button>
+          <Button
+            icon={<EditIcon />}
+            onClick={() => navigate({ to: `/rp-connect/${pipeline.id}/edit` })}
+            variant="secondary-outline"
+          >
+            Edit pipeline
+          </Button>
+          <div aria-hidden className="mx-1 h-6 w-px bg-border" />
```
eblairmckee, line 240: minor: we have a Separator component for this

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+          </DetailLine>
+        ) : null}
+      </div>
+    </header>
+  );
+}
+
+// Edit / create header: name as the title with a mode chip; Docs + Save on the
+// right. The settings the dialog controls are shown below, with "Edit settings"
+// anchored to that block (away from Save).
```
eblairmckee: nit: I beg claude to not add these, we should never add comments above jsx elements, especially when the composition easily speaks for itself

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+            >
+              Docs
+            </Button>
+            <Button disabled={isSaving} onClick={onSave}>
+              Save
+              {isSaving ? <Spinner /> : null}
+            </Button>
+          </div>
+        </div>
+        {nameError ? <p className="pl-11 text-destructive text-sm">{nameError}</p> : null}
```
eblairmckee: major: let's rely on zod + react-hook-form for this, which does pass an `error` prop, which the internal `Input` within `EditableText` will render for you

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+            <TagBadges tags={tags} />
+          </DetailLine>
+        ) : null}
+        {description ? (
+          <DetailLine label="Description">
+            <p className="whitespace-pre-wrap break-words" title={description}>
+              {description}
+            </p>
+          </DetailLine>
+        ) : null}
```
eblairmckee, line 330: minor: can we add a `max-w-2/3` here, docs asked for this recently, and I forgot to relay that to you

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
```diff
@@ -0,0 +1,333 @@
...
+          </div>
+        </div>
+        {nameError ? <p className="pl-11 text-destructive text-sm">{nameError}</p> : null}
+      </div>
+      <div className="flex flex-col items-start gap-2">
+        <MetaStrip items={items} />
+        {tags.length > 0 ? (
+          <DetailLine label="Tags">
+            <TagBadges tags={tags} />
+          </DetailLine>
```
eblairmckee, line 322: major: can we use `BadgeGroup` instead, it handles overflow nicely

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { formatToastErrorMessageGRPC } from 'utils/toast.utils';
+
+const STATE_LABEL: Partial<Record<Pipeline_State, string>> = {
+  [PipelineState.RUNNING]: 'Running',
+  [PipelineState.STARTING]: 'Starting',
+  [PipelineState.STOPPING]: 'Stopping',
+  [PipelineState.STOPPED]: 'Stopped',
+  [PipelineState.ERROR]: 'Error',
+  [PipelineState.COMPLETED]: 'Completed',
+};
```
eblairmckee: nit: did we remove the export from the `constants.ts` file for this mapping? feels like we're duplicating stuff

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+type Tone = 'success' | 'error' | 'muted';
```
eblairmckee, line 40: discuss: this makes me think we should have a `muted` variant for Alert in the registry...

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+// Pill chrome per state: running is a filled green pill, error is red text,
+// everything else (stopped/completed/transitioning) reads as plain muted text.
+const TONE_CLASSES: Record<Tone, string> = {
+  success: 'border-outline-success bg-background-success-subtle text-success',
+  error: 'border-transparent text-destructive',
+  muted: 'border-transparent text-muted-foreground',
+};
```
eblairmckee: major: let's use `cva` for this. highly recommend you use my `molecule/SKILL.md` for things like this, it'll catch a lot of these nuances for you.

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+function getTone(state?: Pipeline_State): Tone {
+  if (state === PipelineState.RUNNING || state === PipelineState.STARTING) {
+    return 'success';
+  }
+  if (state === PipelineState.ERROR) {
+    return 'error';
+  }
+  return 'muted';
+}
```
eblairmckee, line 66: nit: really claude... do we really need a function for this? it's a ternary...

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+      },
+    });
+  }, [pipelineId, stopMutation]);
+
+  // Switch reflects (and drives) the running state. Starting/stopping are
+  // in-flight, so the control is locked until the backend settles.
+  const checked = pipelineState === PipelineState.RUNNING || pipelineState === PipelineState.STARTING;
+  const isTransitioning = pipelineState === PipelineState.STARTING || pipelineState === PipelineState.STOPPING;
+  const isDisabled = !pipelineId || isTransitioning || isStartPending || isStopPending;
+  const label = (pipelineState !== undefined && STATE_LABEL[pipelineState]) || 'Unknown';
```
eblairmckee: nit: optional chain here?
```suggestion
  const label = STATE_LABEL?.[pipelineState] ?? 'Unknown');
```

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+  );
+
+  return (
+    <>
+      <div
+        className={cn(
+          'inline-flex h-9 items-center gap-2 rounded-full border px-3 font-medium text-sm transition-colors',
+          TONE_CLASSES[tone]
+        )}
+      >
```
eblairmckee: minor: why are we setting an text styles here? let's do that on the text fields themselves (if no typography variants suffice)

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+        <Switch
+          aria-label={checked ? 'Stop pipeline' : 'Start pipeline'}
+          checked={checked}
+          className={cn(tone === 'success' && 'data-[state=checked]:bg-success')}
```
eblairmckee, line 141: major: again. `cva`

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+          disabled={isDisabled}
+          onCheckedChange={handleCheckedChange}
+          testId="pipeline-run-toggle"
```
eblairmckee, line 145: discuss: we should upstream this to ui-registry and make it more generalized

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+        />
+        {isTransitioning ? <Spinner className="!size-3.5" /> : null}
```
eblairmckee: minor: 99% sure there's a size prop on spinner

eblairmckee: discuss: and if there isn't.... should add to our backlog to add one and align sizes to button icons, etc. (tokenize all the things!)

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
```diff
@@ -0,0 +1,176 @@
...
+        <span>{label}</span>
```
eblairmckee: major: please use Label from ui-registry

frontend/src/components/pages/rp-connect/pipeline/pipeline-throughput-card.tsx
```diff
@@ -200,32 +199,28 @@ export const PipelineThroughputCard: FC<PipelineThroughputCardProps> = ({ pipeli
...
-              <SelectContent align="end">
-                {TIME_RANGES.map((option) => (
-                  <SelectItem key={option.value} value={option.value}>
-                    {option.label}
-                  </SelectItem>
-                ))}
-              </SelectContent>
-            </Select>
-          </div>
+    <section className="flex flex-col gap-3">
```
eblairmckee, line 202: ty for using semantic html elements, btw.

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -59,22 +57,8 @@ type ButtonConfig = {
...
-  onViewConfig?: () => void;
-  onNameChange?: (name: string) => void;
-  onSave?: () => void;
-  onCancel?: () => void;
-  isSaving?: boolean;
-  isLoading?: boolean;
-  nameError?: string;
-  defaultEditing?: boolean;
+  // Vary emphasis by state: Start is primary (encourage), Stop is neutral.
+  variant?: 'primary' | 'secondary' | 'outline';
```
eblairmckee: major: when it comes to prop type comments we should _always_ use jsdoc comments as they show up nicely in intellisense

eblairmckee: general comment. in this case... it's not worthy of a comment imho

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -103,11 +87,30 @@ export function pipelineStateToVariant(state?: Pipeline_State): { variant: Statu
...
 }
 
+const PIPELINE_STATE_LABEL: Partial<Record<Pipeline_State, string>> = {
+  [PipelineState.RUNNING]: 'Running',
+  [PipelineState.STARTING]: 'Starting',
+  [PipelineState.STOPPING]: 'Stopping',
+  [PipelineState.STOPPED]: 'Stopped',
+  [PipelineState.ERROR]: 'Error',
+  [PipelineState.COMPLETED]: 'Completed',
+};
```
eblairmckee: minor: ah, I believe I mentioned this earlier. we should lift this up to a constant and import it within the `pipeline-status-toggle` so it's reusable/source of truth.

frontend/src/components/pages/rp-connect/pipeline/toolbar.tsx
```diff
@@ -103,11 +87,30 @@ export function pipelineStateToVariant(state?: Pipeline_State): { variant: Statu
...
+
+export function PipelineStatusBadge({ state, size }: { state?: Pipeline_State; size?: StatusBadgeSize }) {
+  const { variant, pulsing } = pipelineStateToVariant(state);
+  return (
+    <StatusBadge pulsing={pulsing} size={size} variant={variant}>
+      {(state !== undefined && PIPELINE_STATE_LABEL[state]) || 'Unknown'}
+    </StatusBadge>
+  );
+}
+
```
eblairmckee: major: errrrrr. remove and replace with pipeline-status-toggle? redundant.

frontend/src/components/pages/rp-connect/template-gallery/slot-fields/secret-slot.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onSecretCreated?: (slotId: string, secretName: string) => void;
+  // When supplied, "Create secret" delegates to the parent instead of opening
+  // the nested AddSecretsDialog (avoids dialog-within-a-dialog). The existing-
+  // secret select stays in the slot regardless.
+  onRequestCreateSecret?: (slotId: string, suggestedName: string | undefined) => void;
+};
+
+export const SecretSlotField = ({ slot, control, onSecretCreated, onRequestCreateSecret }: SecretSlotFieldProps) => {
+  const { data: secretsResponse, refetch } = useListSecretsQuery({}, { enabled: true });
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
```
eblairmckee, line 41: minor: iirc we do have a `useDisclosure` hook for this?

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
eblairmckee: discuss: if this comes up multiple times.. let's revisit the sizes we're setting within the Button component for child svgs

frontend/src/components/pages/rp-connect/pipeline/config-dialog.tsx
SpicyPete: I would worry that "Save" is already on the page to save the Pipeline, and this one is to save the settings part of the pipeline.
It is a bit confusing that we have to press save twice, but at least the before leaving confirmation should help some of the time

frontend/src/components/pages/rp-connect/pipeline/index.tsx
SpicyPete: This was changed in Tailwind V4, the bang now goes at the end.  
https://tailwindcss.com/docs/styling-with-utility-classes#using-the-important-modifier

The border-border! class was on this element already before this PR.  
Though I believe both work while they phase out the V3 pattern.

frontend/src/components/pages/rp-connect/pipeline/list.tsx
SpicyPete: Box is the Box Icon which I'm using as the fallback if there's no connector icon to use, such as branching, or generate, etc.

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
SpicyPete: As we want the Dialog to re-size with it's animation, we want the internal size to change instantly.  Otherwise the interaction of both elements trying to animate breaks the smoothness of animations, so I am only animating the Chevron here

frontend/src/components/pages/rp-connect/pipeline/index.tsx
SpicyPete: `blocker` is an object from the `useBlocker` hook, which has the status in a property like `blocker.status`, as it's an object, and not a boolean I don't think the past tense should apply to this.

I do agree though, `is` prefixed on booleans is always a good pattern to have.

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
SpicyPete: This will still try to index STATE_LABEL with `undefined` which is itself a TS error, 
`Type 'undefined' cannot be used as an index type.`

For this sort of pattern we should guard against undefined in the index.

frontend/src/components/pages/rp-connect/template-gallery/slot-fields/secret-slot.tsx
SpicyPete: We do, in the Chakra library.  Maybe we should add one to the registry if it's useful.

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
SpicyPete: I think for this one we should adjust the registry Switch to have states for success, in a follow up if we want to make this pattern re-usable.

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
SpicyPete: There is not a size-prop on the Spinner component, I created a task to add one.  
https://redpandadata.atlassian.net/browse/UX-1323

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
SpicyPete: IMO this is on the edge of needing it's own function, the equivalent would be
```
const tone = (state === PipelineState.RUNNING || state === PipelineState.STARTING) 
    ? 'success' 
     : state === PipelineState.ERROR 
         ? 'error' 
         : 'muted'
```

Which is a little hard to read, especially if we want to add other pipeline states later, or give starting it's own state.

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
SpicyPete: The separator component needs some adjustments to work here, it only can do 100% height, and cannot override it's height prop yet.  We can use it after we fix it this in the registry

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
eblairmckee: followup: i believe i have a ticket already... we need a ui-registry component for animated chevron

frontend/src/components/pages/rp-connect/pipeline/pipeline-status-toggle.tsx
SpicyPete: https://redpandadata.atlassian.net/browse/UX-1328

frontend/src/components/pages/rp-connect/pipeline/index.tsx
eblairmckee: we should rethink this in a follow

eblairmckee (approved):
conditionally approving assuming we will address major & minor comments either in this PR, in a follow, or create a ticket for them; this is good to ship.

frontend/src/components/pages/rp-connect/pipeline/pipeline-header.tsx
SpicyPete: TagBadges is using BadgeGroup

PR #2478, opened 2026-06-01
Use UI registry components in a topic listing

jvorcak:
- Migrated components from Topic Listing + create modal to a UI registry
- URLs are shareable, filtering is stored in a URL
- Adjusted UX of Topic creation success modal

We're not behind a feature flag for this change 

Screenshot 2026-06-02 at 14 56 17
Screenshot 2026-06-02 at 14 55 54


frontend/src/components/pages/topics/topic-list-new.tsx
malinskibeniamin, line 1: if it's not below FF why can't we just replace the old component instead of creating `-new`?

frontend/src/components/pages/topics/topic-list-new.tsx
```diff
@@ -0,0 +1,616 @@
...
+        </DialogFooter>
+      </DialogContent>
+    </Dialog>
+  );
+}
+
+function hasDeletePrivilege() {
+  // TODO - we will provide ACL for this
+  return true;
+}
```
malinskibeniamin, line 597: is this a blocker?

frontend/src/components/pages/topics/create-topic-dialog.tsx
```diff
@@ -0,0 +1,717 @@
...
+  if (unit === 'default') return;
+  if (value === undefined)
+    throw new Error(`unexpected: value for retention size is 'undefined' but unit is set to ${unit}`);
+  if (unit === 'Bit') return value;
+  if (unit === 'KiB') return value * 1024;
+  if (unit === 'MiB') return value * 1024 * 1024;
+  if (unit === 'GiB') return value * 1024 * 1024 * 1024;
+  if (unit === 'TiB') return value * 1024 * 1024 * 1024 * 1024;
+  if (unit === 'infinite') return -1;
+}
```
malinskibeniamin: seems like at least some of this could live in another file

frontend/src/components/pages/topics/topic-list-new.tsx
```diff
@@ -0,0 +1,616 @@
...
+            {iconClosedEye}
+          </span>
+        </TooltipTrigger>
+        <TooltipContent side="right">{popoverContent}</TooltipContent>
+      </Tooltip>
+    </TooltipProvider>
+  );
+};
+
+function ConfirmDeletionModal({
```
malinskibeniamin: I think we can move this away, this should be a reusable resource-delete-alert-dialog (or similar name) so it does not live in the same file.

frontend/src/components/pages/topics/topic-list-new.tsx
```diff
@@ -0,0 +1,616 @@
...
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <span className="inline-flex text-warning">
+              <svg aria-hidden="true" className="h-4 w-4" fill="currentColor" viewBox="0 0 20 20">
+                <title>Warning</title>
+                <path
+                  clipRule="evenodd"
+                  d="M8.257 3.099c.765-1.36 2.722-1.36 3.486 0l5.58 9.92c.75 1.334-.213 2.98-1.742 2.98H4.42c-1.53 0-2.493-1.646-1.743-2.98l5.58-9.92zM11 13a1 1 0 11-2 0 1 1 0 012 0zm-1-8a1 1 0 00-1 1v3a1 1 0 002 0V6a1 1 0 00-1-1z"
+                  fillRule="evenodd"
+                />
```
malinskibeniamin: I don't think you need this svg baked into JSX

frontend/src/components/pages/topics/topic-list-new.tsx
```diff
@@ -0,0 +1,616 @@
...
+
+  return (
+    <TooltipProvider>
+      {!!leaderlessPartitions && (
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <span className="inline-flex text-destructive">
+              <svg aria-hidden="true" className="h-4 w-4" fill="currentColor" viewBox="0 0 20 20">
+                <title>Error</title>
+                <path
```
malinskibeniamin: same here re: svgs

malinskibeniamin (commented):
Let's make sure that the playwright critical tests in cloud UI are updated when this change lands, because it's likely some e2e tests will break (if they are searching for attributes which may have changed due to this).

frontend/src/components/pages/topics/topic-list-new.tsx
jvorcak: This is the same as in the topic-list.tsx, I think we can address this in a separate issue.

frontend/src/components/pages/topics/topic-list-new.tsx
jvorcak: I've made it this way so that we can always decide whether we can put it behind a feature flag or not, if not (which is what I suggest considering the change), I'll of course remove the suffix.

frontend/src/components/pages/topics/topic-list-new.tsx
jvorcak: Good call, done

frontend/src/components/pages/topics/topic-list-new.tsx
jvorcak: I'm adding a feature flag for this, just to be sure

frontend/src/utils/topic-utils.ts
```diff
@@ -86,6 +86,34 @@ export function convertRetentionSizeToBytes(value: number, unit: RetentionSizeUn
...
+  if (unit === 'default') return;
+  if (value === undefined)
+    throw new Error(`unexpected: value for retention size is 'undefined' but unit is set to ${unit}`);
+  if (unit === 'Bit') return value;
+  if (unit === 'KiB') return value * 1024;
+  if (unit === 'MiB') return value * 1024 * 1024;
+  if (unit === 'GiB') return value * 1024 * 1024 * 1024;
+  if (unit === 'TiB') return value * 1024 * 1024 * 1024 * 1024;
+  if (unit === 'infinite') return -1;
+}
```
malinskibeniamin, line 115: I don't think we need to bail out of biome here.

malinskibeniamin (approved):

PR #2479, opened 2026-06-01
fix(deps): snyk sweep — console frontend (audited overrides + .snyk)

malinskibeniamin:
## Snyk sweep — console frontend

Clears Snyk-flagged vulnerabilities in `frontend` (bun-only) and **audits the whole override set** so we only carry pins that actually do something.

### Override audit

Every simple `overrides`/`resolutions` entry was tested by removing it and re-resolving. **15 were dropped as inert/redundant** — the resolved version is identical with or without them (the lib isn't used directly and its natural resolution already lands on the fixed/latest version), or it isn't in the graph at all:

`mermaid`, `postcss`, `@hono/node-server`, `hono`, `fast-uri`, `webpack`, `elliptic`, `flatted`, `immer`, `immutable`, `lodash`, `lodash-es`, `mdast-util-to-hast`, `react-error-overlay` (not in graph), `serialize-javascript` (not in graph).

Verified: incremental re-resolve leaves every dropped package's resolved version **unchanged**, so removing them is a no-op for the lockfile and the Snyk gate.

### Kept — load-bearing (removing reverts a transitive to an older/vulnerable version)

`ip-address` (express-rate-limit exact-pins `10.1.0`; XSS `GHSA-v2v4-37r5-5v8g` fixed only in 10.1.1), `@tootallnate/once` (1.1.2), `diff` (4.0.4 ReDoS), `prismjs` (1.27), `@remix-run/router` (1.23.1), `qs` (6.13.0), `body-parser` (1.20.3), `sirv` (2.0.4), `ajv` (eslint's 6.x), `picomatch`, `dompurify`, `protobufjs`, plus the path-scoped `**/...` resolutions.

> Several of the kept ones are build-/dev-only (e.g. `sirv` via webpack-bundle-analyzer, `ajv` via eslint, `@tootallnate/once` via http-proxy-agent, `ip-address` via express-rate-limit — none shipped to the browser). Those could be migrated from overrides to `.snyk` ignores in a follow-up; kept as overrides here since they're load-bearing and a dismissal needs Snyk-ID verification.

### `.snyk` dismissals

uuid (v3/v5/v6-buffer only, not reached) and the three Vercel AI SDK advisories (server-side download/validate-URL + unbounded-buffering paths, not invoked in this browser MF remote) — all "no upstream fix / not reachable", with reason + expiry.


c-julin (approved):

malinskibeniamin (commented):
I think the following may not be needed:
- ip-address
- picomatch
- sirv
- **/webpack-bundle-analyzer/sirv
- @tootallnate/once
- **/minimatch/brace-expansion
- **/cosmiconfig/yaml
- **/router/path-to-regexp
- @remix-run/router (we use tanstack router)
- diff
- body-parser
- ajv
- dompurify

please review each carefully 1 by 1 and ensure we either really need them or we can get rid of them.

malinskibeniamin:
@malinskibeniamin reviewed all 13 flagged entries empirically: removed every override/resolution, ran a fresh `bun install` + `snyk test`, and observed exactly which ones are load-bearing. With zero overrides, only **3** vulnerable packages resurface — so the other 10 (plus the 4 selector resolutions) were inert and are now gone. Pushed in 88e25b7d4.

**Kept as real bumps (runtime-shipped, reachable XSS — caret does not float to a safe version on its own):**
- `dompurify` → `^3.4.0`: pulled by monaco-editor / milkdown / mermaid (all browser-rendered). Removing it reverts to 3.1.7 with 9 XSS + prototype-pollution issues. Genuine fix, kept.
- `prismjs` → `^1.30.0`: via `@redpanda-data/ui` > react-syntax-highlighter > refractor (runtime). Reverts to 1.27.0 (code injection). Kept.

**Moved override → `.snyk` ignore (not reachable):**
- `ip-address`: only reaches the graph via `@modelcontextprotocol/sdk` > `express-rate-limit` (a Node-side rate limiter exact-pinning 10.1.0). The XSS sink isn't invoked by this browser MF remote. Replaced the pin with a documented `.snyk` ignore (reason + expiry 2026-08-30).

**Dropped entirely as redundant/inert (removing each reintroduced no Snyk finding — dev/build/test-only transitives whose caret ranges already float to safe versions):**
- `picomatch`, `sirv`, `@tootallnate/once`, `@remix-run/router`, `diff`, `body-parser`, `ajv`, `qs`, `yaml`, `protobufjs`
- selector resolutions: `**/minimatch/brace-expansion`, `**/cosmiconfig/yaml`, `**/router/path-to-regexp`, `**/webpack-bundle-analyzer/sirv`

**Also dropped the mirrored `resolutions` block** — verified a simple bun `overrides` entry already collapses the vulnerable spec ranges to the fixed resolved version in the exported `yarn.lock` (dompurify→3.4.0, prismjs→1.30.0), so the duplicate `resolutions` was unnecessary. The `overrides` block is now just 2 entries.

Result: `snyk test` → **0 open vulnerabilities**; `type:check` and `build` both pass. bun.lock + yarn.lock regenerated consistently.

PR #2486, opened 2026-06-05
Move components to UI registry for topic detail

jvorcak:
I've migrated components to the new UI registry

Screenshot 2026-06-09 at 14 13 58

+ changed the topic configuraion to be more easy to work with.

Before:
Screenshot 2026-06-09 at 14 15 12

After:
Screenshot 2026-06-09 at 14 14 03


jvorcak:
@claude review

frontend/src/components/pages/topics/topic-configuration.tsx
```diff
@@ -65,8 +121,11 @@ const ConfigEditorForm: FC<{
     }
     return entryHasInfiniteValue(editedEntry) ? 'infinite' : 'custom';
   })();
-  const defaultCustomValue =
+  const explicitCustomValue =
     editedEntry.isExplicitlySet && !entryHasInfiniteValue(editedEntry) ? editedEntry.value : '';
+  // For enum-style configs (boolean/select), fall back to the first option when there's
```
malinskibeniamin: This fallback ignores the resolved/inherited value whenever the config is not explicitly set. For a non-explicit BOOLEAN/SELECT with current value `true` or a non-first enum, opening Custom and saving without changing the dropdown writes `false`/the first enum. Please seed from `editedEntry.value`/current/default synonym when present, and only fall back to the first option when no resolved value exists.

frontend/src/components/pages/topics/topic-configuration.tsx
```diff
@@ -140,94 +192,116 @@ const ConfigEditorForm: FC<{
...
-    'DYNAMIC_BROKER_CONFIG',
-    'DYNAMIC_DEFAULT_BROKER_CONFIG',
-    'STATIC_BROKER_CONFIG',
-    'DEFAULT_CONFIG',
-  ];
+  const defaultConfigSynonym = getDefaultConfigSynonym(editedEntry);
 
-  const defaultConfigSynonym = editedEntry.synonyms
-    ?.filter(({ source }) => source !== 'DYNAMIC_TOPIC_CONFIG')
-    .sort((a, b) => SOURCE_PRIORITY_ORDER.indexOf(a.source) - SOURCE_PRIORITY_ORDER.indexOf(b.source))[0];
```
malinskibeniamin, line 153: In the new reset path (`handleReset`, lines 197-205; button at 271-274), `Reset to default` sends `DELETE` immediately, outside the normal Save path and without its pending state. A slow reset followed by Save can race DELETE vs SET, and an accidental click can remove retention/durability overrides. Please route reset through the Default + Save flow, or add confirmation plus shared `isMutating` disabling for Reset/Save/Cancel.

frontend/src/components/pages/topics/topic-configuration.tsx
```diff
@@ -50,13 +57,62 @@ type Inputs = {
...
+  { name: 'Compaction', blurb: 'Log cleanup and key-based compaction behavior.' },
+  { name: 'Replication', blurb: 'Durability and in-sync replica requirements.' },
+  { name: 'Tiered Storage', blurb: 'Offloading topic data to object storage.' },
+  { name: 'Write Caching', blurb: 'Write acknowledgement and caching behavior.' },
+  { name: 'Other', blurb: 'Additional topic configuration.' },
+] as const;
+
+const ALLOWED_CATEGORIES = new Set<string>(CONFIG_CATEGORIES.map((c) => c.name));
+
+function categoryForEntry(entry: ConfigEntryExtended): string {
```
malinskibeniamin, line 111: This allow-list collapses existing backend categories like Iceberg, Schema Registry and Validation, Message Handling, Compression, and Storage Internals into `Other`. The old layout preserved those sections, so the new sidebar loses useful navigation. Please preserve backend categories/order or include the previous display order before falling back to `Other`.

frontend/src/components/pages/topics/topic-configuration.tsx
```diff
@@ -140,94 +192,116 @@ const ConfigEditorForm: FC<{
...
+      </DialogContent>
+    </Dialog>
   );
 };
 
-const ConfigurationEditor: FC<ConfigurationEditorProps> = (props) => {
-  const [filter, setFilter] = useState<string>('');
+const ConfigurationEditorLegacy: FC<ConfigurationEditorProps> = (props) => {
+  const navigate = useNavigate({ from: '/topics/$topicName/' });
+  const { configFilter = '' } = useSearch({ from: '/topics/$topicName/' });
```
malinskibeniamin: This writes raw config search input to the URL. In the legacy path below, `configFilter` still matches `x.value`, so searching a sensitive/internal config value leaves it in browser history and shareable links. Please keep config filters local/session-only, or make URL-backed search strictly name/description-only and not value-backed.

frontend/src/components/pages/topics/tab-consumers.tsx
```diff
@@ -34,34 +42,134 @@ export const TopicConsumers: FC<TopicConsumersProps> = ({ topic }) => {
 
   const rawConsumers = useApiStoreHook((s) => s.topicConsumers.get(topic.topicName));
   const isLoading = rawConsumers === undefined;
-
   const consumers = rawConsumers ?? [];
 
-  const paginationParams = usePaginationParams(consumers.length, uiState.topicSettings.consumerPageSize);
```
malinskibeniamin, line 40: The URL page index is passed straight into React Table with `autoResetPageIndex: false`. A stale `?consumerPage=999` (same pattern in partitions and ACL) renders “No data found” even when rows exist; the previous `usePaginationParams` clamped this. Please clamp/reset pageIndex when data length or pageSize changes and repair the URL.

frontend/src/components/pages/topics/tab-partitions.tsx
```diff
@@ -148,21 +216,19 @@ const PartitionError: FC<{ partition: Partition }> = ({ partition }) => {
...
-      hideCloseButton
-      placement="right-start"
-      size="auto"
-      title="Partition Error"
-    >
-      <Box>
-        <WarningIcon color="orange" size={20} />
-      </Box>
+    <Popover>
+      <PopoverTrigger asChild>
```
malinskibeniamin, line 168: This is a new icon-only button without an accessible name. Screen reader users only get an unlabeled button for partition error details. Please use the registry `Button` icon size (or add `aria-label="Show partition error details"`).

frontend/tests/shared/global-setup.mjs
```diff
@@ -940,18 +951,38 @@ async function startBackendServerWithConfig(
...
         const inspectJson = JSON.parse(inspect);
+        console.error(`Container ${containerId} exit code:`, inspectJson[0].State.ExitCode);
         console.error('Container state:', JSON.stringify(inspectJson[0].State, null, 2));
-        console.error('Container logs:', logs);
-      } catch (logError) {
-        console.error('Could not fetch container diagnostics:', logError.message);
+      } catch (inspectError) {
+        console.error('Could not inspect container (likely already removed):', inspectError.message);
       }
     }
```
malinskibeniamin, line 953: The new log consumer dumps all backend container output on startup failure. These e2e runs can include enterprise config/license material and auth/SASL/JWT errors, so full logs can leak secrets into CI output. Please redact token/password/license-like values and cap this to a safe tail before printing.

malinskibeniamin (changes requested):
## Review

Thanks for the migration. Major value is clear: topic detail moves off legacy UI pieces and the new Configuration tab improves discoverability. Requesting changes for a few regressions before merge.

Main blockers: config edits can silently write/reset the wrong value, URL-backed state creates stale/secret-leak paths, pagination can show false empty tables, and some categories/accessibility/security guards regressed.

One additional issue not attached inline: `topic-details.tsx` tab switching uses `historyReplace(`${loc.pathname}#${loc.hash}`)`, which drops the search string. Since this PR makes `configFilter`/`configScope` URL-backed, switching tabs and returning loses the Configuration search/scope state. Please preserve `loc.search`/`searchStr` or navigate through the router.

frontend/tests/shared/global-setup.mjs
jvorcak: We've now reverted all changes in global-setup

graham-rp (approved):
LGTM for self-hosted

SpicyPete (commented):
Looking great for the most part, just some feedback on some things I'm noticing.

1. ACL's tab, not sure if it's been migrated yet.
2. Minor: [a] I like this UX, however when we scroll to the bottom, only the top section is highlighted on the left.  As often many sections are in view, perhaps we could highlight all of the sections currently in view.  Might require larger changes, so not needed for this PR.
3. Minor: [b] Some places are missing spacing, so items are too close
4. [c] - There are more configuration below, but it's not obvious we can scroll down more, perhaps we use the registry shadow indicators for this, or some other way to indicate in the UI there's more to see below
5. Topics Messages tab - Doesn't appear to be migrated yet, nor the dialogs on this tab/page
6. [d] The top area still seems to be on Chakra components, is this in scope?

#### Images

##### [a]
 
Image

##### [b]

Image

#####  [c]

Image

##### [d]

Image

jvorcak:
@SpicyPete some things are still not migrated indeed. As for the other stuff, ACL and messages should be migrated, I've removed the feature flag becuase it was too complex to manage. Can you try again? 

PS: I couldn't reproduce the state you're describing in cloud UI nor in Console.

malinskibeniamin (approved):

PR #2495, opened 2026-06-11
UX-1330 - feat(sql): add SQL query workspace with Iceberg bridge

c-julin:
## Summary

Adds a SQL query workspace to Console for querying Redpanda SQL: a catalog tree, SQL editor, results view, and an Iceberg bridge wizard.

Refs: UX-1330

## Changes

- New `frontend/src/components/pages/sql/` module: `sql-workspace`, `sql-editor`, `sql-results`, `catalog-tree`, `sql-wizard`, shared `sql-types`
- New `/sql` route and `react-query/api/sql` data layer
- Follow-up styling passes: design tokens instead of off-token colours/ad-hoc classes, inline styles moved to classes
- Vendored redpanda-ui registry component updates and supporting dependency bumps; zod pinned to ^4.3.6

## Notes

- Backend (Go) changes for SQL auth/config are intentionally excluded — they ship in a separate PR.
- Draft: screenshots and final QA to follow.


c-julin:
@claude review

c-julin:
@claude review

frontend/src/routes/sql.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { createFileRoute } from '@tanstack/react-router';
+import { SqlWorkspace } from 'components/pages/sql/sql-workspace';
+import { Database } from 'lucide-react';
+
+export const Route = createFileRoute('/sql')({
```
malinskibeniamin, line 18: **P1 — `/sql` route is not feature-guarded; direct navigation bypasses the SQL gate**

The `/sql` route has no `beforeLoad`. The `enableSqlInConsole` / `AppFeatures.REDPANDA_SQL` check in `route-utils.tsx` only hides the *sidebar link* — typing `/sql` (or `/clusters/<id>/sql` when embedded) still mounts `SqlWorkspace` and fires `ListCatalogs` / `ListTables` / `ExecuteQuery` even when the feature is off or unlicensed. Every other gated route uses a `beforeLoad` redirect (see `frontend/src/routes/observability/index.tsx:23`).

Severity hinges on backend enforcement (the PR says SQL auth ships separately): if the backend independently enforces the license/flag this is consistency + defense-in-depth; if not, an unlicensed tenant reaches the SQL endpoint. Flagged P1 because it diverges from the established guard pattern and the backend gate is out of this PR.

**Suggested fix:** add a `beforeLoad` mirroring `observability/index.tsx` that does `throw redirect({ to: '/', replace: true })` when `!(isEmbedded() ? isFeatureFlagEnabled('enableSqlInConsole') : AppFeatures.REDPANDA_SQL)`.

**One-shot:** In repo console `frontend/src/routes/sql.tsx`, add a `beforeLoad` to `createFileRoute('/sql')` mirroring `routes/observability/index.tsx` that redirects to `/` when the SQL feature is disabled (embedded → `!isFeatureFlagEnabled('enableSqlInConsole')`, standalone → `!AppFeatures.REDPANDA_SQL`); verify with `cd frontend && bun run type:check && bun run lint`.

frontend/src/components/pages/sql/sql-workspace.tsx
```diff
@@ -0,0 +1,657 @@
...
+    (sql: string) => {
+      const token = nextRunToken();
+      const kw = firstKeyword(sql);
+      // Drive the bridge indicator off the executed query (single tiered topic),
+      // not the catalog click — so it only shows for the topic actually queried.
+      const nextBridgeTopic = kw === 'SELECT' ? bridgeTopicForQuery(sql, completionCatalogs) : null;
+      setBridgeTopic(nextBridgeTopic);
+      setBridgeRunAt(nextBridgeTopic ? Date.now() : null);
+
+      if (kw !== 'SELECT') {
```
malinskibeniamin: **P1 — SELECT-only gate is both over-restrictive and trivially bypassable**

`doRun` gates on `firstKeyword(sql) === 'SELECT'`. Two problems:

- **Over-restrictive:** `firstKeyword('WITH x AS (…) SELECT …')` returns `WITH`, and `EXPLAIN SELECT …` returns `EXPLAIN`. Both are read-only and a canonical way to write many queries, yet both are rejected with "Only SELECT statements are supported." A SQL studio that can't run a CTE will read as broken.
- **Bypassable:** only the *first* keyword is checked; the full `statement` is sent unmodified, so `SELECT 1; DROP TABLE x;` passes the gate. This is acceptable only if the backend enforces read-only (out of this PR) — it is not a client-side security control.

**Suggested fix:** switch to a write/DDL/DCL **denylist** on the first keyword (INSERT/UPDATE/DELETE/CREATE/DROP/ALTER/GRANT/REVOKE/TRUNCATE…), so read-shaped statements (SELECT/WITH/EXPLAIN) pass and the server rejects what it can't run. Keep the CREATE→wizard and GRANT/REVOKE→Security hints. Confirm the backend independently enforces read-only.

**One-shot:** In `frontend/src/components/pages/sql/sql-workspace.tsx` `doRun` (and `sql.tsx` `firstKeyword`), replace the `=== 'SELECT'` allowlist with a write-keyword denylist so WITH/EXPLAIN/SELECT are permitted and INSERT/UPDATE/DELETE/CREATE/DROP/ALTER/GRANT/REVOKE/TRUNCATE are blocked; add tests in `sql.test.tsx` for WITH and EXPLAIN passing while CREATE/GRANT stay blocked.

frontend/src/components/pages/sql/sql-workspace.tsx
```diff
@@ -0,0 +1,657 @@
...
+  // to 0x0, which fires the ResizeObserver, and we release until shown again.
+  const isVisible = () => el.getClientRects().length > 0;
+  let active = false;
+  const sync = () => {
+    if (isVisible() && !active) {
+      layout();
+      lockAll();
+      active = true;
+    } else if (!isVisible() && active) {
+      unlockAll();
```
malinskibeniamin, line 245: **P1 — Full-mode studio strands `data-page-expanded` on the shared `<html>` when embedded cloud-ui hides (not unmounts) the remote**

`setPageExpanded(false)` runs only on React unmount (`attachOverlay(null)`) and on `toggleMode → boxed`. But the comment at L228–232 documents that cloud-ui keeps Console *mounted + `display:none`* while on its own routes — and the `sync()` path written for exactly that case calls `unlockAll()` but never clears `PAGE_EXPANDED_ATTR`.

Repro (embedded): enter fullscreen → navigate to a cloud-ui page. The studio goes `display:none`, scroll locks release, but `data-page-expanded=""` stays on `document.documentElement`. Per the contract at L102–106, cloud-ui's layout keys `expandableWidth` off that attribute, so every other cloud-ui page renders expanded full-width until the studio is shown and toggled back.

**Suggested fix:** tie the host attribute to the same on-screen lifecycle as the scroll locks: in `sync()`, clear `setPageExpanded(false)` when becoming hidden and re-apply `setPageExpanded(modeRef.current === 'full')` when becoming visible.

**One-shot:** In `frontend/src/components/pages/sql/sql-workspace.tsx` `setupOverlayLayout`'s `sync()`, also call `setPageExpanded(false)` when the overlay becomes hidden and `setPageExpanded(modeRef.current === 'full')` when it becomes visible; add a test mounting in full mode, firing the ResizeObserver with a 0×0 rect, and asserting `document.documentElement` has no `data-page-expanded`.

frontend/src/components/pages/sql/sql-workspace.tsx
```diff
@@ -0,0 +1,657 @@
...
+  type QueryRun,
+  type ResultRow,
+  type SqlRole,
+  type TableRef,
+} from './sql-types';
+import { SqlWizard, type WizardTopic } from './sql-wizard';
+
+const INITIAL_QUERY = 'SELECT name, type\nFROM system.catalogs\nORDER BY name;';
+
+let RUN_TOKEN = 0;
```
malinskibeniamin: **P2 — Stale-result guard uses a module-global `RUN_TOKEN` instead of a per-instance ref**

`RUN_TOKEN` / `nextRunToken()` are module-level, so the `if (RUN_TOKEN !== token) return` guards in `executeQuery`'s `onSuccess`/`onError` compare against state shared by *every* `SqlWorkspace` instance. With two concurrently-mounted instances (embedded cloud-ui keeps Console mounted while a second studio mounts; dev StrictMode double-invoke) instance B's run advances the shared counter and silently drops instance A's in-flight result, leaving A stuck on "running". Single-instance use is unaffected, but it's a latent correctness hazard and the per-run token already lives in `run` state.

**Suggested fix:** replace the module global with `const latestRunToken = useRef(0)` inside `SqlWorkspace`; issue via `const token = (latestRunToken.current += 1)` and guard with `latestRunToken.current !== token`.

**One-shot:** In `frontend/src/components/pages/sql/sql-workspace.tsx`, remove the module-level `RUN_TOKEN`/`nextRunToken` and use a `useRef(0)` inside `SqlWorkspace` for token issuance and the onSuccess/onError guards; verify with `cd frontend && bun run type:check && bun run test`.

frontend/src/components/pages/sql/sql-workspace.tsx
```diff
@@ -0,0 +1,657 @@
...
+    short: column.type.toLowerCase(),
+    isArray: isArrayPgType(column.type),
+  };
+}
+
+function cellValueForColumn(value: SqlValue, column: ColumnDef): CellValue {
+  let cell: CellValue = value.nullValue ? null : (value.value ?? null);
+  // Arrays keep their raw string form — only scalar bools coerce.
+  if (cell !== null && column.kind === 'bool' && !column.isArray) {
+    cell = cell === 'true' || cell === 't';
```
malinskibeniamin: **P2 — Boolean cell coercion is case-sensitive; `'TRUE'`/`'T'` render as false**

`cellValueForColumn` does `cell = cell === 'true' || cell === 't'`. PostgreSQL's wire format is lowercase `t`/`f`, but if the Oxla/PG driver ever emits `TRUE` / `T` / `True`, a true value silently renders as a red `false` — a data-display integrity bug, not just cosmetic.

**Suggested fix:** lowercase before comparing: `const v = cell.toLowerCase(); cell = v === 'true' || v === 't';`

**One-shot:** In `frontend/src/components/pages/sql/sql-workspace.tsx` `cellValueForColumn`, lowercase the value before the bool comparison and add a unit test covering `'TRUE'`/`'T'`/`'false'` plus the bool-array passthrough; verify with `cd frontend && bun run test`.

malinskibeniamin (commented):
## Review — SQL query workspace + Iceberg bridge (#2495)

Multi-hat review (ponytail, security, resilience, regular, adversarial, visual, test-perf, thermo-nuclear) against merge-base `44fc8fe2`. This is a large, well-structured feature (~4k LOC) with good test coverage on the leaf components. Inline comments cover the actionable findings; this summary holds the value gate, the open question for the backend reviewer, what was cleared, and P3/follow-ups.

### Open question (sets final severity of two inline findings)
The PR notes backend SQL auth/config ships separately. **Does `SQLService.ExecuteQuery` independently enforce (a) read-only and (b) the SQL license / `enableSqlInConsole` gate server-side?** If yes, the `/sql` route guard and the SELECT-only gate are defense-in-depth/UX; if no, the client gate is the only barrier and both rise in severity.

### Cleared with evidence (not issues)
- **No XSS in results rendering** — all cells render as React text children; no `dangerouslySetInnerHTML`/`setHTML`; CSV/JSON export goes through `Blob`.
- **CREATE TABLE interpolation is safe** — `tableName` is zod-validated `^[a-z_][a-z0-9_]*$`; `topic` comes from the server topic list and Kafka topic names can't contain a quote.
- **`zod@^4.3.6`** is pre-existing on master, not introduced here.
- **`routeTree.gen.ts`** is genuine codegen output.
- Editor `outline: none` is paired with a caret + active-line highlight (conventional focus indication); icon buttons' `title` provides a fallback accessible name — minor, not posted.

### P3 / follow-ups (not posted inline)
- `sql-workspace.tsx` (657 LOC — the core logic: `doRun` gate, run-token guard, `useStudioMode` persistence, helpers) has **no test file**, while every leaf component does. Add coverage for the gate branches and `cellValueForColumn`/`resultRowFromProto`.
- The `bridge` lag `useMemo` can briefly pair the new topic name with the previous topic's lag during a topic switch (transient display inaccuracy) — gate on `isFetching`.
- `completionCatalogs` → editor `extensions` rebuild on every `redpandaTablesData` refetch (new object ref) → editor reconfiguration churn; stabilize the schema memo.
- `bun.lock` dropped the top-level `configVersion: 0` — confirm the bun version matches CI to avoid frozen-install drift.
- `catalog-tree.tsx` uses `biome-ignore noExcessiveCognitiveComplexity` on `TableRow`/`NamespaceNode` rather than extracting sub-components.
- `globals.css` carries a UX-1259 dark-mode border fix inside this UX-1330 PR — note it in the description for traceability.
- A11y polish: disabled "Run selection" button has no tooltip explaining why; `DataGrid` and the table-specific icon buttons would benefit from explicit `aria-label`s.
- `sqlRole` defaults to `viewer` standalone (admin injected by cloud-ui via federation) — a one-line comment would stop it reading as missing wiring.

### Value gate
Major improvement: a full SQL query workspace (catalog tree, CodeMirror editor, results grid, Iceberg bridge wizard) — net-new product capability. **Value: HIGH.** Steelman not needed.

### Counts
Posted inline: **P1 ×3, P2 ×7, P3 ×2.** Plus ~8 P3/follow-ups in this summary. No P0.

🤖 Multi-hat review via Claude Code

frontend/src/components/pages/sql/sql-types.ts
```diff
@@ -0,0 +1,162 @@
...
+// literal "json"/"json[]" (the backend parses structure into Column.fields), so
+// they fall through to the JSON branch. Arrays map to their element kind;
+// anything unrecognized defaults to a string.
+export function columnKindForPgType(pgType: string): ColumnKind {
+  const element = arrayElementPgType(pgType);
+  if (element !== null) {
+    return columnKindForPgType(element);
+  }
+  const t = pgType.toUpperCase();
+  if (NUMERIC_TYPE.test(t)) {
```
malinskibeniamin, line 149: **P2 — `INTERVAL` (and geometric `POINT`) columns misclassified as numeric**

`columnKindForPgType` tests `NUMERIC_TYPE = /(INT|FLOAT|…)/` *before* `TEMPORAL_TYPE`, and the regex matches substrings — so `INTERVAL` matches `INT` → returns `'num'` (right-aligned, numeric styling) instead of `'time'`. `POINT` likewise matches `INT`.

**Suggested fix:** test `TEMPORAL_TYPE` before `NUMERIC_TYPE`, and/or anchor the numeric match to whole type tokens so `INTERVAL`/`POINT` don't match `INT`.

**One-shot:** In `frontend/src/components/pages/sql/sql-types.ts` `columnKindForPgType`, check `TEMPORAL_TYPE` before `NUMERIC_TYPE` and tighten `NUMERIC_TYPE` to avoid matching `INTERVAL`/`POINT`; add cases for `INTERVAL`, `POINT`, `TIMESTAMP`, `INT8` in `sql-types.test.ts`; verify with `cd frontend && bun run test`.

frontend/src/components/pages/sql/catalog-tree.tsx
```diff
@@ -0,0 +1,580 @@
...
+// biome-ignore lint/complexity/noExcessiveCognitiveComplexity: Leaf row combines grant, selection, tiering and query affordance state.
+function TableRow({ catalog, table }: TableRowProps) {
+  const { activeTableId, openTables, toggleTable, onQueryTable } = useCatalogTree();
+  const allowed = table.allowed !== false;
+  const isOpen = Boolean(openTables[table.id]);
+  const isActive = activeTableId === table.id;
+  const isIceberg = catalog.engine === 'iceberg' || table.iceberg === true;
+  // A Redpanda-catalog table is Iceberg-tiered when its backing topic has
+  // `redpanda.iceberg.mode` enabled (read from the Kafka topic config).
+  const { isIceberg: topicTiered } = useTopicIcebergQuery(table.topicName ?? '', {
```
malinskibeniamin, line 314: **P2 — N+1: one `GetTopicConfigurations` request per visible table row**

`TableRow` calls `useTopicIcebergQuery(table.topicName, …)` once per row (enabled for every Redpanda-catalog table). Expanding a namespace renders up to `CAT_LIMIT` (20) rows → up to 20 concurrent `GetTopicConfigurations` RPCs, and each "load more" fires another batch. On large clusters this is a lot of redundant traffic just to render a tiering badge.

**Suggested fix:** resolve the tiering flag once at the catalog/namespace level (where `useListTablesQuery` already runs) and pass `isIceberg` down as a prop, or batch via a single configs call if the API supports it. React Query dedupes identical topics, but distinct topics each fire.

**One-shot:** In `frontend/src/components/pages/sql/catalog-tree.tsx`, lift the `useTopicIcebergQuery` call out of `TableRow` into the namespace/catalog level, build a `Map<topic, boolean>`, and pass `isIceberg` as a prop to `TableRow`; add a test asserting the query fires once per table (not per render).

frontend/src/components/pages/sql/sql-workspace.tsx
```diff
@@ -0,0 +1,657 @@
...
+
+  const closeWizard = useCallback(() => {
+    setWizardOpen(false);
+    setWizardError(undefined);
+  }, []);
+
+  const onCreateTable = useCallback(
+    ({ topic, tableName }: { topic: string; tableName: string }) => {
+      setWizardError(undefined);
+      const statement = `CREATE TABLE default_redpanda_catalog=>${tableName}\n  WITH (topic='${topic}');`;
```
malinskibeniamin: **P2 — CREATE TABLE statement is built twice; the wizard preview can drift from what runs**

`onCreateTable` here builds `` `CREATE TABLE default_redpanda_catalog=>${tableName} WITH (topic='${topic}')` `` inline, while `sql-wizard.tsx:70` `createTableSql()` builds the same statement independently for the "This will run" preview. Two sources of truth means the preview shown to the user can silently diverge from the executed statement if either changes.

**Suggested fix:** export `createTableSql` from the wizard (or a shared `sql` helper) and call it in both the preview and `onCreateTable`.

**One-shot:** In repo console, export `createTableSql` from `frontend/src/components/pages/sql/sql-wizard.tsx` and reuse it in `sql-workspace.tsx` `onCreateTable` so the previewed and executed CREATE TABLE statements share one builder; verify with `cd frontend && bun run type:check && bun run test`.

frontend/src/components/pages/sql/sql.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const word = stmt.replace(LEADING_COMMENTS, '').match(FIRST_KEYWORD_RE);
+  return word ? word[0].toUpperCase() : '';
+}
+
+// Oxla addresses catalog tables as `catalog=>table`. A bridge indicator is
+// only meaningful for a single Redpanda-catalog table reference.
+const BRIDGE_REF_RE = /([A-Za-z_][\w$]*)\s*=>\s*"?([a-zA-Z0-9._-]+)/g;
+
+export function bridgeTopicForQuery(stmt: string, catalogs: Catalog[]): string | null {
+  const matches = [...stmt.matchAll(BRIDGE_REF_RE)];
```
malinskibeniamin: **P2 — Bridge-topic detection matches `=>` inside comments/strings and ignores multi-ref queries**

`bridgeTopicForQuery` runs `BRIDGE_REF_RE` over the raw statement with no comment/string stripping (unlike the editor's completion source, which skips Comment/String nodes). So `SELECT * FROM cat=>cars -- TODO cat=>orders` yields two matches → returns `null` → the Iceberg bridge lag indicator disappears on a valid single-table query; and a `=>` inside a string/comment on a non-bridge query can point the lag query at the wrong topic. It also returns `null` for any legitimate two-table-ref query (joins across tiered tables).

**Suggested fix:** strip `--` line comments, `/* */` block comments, and single-quoted strings before matching (reuse the `LEADING_COMMENTS` approach or walk the syntax tree). Decide and document the intended multi-ref behavior.

**One-shot:** In `frontend/src/components/pages/sql/sql.tsx` `bridgeTopicForQuery`, strip SQL comments and single-quoted string literals before running `BRIDGE_REF_RE`; add tests where a `=>` in a comment is ignored and a string-only `=>` returns `null`.

frontend/package.json
```diff
@@ -113,11 +120,12 @@
     "react": "^18.3.1",
     "react-beautiful-dnd": "^13.1.1",
     "react-compiler-runtime": "^1.0.0",
+    "react-data-grid": "7.0.0-beta.47",
```
malinskibeniamin, line 123: **P2 — New dependency `react-data-grid@7.0.0-beta.47` is a pre-release**

The entire results grid depends on a 7.0 *beta*. Betas can ship breaking changes between prereleases or be yanked, and this is a brand-new dependency for the repo. Worth confirming there isn't a stable line that works, or documenting/tracking the beta pin.

**Suggested fix:** confirm whether a stable `react-data-grid` release covers the needs (virtualization, frozen row-number column, resize); if staying on the beta, pin it exactly (no caret) and add an upgrade-tracking note.

**One-shot:** In repo console `frontend/package.json`, confirm the `react-data-grid@7.0.0-beta.47` pin is intentional; if so, document the rationale and keep it exact-pinned, else evaluate the latest stable release; verify with `cd frontend && bun install && bun run build`.

frontend/src/utils/fullscreen-routes.ts
```diff
@@ -0,0 +1,95 @@
...
+  return paths;
+}
+
+/**
+ * True when `pathname` is (or is nested under) one of `paths`. Matches the path
+ * as a `/`-bounded segment so it tolerates host prefixes when embedded — e.g.
+ * Cloud UI's `/clusters/<id>/sql` — while not matching `/mysql` or `/sqlx`.
+ */
+export function matchesFullscreenPath(pathname: string, paths: string[]): boolean {
+  return paths.some((path) => {
```
malinskibeniamin: **P3 — `matchesFullscreenPath` matches the fullscreen path anywhere in the pathname (interior segments included)**

`matchesFullscreenPath` uses `pathname.indexOf(path)` and only checks the char *after* the match, so with `paths=['/sql']` a pathname like `/clusters/sql/overview` also returns true. Only `/sql` is fullscreen today and no real route has an interior `sql` segment, so this is latent — but any future route/segment literally named `sql` would falsely strip page chrome.

**Suggested fix:** require the match to be the trailing segment (e.g. `pathname === path || pathname.endsWith(path)` with a `/` boundary before it) rather than `indexOf` anywhere.

**One-shot:** In `frontend/src/utils/fullscreen-routes.ts` `matchesFullscreenPath`, only match when the fullscreen path is the trailing segment of `pathname`; add tests for `/clusters/sql/overview` (false) vs `/clusters/abc/sql` (true).

frontend/src/components/pages/sql/catalog-tree.tsx
```diff
@@ -0,0 +1,580 @@
...
+          disabled={!allowed}
+          onClick={() => toggleTable(table.id)}
+          role="treeitem"
+          type="button"
+        >
+          <Chevron className="shrink-0 text-disabled" size={13} />
+          <LucideTable className={tableIcoClass} size={13} />
+          <span className={LABEL}>{table.name}</span>
+          {isIceberg ? (
+            <Badge className="uppercase tracking-wide" size="sm" variant="info-inverted">
```
malinskibeniamin: **P3 — Iceberg badge is force-uppercased here but not in the wizard (inconsistent product-name casing)**

This row renders the Iceberg badge with `className="uppercase tracking-wide"` → "ICEBERG", while `sql-wizard.tsx`'s `IcebergBadge` renders it normally ("Iceberg"). "Iceberg" is a product name; the two entry points should match.

**Suggested fix:** drop `uppercase tracking-wide` from the two badges here (≈L337 and L343) to match the wizard, or apply the same styling in both places.

**One-shot:** In `frontend/src/components/pages/sql/catalog-tree.tsx`, remove `uppercase tracking-wide` from the two Iceberg `<Badge>`s (≈L337, L343) so the casing matches `sql-wizard.tsx`'s `IcebergBadge`; verify with `cd frontend && bun run lint`.

frontend/src/components/layout/header.tsx
```diff
@@ -69,8 +69,14 @@ function BreadcrumbHeaderRow({ useNewSidebar, breadcrumbItems }: BreadcrumbHeade
...
 }
 
-function AppPageHeader() {
+function AppPageHeader({ breadcrumbOnly = false }: { breadcrumbOnly?: boolean }) {
   useApiStoreHook((s) => s.userData); // re-render when userData changes
+  // Fullscreen routes (e.g. the SQL studio) carry their own title bar/toolbar, so
+  // they never want the title+actions row — only the breadcrumb. Robust to which
+  // layout branch renders the header (standalone vs embedded misdetection).
+  const matches = useMatches();
+  const isFullscreenRoute = matches.some((m) => m.staticData.fullscreen);
```
malinskibeniamin, line 78: nice 👍

frontend/package.json
```diff
@@ -53,20 +53,26 @@
...
     "@chakra-ui/object-utils": "^2.1",
     "@chakra-ui/popper": "^2.1",
     "@chakra-ui/portal": "^2.1",
     "@chakra-ui/react-use-disclosure": "^2.1",
     "@chakra-ui/system": "^2.1",
+    "@codemirror/autocomplete": "^6.20.3",
+    "@codemirror/lang-sql": "^6.10.0",
+    "@codemirror/language": "^6.12.3",
+    "@codemirror/state": "^6.6.0",
+    "@codemirror/view": "^6.43.1",
```
malinskibeniamin, line 68: What's our bundle size looking like after this?

frontend/package.json
```diff
@@ -193,6 +202,13 @@
     "vitest-browser-react": "^2.2.0"
   },
   "overrides": {
+    "@codemirror/autocomplete": "^6.20.3",
+    "@codemirror/commands": "^6.10.3",
+    "@codemirror/language": "^6.12.3",
+    "@codemirror/lint": "^6.9.7",
+    "@codemirror/search": "^6.7.0",
+    "@codemirror/state": "^6.6.0",
+    "@codemirror/view": "^6.43.1",
```
malinskibeniamin: Not sure if we need the overrides if it just got added. Maybe it was overkill, I think there should be no security updates for packages that are modern/brand new/recently included

.gitignore
```diff
@@ -17,7 +18,14 @@ requests.txt
...
 
-.cursor
+# Go workspace (local dev only)
+go.work
+go.work.sum
+
+# Claude Code local state
+.claude/worktrees/
+.claude/settings.local.json
 .claude/worktrees/
```
malinskibeniamin: I think it's already in gitignore from before (fine either way)

malinskibeniamin (approved):
Left some comments

frontend/package.json
c-julin: the code mirror and react grid alone is about 150 kb extra according to claude gzipped

frontend/package.json
c-julin: this version is about a year old, has what we need anything stable is like 6 years old i think the beta is the right one to use can use the latest beta which is about 6 months old

PR #2497, opened 2026-06-11
fix(frontend): fetch all pages in aigw MCP server and LLM provider lists

birdayz:
## What

Make the aigw v2 list hooks — `useListAigwMCPServersQuery` and `useListLLMProvidersQuery` — walk `next_page_token` until exhausted instead of reading a single page. Applies to every consumer: the AI agent list page, the create page, and the agent configuration tab.

## Why

aigw pages its list endpoints (default 50, ordered `created_at` desc) and returns a `next_page_token`. Both hooks did one call with `pageSize: 50` and threw the token away. Once a tenant crosses 50 MCP servers, everything older than the newest page silently disappears: the agent configuration tab renders the agent's stale server references as disabled remotes (the lookup against the truncated list fails), and the add-server picker can't offer the missing servers at all.

Surfaced on a tenant with 82 MCP servers: `servicenow-catalog` and `servicenow-knowledge` at positions 74/75 in newest-first order, permanently outside the first page.

Silently truncating a list is worse than failing. The server was saying there were more pages; the client didn't listen.

## Implementation details

**Reuse the existing helper.** Both hooks switch to `useInfiniteQueryWithAllPages` — the same auto-paginating wrapper the dataplane MCP list (`react-query/api/remote-mcp.tsx`) already uses — and flatten the pages. Consumers keep the same `data.mcpServers` / `data.llmProviders` shape.

**page_size stays unset.** The server owns page-size policy (default 50, oversized requests clamped); the token walk is the correctness mechanism. If the backend tunes its default, clients follow for free.

**Loading semantics.** `isLoading` holds until the last page lands, so consumers never render a truncated list as if it were complete. A failing page stops the walk and surfaces the error instead of retry-looping.

**Tests.** Hook-level suites pin the full contract for both endpoints: multi-page walk and flatten order, token echo sequence, page_size unset on every request, filter forwarded per page, single-page and empty-tenant cases, disabled query, mid-walk failure, and loading-until-final-page. Plus a page-level regression on the agent list: an agent whose MCP server lives beyond page one resolves its tools.

## References

Same bug fixed in cloudv2 adp-ui (agents v2): redpanda-data/cloudv2#27046

### DIFF
--- a/frontend/src/components/pages/agents/list/ai-agent-list-page.test.tsx
+++ b/frontend/src/components/pages/agents/list/ai-agent-list-page.test.tsx
@@ -220,6 +220,64 @@ describe('AIAgentsListPage', () => {
     expect(screen.getByText('gpt-3.5-turbo')).toBeVisible();
   });
 
+  test('should resolve MCP tools for servers beyond the first page of the aigw list', async () => {
+    const agent = create(AIAgentSchema, {
+      id: 'agent-1',
+      displayName: 'Test Agent 1',
+      description: 'Description 1',
+      state: AIAgent_State.RUNNING,
+      provider: {
+        provider: {
+          case: 'openai',
+          value: {
+            apiKey: 'secret-1',
+          },
+        },
+      },
+      model: 'gpt-4',
+      systemPrompt: 'You are helpful',
+      // The referenced server is NOT on the first page of the MCP list —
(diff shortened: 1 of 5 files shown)


birdayz:
@claude review this PR

malinskibeniamin (approved):

PR #2498, opened 2026-06-11
proto(sql): add catalog_namespace to table RPCs

r-vasquez:
Catalog names are only unique within a
namespace, so resolving a catalog by name alone
is ambiguous when the same name exists in more
than one namespace. Add an optional
catalog_namespace field to ListTablesRequest and
DescribeTableRequest to disambiguate. When unset,
ListTables returns tables from all matching
catalogs (rows carry Table.namespace_name) while
DescribeTable fails with INVALID_ARGUMENT and a
documented 400 response.

Also reword Column.type docs to the
engine-reported names (e.g. "integer", "text[]")
instead of Postgres driver names.


proto/redpanda/api/dataplane/v1alpha3/sql.proto
```diff
@@ -128,12 +142,15 @@ message Catalog {
...
+// Table mirrors a row from `SHOW REDPANDA TABLES <catalog>`. Backing-specific
 // fields (connection_name, topic_name, *_policy) are only populated for
 // Kafka-backed tables.
 message Table {
   reserved 1;
   reserved "database_name";
+  // Namespace of the catalog that owns this table. Namespaces attach to
+  // catalogs (connections), not to tables; tables of same-named catalogs in
+  // different namespaces are distinguished by this field.
   string namespace_name = 2;
```
r-vasquez: Wondering if this should be called catalog_namespace_name @c-julin, wdyt? Are breaking changes allowed in alphav3, or will the CI break?

r-vasquez:
Commit: 3b440b2f41713c6f57af43ee608c43bae0f52c3e includes the breaking changes. It's an alpha version so we should be fine as there are no current users of this one.

c-julin (approved):

PR #2502, opened 2026-06-13
fix(frontend): bound long-lived-tab renderer memory growth

malinskibeniamin:
Bounds renderer-memory growth on long-lived Console tabs (the BYOC Chrome-OOM class). Code-only; investigation notes are kept out of the PR.

Landed:
- **api.errors ring buffer** — was an unbounded append from ~20 legacy refresh paths; on a long-lived tab with auto-refresh + a failing endpoint it retained an Error (stack + Response body) per failure forever. Now capped (50).
- **live-tail / filtered message buffer sliding window** — the `End`-origin / push-down-filter stream (30-min window) appended to `messages[]` with no cap; now a 50k sliding window, gated to the live path so pagination is untouched.
- New unit-tested `bounded-array` util (`boundedAppend`, `trimToLast`).

More memory fixes (log-view buffers, module caches, Monaco disposal, store/persist bounds, dependency dedupe/drop) land in follow-up commits.

### DIFF
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -51,7 +51,6 @@
         "cmdk": "^1.1.1",
         "date-fns": "^4.1.0",
         "dexie": "^4.2.1",
-        "dexie-react-hooks": "^4.2.0",
         "dotenv": "^17.2.3",
         "es-cookie": "^1.5.0",
         "framer-motion": "^7.10.3",
@@ -1933,8 +1932,6 @@
     "devlop": ["devlop@1.1.0", "", { "dependencies": { "dequal": "^2.0.0" } }, "sha512-RWmIqhcFf1lRYBvNmr7qTNuyCt/7/ns2jbpp1+PalgE/rDQcBT0fioSMUpJ93irlUhC5hrg4cYqe6U+0ImW0rA=="],
 
     "dexie": ["dexie@4.2.1", "", {}, "sha512-Ckej0NS6jxQ4Po3OrSQBFddayRhTCic2DoCAG5zacOfOVB9P2Q5Xc5uL/nVa7ZVs+HdMnvUPzLFCB/JwpB6Csg=="],
-
-    "dexie-react-hooks": ["dexie-react-hooks@4.2.0", "", { "peerDependencies": { "@types/react": ">=16", "dexie": ">=4.2.0-alpha.1 <5.0.0", "react": ">=16" } }, "sha512-u7KqTX9JpBQK8+tEyA9X0yMGXlSCsbm5AU64N6gjvGk/IutYDpLBInMYEAEC83s3qhIvryFS+W+sqLZUBEvePQ=="],
 
     "diff": ["diff@4.0.4", "", {}, "sha512-X07nttJQkwkfKfvTPG/KSnE2OMdcUCao6+eXF3wmnIQRn2aPAHH3VxDbDOdegkd6JbPsXqShpvEOHfAT+nCNwQ=="],
 
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -61,7 +61,6 @@
         "jwt-decode": "^4.0.0",
         "lottie-react": "^2.4.1",
         "lucide-react": "^1.17.0",
-        "moment": "^2.30.1",
         "monaco-editor": "^0.54.0",
         "monaco-editor-webpack-plugin": "^7.1.1",
         "monaco-yaml": "^5.5.0",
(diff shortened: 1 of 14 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Thanks for the review — addressed the actionable items:

- **🟡 `trimToLast` O(n) per message** (`ad2b4acd9`): trims only once the buffer drifts a slack margin (`LIVE_TAIL_TRIM_SLACK = 1024`) past the cap, so the `splice` is amortized to ~once per slack messages instead of every message past the cap. Memory ceiling stays cap + slack.
- **BSL headers on util tests** (`3267655ae`): added to `bounded-array.test.ts` and `lazy-map.test.ts` for consistency.
- **`LazyMap.has()` LRU note** (`3267655ae`): documented that only `get()` refreshes recency.

Not changed (by design, as you noted):
- **LRU vs in-flight dedup** — agreed it's negligible: `get()` bumps recency on every access, so eviction of a *pending* entry needs 500 other distinct URLs within one request's lifetime. Leaving as-is.
- **`evictOldest` `while` vs `if`** — kept the loop (robust against future `maxSize` changes), as you suggested.

`type:check` + unit/integration tests green locally; CI running.

malinskibeniamin:
All three review items are addressed in the current branch:

1. **`trimToLast` O(n) per message** — amortized: `LIVE_TAIL_TRIM_SLACK = 1024` (`backend-api.ts:2709`) and the trim now only runs when `messages.length > LIVE_TAIL_MAX_MESSAGES + LIVE_TAIL_TRIM_SLACK` (`:2855`), so the splice is ~once per slack window. Memory ceiling stays cap + slack.
2. **Missing BSL headers** — `bounded-array.test.ts` and `lazy-map.test.ts` both carry the copyright header now.
3. **`LazyMap.has()` recency note** — the class doc (`lazy-map.ts:20`) states recency is refreshed by `get()` only, not `has()`.

The LRU-eviction-vs-dedup and `evictOldest` while-loop notes were flagged as awareness-only — no change. CI is green.

malinskibeniamin:
## `/review` — 7-hat audit (vs `origin/master`)

**Verdict: not blocking.** Standards ✓, Security/privacy ✓ (all changes *reduce* retained surface), Visual ✓ (no rendered-output change), Spec n/a. Value: **HIGH**. 2 medium + 2 low findings, all in new code.

**🟠 M1 — F4 `perTopicSettings` cap is incomplete + mislabeled** (3 hats agree)
`ui-state.ts` caps the `setCurrentTopicName` path, but the **persisted dominant store is the zustand `useTopicSettingsStore`** (`topic-settings-store.ts`, key `uiSettings-v3`), which appends **uncapped** at 6+ sites (sort/preview changes) and whose `subscribe → syncPerTopicSettingsToMobX` overwrites the MobX array — so the cap can be re-inflated and the persisted/localStorage growth is unbounded. Also `boundedAppend` is FIFO-by-first-visit, not the LRU the comment implies. → **fixing**: cap the zustand store path + correct wording.

**🟠 M2 — the headline live-tail trim is untested**
`backend-api.ts` `createMessageSearch` streaming loop (`isLiveTail` gate + 50k+1024 slack trim) is the dominant memory fix, but only the `trimToLast` primitive is unit-tested. → **fixing**: add an integration test driving the stream.

**🟡 L1 — LazyMap can evict a pending `CacheEntry`** → benign duplicate refetch (needs 500+ distinct URLs, self-heals). Optional follow-up.
**🟡 L2 — coverage gaps**: LazyMap↔`cachedApiRequest`, the `perTopicSettings` store path, Monaco disposal, `refreshTopics` error/empty. Optional follow-ups.

Applying M1 + M2 now; leaving the lows as optional.

malinskibeniamin:
Both medium findings from the `/review` are applied:

- **M1 — persisted topic-settings store now capped** (`412150242`): the dominant vector was the zustand `useTopicSettingsStore` (key `uiSettings-v3`), not just the `setCurrentTopicName` path. Added a trim-subscriber that bounds `perTopicSettings` to `MAX_PER_TOPIC_SETTINGS` (200) on every change — registered after the MobX sync so the mirror ends up trimmed; the length guard makes the re-entrant `setState` terminate. New `topic-settings-store.test.tsx` verifies the cap (oldest dropped, newest kept). Clarified the `ui-state.ts` comment as FIFO.
- **M2 — live-tail trim now tested** (`1dfc546cf`): extracted the push + amortized-slack-trim into `appendWithSlackCap` (`bounded-array.ts`) with direct unit tests (settles at cap, keeps newest, only trims past `max+slack`); the `createMessageSearch` `data` case uses it. Behavior-neutral (same 50k + 1024 ceiling).

Lows (L1 pending-evict refetch, L2 remaining coverage) left as optional follow-ups as noted.

`type:check` clean, 799 unit + integration tests green locally; CI running.

jvorcak (approved):

PR #2503, opened 2026-06-13
fix(frontend): construct protobuf messages with create(Schema)

malinskibeniamin:
## What & why

Audit finding **DATA-01**. Three call sites hand-built protobuf messages as object
literals carrying a hardcoded `$typeName` string instead of the protobuf-es **v2**
`create(Schema, …)` API:

- `transcript-list-page.tsx` — `AttributeFilter[]` (4 presets + user filters) and the returned `ListTracesRequest_Filter`
- `acl-model.tsx` — `convertRulesToCreateACLRequests` built `CreateACLRequest` objects from a `baseRule` Pick that hardcoded `$typeName`

A proto rename or package-version bump (`v1alpha3` → `v1`) would leave these literals structurally valid to `tsgo` while the runtime message no longer matched the registered schema — breaking `anyPack`/serialization/equality — and omitted fields skipped proto defaults.

## Change

Use `create(AttributeFilterSchema, …)`, `create(ListTracesRequest_FilterSchema, …)`, and `create(CreateACLRequestSchema, …)` (protobuf-es v2). `$typeName` is dropped from the literals and from the `baseRule` shape; messages now carry it from the schema.

## Verification
- `bun run type:check` ✅
- `bun run lint:check` ✅
- `acl-model` / `create-acl` / `acl-detail` tests (29) ✅ and `transcript-list-page` tests (6) ✅
- `grep '$typeName'` in both files → 0

### DIFF
--- a/frontend/src/components/pages/security/shared/acl-model.tsx
+++ b/frontend/src/components/pages/security/shared/acl-model.tsx
@@ -1,3 +1,4 @@
+import { create } from '@bufbuild/protobuf';
 import type { ConnectError } from '@connectrpc/connect';
 import {
   ACL_Operation,
@@ -5,6 +6,7 @@ import {
   ACL_ResourcePatternType,
   ACL_ResourceType,
   type CreateACLRequest,
+  CreateACLRequestSchema,
   type ListACLsResponse,
 } from 'protogen/redpanda/api/dataplane/v1/acl_pb';
 import type { ReactNode } from 'react';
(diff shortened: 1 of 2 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Addressed the optional nits from the review:

- **Redundant default fields** — dropped `value: ''` on the IN filter and `values: []` on the EQUALS/NOT_EQUALS `AttributeFilter`s, since `create(AttributeFilterSchema, …)` supplies those proto defaults. Kept the IN filter's real `values: ['chat', 'text_completion']`.
- **Audit completeness** — swept the rest of `frontend/src` for other object-literal `$typeName:` messages: **0 remaining** in non-generated code, so this PR closes the class for the frontend.

`type:check` + `lint` + transcript/acl tests pass.

c-julin (approved):

PR #2504, opened 2026-06-13
fix(frontend): add accessible names to icon-only buttons

malinskibeniamin:
## What & why

Audit finding **A11Y-02**. Icon-only buttons were announced as a bare "button" to
screen readers (WCAG 4.1.2 Name, Role, Value). The codebase already has the correct
pattern elsewhere (`secrets-store-actions` uses `sr-only`, `schema-list` uses
`aria-label`), so these are inconsistent omissions, not a missing capability.

## Change

Added `aria-label` to the verified-cited icon-only buttons:

| Location | Button | Label |
|----------|--------|-------|
| `transcripts/components/transcript-filter-bar` | refresh | "Refresh" |
| `security/tabs/acls-tab` | ACL delete trigger | "Delete ACL" |
| `security/tabs/users-tab` | user actions menu | "Open user actions" |
| `shadowlinks/details/tasks-table` | refresh | "Refresh tasks" |
| `shadowlinks/details/shadow-topics-table` | clear filter / refresh | "Clear filter" / "Refresh topics" |
| `rp-connect/pipeline` | back | "Go back" |

## Verification
- `bun run type:check` ✅
- `bun run lint:check` ✅
- `rp-connect/pipeline` integration tests (23) ✅

## Follow-up (not in this PR)
A broader sweep found ~77 icon-only buttons across the app lacking an accessible
name. This PR fixes the audit-cited, highest-traffic controls; the remaining sweep
(plus a lint rule to enforce it) is a separate change so this stays reviewable.

### DIFF
--- a/frontend/src/components/pages/rp-connect/pipeline/index.tsx
+++ b/frontend/src/components/pages/rp-connect/pipeline/index.tsx
@@ -970,7 +970,7 @@ function PipelinePageContent() {
       ) : null}
       {mode === 'view' && !pipeline ? (
         <div className="flex items-center gap-2">
-          <Button className="-ml-3.5 shrink-0" onClick={handleCancel} size="icon" variant="ghost">
+          <Button aria-label="Go back" className="-ml-3.5 shrink-0" onClick={handleCancel} size="icon" variant="ghost">
             <ArrowLeftIcon className="h-5 w-5" />
           </Button>
           <Skeleton variant="text" width="md" />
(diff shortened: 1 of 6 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Addressed the review:

- **Per-row unique labels (the "worth fixing" point)** — `acls-tab` Delete button is now `aria-label={`Delete ACL for ${record.principalName}`}` and `users-tab` actions trigger is `aria-label={`Open actions for ${user.name}`}`, matching the `roles-tab.tsx` convention so screen-reader users can distinguish rows.
- **`users-tab` placement** — confirmed; the `aria-label` stays on the inner `<button>` that `Button asChild` renders.
- **Tests / tooltips / label-consistency** — left for the broader follow-up sweep as you suggested (these files have no existing test harness, and the sighted-tooltip gap is cheapest to fold into that sweep). Label variation ("Refresh" vs "Refresh topics") is intentional disambiguation.

`type:check` + `lint` pass.

c-julin (approved):

PR #2505, opened 2026-06-13
fix(frontend): add rel=noopener noreferrer to target=_blank links

malinskibeniamin:
## What & why

Audit finding **SECURITY-03**. Links opened with `target="_blank"` but no `rel`
expose **reverse-tabnabbing** (the opened page can drive `window.opener` to a
phishing URL) and referrer leakage. Practical risk is low here (destinations are
the trusted `docs.redpanda.com` domain and modern browsers imply `noopener`), but
it's a latent footgun the moment a `target="_blank"` href becomes dynamic.

## Change

Added `rel="noopener noreferrer"` to the **16** external-link sites (custom
`Link` / `UILink` / router-link components) across 8 files that were missing it.

Native `<a target="_blank">` are already covered by biome's recommended
`noBlankTarget` rule — but that rule only inspects native anchors, so the registry
`Link`/`UILink` consumers needed explicit `rel`. The vendor registry component
(`redpanda-ui/.../auto-form/field-wrapper.tsx`) was left untouched (it already had
`rel` anyway).

## Verification
- `bun run type:check` ✅
- `bun run lint:check` ✅
- Independent sweep: **0** `target="_blank"` elements remain without a `rel` (excluding vendor/generated).

## Follow-up (not in this PR)
Biome's `noBlankTarget` can't enforce `rel` on custom components. A small custom lint
rule (or defaulting `rel` in the upstream registry `Link`) would prevent regressions —
tracked separately.

### DIFF
--- a/frontend/src/components/pages/connect/overview.tsx
+++ b/frontend/src/components/pages/connect/overview.tsx
@@ -163,7 +163,7 @@ class KafkaConnectOverview extends PageComponent<{
               {this.props.isKafkaConnectEnabled
                 ? 'Redpanda Connect is an alternative to Kafka Connect. Choose from a growing ecosystem of readily available connectors.'
                 : 'Redpanda Connect is a data streaming service for building scalable, high-performance data pipelines that drive real-time analytics and actionable business insights. Integrate data across systems with hundreds of prebuilt connectors, change data capture (CDC) capabilities, and YAML-configurable pipelines.'}{' '}
-              <Link href="https://docs.redpanda.com/redpanda-connect/home/" target="_blank">
+              <Link href="https://docs.redpanda.com/redpanda-connect/home/" rel="noopener noreferrer" target="_blank">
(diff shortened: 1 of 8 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Thanks for the review. No code change needed here:

- **Consistency (`noopener noreferrer` vs `noreferrer`)** — both are safe (`noreferrer` implies `noopener`); leaving the explicit form. No action, as noted.
- **Registry default `rel`** — agreed that's the better long-term fix, but it lives in the CLI-installed vendor component (`redpanda-ui/.../typography.tsx`), so it belongs upstream in the registry repo rather than a local edit here. Tracked as the follow-up in the PR description.

The "0 unprotected `target="_blank"` sites" claim was re-verified independently. CI is green.

c-julin (approved):

PR #2506, opened 2026-06-13
test(frontend): make unit + integration test suites warning-free

malinskibeniamin:
## What

Eliminates **every** console warning emitted during `bun run test:unit` and `bun run test:integration` so the suites run clean and the `test-warning-check` / `ci-warning-audit` hooks pass without `TEST_WARNINGS_ALLOW`.

- **Unit:** 51 files / 799 tests — 0 warnings
- **Integration:** 103 files / 1124 tests — 0 warnings

Split out of #2502 (already merged) as requested — pure test-hygiene, no feature behavior change.

### Source fix
- **`tsx-utils` `Button`**: destructure `disabledReason` out of the spread in **both** branches. The no-tooltip branch previously spread the whole prop bag, so the (falsy) `disabledReason` key still reached `RpButton` and leaked to the DOM element — React warns about it as an unknown attribute. This also fixes the leak in the real app, not just tests.

### Test-side `act()` fixes (no source/behavior change)
- **`create-acl.test`**: 5 synchronous tests render an RHF form whose default-value / resolver settling landed after the sync body but before unmount, outside `act()`. Made async and switched the primary assertion to `findByTestId` (polls inside `act`, flushing the pending update).
- **`shadowlink-create-page.test`**: the page's `useForm` uses `mode:'onChange'` + async `zodResolver`; each synchronous `fireEvent.input` kicked off validation that re-rendered subscribers a microtask later, outside `act()`. Await each input's settled value, plus the final `toast.success`, so updates flush inside `act()`.
- **`group-details.test`**: the file's own `afterEach` store reset ran **before** RTL `cleanup()` (vitest hooks are LIFO), re-rendering the still-mounted class component outside `act()`. Wrapped the reset in `act()` and switched initial waits to `findByTestId`.

### Suppressed — external/vendored, not fixable in-repo (each with rationale in `vitest.setup.integration.ts`)
- **`` `ref` is not a prop ``**: React 18.3 warning getter read by the CLI-installed registry `Slot` during `asChild` composition; resolves in React 19.
- **`isInPortal`**: valid `@redpanda-data/ui` Popover prop the component forwards to a DOM node in some render paths.
- **`<button>` descendant of `<button>`**: `@redpanda-data/ui` Accordion renders `heading` inside its trigger button; the consumer-group topic view **intentionally keeps its action controls in that header** — suppressed rather than relocate them into the collapsed panel (a UX regression).

## Test plan
- [x] `bun run test:unit` — 799 passed, 0 warnings
- [x] `bun run test:integration` — 1124 passed, 0 warnings
- [x] `bun run type:check` — clean
- [x] `bun run lint` — no changes produced (pre-existing `vitest.setup` findings unchanged from master)

### DIFF
--- a/frontend/src/components/pages/consumers/group-details.test.tsx
+++ b/frontend/src/components/pages/consumers/group-details.test.tsx
@@ -9,7 +9,7 @@
  * by the Apache License, Version 2.0
  */
 
-import { screen, waitFor } from 'test-utils';
+import { act, screen } from 'test-utils';
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
 
 const { refreshConsumerGroupMock, refreshConsumerGroupAclsMock } = vi.hoisted(() => ({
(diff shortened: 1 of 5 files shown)


malinskibeniamin:
@claude review

c-julin (approved):
🫡

PR #2507, opened 2026-06-16
Migrate produce record to UI registry

jvorcak:
Needs to merge https://github.com/redpanda-data/console/pull/2486 first

Screenshot 2026-07-17 at 15 30 55


frontend/src/components/pages/topics/DeleteRecordsModal/delete-records-modal.tsx
```diff
@@ -72,6 +77,44 @@ function TrashIcon() {
   );
 }
 
+function DeleteRecordsIntro({ children }: { children: React.ReactNode }) {
+  return (
+    <div className="flex items-center gap-4">
+      <TrashIcon />
+      <Text variant="muted">{children}</Text>
```
SpicyPete: 1 general comment across the entire PR.
Now that `<Text>` and `<Heading>` are deprecated, lets ask Claude to replace them with the new text-utility class pattern

frontend/src/components/pages/topics/topic-produce.tsx
```diff
@@ -341,364 +362,344 @@ const PublishTopicForm: FC<{ topicName: string }> = ({ topicName }) => {
...
-        <Box my={6}>
-          <PublishTopicForm topicName={this.props.topicName} />
-        </Box>
-      </Box>
-    );
-  }
-}
+  return (
+    <div className="flex flex-col gap-2">
+      <Heading level={1}>Produce Kafka record</Heading>
```
SpicyPete: Minor: In the UI we are saying "Produce" then "Produce Kafka record" right below, which feels like we can maybe remove the first "Produce" line.

Image

SpicyPete (commented):
1. Lets not add new <Text and <Heading components in this and prefer text-utility classes
2. The Delete Records model is a bit confusing to me.
  2a. It's not clear this is a multi-step dialog, and since the action is destructive red it seems like it may cause a destructive action.  Suggest we make the steps more clear (or that there are multiple steps) at a  dialog level
  2b. Perhaps we can show the selected Topic Name in the dialog text itself as opposed to always "your topic"
3.
  3a. The manual offset only has the low number labeled, and some small alignment issues.  
  3b. In the top text we are saying "choose the new low offset for your selected partitions", but in the previous step we can only select 1 partition, so maybe shouldn't be plural?
  3c. Showing the raw timestamp is likely correct, but it might be better for usability if we also show the human readable date.
4.  Minor:  The produce record page feels very spaced out, but redoing this page is likely beyond scope of the PR.

Just checked this against the current version, and it's much better now, even without the above requests

#### Figures
##### 2a / 2b
Image

##### 3a
Image

##### 3c
Image

jvorcak:
Screenshot 2026-07-20 at 14 07 46
 
@SpicyPete I've tried to change it a bit, even though I think it could be completely re-thought in a different PR.

SpicyPete:
Screenshot 2026-07-20 at 8 54 51 AM
 
 Better, but I think it could use a bit more care, for example the slider isn't showing the range labels for me, showing that there are multiple steps as good.
 
 A bit more polish on this, then I agree further updates can be done as follow-up work

frontend/src/components/pages/topics/DeleteRecordsModal/delete-records-modal.tsx
```diff
@@ -243,100 +268,119 @@ const ManualOffsetContent = ({
...
   const { marks, min, max } = getMarks(partition);
+  // Default to the partition's low watermark — offset 0 is below the valid range.
+  const [sliderValue, setSliderValue] = useState(min);
+
+  // Report the low-watermark default to the parent so submitting without moving the
+  // slider uses a valid in-range value instead of 0.
+  // biome-ignore lint/correctness/useExhaustiveDependencies: report the initial default once on mount
+  useEffect(() => {
+    onOffsetSpecified(min);
+  }, []);
```
eblairmckee, line 336: minor: I feel like we could get around having to use a useEffect here by just passing more things into this component via props

frontend/src/components/pages/topics/DeleteRecordsModal/delete-records-modal.tsx
```diff
@@ -243,100 +268,119 @@ const ManualOffsetContent = ({
...
-        value={sliderValue}
-      />
-    </Flex>
+              </span>
+            ))}
+          </div>
+        ) : null}
+      </div>
+      <div className="flex w-28 shrink-0 flex-col gap-1">
+        <label className="text-caption text-muted-foreground" htmlFor="manual-offset-input">
```
eblairmckee, line 358: nit: before doing too many more of these I would update the registry and start using the new typography utility classes https://redpanda-ui-registry.netlify.app/docs/typography#usage

eblairmckee (approved):

PR #2513, opened 2026-06-17
Ux 1341 remove rest for topic listing

jvorcak:


malinskibeniamin (approved):

c-julin (changes requested):
Not to be a pain but this is a change we dont want to make on the dataplane api :( adding log dirs. Its been a long time ticket which i haven't done yet to split this out, if we can wait until next week i'll try prioritise it

proto/redpanda/api/dataplane/v1/topic.proto
```diff
@@ -143,6 +143,26 @@ message ListTopicsResponse {
...
+    repeated ReplicaError replica_errors = 2;
+    // Optional human-readable hint explaining a partial or missing result.
+    string hint = 3;
+  }
+  // ReplicaError describes a log dir query error for a single broker.
+  message ReplicaError {
+    // Broker that returned the error.
+    int32 broker_id = 1;
+    // Error message returned by the broker.
+    string error = 2;
```
malinskibeniamin: can we make this proto have more information, so it's properly validated, maybe CEL expressions too, and using correct types like `cleanup_policy` I imagine can be more strict than a string

jvorcak:
> Not to be a pain but this is a change we dont want to make on the dataplane api :( adding log dirs. Its been a long time ticket which i haven't done yet to split this out, if we can wait until next week i'll try prioritise it

@c-julin no problem feel free to update me or we can get in touch in DM

PR #2520, opened 2026-06-19
config: inject Bearer token on config.fetch

r-vasquez:
REST helpers that call config.fetch directly,
rather than going through the rest<T>() wrapper,
sent no Authorization header. Under Module
Federation v2 the Cloud UI host hands Console the
access token via config.jwt instead of a
pre-authenticated fetch, so those requests went
out unauthenticated.

Wrap the base fetch so every config.fetch call
attaches the same Bearer token the gRPC
interceptor uses. The token is read lazily so
in-place host refreshes are picked up, and an
Authorization header already set by a legacy
host-provided (V1) fetch is left untouched.

### DIFF
--- a/frontend/src/config.test.tsx
+++ b/frontend/src/config.test.tsx
@@ -1,0 +1,82 @@
+/**
+ * Copyright 2026 Redpanda Data, Inc.
+ *
+ * Use of this software is governed by the Business Source License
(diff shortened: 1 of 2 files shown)


Mateoc (approved):
LGTM

PR #2523, opened 2026-06-22
proto(sql): add GetSqlIdentity RPC to v1alpha3 SQLService [UX-1330]

c-julin:
## What

Adds a `GetSqlIdentity` RPC to the v1alpha3 `SQLService` plus regenerated Go/TS clients and v1alpha3 OpenAPI.

```proto
rpc GetSqlIdentity(GetSqlIdentityRequest) returns (GetSqlIdentityResponse);

message GetSqlIdentityResponse {
  string current_user = 1; // engine username (SQL current_user)
  bool   is_admin = 2;     // superuser? sourced from pg_roles.rolsuper
}
```

Gated on `PERMISSION_VIEW` / `API_REDPANDA_SQL`, same as the other SQL RPCs; HTTP `GET /v1alpha3/sql/identity`.

## Why

The SQL editor UI needs to distinguish a SQL **admin** from a **viewer** to show/hide write actions (e.g. the create-table wizard). There is currently no signal for this — the frontend hardcodes `sqlRole = 'viewer'`. The engine (Oxla) is the source of truth: admin == `rolsuper`. This RPC surfaces that so the UI can gate the button on a real capability rather than assuming a role. The engine still enforces privileges on execution regardless.

Validated against a live Oxla cluster: `SELECT rolname, rolsuper FROM pg_roles WHERE rolname = current_user` returns `t` for an admin (`oxla`) and `f` for a created viewer.

## Notes

- Generated files only; the handler is implemented enterprise-side (console-enterprise PR, stacked on the in-flight SQL work).
- Regeneration via `task proto:generate`; unrelated whitespace churn in v1alpha2 OpenAPI examples was reverted (CI uses `git diff -w`).

Ref UX-1330.


c-julin:
Superseded by https://github.com/redpanda-data/console/pull/2524 (renamed branch to `sql-get-identity` to match the enterprise PR and the `sql-*` branch family).

PR #2525, opened 2026-06-23
feat(frontend): migrate to React 19 via Module Federation bridge

malinskibeniamin:
> Draft for the UX team to review first.

## What

Upgrade **Console's frontend** from **React 18.3.1 → React 19.2.x** while keeping **Module Federation** working with the **Cloud UI** host, which stays on **React 18**.

## Why the obvious upgrade breaks MF

`rp_console` shares `react`/`react-dom`/`@tanstack/*`/`react-hook-form`/`zod` as **singleton** modules with Cloud UI. With a singleton, MF picks **one** React for the whole page — the React-18 host wins and feeds it to the remote. React-19-compiled remote code then runs against the React-18 runtime → "invalid hook call". **You cannot run two React majors under `singleton: true`.**

This is the same problem adp-ui solved in [cloudv2#27303](https://github.com/redpanda-data/cloudv2/pull/27303); admin-ui ([cloudv2#27324](https://github.com/redpanda-data/cloudv2/pull/27324)) is the simpler standalone reference.

## The fix: `@module-federation/bridge-react@2.5.1`

The bridge changes the host↔remote contract from "remote returns a React element rendered into the host tree" to "remote returns a `{ render, destroy }` object that mounts into a host-supplied DOM node using its **own** React." That decouples the two React majors.

- **`module-federation.config.ts`**: `shared: {}` (nothing shared) so Console bundles its own React 19, react-query, router, etc.
- **New `./BridgeApp` expose** (`src/federation/console-federated-bridge.tsx`) — `createBridgeComponent` from `@module-federation/bridge-react/v19` (the default entry calls the legacy `render` API and throws on React 19). Wraps the unchanged `ConsoleApp`; same `ConsoleAppProps` contract.
- **`./App` becomes a React-18-safe legacy shim** (`src/federation/console-legacy-app.tsx`, ported from adp-ui). It returns a React-18 element shape whose ref mounts the real React 19 tree via Console's own `createRoot` — **so the current Cloud UI host keeps rendering `rp_console/App` with no Cloud UI change**. New hosts move to `./BridgeApp`.
- React Compiler `target: '18' → '19'`.

## The non-obvious parts (Console-specific, not present in adp-ui)

adp-ui uses the vendored UI registry; Console uses the `@redpanda-data/ui@4.2.0` package and a few legacy deps, which surfaced extra React-19 work:

1. **`react-router-dom` alias override.** Installing `@module-federation/bridge-react` auto-activates a webpack plugin that aliases `react-router-dom$` to its own router shim. Because Console declares no direct `react-router-dom`, the plugin falls back to its **v6 shim** (only exports `BrowserRouter`/`RouterProvider`) and breaks `@redpanda-data/ui`, which imports `NavLink`/`Link` from the real react-router-dom@7. Console doesn't federate routing (it uses `@tanstack/react-router`), so `rsbuild.config.ts` points `react-router-dom$` back at the real package (the plugin spreads the user alias last, so it wins).
2. **`react-beautiful-dnd` → `@hello-pangea/dnd`.** rbd pulls `react-redux@7`, which calls `ReactDOM.unstable_batchedUpdates` (removed in React 19) at module init → DnD would crash at runtime. `@hello-pangea/dnd` is the maintained drop-in (identical API). 2 files.
3. **`framer-motion@7` → `motion/react`.** framer-motion@7's `motion.*` typings break under `@types/react@19` (props resolve to `unknown`). Console already ships `motion@12`, so the 7 `framer-motion` imports move to `motion/react`.
4. **`react-onclickoutside` build shim.** Statically imports `findDOMNode` (removed in React 19) via `@redpanda-data/ui → react-datepicker`; breaks ESM linking. Console renders no datepicker → identity-HOC shim (mirrors admin-ui).
5. **Global `JSX` namespace shim.** `@types/react@19` removed the global `JSX` namespace. `@redpanda-data/ui` bundles `react-markdown@8`, whose `complex-types.ts` ships as source (so `skipLibCheck` can't skip it) and uses `keyof JSX.IntrinsicElements`. A type-only ambient declaration restores `JSX` as an alias of `React.JSX`.
6. **Dedupe `@types/react`/`memoize-one`** via `overrides`/`resolutions` (the dnd swap had re-hoisted an untyped `memoize-one@4`; a stray `@types/react@18` lingered under two `@types/*` deps).
7. App-code React-19 type fixes via `types-react-codemod` (scoped JSX, `useRef`, `RefObject`).

## Verification (all Console `frontend/` CI gates, run locally)

| Gate | Result |
|---|---|
| `bun install --frozen-lockfile` | ✅ bun.lock + yarn.lock synced |
| `bun run type:check` (tsgo) | ✅ 0 errors |
| `bun run lint` (ultracite) | ✅ exit 0, idempotent (clean re-run) |
| `bun run test:unit` | ✅ 832 passed (54 files) |
| `bun run test:integration` | ✅ 1179 passed (111 files); zero **new** warnings (the one `act()` warning in `catalog-tree.test.tsx` pre-exists on `master`) |
| `bun run build` (rsbuild) | ✅ manifest exposes `./App` + `./BridgeApp`, `shared: []`; no findDOMNode link error |
| `bun run doctor` (react-doctor) | ✅ exit 0, detects React ^19.2.0 |
| new federation tests | ✅ 9 passed (config contract + legacy-shim lifecycle) |

## Cloud UI coordination (please read before merge)

- **`rp_console/App`** keeps working unchanged via the legacy shim — the safety net, no hard cutover.
- **`rp_console/BridgeApp`** is the new primary path. A coordinated **cloudv2 / Cloud UI** PR switches `console-loader` to consume it via `createRemoteAppComponent` (mirrors the adp-loader change). _(Companion PR follows.)_
- ⚠️ **`rp_console/connect-tiles`**: the serverless onboarding wizard renders `ConnectTiles` directly in Cloud UI's React-18 tree **and holds an imperative `triggerSubmit()` ref**, which the element-shape shim can't carry. It is gated behind the **default-off** `enable-serverless-onboarding-wizard` flag, so this draft is safe to review without it; the fix is tracked as a scoped follow-up in the companion PR.

## Known follow-ups
- `react-compiler-runtime` is now unused with `target: '19'` (React 19 ships `react/compiler-runtime`) — left in place to keep this diff focused.
- Pre-existing `act()` warning in `catalog-tree.test.tsx` (SQL studio) is unrelated to this change.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -70,12 +70,14 @@
     "@connectrpc/connect-query": "^2.2.0",
     "@connectrpc/connect-web": "^2.1.0",
     "@emotion/css": "^11.13.5",
+    "@hello-pangea/dnd": "^18.0.1",
     "@hookform/resolvers": "^5.2.2",
     "@icons-pack/react-simple-icons": "^13.8.0",
     "@lezer/highlight": "^1.2.3",
     "@milkdown/kit": "^7.18.0",
     "@milkdown/react": "^7.18.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
+    "@module-federation/bridge-react": "2.5.1",
     "@module-federation/runtime": "^2.5.1",
     "@monaco-editor/react": "^4.7.0",
     "@redpanda-data/ui": "^4.2.0",
@@ -101,7 +103,6 @@
     "dexie": "^4.2.1",
     "dotenv": "^17.2.3",
     "es-cookie": "^1.5.0",
-    "framer-motion": "^7.10.3",
     "hast": "^1.0.0",
     "hast-util-to-jsx-runtime": "^2.3.6",
     "js-base64": "^3.7.8",
@@ -117,12 +118,11 @@
     "pretty-bytes": "^5.6.0",
     "pretty-ms": "^7.0.1",
     "prismjs": "^1.30.0",
-    "react": "^18.3.1",
-    "react-beautiful-dnd": "^13.1.1",
+    "react": "^19.2.0",
     "react-compiler-runtime": "^1.0.0",
     "react-data-grid": "7.0.0-beta.47",
     "react-day-picker": "^9.14.0",
-    "react-dom": "^18.3.1",
+    "react-dom": "^19.2.0",
     "react-dropzone": "^15.0.0",
     "react-highlight-words": "^0.21.0",
     "react-hook-form": "^7.76.1",
@@ -173,9 +173,8 @@
     "@testing-library/user-event": "^14.6.1",
     "@types/json-bigint": "^1.0.4",
     "@types/node": "^22.19.1",
-    "@types/react": "^18.3.26",
-    "@types/react-beautiful-dnd": "^13.1.8",
-    "@types/react-dom": "^18.3.7",
+    "@types/react": "^19.2.17",
+    "@types/react-dom": "^19.2.2",
     "@types/react-highlight-words": "^0.20.0",
     "@types/react-syntax-highlighter": "^15.5.13",
     "@typescript/native-preview": "^7.0.0-dev.20260108.1",
@@ -204,10 +203,12 @@
   "overrides": {
     "dompurify": "^3.4.0",
     "prismjs": "^1.30.0",
-    "baseline-browser-mapping": "2.10.33"
+    "baseline-browser-mapping": "2.10.33",
+    "memoize-one": "^6.0.0"
   },
   "resolutions": {
-    "baseline-browser-mapping": "2.10.33"
+    "baseline-browser-mapping": "2.10.33",
+    "memoize-one": "^6.0.0"
   },
   "patchedDependencies": {
     "@tanstack/react-router@1.170.15": "patches/@tanstack%2Freact-router@1.170.15.patch"
--- a/frontend/src/components/pages/connect/dynamic-ui/list.tsx
+++ b/frontend/src/components/pages/connect/dynamic-ui/list.tsx
@@ -9,11 +9,11 @@
  * by the Apache License, Version 2.0
  */
 
+import { DragDropContext, Draggable, Droppable, type DropResult, type ResponderProvided } from '@hello-pangea/dnd';
 import { Button, Input, Tooltip } from '@redpanda-data/ui';
 import { arrayMoveMutable } from 'array-move';
 import { CloseIcon, MenuIcon } from 'components/icons';
-import { useEffect, useRef, useState } from 'react';
-import { DragDropContext, Draggable, Droppable, type DropResult, type ResponderProvided } from 'react-beautiful-dnd';
+import { type JSX, useEffect, useRef, useState } from 'react';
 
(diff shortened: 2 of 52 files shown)


malinskibeniamin:
Cloud UI companion PR (consumes `rp_console/BridgeApp`): https://github.com/redpanda-data/cloudv2/pull/27360

malinskibeniamin:
@claude review

frontend/src/components/pages/connect/dynamic-ui/list.tsx
```diff
@@ -9,11 +9,11 @@
  * by the Apache License, Version 2.0
  */
 
+import { DragDropContext, Draggable, Droppable, type DropResult, type ResponderProvided } from '@hello-pangea/dnd';
```
malinskibeniamin, line 12: why?

frontend/package.json
```diff
@@ -204,10 +203,16 @@
...
+    "memoize-one": "^6.0.0",
+    "@types/react": "^19.2.17",
+    "@types/react-dom": "^19.2.2"
   },
   "resolutions": {
-    "baseline-browser-mapping": "2.10.33"
+    "baseline-browser-mapping": "2.10.33",
+    "memoize-one": "^6.0.0",
+    "@types/react": "^19.2.17",
+    "@types/react-dom": "^19.2.2"
```
malinskibeniamin: not sure if we need either of the overrides/resolutions

frontend/src/components/pages/connect/dynamic-ui/list.tsx
malinskibeniamin: `react-beautiful-dnd` is unmaintained and breaks under React 19: it pulls in `react-redux@7`, which calls `ReactDOM.unstable_batchedUpdates` at module init — that API was removed in React 19, so any drag-and-drop mount would throw at runtime. `@hello-pangea/dnd` is the maintained drop-in fork with an identical API (`DragDropContext`/`Droppable`/`Draggable`/`DropResult`/`ResponderProvided`) and React 19 support, so the only change is the import specifier (here and in `preview-settings.tsx`).

frontend/package.json
malinskibeniamin: Good instinct — I tested it. Only `memoize-one` is actually needed: removing `react-beautiful-dnd` (which pinned `memoize-one@^5`) lets bun re-hoist the untyped `memoize-one@4.1.0` from `react-highlight-words`, and `src/config.ts` imports it directly → `TS7016` (no bundled types). Pinning `memoize-one@^6` (typed; the version `@hello-pangea/dnd` and `react-select` already want) fixes it.

The `@types/react` / `@types/react-dom` overrides turned out **not** to be needed after the rebase — a single `@types/react@19.2.17` now resolves on its own — so I dropped those. Kept `memoize-one` in both `overrides` and `resolutions` to match the existing `baseline-browser-mapping` pin (npm reads `overrides`, the yarn.lock side reads `resolutions`).

c-julin (approved):

PR #2537, opened 2026-06-26
Migrate Quotas to the new UI registry

jvorcak:
Before:

Screenshot 2026-06-26 at 12 46 03

After: 

Screenshot 2026-06-26 at 12 46 38
Screenshot 2026-06-26 at 12 46 34


frontend/src/components/redpanda-ui/components/alert.tsx
```diff
@@ -5,23 +5,17 @@ import React from 'react';
 import { cn, type SharedProps } from '../lib/utils';
 
 const alertVariants = cva(
-  'relative grid w-full grid-cols-[0_1fr] items-start gap-y-0.5 rounded-lg border px-4 py-3 text-sm has-[>svg]:grid-cols-[calc(var(--spacing)*4)_1fr] has-[>svg]:gap-x-3 [&>svg]:size-4 [&>svg]:translate-y-0.5 [&>svg]:text-current',
+  // Body text is neutral high-contrast; the tone lives in the surface, border, and icon.
+  // NOTE: the dark-mode palette is provisional and not yet contrast-tested.
+  'relative grid w-full grid-cols-[0_1fr] items-start gap-y-0.5 rounded-lg border px-4 py-3 text-grey-900 text-sm has-[>svg]:grid-cols-[calc(var(--spacing)*4)_1fr] has-[>svg]:gap-x-3 *:data-[slot=alert-description]:text-grey-900 dark:text-grey-50 dark:*:data-[slot=alert-description]:text-grey-50 [&>svg]:size-4 [&>svg]:translate-y-0.5',
```
malinskibeniamin, line 10: **P1 — Shared `Alert` primitive restyled app-wide inside a feature PR, with author-admitted untested dark contrast**

**What:** This PR rewrites the base + all four variants of the shared registry primitive `alert.tsx`. The inline NOTE (line 9) concedes the dark-mode palette is "provisional and not yet contrast-tested."

**Why:** `alert.tsx` is consumed by every `Alert` across Console, so this is an app-wide visual change landing inside a "Migrate Quotas" PR (blast radius + scope creep). Shipping an explicitly-unverified dark palette risks WCAG 1.4.3 contrast regressions everywhere. Also: the base sets description color via `*:data-[slot=alert-description]:text-grey-900/50` (this line) while `AlertDescription` still carries `text-muted-foreground` (line 67) — confirm which wins per variant/theme so description text isn't accidentally muted on the tinted surfaces.

**Fix:** Split the `alert.tsx` change into its own registry-scoped PR and verify ≥4.5:1 contrast for title + description on each `bg-background-*-subtle` surface in both themes before merge; keep this PR limited to the quotas page + route.

**One-shot:** Extract the alert.tsx variant restyle into a separate registry PR and contrast-test all four variants in light+dark; Quotas should consume `Alert`, not redefine it.

frontend/src/components/pages/quotas/quotas-list.tsx
```diff
@@ -58,42 +205,61 @@ const QuotasList = () => {
...
-    const value = values.find((v) => v.valueType === valueType)?.value;
-    return value ? (
-      prettyBytes(value)
-    ) : (
-      <span style={{ opacity: 0.3 }}>
-        <SkipIcon />
-      </span>
-    );
+  const pagination: PaginationState = {
+    pageIndex: search.page ?? 0,
```
malinskibeniamin: **P2 — A stale / too-large `?page=` shows "No quotas configured" even when quotas exist**

**What:** `pageIndex: search.page ?? 0` feeds a controlled table with `autoResetPageIndex: false`, and the route schema bounds `page` only by `min(0)` (no upper clamp vs the data).

**Why:** With `autoResetPageIndex:false`, an out-of-range page index yields an empty row model that falls through to the `No quotas configured.` empty state — indistinguishable from a genuinely empty list (e.g. a bookmarked `/quotas?page=9`, or the quota count shrinking).

**Fix:** Clamp `pageIndex` to `table.getPageCount() - 1` (or drop `autoResetPageIndex:false`).

**One-shot:** Clamp the controlled `pageIndex` to `getPageCount()-1` so an out-of-range `?page=` can't render a false empty state.

frontend/src/components/pages/quotas/quotas-list.tsx
```diff
@@ -43,12 +76,126 @@ const mapEntityTypeToDisplay = (entityType: Quota_EntityType): 'client-id' | 'us
...
+            <span className="sr-only">{NOT_CONFIGURED_LABEL}</span>
+          </span>
+        }
+      />
+      <TooltipContent>{NOT_CONFIGURED_LABEL}</TooltipContent>
+    </Tooltip>
+  </TooltipProvider>
+);
+
+const formatBytes = (value?: number) => (value ? prettyBytes(value) : <NotConfigured />);
```
malinskibeniamin: **P3 — A quota explicitly set to `0` renders as "No limit configured"**

**What:** `formatBytes`/`formatRate` (lines 149–150) use `value ? … : <NotConfigured/>`, so a rate of `0` is treated as unset.

**Why:** `getRate` returns `undefined` when a value is genuinely absent, so `0` is a real, maximally-restrictive limit — rendering it as "No limit configured" inverts the meaning. (Behavior is carried over from `master`, but you rewrote these helpers here, so it's cheap to fix in place.)

**Fix:** Branch on `value === undefined` instead of truthiness.

**One-shot:** In quotas-list.tsx:149–150, guard `formatBytes`/`formatRate` on `value === undefined` so a `0` quota renders as `0`.

frontend/src/components/pages/quotas/quotas-list.tsx
```diff
@@ -43,12 +76,126 @@ const mapEntityTypeToDisplay = (entityType: Quota_EntityType): 'client-id' | 'us
...
+      </Button>
+      {tooltip ? <InfoTooltip>{tooltip}</InfoTooltip> : null}
+    </div>
+  );
+}
+
+/** Renders for a rate that has no quota set — an em-dash with a tooltip explaining there is no limit. */
+const NOT_CONFIGURED_LABEL = 'No limit configured';
+
+const NotConfigured = () => (
```
malinskibeniamin, line 115: **P3 — A `TooltipProvider` is mounted per cell**

**What:** `NotConfigured` (line 132) and `InfoTooltip` (line 93) each wrap their own `<TooltipProvider>`, so a 50-row table can mount 100+ providers.

**Why:** `TooltipProvider` is meant to be hoisted once near the root; per-instance providers are unnecessary context churn and not the registry's intended composition.

**Fix:** Wrap the page/table in a single `<TooltipProvider>` and drop the per-component ones, keeping only `Tooltip`/`TooltipTrigger`/`TooltipContent`.

**One-shot:** Hoist one `TooltipProvider` to the page root and remove the per-cell providers in `NotConfigured`/`InfoTooltip`.

frontend/src/components/pages/quotas/quotas-list.tsx
```diff
@@ -104,106 +270,75 @@ const QuotasList = () => {
...
 
     return (
       <PageContent>
-        <Section>
-          <Alert status="warning" style={{ marginBottom: '1em' }} variant="solid">
-            <AlertIcon />
-            {error.message || 'Failed to load quotas'}
-          </Alert>
-        </Section>
+        <Alert variant="destructive">
```
malinskibeniamin: **P3 — The destructive error `Alert` shows the default info icon**

**What:** `<Alert variant="destructive">` (line 295) passes no `icon`, and `Alert` defaults `icon` to `<InfoIcon/>` — so a load error renders a red-tinted "i".

**Why:** An info glyph on an error alert is a semantic/visual mismatch (the sibling overview/shadowlink alerts pass `TriangleAlertIcon`/`Lock`).

**Fix:** Pass an error icon, e.g. `icon={<TriangleAlertIcon />}`.

**One-shot:** Add an error icon to the destructive `Alert` at quotas-list.tsx:295.

malinskibeniamin (commented):
## /review — PR #2537 (Migrate Quotas to the new UI registry)

Multi-hat review (ponytail · resilience · regular · adversarial · visual · test-perf · security/privacy). Inline comments below, each tagged P1–P3.

**P2 — New logic ships without Vitest unit tests.** The `quotas/` dir contains only `quotas-list.tsx`; the new `getRate` / `mapEntityTypeToDisplay` / `format*` helpers and the URL↔TanStack sort/pagination mapping have only Playwright e2e coverage. `AGENTS.md` requires Vitest for new code — a `quotas-list.test.tsx` (or extracted pure helpers) would pin the sort/pagination/format logic.

**Pre-existing (not introduced here, so no inline):** the permission branch keys off `error.message.includes('permission')` rather than `error.code === Code.PermissionDenied`. It's carried over from `master`, but since this PR refactors that exact block, switching to the ConnectError code (as sibling PR #2540 does) would be a cheap robustness win.

**PR value:** HIGH — legitimate legacy→registry migration that removes `@redpanda-data/ui`.

**Needs attention:** 1×P1, 2×P2, 3×P3. The P1 (app-wide shared-`Alert` restyle with untested dark contrast) should be resolved or split out before merge.

frontend/src/components/redpanda-ui/components/alert.tsx
malinskibeniamin, line 1: is this upstreamed from the UI registry?

malinskibeniamin (approved):

PR #2546, opened 2026-07-01
refactor(schema-registry): align list with topic/security listings

jvorcak:
- Replace legacy Section box with ListLayout scaffolding so the table renders flush like the topics/security lists
- Add faceted Type and Compatibility filters (DataTableFacetedFilter)
- Move the row delete action into a three-dots kebab menu
- Use shared DataTablePagination
- Load all subject details up front (useSchemaDetailsByNameQuery, shared cache) to enable list-wide type/compatibility filtering
- Update schema e2e helpers/spec for the new kebab delete flow
- Uses new stats bar https://redpanda-ui-registry.netlify.app/docs/stat

Before:
Screenshot 2026-07-16 at 12 02 23

After:
Screenshot 2026-07-16 at 12 00 30


frontend/src/components/redpanda-ui/components/dropdown-menu.tsx
```diff
@@ -236,6 +236,8 @@ function DropdownMenuContent({
 type DropdownMenuItemProps = React.ComponentProps<typeof DropdownMenuPrimitive.Item> & {
   inset?: boolean;
   variant?: 'default' | 'destructive';
+  /** @deprecated Base UI menu items have no `onSelect`. Use `onClick` (add `closeOnClick={false}` to keep the menu open). */
+  onSelect?: never;
```
malinskibeniamin, line 240: it sounds like a UI registry problem to fix

frontend/src/components/pages/quotas/quotas-list.tsx
```diff
@@ -317,9 +316,7 @@ const QuotasList = () => {
           </TableBody>
         </Table>
 
-        <ListLayoutPagination>
-          <DataTablePagination table={table} />
-        </ListLayoutPagination>
+        <DataTablePagination table={table} />
```
malinskibeniamin, line 320: what's the difference in how it looks? was it just leftover pattern we did not use?

frontend/src/components/redpanda-ui/components/dropdown-menu.tsx
jvorcak: I think this was done by syncing the registry.

frontend/src/components/pages/quotas/quotas-list.tsx
jvorcak: We removed the border around the table, because it's not used like this on other pages.

frontend/src/components/pages/schemas/schema-list.tsx
```diff
@@ -12,30 +12,45 @@
...
+} from 'components/redpanda-ui/components/dropdown-menu';
+import { ListLayout, ListLayoutFilters, ListLayoutSearchInput } from 'components/redpanda-ui/components/list-layout';
 import { Skeleton } from 'components/redpanda-ui/components/skeleton';
 import { Spinner } from 'components/redpanda-ui/components/spinner';
+import { Stat, StatGroup } from 'components/redpanda-ui/components/stat';
 import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from 'components/redpanda-ui/components/table';
 import { Tooltip, TooltipContent, TooltipTrigger } from 'components/redpanda-ui/components/tooltip';
-import { ChevronLeftIcon, ChevronRightIcon, ChevronsLeftIcon, ChevronsRightIcon, SearchIcon } from 'lucide-react';
-import { parseAsBoolean, parseAsInteger, parseAsString, useQueryState } from 'nuqs';
+import { Text } from 'components/redpanda-ui/components/typography';
```
SpicyPete: Now that `Text` and `Heading` are deprecated, lets use text-utilities in this PR instead.

frontend/src/components/pages/schemas/schema-list.tsx
```diff
@@ -278,404 +362,399 @@ const SchemaList: FC = () => {
...
+              title="Schema search help"
+              titleTestId="schema-help-title"
+            >
+              <section aria-labelledby="filtering-heading" className="space-y-3">
+                <h3 className="font-semibold text-gray-900" id="filtering-heading">
+                  Filtering schemas
                 </h3>
-                <p className="text-gray-600 text-sm leading-relaxed">
-                  If a number matches a schema ID, the results include all subjects referencing that schema.
+                <p className="text-base text-gray-600 leading-relaxed">
```
SpicyPete: Is there a semantic colour we can use instead of raw palette colours?  Seeing gray shades being used many times, so this comment is for all of them

frontend/src/components/pages/schemas/schema-list.tsx
jvorcak: Done — replaced the deprecated `Text` wrapper with a plain `<div>` using the text utilities.

frontend/src/components/pages/schemas/schema-list.tsx
jvorcak: Good catch — swapped all the raw gray shades for semantic tokens: `text-gray-900` → `text-foreground`, `text-gray-600`/`text-gray-400` → `text-muted-foreground`, and the inline `dimgrey` icon color → `text-muted-foreground`.

frontend/src/components/pages/schemas/schema-list.tsx
```diff
@@ -278,404 +361,399 @@ const SchemaList: FC = () => {
...
+                  <h3 className="font-semibold text-base text-foreground" id="schema-id-heading">
+                    Schema ID
+                  </h3>
+                  <p className="text-muted-foreground text-sm leading-relaxed">
+                    If a number matches a schema ID, the results include all subjects referencing that schema.
+                  </p>
+                </section>
+
+                <section aria-labelledby="subject-name-heading" className="space-y-2">
+                  <h3 className="font-semibold text-base text-foreground" id="subject-name-heading">
```
SpicyPete: If we are adding semibold, we can probably just use `text-heading-sm` (or other size)

frontend/src/components/pages/schemas/schema-list.tsx
```diff
@@ -278,404 +361,399 @@ const SchemaList: FC = () => {
...
+                    Subject name
+                  </h3>
+                  <p className="text-muted-foreground text-sm leading-relaxed">
```
SpicyPete: `text-body`, and we probably don't want to change the `leading` unless it's intentional as the util functions should be setting this

frontend/src/globals.css
```diff
@@ -34,6 +34,19 @@
...
+   size is definite, so the fixed-height dividers inside filter chips (`h-4`) pin to the
+   top of the button and stop halfway down. Center those.
+   Scoped to a button's own children: every vertical Separator inside a button sets a
+   height, while the ones relying on stretch (data-table-filter's `h-7` row) never do,
+   so this can't collapse them to zero. Lives here (app entry CSS) rather than the
+   registry-synced component so it survives re-syncs — same reasoning as UX-1259 above.
+   Remove once the registry centers height-bearing vertical separators itself. */
+button > [data-slot="separator"][data-orientation="vertical"] {
+  align-self: center;
+}
```
SpicyPete: This seems like excess explanation and very targeted, so probably something we should fix upstream or we'll see again.
Though the Separator was recently updated so this may be already fixed.

Edit: I am seeing some weird separator behavior in other branches, so this may be an important fix to upstream later.  Can keep in this PR for now.

SpicyPete (commented):
Looking great for the most part, just a few minor comments

frontend/src/globals.css
jvorcak: Sorry, this was supposed to be removed before the push

frontend/src/components/pages/schemas/schema-list.tsx
```diff
@@ -278,404 +361,399 @@ const SchemaList: FC = () => {
...
     onSortingChange: setSorting,
     onPaginationChange: handlePaginationChange,
+    onColumnFiltersChange: handleColumnFiltersChange,
     autoResetPageIndex: false,
     getCoreRowModel: getCoreRowModel(),
     getSortedRowModel: getSortedRowModel(),
+    getFilteredRowModel: getFilteredRowModel(),
+    getFacetedRowModel: getFacetedRowModel(),
+    getFacetedUniqueValues: getFacetedUniqueValues(),
     getPaginationRowModel: getPaginationRowModel(),
```
SpicyPete, line 494: if we add ` enableRowSelection: false` we won't show the "N of M rows selected" on this table which has no selectable rows

SpicyPete (approved):
Looks good, just some minor details and 1 note on the table filtering.
The further pages in schema seem to need some love, but that's out of scope for this PR.


I am seeing a small issue in a data table separator alignment, so I fixed that at the registry level, next time the data-table component updates that should get fixed automagically.

r-vasquez:
The original idea of having a Context selector is that users wanted to interact with the whole registry settings based on which context they are on, not only to filter the table.

So if I moved from context `Default` to `.prod` it changed the Mode, Compatibility AND it filters the table with the schemas that belong to that context.

On top of that, if you clicked `Create new schema` you get moved to a form where the Context was already selected for you.

I'm not sure if this new selector of Context in the table filters is intuitive enough for users to realize that. Can you confirm that you tested with different contexts and modes/compat and checked with product on this change?

jvorcak:
@r-vasquez what about this? 
Screenshot 2026-07-31 at 15 32 40

PR #2553, opened 2026-07-07
UX-1338 — Redpanda Connect Pipeline (RPCN) Visual View & Editor

SpicyPete:
## Summary

Adds a full **visual editor** for Redpanda Connect (RPCN) pipelines: an interactive, node-based canvas that renders a pipeline's YAML as a directed graph and lets you inspect and edit nodes without hand-writing YAML. It sits alongside the existing YAML editor as a new **"Visual" lane**.

The editor is **feature-flagged** and currently scoped to the embedded (Cloud) context — it activates only when `enableRpcnVisualEditor` **and** `enablePipelineDiagrams` are on and the page is embedded. When enabled, editing opens on the Visual lane by default; otherwise the experience is unchanged (YAML lane).

### Canvas & layout
- **Flow canvas** (`pipeline-flow-canvas.tsx`, `pipeline-flow-canvas-nodes.tsx`) — renders the pipeline as a left-to-right DAG using **React Flow + Dagre** (`@dagrejs/dagre`, new dependency): input → processors → output, with control-flow constructs (`switch`, `branch`, `try`/`catch`, `broker`, `fallback`) drawn as split → case-lanes → merge, plus conditional/error/reference edge styling.
- **Scope regions** — each control-flow construct gets a faint dashed enclosure. These are traced as a single **connected outline** that hugs the construct's actual (L-shaped) footprint per Dagre column, so a fan-out no longer paints over unrelated neighbor cards. Tall regions tile their fill under the browser's max paintable-element size so the background never silently drops.
- **Flow parser & layout** (`pipeline-flow-parser.ts`, `pipeline-flow-meta.ts`) — parses pipeline YAML into a flow tree and computes the layout with source-aligned ranks. Replaces the old sidebar diagram (`pipeline-flow-diagram.tsx` removed).

### Editing
- **Node inspector** (`node-inspector.tsx`, `node-config-form.tsx`) — side panel to view/edit a selected node's config via Registry Field forms (react-hook-form + Zod), drill into control-flow children (switch cases / steps), edit a switch case's routing condition inline, and edit a node's raw YAML in place.
- **Editor store** (`use-pipeline-editor-store.ts`) — Zustand store with undo/redo; pending visual edits are flushed to YAML at save time.
- **Component picker / command palettes** (`pipeline-canvas-command-palette.tsx`, `onboarding/connect-command-palette.tsx`) — quick actions (view in YAML, undo/redo, jump to node) and a connector/processor inserter with curated defaults, alias-aware search (e.g. "queue" → `kafka_franz`; aliases live in `connect-command-palette-utils.ts`), and de-emphasis of deprecated/experimental components.

### Feedback surfaces
- **Structure tree** (`pipeline-structure-tree.tsx`) — sidebar outline in document order with collapsible groups, roving-tabindex keyboard navigation, and click-to-reveal.
- **Problems panel** (`pipeline-problems-panel.tsx`, `pipeline-lint.ts`) — floating chip listing lint problems (mapped from backend `LintHint`s by YAML line range, merged with save-error hints) and detected secret references; click a problem to jump to its node.
- **Unsaved panel** (`pipeline-unsaved-panel.tsx`, `pipeline-diff.ts`) — floating chip listing nodes whose config differs from the last save (signature-based diff); click to jump.
- **Tips bar & template CTA** (`editor-tips-bar.tsx`, `template-cta.tsx`) — platform-aware keyboard-shortcut hints and a "start from a template" entry point for empty pipelines.

## Testing

Unit + integration tests accompany the new modules — flow parser/layout, lint mapping, dirty-diff, canvas render (incl. tall-region paint-cap), inspector, config form, structure tree, problems/unsaved panels, and command palettes. `type:check`, `lint`, and the test suite are green.

## Known limitations / deferrals
- **Delete-guard** (confirming deletion of a node with dependents) is a TODO.

## How to test
Enable `enableRpcnVisualEditor` + `enablePipelineDiagrams`, open a pipeline in the embedded editor, and confirm the Visual lane loads. Verify node selection/edit, undo/redo, save round-trips to YAML, and the problems/unsaved chips.

### Sidelane / YAML enhancements
Screenshot 2026-07-07 at 9 31 10 AM
Screenshot 2026-07-07 at 9 31 41 AM

### Visual view
Screenshot 2026-07-07 at 9 32 07 AM
Screenshot 2026-07-07 at 9 32 38 AM


SpicyPete:
@claude review

SpicyPete:
https://github.com/redpanda-data/console/pull/2554
This PR pulls out a bunch of the non-core updates to this PR

frontend/src/components/pages/rp-connect/onboarding/add-connector-dialog.tsx
```diff
@@ -44,34 +34,22 @@ export const AddConnectorDialog = ({
...
-          <ConnectTiles
-            className="px-0 py-0"
-            components={components}
-            componentTypeFilter={typeFilter}
-            gridCols={3}
-            hideHeader
-            onChange={onAddConnector}
-            searchPlaceholder={searchPlaceholder}
-            variant="ghost"
-          />
```
eblairmckee, line 73: we should make an action item that once this is GA and well tested we remove the old RPCN UXs. potentially Q3 or Q4

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.test.tsx
```diff
@@ -0,0 +1,91 @@
+import { ComponentStatus } from 'protogen/redpanda/api/dataplane/v1/pipeline_pb';
+import { describe, expect, it } from 'vitest';
+
+import { asciidocToMarkdown, buildEmptyMessage, byProminence } from './connect-command-palette';
+import type { ConnectComponentSpec } from '../types/schema';
+
+const spec = (name: string, status: ComponentStatus = ComponentStatus.STABLE): ConnectComponentSpec =>
+  ({ name, type: 'processor', status }) as ConnectComponentSpec;
+
+const sorted = (specs: ConnectComponentSpec[]) => [...specs].sort(byProminence).map((s) => s.name);
```
eblairmckee: I'm seeing a lot of unit tests in here. can we move those other to test the actual sort and spec helpers (presuming they are in a separate .ts file?)
and scope this test file to only integration test expectations

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+      // Section titles (`==`/`===`/… Title) → small heading.
+      .replace(/^=+\s+(.{1,60})$/gm, '#### $1')
+      // Strip leftover markers from over-long titles.
+      .replace(/^=+\s+/gm, '')
+      // List markers → bullets.
+      .replace(/^\*\s+/gm, '- ')
+      .replace(/\n{3,}/g, '\n\n')
+      .trim()
+  );
+}
```
eblairmckee: can we move all of these to a utils file, please. goes along with my comment on the test file for this

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+  a: ({ href, children }) => (
+    <Link href={href} rel="noopener noreferrer" target="_blank">
+      {children}
+    </Link>
+  ),
+  code: ({ children }) => <code className="rounded bg-muted px-1 py-0.5 font-mono text-xs">{children}</code>,
+  ul: ({ children }) => <ul className="list-disc space-y-0.5 pl-4 text-foreground text-sm">{children}</ul>,
+  ol: ({ children }) => <ol className="list-decimal space-y-0.5 pl-4 text-foreground text-sm">{children}</ol>,
+  li: ({ children }) => <li className="leading-relaxed">{children}</li>,
+};
```
eblairmckee, line 73: question: is this the only place in RPCN where we render markdown? if so, fine to keep here, but I'd rather make this more reusable and have it live in its own shared place

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+    return <FormattedDescription markdown={markdown} />;
+  }
+  return (
+    <div className="flex flex-col gap-1">
+      <div className={expanded ? '' : 'relative max-h-44 overflow-hidden'}>
+        <FormattedDescription markdown={markdown} />
+        {expanded ? null : (
+          <div className="pointer-events-none absolute inset-x-0 bottom-0 h-10 bg-gradient-to-t from-background to-transparent" />
+        )}
+      </div>
```
eblairmckee: can we use collapsible from registry?

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+      <button
+        className="self-start font-medium text-primary text-xs hover:underline"
+        onClick={() => setExpanded((v) => !v)}
+        type="button"
+      >
+        {expanded ? 'Show less' : 'Show more'}
+      </button>
```
eblairmckee: this seems like a Button variant ghost

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+    return <span className="font-mono font-semibold text-sm">{name}</span>;
+  }
+  return (
+    <span className="font-mono font-semibold text-sm">
+      {name.slice(0, idx)}
+      <span className="bg-primary/15 text-primary">{name.slice(idx, idx + query.length)}</span>
+      {name.slice(idx + query.length)}
+    </span>
+  );
+}
```
eblairmckee, line 112: nit: oh I would love to see this upstreamed to the registry command component

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+
+// "processors", "caches or rate limits" — plural label for the slot's allowed types.
+function pluralTypeLabel(types?: ConnectComponentType[]): string {
+  if (!types || types.length === 0) {
+    return 'components';
+  }
+  return types.map((type) => `${type.replace(/_/g, ' ')}s`).join(' or ');
+}
```
eblairmckee: 99% sure we have a helper/util for this already

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+function withArticles(types: ConnectComponentType[]): string {
+  const labeled = types.map((type) => {
+    const label = type.replace(/_/g, ' ');
+    return `${STARTS_WITH_VOWEL.test(label) ? 'an' : 'a'} ${label}`;
+  });
+  if (labeled.length <= 1) {
+    return labeled[0] ?? '';
+  }
+  return `${labeled.slice(0, -1).join(', ')} or ${labeled.at(-1)}`;
+}
```
eblairmckee: feels like a `utils/plural.ts` would be a lovely home for this and the prev function

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+}) {
+  const category = component.categories?.[0];
+  // Fall back to the humanized type (cache / rate_limit / …) so same-named components of different
+  // kinds — e.g. the redis cache vs the redis rate-limit — are distinguishable in the list.
+  const badgeLabel = category ? getCategoryDisplayName(category) : component.type.replace(/_/g, ' ');
+  return (
+    <CommandItem
+      className="gap-3 hover:bg-accent/50"
+      key={`${component.type}-${component.name}`}
+      onDoubleClick={() => onCommit(component)}
```
eblairmckee, line 132: nice

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+        <DescriptionBlock
+          collapsible={(component.description?.length ?? 0) > 700}
+          key={`${component.type}:${component.name}`}
+          markdown={descriptionMd}
+        />
+      ) : null}
+
+      {categories.length > 0 ? (
+        <div className="flex flex-col gap-1.5">
+          <Text className="font-medium text-muted-foreground text-xs uppercase tracking-wide">Categories</Text>
```
eblairmckee, line 194: note for later: we use this kind of label style so often, would be nice to include this in the new typography system I'm ruminating on

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+          <div className="flex flex-wrap gap-1.5">
+            {categories.map((category) => (
+              <Badge key={category} size="sm" variant="neutral-inverted">
+                {category}
+              </Badge>
+            ))}
+          </div>
```
eblairmckee: nit: can use BadgeGroup which will also handle overflow if a max number is hit and can render in a tooltip

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+    [recents, byName]
+  );
+
+  const suggested = useMemo(() => {
+    const names = new Set((allowedTypes ?? []).flatMap((t) => SUGGESTED_BY_TYPE[t] ?? []));
+    const recentNames = new Set(recentComponents.map((c) => c.name));
+    return [...names]
+      .map((n) => byName.get(n))
+      .filter((c): c is ConnectComponentSpec => Boolean(c) && !recentNames.has((c as ConnectComponentSpec).name));
+  }, [allowedTypes, byName, recentComponents]);
```
eblairmckee: nit: can make this a util and pull out of component body

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
```diff
@@ -0,0 +1,708 @@
...
+              </CommandGroup>
+            ) : browseItems ? (
+              <CommandGroup>
+                {browseItems.map((component) => (
+                  <Row
+                    component={component}
+                    key={`${component.type}-${component.name}`}
+                    onCommit={handleCommit}
+                    onPreview={handlePreview}
+                    query=""
```
eblairmckee: nit: could make query optional and default to `""`

frontend/src/components/pages/rp-connect/pipeline/floating-chip-panel.tsx
```diff
@@ -0,0 +1,112 @@
...
+  /** Chip text (e.g. "3 problems"). */
+  label: React.ReactNode;
+  /** Tone/color classes appended to the chip button (border/text/hover). */
+  chipClassName?: string;
+  chipTestId?: string;
+  /** Width/extra classes for the expanded list container. */
+  listClassName?: string;
+  listTestId?: string;
+  /** Expanded-list content; call `close` after acting on a row to collapse the panel. */
+  children: (close: () => void) => React.ReactNode;
```
eblairmckee, line 63: ❤️

eblairmckee: all of this is great. very reusable. looks like a molecule 😏 ⚛️

frontend/src/components/pages/rp-connect/pipeline/floating-chip-panel.tsx
```diff
@@ -0,0 +1,112 @@
...
+};
+
+/**
+ * A floating, bottom-anchored chip on the Visual canvas that expands to a scrollable list on click.
+ * Owns open/close (outside-click + Escape dismissal); callers supply the chip's tone, label, and rows.
+ */
+export function FloatingChipPanel({
```
eblairmckee, line 70: can you attach some screenshots of what this looks like. I am having a hard time picturing this

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -579,8 +585,6 @@ function ViewModePanel({ pipeline }: { pipeline: Pipeline | undefined }) {
 
 function EditorPanel({
   isServerlessInitializing,
-  slashTipVisible,
-  onDismissSlashTip,
```
eblairmckee, line 583: with the feature flag for the visual editor turned off, do we no longer show the slash menu?

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -653,41 +638,138 @@ function EditorPanel({
...
+  // Node → YAML line ranges.
+  const nodeRanges = useMemo(() => {
+    try {
+      return nodeLineRanges(deferredYaml);
+    } catch {
+      return [];
+    }
+  }, [deferredYaml]);
+  // Latest ranges for the long-lived cursor listener, without re-subscribing per keystroke.
+  const nodeRangesRef = useRef(nodeRanges);
```
eblairmckee, line 676: nit: if we do want to phase this out, I am noticing A LOT of functionality that is not behind the new visual editor feature flag, that easily could be ripped out into a separate PR. 
just noting in case at some point we do want to rescope this into smaller PRs

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -653,41 +638,138 @@ function EditorPanel({
...
+      requestRevealNode(null);
+      return;
+    }
+    // Bounded retry: allow a few re-runs (editor mount, ranges catch-up), then drop the request
+    // so an id that never resolves can't fire as a surprise jump much later.
+    revealAttemptRef.current.count += 1;
+    if (revealAttemptRef.current.count > MAX_REVEAL_ATTEMPTS) {
+      requestRevealNode(null);
+    }
+  }, [revealNodeId, editorInstance, nodeRanges, revealNodeInEditor, requestRevealNode]);
```
eblairmckee, line 759: nit: things like this I typically like to pull out into their own hook files with their own unit tests

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -653,41 +638,138 @@ function EditorPanel({
...
+      <div className="relative min-h-0 flex-1 overflow-hidden">
+        <ScrollShadow className="h-full overflow-x-hidden">
+          {showStructureTree ? (
+            <PipelineStructureTree
+              configYaml={deferredYaml}
+              errorNodeIds={errorNodeIds}
+              onAddConnector={canEdit ? (section) => onAddConnector(section as ConnectComponentType) : undefined}
+              onSelectNode={handleSelectNode}
+              selectedNodeId={activeNodeId}
+              unsavedNodeIds={unsavedNodeIds}
```
eblairmckee, line 774: 👍  I like that we're passing around ids rather than more config objects

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -653,41 +638,138 @@ function EditorPanel({
...
+            />
+          ) : null}
+        </ScrollShadow>
+        {showStructureTree && onBrowseTemplates ? (
+          <TemplateGalleryCta onBrowseTemplates={onBrowseTemplates} show={showTemplateCta} />
```
eblairmckee, line 779:
```suggestion
          showTemplateCta ? <TemplateGalleryCta onBrowseTemplates={onBrowseTemplates}/> : null
```

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -859,16 +949,67 @@ function PipelinePageContent() {
...
+    () => new Set(mapLintHintsToNodes(deferredYamlContent, Object.values(lintHints)).keys()),
+    [deferredYamlContent, lintHints]
+  );
+  const unsavedNodeIds = useMemo(
+    () =>
+      mode !== 'view' && initialYaml !== null
+        ? new Set(changedNodeIds(initialYaml, deferredYamlContent))
+        : EMPTY_NODE_IDS,
+    [mode, initialYaml, deferredYamlContent]
+  );
```
eblairmckee, line 984: any reason why this isn't stored in useState instead of useMemo?

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -859,16 +949,67 @@ function PipelinePageContent() {
   // Guard against losing unsaved edits when navigating away from the editor (edit or create).
   const yamlDirty = initialYaml !== null && yamlContent !== initialYaml;
   const hasUnsavedChanges = mode !== 'view' && (form.formState.isDirty || yamlDirty);
+
+  // Guard-time dirty check: flush any in-progress inspector draft into the store first (the
+  // rendered `hasUnsavedChanges` above can't see a pending draft), then re-read fresh state.
+  const hasUnsavedChangesFresh = useCallback(() => {
```
eblairmckee: nit: rename to `flushInProgressDraftToStore` or something more like a function name > boolean const

frontend/src/components/pages/rp-connect/pipeline/index.tsx
```diff
@@ -859,16 +949,67 @@ function PipelinePageContent() {
...
+      }
+      e.preventDefault();
+      if (!isSaving) {
+        handleSave();
+      }
+    };
+    window.addEventListener('keydown', onKeyDown);
+    return () => window.removeEventListener('keydown', onKeyDown);
+  }, [mode, isSaving, handleSave]);
+
```
eblairmckee: iirc we have a useKeyboardEvent or useHotKey hook for this. we may need to pass a query selected element to the hook, though, not sure what the focus trap looks like in that hook

frontend/src/components/pages/rp-connect/pipeline/index.tsx
SpicyPete: The slash menu on the YAML editor still works/appears with the visual editor turned off

frontend/src/components/pages/rp-connect/pipeline/invalid-config-notice.tsx
```diff
@@ -0,0 +1,33 @@
...
+      className={cn(
+        'flex items-start gap-2 rounded-md border border-warning/40 bg-warning-subtle text-foreground',
+        className
+      )}
+    >
+      <TriangleAlert className="mt-px size-4 shrink-0 text-warning" />
+      <span>{children}</span>
+    </output>
+  );
+}
```
eblairmckee, line 28: can we use an Alert from registry

frontend/src/components/pages/rp-connect/pipeline/index.tsx
SpicyPete: We are doing animations when this components mounts/unmounts, so we could do this change, but we'll need to move the animation logic up to this component instead.

frontend/src/components/pages/rp-connect/onboarding/connect-command-palette.tsx
SpicyPete: Yes, within RPCN this is the only place for now, I agree we should change this if we use markdown elsehwere in RPCN

frontend/src/components/pages/rp-connect/pipeline/index.tsx
SpicyPete: Noted, but in this instance it's pretty coupled with the Monaco editor

frontend/src/components/pages/rp-connect/pipeline/index.tsx
SpicyPete: Simpler logic as it's derived from the watched values, keeps it in sync easier and React can render useMemos efficiently vs setting multiple states.

frontend/src/components/pages/rp-connect/pipeline/floating-chip-panel.tsx
SpicyPete: Screenshot 2026-07-10 at 2 06 43 PM
Screenshot 2026-07-10 at 2 07 01 PM

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  if (item.isDefault) {
+    return 'default';
+  }
+  if (item.isErrorPath) {
+    return 'on error';
+  }
+  return;
+};
+
+// Routing-condition text color: red for error paths, muted for default, the warning accent otherwise.
```
eblairmckee: nit: I avoid calling out any colors or styles in comments, since it increases maintenance surface area if a return value changes. this is so short and scannable anyways, it's unnecessary

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  if (item.isErrorPath) {
+    return 'text-destructive';
+  }
+  if (item.isDefault) {
+    return 'text-muted-foreground';
+  }
+  return 'text-warning';
+};
+
+// One clickable child row (routing condition, component name, lint count, chevron). Selecting it
```
eblairmckee: dying to know if you're using sonnet 4.8... these style related comments looks like him..

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  item: InspectorChildItem;
+  onSelect: (item: InspectorChildItem) => void;
+}) => {
+  const condText = childItemConditionText(item);
+  return (
+    <button
+      className="flex w-full cursor-pointer items-center gap-2 rounded-md border border-border px-3 py-2 text-left transition-colors hover:border-primary/50 hover:bg-muted/40"
+      onClick={() => onSelect(item)}
+      type="button"
+    >
```
eblairmckee, line 96: can we use a ui registry button please

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+      <span className="flex min-w-0 flex-1 flex-col">
```
eblairmckee: nit: flex layout elements should be divs not spans. spans are inline/children of layouts

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+        {condText ? (
+          <span
+            className={cn('truncate font-medium font-mono text-caption-sm', childItemCondColor(item))}
```
eblairmckee: nit: Text component + variant please

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+            title={condText}
+          >
+            {condText}
+          </span>
+        ) : null}
+        <span className="truncate font-medium text-sm" title={item.name}>
+          {item.name}
+        </span>
```
eblairmckee: Text as span

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+      </span>
+      {item.lintCount ? (
+        <span className="inline-flex shrink-0 items-center gap-1 rounded border border-destructive/40 bg-destructive/5 px-1.5 py-0.5 font-medium text-caption-sm text-destructive">
+          <AlertCircle className="size-3" />
+          {item.lintCount}
+        </span>
+      ) : null}
```
eblairmckee: you can use dot-count from the registry instead.

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  items,
+  onSelect,
+  label,
+}: {
+  items: InspectorChildItem[];
+  onSelect: (item: InspectorChildItem) => void;
+  label: string;
+}) => (
+  <div className="flex flex-col gap-1.5">
+    <Label className="font-medium text-sm">{label}</Label>
```
eblairmckee: nit: do we need an aria htmlFor to associate the label with the child item

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  id,
+}: {
+  kind: ResourceKind;
+  value: string;
+  onChange: (value: unknown) => void;
+  id?: string;
+}) => {
+  const { labels, onCreateResource } = useContext(ResourceFieldContext);
+  const options = labels[kind];
+  const current = value ?? '';
```
eblairmckee: nit: could just set the initial value on line 170 `value = ''`

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+        }
+        onChange(v);
+      }}
+      value={current}
+    >
+      <SelectTrigger id={id}>
+        <SelectValue placeholder="Select a resource…" />
+      </SelectTrigger>
+      <SelectContent>
+        {isMissing ? <SelectItem value={current}>{current} (missing)</SelectItem> : null}
```
eblairmckee: what's the edge case for a missing item? not clear to me

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
SpicyPete: Image

These are these buttons, which nothing in the registry would map to well.  We could overload it's classNames a  bunch, but then it's simpler to just make a one-off

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+            {label}
+          </SelectItem>
+        ))}
+        {onCreateResource ? (
+          <SelectItem value={CREATE_RESOURCE_VALUE}>
+            <span className="flex items-center gap-1.5 text-primary">
+              <Plus className="size-3.5" />
+              Create new {kind === 'cache' ? 'cache' : 'rate limit'}…
+            </span>
+          </SelectItem>
```
eblairmckee, line 210: 99% sure we have a component for this.. I would look at what ui registry combobox does... it may or may not be exported as a subcomponent

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+        ) : null}
+      </SelectContent>
+    </Select>
+  );
+};
+
+const SCALAR_TYPES = new Set(['string', 'int', 'float', 'bool']);
```
eblairmckee: i feel like a constant exists for this in schema.ts (one of them...)

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+
+// A single editable value: a primitive (string/int/float/bool), an enum select, or a
+// resource reference (stored as a string).
+function isScalarField(spec: RawFieldSpec): boolean {
+  return (
+    Boolean(spec.name) &&
+    spec.kind === 'scalar' &&
+    (SCALAR_TYPES.has(spec.type) || hasOptions(spec) || isResourceRefField(spec))
+  );
+}
```
eblairmckee: same here. when we parse the rpcn schema we have a few utils and consts for determining field types. would love to see those reused, or move these utils to schema.ts

eblairmckee: that said, since I'd love to see these utils generalized in schema.ts we should remove these unnecessary comments. when in doubt, give a util a better name, don't vomit comments unnecessarily (I am look at your, claude)

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  'input',
+  'output',
+  'processor',
+  'cache',
+  'rate_limit',
+  'buffer',
+  'metrics',
+  'tracer',
+  'scanner',
+]);
```
eblairmckee: also exists in schema.ts

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+}
+
+// Drop objects that became empty after clearing their fields, so the YAML stays tidy.
+function pruneEmptyObjects(obj: Record<string, unknown>): void {
+  for (const [key, val] of Object.entries(obj)) {
+    if (val && typeof val === 'object' && !Array.isArray(val)) {
+      pruneEmptyObjects(val as Record<string, unknown>);
+      if (Object.keys(val).length === 0) {
+        delete obj[key];
+      }
```
eblairmckee, line 257: these seem like lodash util rewrites... I would be shocked if we didn't already have an `utils/object` file with helpers for this. if not... let's add one

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+function coerceArrayItems(spec: RawFieldSpec, text: string): unknown[] {
+  const lines = text
+    .split('\n')
+    .map((l) => l.trim())
+    .filter((l) => l !== '');
+  if (spec.type === 'int' || spec.type === 'float') {
+    return lines.map(Number).filter((n) => !Number.isNaN(n));
+  }
+  return lines;
+}
```
eblairmckee, line 303: again.. more schema.ts rewrites

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
SpicyPete: Opus 4.8 mostly, but it does use sonnet for sub-tasks often

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+      </span>
+    ) : null}
+    {spec.type && spec.type !== 'string' ? <span className="text-muted-foreground text-xs">{spec.type}</span> : null}
+    {spec.defaultValue ? (
+      <span className="text-muted-foreground text-xs">
+        default: <span className="font-mono">{spec.defaultValue}</span>
+      </span>
+    ) : null}
+  </div>
+);
```
eblairmckee, line 468: iirc we have a component for Required... but then again that may have been adp-ui... 
regardless.. worth elevating this to a shared components directory rather than one-offing this. would be very useful in react-hook-form instances.

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+
+const FieldDescription = ({ spec }: { spec: RawFieldSpec }) =>
+  spec.description ? (
+    <Text className="text-muted-foreground" variant="bodySmall">
+      {spec.description}
+    </Text>
+  ) : null;
```
eblairmckee, line 475: same comment as above.

eblairmckee: worth noting I like to have both field-description as well as field-helper-text (the latter being positioned below the field, mutually exclusive with field-error)
this is pseudo code for the sake of discussion. would be nice to elevate these kinds of things to be common components so future console UIs can reuse them. and eventually upstream to ui registry as typography utility classes 😏

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+          size="icon-sm"
+          type="button"
+          variant="ghost"
+        >
+          {show ? <EyeOff className="size-3.5" /> : <Eye className="size-3.5" />}
+        </Button>
+      )}
+    </div>
+  );
+};
```
eblairmckee: we have a `Input type="password"` in the registry already, prefer using that... or if we need something more custom this deserves to be its own component. 

Actually... I think the registry also has a `SecretSelector` component, too, that's compatible with auto-form/react-hook-form... worth investigating.

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+const ScalarControl = ({
+  spec,
+  value,
+  onChange,
+  id,
+  invalid,
+}: {
+  spec: RawFieldSpec;
+  value: string | boolean;
+  onChange: (value: unknown) => void;
```
eblairmckee, line 532: seems a little off.... should was have a `Value: string | boolean` and reuse for the `onChange` value type?

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+    return null;
+  }
+  if (spec.type === 'int' && !INT_VALUE_RE.test(text)) {
+    return "Not a valid integer — this change won't be saved until fixed.";
+  }
+  if (spec.type === 'float' && Number.isNaN(Number(text))) {
+    return "Not a valid number — this change won't be saved until fixed.";
+  }
+  return null;
+}
```
eblairmckee, line 584: again with reusability. these could be lovely helpers for auto-form instances

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+
+// biome-ignore lint/suspicious/noTemplateCurlyInString: a literal Connect secret reference, not a JS template
+const SECRET_REF_EXAMPLE = '${secrets.MY_SECRET}';
```
eblairmckee: also pretty sure we have a function for this already

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  children: React.ReactNode;
+}) => (
+  <Collapsible className="rounded-md border border-border/60" defaultOpen={defaultOpen}>
+    <CollapsibleTrigger className="group flex w-full items-center justify-between px-3 py-2 text-left">
+      <Text variant="bodyStrongMedium">{label}</Text>
+      <ChevronDown
+        className={cn('size-4 text-muted-foreground transition-transform group-data-[panel-open]:rotate-180')}
+      />
+    </CollapsibleTrigger>
+    <CollapsibleContent className="flex flex-col gap-4 px-3 pt-1 pb-3">{children}</CollapsibleContent>
```
eblairmckee, line 667: this is basically ui registry Accordion. worth looking at whether it's a viable substitute rather than yet another one-off

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+  fields: RawFieldSpec[];
+  path: string[];
+  control: Control<FormValues>;
+}) => {
+  const formFields = fields.filter(isFormField);
+  const ordered = [
+    ...formFields.filter((f) => checkRequired(f) && !f.advanced),
+    ...formFields.filter((f) => !(checkRequired(f) || f.advanced)),
+    ...formFields.filter((f) => f.advanced),
+  ];
```
eblairmckee: nit: a flapMap + predicate util would be lovely to use here.

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
SpicyPete: This was mainly for the styling, changed to a Text component so its not a dangling label

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+                        />
+                      </div>
+                      {invalid ? (
+                        <Text className="text-destructive" variant="bodySmall">
+                          Invalid YAML — these settings won't be saved until fixed.
+                        </Text>
+                      ) : null}
+                    </div>
+                  );
+                }}
```
eblairmckee, line 936: this would be a lot more readable if we hoisted this to a constant and returned most -> least specific conditionals.

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+          {!isListValued && showRaw ? (
+            <FieldGroup defaultOpen={false} label="Other settings (YAML)">
+              <Controller
+                control={control}
+                name="raw"
+                render={({ field }) => {
+                  const invalid = field.value.trim() !== '' && parseRawSection(true, field.value) === null;
+                  return (
+                    <div className="flex flex-col gap-1.5">
+                      <div className="h-[200px] overflow-hidden rounded-md border border-border">
```
eblairmckee: why `overflow-hidden`? seems like a broken UI for longer (over 200px high) yaml configs

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -0,0 +1,1014 @@
...
+
+          {!isListValued && componentFields.length > 0 && hasChildList ? (
+            <ChildItemsList
+              items={childItems as InspectorChildItem[]}
+              label="Steps"
+              onSelect={onSelectChild as (item: InspectorChildItem) => void}
+            />
+          ) : null}
+          {!isListValued && componentFields.length > 0 && !hasChildList ? (
+            <div className="rounded-md border border-border/60 border-dashed px-3 py-2">
```
eblairmckee: I'd be curious to see a screenshot of this.... dashed borders are typically used for placeholder/empty/you-could-add-something-optional form sections

frontend/src/components/pages/rp-connect/pipeline/node-inspector.tsx
```diff
@@ -0,0 +1,842 @@
...
+        onClose={onClose}
+        onDelete={handleDelete}
+        onOpenInYaml={onOpenInYaml}
+        usedByCount={resourceLabel ? usedByCount : undefined}
+      />
+      {lintHints && lintHints.length > 0 ? <InspectorLintErrors hints={lintHints} /> : null}
+      {danglingRef && !readOnly ? (
+        <DanglingRefBanner onCreate={handleCreateMissingResource} refLabel={danglingRef.ref} />
+      ) : null}
+      {(() => {
```
eblairmckee: minor: can we lift this to a constant... I lean against inline function calls just to render more jsx... screams that it should be its own subcomponent.. or const within component body

frontend/src/components/pages/rp-connect/pipeline/node-inspector.tsx
```diff
@@ -0,0 +1,842 @@
...
+          ) : null;
+        if (readOnly) {
+          return (
+            <>
+              {conditionSection}
+              <ReadOnlyComponent component={component} />
+            </>
+          );
+        }
+        if (useForm && spec) {
```
eblairmckee: my earlier comment about early returns for most -> least specific conditions... this is a great example of that. let's hoist this to outside the render

frontend/src/components/pages/rp-connect/pipeline/node-inspector.tsx
```diff
@@ -0,0 +1,842 @@
...
+    <ul className="flex flex-col gap-1">
+      {hints.map((hint, i) => (
+        <li className="text-destructive text-xs" key={`${hint.line}-${hint.column}-${i}`}>
+          {hint.hint}
+          {hint.line > 0 ? <span className="text-destructive/70"> (line {hint.line})</span> : null}
+        </li>
+      ))}
+    </ul>
+  </div>
+);
```
eblairmckee, line 489: looks like it could be a registry Alert

frontend/src/components/pages/rp-connect/pipeline/node-inspector.tsx
```diff
@@ -0,0 +1,842 @@
...
+
+// Write a component-config draft at `target`, cascading a resource-label rename to every component
+// that references it (so the link is never silently broken).
+function applyComponentDraft(
```
eblairmckee, line 493: I am dying to know how claude decides whether a function should be a helper outside of the component body... vs a callback... this screams "should've been a callback"

frontend/src/components/pages/rp-connect/pipeline/node-inspector.tsx
SpicyPete: Screenshot 2026-07-10 at 4 03 37 PM

Maybe, but in this instance then we'll have more boxes-in-boxes, so a custom full-width alert area makes sense to me

frontend/src/components/pages/rp-connect/pipeline/node-inspector.tsx
SpicyPete: This is a pure function which doesn't rely on anything in the component, so keeping it out should be okay

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
SpicyPete: This is for the extra yaml field within the side-panel, so it needs some height set and overflow-hidden, I will increase the height so more can fit in, but the extra yaml field for this is likely rarely used

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
SpicyPete: An accordion is for multiple expanding areas, so if the look of an accordion changes, it will unexpectedly break here.  A generic Collapsible is less likely to lead to future breakage

SpicyPete:
@claude review

eblairmckee:
@claude conduct an adversarial review using this skill: https://claude.ai/customize/skills?selectedId=skill_01U57srusYSMerLEneHv2hgM
specifically, I want you to focus on:
1. the various utils modified for parsing redpanda connect schemas and yaml. 
2. any not feature flagged change is well tested and regardless of whether visual editor flag is on/off everything will still work.

eblairmckee (approved):

PR #2557, opened 2026-07-14
UX: New Schema Registry Sync options in Shadow Link (Create Only)

r-vasquez:
This PR adds the new Schema Registry Sync option (sync over API) to the Shadow Link create form. This is gated behind a backend-driven change that looks for the first reported version of the cluster, in which the cluster must be of a version > 26.2.0. Older clusters won't see this change.

A follow-up PR will have the bits for handling the Edit and the View page with these new fields.

**Example**

image


backend/pkg/api/connect/service/license/service_test.go
```diff
@@ -1,40 +0,0 @@
-// Copyright 2024 Redpanda Data, Inc.
```
r-vasquez, line 1: Moved this to pkg/redpanda/version_test.go

backend/pkg/redpanda/version.go
```diff
@@ -0,0 +1,67 @@
...
+type Version struct {
+	Major   int
+	Feature int
+	Patch   int
+}
+
+// versionRegex extracts the version core from a broker-reported string such as
+// "v26.2.1 - 491e56900d2316fcbb22aa1d37e7195897878309", "v26.2.0-rc1", or
+// "v26.2.0-dev".
+var versionRegex = regexp.MustCompile(`^v?(\d+)\.(\d+)\.(\d+)(?:\s|-rc\d+|-dev|-nightly|$)`)
```
r-vasquez, line 31: Taken from https://github.com/redpanda-data/redpanda/blob/dev/src/go/rpk/pkg/redpanda/version.go#L23

frontend/package.json
```diff
@@ -87,7 +87,7 @@
     "@stepperize/react": "^5.1.9",
     "@tanstack/react-form": "^1.12.3",
     "@tanstack/react-query": "^5.99.2",
-    "@tanstack/react-router": "^1.170.17",
+    "@tanstack/react-router": "^1.170.7",
```
r-vasquez: Ignore, I need to rebase with `master`

r-vasquez (commented):
In hindisight I should have done first the Edit and View pages as this PR has some 'weird' logic to gate the view/edit page to the old switch. However, this was easier to port while we merge the cloud-ui PR.

Also, we need the Cloud UI PR in first before we do view/edit.

c-julin (approved):

PR #2558, opened 2026-07-14
shadowlink: survive slow status aggregation on large clusters

r-vasquez:
Redpanda builds shadow link status by fanning out
to every broker, so on large clusters any shadow
link request can exceed rpadmin's default 10s
client timeout. The pester client then retried
the expensive aggregation two more times, and the
route loaders awaited the result with no pending
UI and rethrew the failure, so users saw a blank
page followed by the raw error boundary.

The client factory now accepts per-request
timeout and max-attempt options that map onto
rpadmin options; console-enterprise consumes them
for shadow link calls, which is why they have no
in-repo caller. The router renders a pending
fallback while loaders run, and the shadowlink
loaders swallow Connect errors so the pages can
render their own error states with retry instead
of failing the route.

Example of the loading state:

image

### DIFF
--- a/backend/pkg/factory/redpanda/redpanda.go
+++ b/backend/pkg/factory/redpanda/redpanda.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"time"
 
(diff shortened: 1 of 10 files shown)


eblairmckee (approved):
this reminds me we need to figure out what the loading UI should look like for a whole route that's being loaded... another day.

PR #2560, opened 2026-07-15
RPCN - Support for optional fields

SpicyPete:
# RPCN: correct optional-field support, lint-anchored editing, and topic selection in the visual editor

## Background

cloudv2 #27724 (merged 2026-07-14) bumped `benthos/v4` 4.53 → 4.73 in `redpanda-connect-api`, which re-serializes the Connect config schema before serving it. Benthos only began emitting per-field `is_optional` / `is_advanced` / `is_secret` / `is_deprecated` at v4.59, so until that bump the console never received them and the frontend compensated with a defensive heuristic ladder (`checkRequired`) that guessed required-ness from proto flags and name patterns.

The two RPCs still differ in what they carry (verified against cloudv2 source and the embedded RPCN 4.100 schema):

| Signal | `ListComponents` (proto `FieldSpec`) | `GetPipelineServiceConfigSchema` (raw JSON) |
|---|---|---|
| `optional` / `advanced` / `deprecated` | ✅ | ✅ |
| `is_secret` | ❌ no field in the proto at all | ✅ |
| defaults | ⚠️ string defaults only — int/bool/collection defaults arrive as `''` | ✅ exact `required` arrays computed by the backend with full default knowledge |

So required/secret truth must come from the raw JSON path; the proto alone cannot distinguish "required" from "default lost in serialization".

This branch adopts those restored signals end-to-end, then fixes the editing friction they exposed: lint errors referencing invisible line numbers, edits that only saved on node-deselect, and topic fields that were plain text boxes.

### Schema-signal enrichment
- New `enrichComponentsWithConfigSchema` (`utils/schema-enrichment.ts`) stamps `requiredBySchema` and `secret` onto proto-derived specs by walking the raw config-schema JSON (object/array/map wrappers unwrapped so JSON-Schema nodes line up with proto children).
- The enrichment gate keys on the presence of `required` arrays (long-standing in the JSON) rather than `is_optional` (4.59+ only), so mid-generation dataplanes that serve `required` arrays without per-field flags still get correct required/optional grouping.
- Curated `LINT_REQUIRED_FIELDS` overrides for requirements the schema can't express (`input:redpanda` `topics`/`consumer_group` are `.Optional()` in the engine but enforced conditionally by lint) — the form shows them as required instead of contradicting the linter.
- Degrades gracefully: a schema that is missing, unparsable, or pre-enrichment leaves specs unstamped.
- New `useEnrichedComponents` hook consolidates the recipe for all three YAML-generating consumers (pipeline editor, onboarding wizard, template form panel) and keeps `components` empty until the schema query settles, so YAML is never generated from un-enriched specs. Display-only consumers (gallery grid, tiles, command palette) intentionally stay on plain `parseSchema`.

### Required-field detection
- `checkRequired` is now: deprecated guard → trust `requiredBySchema` stamp → documented degraded proto ladder (for pre-4.59 dataplanes only). Dead heuristic branches deleted.
- `generateDefaultValue` no longer zero-fills fields whose default was dropped by the proto — they're omitted so engine defaults apply (previously e.g. `auto_replay_nacks` defaulting to `true` was rendered as `false`).
- Field-less components (`drop`, etc.) generate `{}` instead of nothing, so they render as canvas nodes and produce valid YAML.

### Lint errors anchored to form fields
- New `lint-field-mapping.ts` maps a node's lint hints onto the inspector's form fields via three tiers: explicit `field x` tokens in the message, the hint's line number resolved through the parsed document's node ranges (deepest field wins, walking up to the nearest rendered ancestor), and loose name mentions (`consumer group` finds `consumer_group`).
- Errors render under the field they're about — including group-level errors like `expected object value, got !!null (batching)` inside their section — and are hidden while the field is dirty so stale complaints don't fight live edits.
- Collapsed groups (including Advanced) force-open when new errors anchor inside them; a manual collapse sticks until the error set actually changes.
- Hints that anchor to nothing the form renders stay in the banner, labeled with the field path instead of a line number the Visual view can't show.
- Empty configs skip the server lint round-trip entirely (it could only say "config_yaml: value is required").

### Field-level commits
- Edits now flush to the YAML on field blur, ⌘⏎, or an explicit Apply affordance in a pending-edits footer — not just on node-deselect — so the canvas, structure tree, and lint react while you're still in the field.
- Commits keep the form mounted (focus and scroll survive); external YAML writes re-sync a clean form in place and rebuild pending drafts from the fresh value so nothing clobbers concurrent writes (e.g. a topic created mid-edit).
- Malformed numeric edits and invalid raw-YAML sections are excluded from commits and visibly stay dirty instead of silently reverting.
- Undo/redo remounts the form so stale edit state can't resurrect undone values.

### Topic selection on topic fields
- `topic`/`topics` fields on Redpanda-family components (`kafka`, `kafka_franz`, `redpanda*` — same gate as the template flow) get cluster-aware pickers: a creatable combobox on scalar fields (existing topics, free text for interpolations/`topic:partition`) and an append-only picker under the topics textarea that preserves free-text lines.
- A "Create new topic" affordance opens the Add-topic dialog; the created topic is patched into the component's YAML — appending to an existing `topics` list in place (dedup, comments and non-string entries preserved).
- Foreign connectors' topic fields (`mqtt`, `gcp_pubsub`, `pulsar`, …) stay plain inputs — their topics are not cluster topics.

### Editor polish
- Minimap no longer clips flow edges after adding nodes (world bounds union the viewport).
- Unsaved-state indicators (canvas chip, node dots, structure tree, header, inspector footer) switched from warning-orange to the informative tone so they read as state, not a second tier of error.
- Overlong default-value hints truncate with a title tooltip; "Optional" grouping is dropped when a component has no required fields (everything rendered plainly).
- Create-topic dialog steps in the template gallery and editor sized to their content (no dead space / floating buttons); add-topic step describes exactly what the current mode allows.

### Template fixes (drift caught by this work)
Four templates referenced components/fields the engine never shipped or has since renamed — invisible because the test snapshot was frozen at RPCN 4.65 with the components allowlisted as known-missing:
- **dynamodb-cdc**: `aws_dynamodb_stream` → `aws_dynamodb_cdc` (`tables` list, `start_from`/`snapshot_mode`)
- **oracle-cdc**: `oracle_cdc` → `oracledb_cdc` (`connection_string`, `include`)
- **sqlserver-cdc**: `sql_server_cdc` → `microsoft_sql_server_cdc` (`connection_string`, `include`)
- **iceberg**: `catalog_uri`/`warehouse` → `catalog.url` / optional `catalog.warehouse` — this one was user-visible broken: the component exists, so the template showed in the gallery and its YAML failed dataplane lint on every deploy
- Refreshing `assets/rp-connect-schema-full.json` 4.65 → RPCN 4.100 (the cloud schema cloudv2 actually serves) also caught `redpanda_migrator` having dropped its `batching` field; removed from the mirror template.
- `KNOWN_MISSING_COMPONENTS` is now **empty** — every template is fully validated by the regression tests (schemaField paths resolve, stitched YAML contains only known fields, required fields always emitted).

### Docs links module
- All `docs.redpanda.com` URLs across the app now live in `src/utils/docs-links.ts` (~25 call sites migrated). Every entry is canonical against the restructured docs site (direct 200s, no redirects).

## Behavior changes to be aware of
- The redpanda input's `seed_brokers` is no longer marked required (genuinely `.Optional()` in the engine); `topics`/`consumer_group` remain required via the curated lint override. The wizard still auto-populates them.
- Optional fields whose defaults don't survive the proto are omitted from starter YAML instead of being zero-filled.
- Hardcoded template slot `required` values were audited against the served schema: templates stricter than the engine (e.g. S3 region/credentials) are deliberate curation; the one looser case (postgres-cdc `slot_name`) is safe because `defaultWhenBlank` always supplies a value.
- Creating a topic from the editor now *appends* to an existing `topics` list instead of replacing it.


SpicyPete:
@claude review

SpicyPete:
@claude review

frontend/src/components/pages/rp-connect/onboarding/add-topic-step.tsx
```diff
@@ -291,8 +291,15 @@ export const AddTopicStep = forwardRef<BaseStepRef<AddTopicFormData>, AddTopicSt
...
             <FormLabel>Topic name</FormLabel>
+            {/* Only offer what the mode can actually do — "create a new topic" with no input reads as broken. */}
             <FormDescription>
-              Choose an existing topic to read or write data from, or create a new topic.
+              {
+                {
+                  new: 'Enter a name for the new topic.',
+                  existing: 'Choose an existing topic to read or write data from.',
+                  both: 'Choose an existing topic to read or write data from, or create a new topic.',
+                }[selectionMode]
```
eblairmckee: nit: why not hoist? kind of weird to see an object in the render body like this

frontend/src/components/pages/rp-connect/onboarding/onboarding-wizard.tsx
```diff
@@ -61,11 +60,7 @@ export const ConnectOnboardingWizard = ({
 }: ConnectOnboardingWizardProps = {}) => {
   const navigate = useNavigate();
 
-  const { data: componentListResponse, isLoading: isComponentListLoading } = useListComponentsQuery();
-  const components = useMemo(
-    () => (componentListResponse?.components ? parseSchema(componentListResponse.components) : []),
-    [componentListResponse]
-  );
+  const { components, componentList, isLoading: isComponentListLoading } = useEnrichedComponents();
```
eblairmckee, line 63: nice

frontend/src/components/pages/rp-connect/pipeline/lint-field-mapping.ts
```diff
@@ -0,0 +1,206 @@
...
+import type { LintHint } from '@buf/redpandadata_common.bufbuild_es/redpanda/api/common/v1/linthint_pb';
+import { isMap, isScalar, LineCounter, parseDocument } from 'yaml';
+
+import { type EditTarget, editTargetPath } from '../utils/yaml';
+
+/**
+ * Anchors a node's lint hints to the inspector form fields they belong to, so an error renders
+ * under the field (or inside the group section) it's about instead of as "(line 12)" in a banner
+ * the user has to decode. See `resolveHint` for the anchoring tiers.
+ */
```
eblairmckee: orphaned?

frontend/src/components/pages/rp-connect/pipeline/lint-field-mapping.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/** Field/group key (`path.join('/')`) → the lint messages anchored to it. */
```
eblairmckee, line 17: i'm gonna guess something weird happened wrt comments in this file?

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -284,11 +294,14 @@ function coerceScalar(spec: RawFieldSpec, raw: string | boolean): string | numbe
   return text;
 }
 
-function coerceArrayItems(spec: RawFieldSpec, text: string): unknown[] {
-  const lines = text
+const arrayFieldLines = (text: string): string[] =>
```
eblairmckee, line 299: nit: for the functions in this file I wonder if they already exist in a util file. some look a little familiar

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -482,6 +573,93 @@ const SecretInput = (props: { id?: string; value: string; onChange: (value: unkn
...
+  const { onCreateTopic } = useContext(ResourceFieldContext);
+  if (!onCreateTopic) {
+    return null;
+  }
+  return (
+    <Button className="self-start" onClick={onCreateTopic} size="sm" type="button" variant="outline">
+      <Plus className="size-3.5" /> Create new topic
+    </Button>
+  );
+};
```
eblairmckee, line 597: [code smell]: any particular reason why you're using context to pass ar[ound event handlers like this?

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -482,6 +573,93 @@ const SecretInput = (props: { id?: string; value: string; onChange: (value: unkn
...
+      />
+      <CreateTopicButton />
+    </div>
+  );
+};
+
+// A select can't be cleared by deleting text the way inputs can, so without an explicit unset
+// item a value (often seeded by the insert template) is stuck in the YAML forever. Maps to ''
+// which the save path treats as "remove the key". Sentinel because '' can't be an item value.
+const UNSET_OPTION = '__unset__';
```
eblairmckee: [nit]: a lot of these comments discuss business logic but seem to live in the wrong places... like a sentinel like this is self-explanatory based on the constant name... hwo it's used should not be relevant here, rather when rendering it if there are nuances to comment about that would imrpove future contributors' devex.. then comment verbosely

frontend/src/components/pages/rp-connect/pipeline/node-config-form.tsx
```diff
@@ -809,121 +1104,188 @@ export function NodeConfigForm({
...
-                    </div>
-                  );
-                }}
-              />
-            </FieldGroup>
+                        {invalid ? (
+                          <div className="text-body-sm text-destructive">
+                            Invalid YAML — these settings won't be saved until fixed.
+                          </div>
+                        ) : null}
```
eblairmckee, line 1266: nice

frontend/src/components/pages/rp-connect/template-gallery/known-missing-components.ts
```diff
@@ -14,9 +14,4 @@
  * (enterprise-only or pending a snapshot refresh). Schema regression tests skip these;
  * prune as components land in the snapshot.
  */
-export const KNOWN_MISSING_COMPONENTS: ReadonlySet<string> = new Set([
-  'aws_dynamodb_stream',
-  'oracle_cdc',
-  'sql_server_cdc',
-  'iceberg',
-]);
+export const KNOWN_MISSING_COMPONENTS: ReadonlySet<string> = new Set([]);
```
eblairmckee: [minor]: remove if orphaned? or no use in future?

frontend/src/components/pages/rp-connect/types/schema.ts
```diff
@@ -4,6 +4,12 @@ import type { ComponentSpec, ComponentStatus, FieldSpec } from 'protogen/redpand
 export interface RawFieldSpec extends Omit<FieldSpec, 'children'> {
   comment?: string; // Keep for internal use (YAML generation)
   children?: RawFieldSpec[];
+  // Stamped by enrichComponentsWithConfigSchema from the raw config schema JSON, which carries
+  // signals the proto FieldSpec drops (is_secret entirely; required-ness for fields whose
+  // non-string default is lost in serialization). Both stay undefined when the dataplane's
+  // schema predates flag serialization (benthos < 4.59) — consumers then fall back to proto flags.
```
eblairmckee: wrong line?

frontend/src/components/pages/rp-connect/utils/schema-enrichment.ts
```diff
@@ -0,0 +1,164 @@
...
+    };
+  });
+}
+
+// Stamps from the raw schema; returns the input untouched when the schema is missing, unparsable,
+// or carries no signal. `required` arrays are the load-bearing signal and benthos has emitted them
+// (computed knowing every default) far longer than the per-field flags (`is_optional`/`is_secret`,
+// 4.59+) — so a mid-generation document still fixes required-ness even though it can't flag
+// secrets (the name heuristic still covers those). A document with no `required` keys at all
+// predates the arrays; stamping `requiredBySchema: false` everywhere from it would be a lie.
```
eblairmckee: [nit]: would be nice to stick to jsdoc comments

eblairmckee (approved):
lgtm. left some minor nits that can be addressed in a follow

PR #2563, opened 2026-07-17
Fixing a key collision in RPCN pipeline list

SpicyPete:
## Fix duplicate React keys in the RPCN pipeline list

The Input/Processors/Output columns keyed connector badges by component name, but `parseConfigComponents` intentionally preserves duplicates (e.g. a `broker` input with two `redpanda` children yields `['redpanda', 'redpanda']`), so pipelines reusing a component triggered React's duplicate-key warning (`Encountered two children with the same key, '.$redpanda'`) on every render.

### Changes

- Add a `toKeyedNames` helper that suffixes each name with its occurrence count (`redpanda-0`, `redpanda-1`) and use it for the badge cells and their "+N more" overflow lists. Occurrence counts are used instead of array indexes to keep keys stable and satisfy Biome's `noArrayIndexKey`.
- Suppress the pre-existing `noArrayIndexKey` warning on the static skeleton rows, leaving the file lint-clean.

No behavior change — duplicate badges still render; they just get unique keys.

### DIFF
--- a/frontend/src/components/pages/rp-connect/pipeline/list.tsx
+++ b/frontend/src/components/pages/rp-connect/pipeline/list.tsx
@@ -102,7 +102,14 @@ const transformAPIPipeline = (apiPipeline: APIPipeline): Pipeline => {
   };
 };
 
-// Names can repeat (e.g. two `redpanda` inputs), so suffix each occurrence for unique keys.
+/**
+ * Pairs each name with a unique React key by suffixing its occurrence index,
+ * since component names can repeat (e.g. two `redpanda` inputs).
+ *
+ * @param names - Component names, possibly containing duplicates.
+ * @returns One entry per input name, e.g. `["redpanda", "redpanda"]` →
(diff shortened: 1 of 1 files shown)


frontend/src/components/pages/rp-connect/pipeline/list.tsx
```diff
@@ -102,6 +102,16 @@ const transformAPIPipeline = (apiPipeline: APIPipeline): Pipeline => {
   };
 };
 
+// Names can repeat (e.g. two `redpanda` inputs), so suffix each occurrence for unique keys.
+const toKeyedNames = (names: string[]): { name: string; key: string }[] => {
```
jvorcak, line 113: shall we provide proper JSdoc + unit tests so that we know what this function does at the first look?

SpicyPete: I updated to a JSDoc format.  I think a unit test is a bit overkill as it's a 5 line local function

jvorcak (approved):

PR #2564, opened 2026-07-17
fe/shadowlink: add read-only schema registry section

r-vasquez:
The details view only showed an Enabled/Disabled badge for _schemas topic shadowing. Schema Registry can now replicate over the source REST API, so add a read-only card for that mode and map the new options through both the dataplane and controlplane paths.

Supporting the controlplane case meant upgrading the redpandadata/cloud buf SDK, which also dropped the unused resource_group_id field from the ShadowLink, so remove it from the model too.

### **Examples:**
**Over API**
image

**_schemas topic**
image

**sync disabled**
image


frontend/src/components/pages/shadowlinks/details/config/configuration-schema-registry.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+    expect(screen.getByTestId('schema-registry-mode-badge')).toHaveTextContent('Sync over API');
+    expect(screen.getByTestId('schema-registry-status-badge')).toHaveTextContent('Enabled');
+    expect(screen.getByTestId('sr-config-source-url-value')).toHaveTextContent(
+      'https://psrc-x1234.us-east-1.aws.confluent.cloud'
+    );
+
+    expect(screen.getByTestId('sr-config-auth-type-value')).toHaveTextContent('HTTP Basic');
+    expect(screen.getByTestId('sr-config-auth-username-value')).toHaveTextContent('sr-replicator');
+    expect(screen.getByTestId('sr-config-auth-password-value')).toHaveTextContent('Last changed Jul 1, 2026 at 12:00');
```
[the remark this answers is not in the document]
r-vasquez: We pin TZ=GMT to all of our tests: https://github.com/redpanda-data/console/blob/master/frontend/package.json#L25-L34

frontend/src/components/pages/shadowlinks/details/config/configuration-schema-registry.tsx
```diff
@@ -0,0 +1,384 @@
...
+  testId,
+  value,
+}: {
+  label: string;
+  mono?: boolean;
+  narrow?: boolean;
+  testId: string;
+  value: React.ReactNode;
+}) => (
+  <div className={cn('grid items-baseline gap-x-4', narrow ? 'grid-cols-[120px_1fr]' : 'grid-cols-[200px_1fr]')}>
```
graham-rp, line 104: [question] I may be slightly out of the loop, but I feel like we should have a component for this type of thing, right?

graham-rp (approved):
I wasn't able to replicate the whole thing locally, but looks good overall

frontend/src/components/pages/shadowlinks/details/config/configuration-schema-registry.tsx
r-vasquez: The closest I could find was https://redpanda-ui-registry.netlify.app/docs/item, but it does not align with the `label:  value` thing we are doing here.

I asked Claude to look for an alternative and pointed me that it took inspiration from https://github.com/redpanda-data/console/blob/master/frontend/src/components/pages/rp-connect/pipeline/details-dialog.tsx when porting the design.

I will leave the comment open if anyone finds a better alternative, though. Thanks!

PR #2569, opened 2026-07-20
fe/shadowlink: make schema registry section editable

r-vasquez:
The edit page only exposed the legacy _schemas switch; API-mode links were read-only. Render the redesigned section on edit, behind the same feature gate as the create wizard.

Secrets are never seeded into the form: sending them empty on update keeps the stored values. Fields without a UI (paused, SNI opt-out, file-path TLS) are round-tripped so a message rebuild does not reset them. Direct topic<->api switches are unsupported by the backend, so those tabs are locked.

While the gate is closed the SR validation rules are skipped so the api-mode slice does not block saving other fields.

**Before:**

image

**Now:**
image

^ Note that the _schemas tab is unavailable to prevent users from switching modes as stated above.

### DIFF
--- a/frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/source-connection-section.tsx
+++ b/frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/source-connection-section.tsx
@@ -20,11 +20,26 @@ import {
 import { Input } from 'components/redpanda-ui/components/input';
 import { Tabs, TabsList, TabsTrigger } from 'components/redpanda-ui/components/tabs';
 import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from 'components/redpanda-ui/components/tooltip';
+import { SecretSelector, type SecretSelectorCustomText } from 'components/ui/secret/secret-selector';
+import { isEmbedded } from 'config';
 import { InfoIcon } from 'lucide-react';
+import { Scope } from 'protogen/redpanda/api/dataplane/v1/secret_pb';
+import { useMemo } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
+import { useListSecretsQuery } from 'react-query/api/secret';
 
 import { SrTlsConfiguration } from './sr-tls-configuration';
+import { extractSecretId, toSecretReference } from '../../connection/secret-reference';
 import { type FormValues, SR_AUTH_METHOD } from '../../model';
+
+const SR_PASSWORD_SECRET_TEXT: SecretSelectorCustomText = {
+  dialogDescription:
+    'Create a new secret for your Schema Registry HTTP Basic password. The secret will be stored securely.',
+  secretNamePlaceholder: 'e.g., SCHEMA_REGISTRY_PASSWORD',
+  secretValuePlaceholder: 'Enter password...',
+  secretValueDescription: 'Your Schema Registry HTTP Basic password',
+  emptyStateDescription: 'Create a secret to securely store your Schema Registry password',
+};
 
 const InfoTooltip = ({ content }: { content: string }) => (
   <TooltipProvider>
@@ -40,6 +55,17 @@ const InfoTooltip = ({ content }: { content: string }) => (
 export const SourceConnectionSection = () => {
   const { control, setValue } = useFormContext<FormValues>();
   const authMethod = useWatch({ control, name: 'schemaRegistry.authMethod' });
+
+  // Fetch secrets for SecretSelector (embedded mode only)
+  const { data: secretsData } = useListSecretsQuery({}, { enabled: isEmbedded() });
+  const availableSecrets = useMemo(() => {
+    if (!secretsData?.secrets) {
+      return [];
+    }
+    return secretsData.secrets
+      .filter((secret): secret is NonNullable<typeof secret> & { id: string } => Boolean(secret?.id))
+      .map((secret) => ({ id: secret.id, name: secret.id }));
+  }, [secretsData]);
 
   const handleAuthMethodChange = (next: string) => {
     if (next !== SR_AUTH_METHOD.NONE && next !== SR_AUTH_METHOD.BASIC) {
@@ -115,7 +141,21 @@ export const SourceConnectionSection = () => {
                   <InfoTooltip content="For Confluent Cloud, this is the API secret." />
                 </FormLabel>
                 <FormControl>
-                  <Input {...field} testId="sr-basic-password-input" type="password" />
+                  {isEmbedded() ? (
+                    <SecretSelector
+                      availableSecrets={availableSecrets}
+                      customText={SR_PASSWORD_SECRET_TEXT}
+                      minValueLength={1}
+                      onChange={(secretId) => {
+                        field.onChange(secretId ? toSecretReference(secretId) : '');
+                      }}
+                      placeholder="Select or create password secret"
+                      scopes={[Scope.REDPANDA_CLUSTER]}
+                      value={extractSecretId(field.value)}
+                    />
+                  ) : (
+                    <Input {...field} testId="sr-basic-password-input" type="password" />
+                  )}
                 </FormControl>
                 <FormMessage />
               </FormItem>
--- a/frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/sr-tls-configuration.tsx
+++ b/frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/sr-tls-configuration.tsx
@@ -10,21 +10,36 @@
  */
 
 import { Badge } from 'components/redpanda-ui/components/badge';
+import { Button } from 'components/redpanda-ui/components/button';
 import { Collapsible, CollapsibleContent, CollapsibleTrigger } from 'components/redpanda-ui/components/collapsible';
 import { Dropzone, DropzoneContent, DropzoneEmptyState } from 'components/redpanda-ui/components/dropzone';
 import { FormControl, FormField, FormItem, FormLabel, FormMessage } from 'components/redpanda-ui/components/form';
 import { Switch } from 'components/redpanda-ui/components/switch';
 import { cn } from 'components/redpanda-ui/lib/utils';
+import { SecretSelector, type SecretSelectorCustomText } from 'components/ui/secret/secret-selector';
+import { isEmbedded } from 'config';
 import { Check, ChevronRight, Trash2, UploadCloud } from 'lucide-react';
+import { Scope } from 'protogen/redpanda/api/dataplane/v1/secret_pb';
 import { useCallback, useMemo, useState } from 'react';
 import { useFormContext, useFormState, useWatch } from 'react-hook-form';
+import { useListSecretsQuery } from 'react-query/api/secret';
 import { docsLinks } from 'utils/docs-links';
 
-import { type FormValues, hasSrCertificate } from '../../model';
+import { extractSecretId, toSecretReference } from '../../connection/secret-reference';
+import { type FormValues, hasSrCertificate, hasSrClientKeyMaterial, type SchemaRegistryFormValues } from '../../model';
 
 const TLS_DOCS_URL = docsLinks.selfManaged.shadowingNetworkSetup;
 
 type SrCertificateType = 'ca' | 'clientCert' | 'clientKey';
+
+const SR_CLIENT_KEY_SECRET_TEXT: SecretSelectorCustomText = {
+  dialogDescription:
+    'Create a new secret for your Schema Registry mTLS client private key. The secret will be stored securely.',
+  secretNamePlaceholder: 'e.g., SR_MTLS_CLIENT_KEY',
+  secretValuePlaceholder: 'Paste PEM-encoded private key...',
+  secretValueDescription: 'Your Schema Registry mTLS client private key (PEM format)',
+  emptyStateDescription: 'Create a secret to securely store your Schema Registry mTLS client private key',
+};
 
 const CERTIFICATE_ACCEPT = {
   'application/x-pem-file': ['.pem'],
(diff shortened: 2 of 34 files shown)


frontend/src/components/pages/shadowlinks/shadowlink-test-helpers.ts
```diff
@@ -12,6 +12,26 @@
...
+  useSupportedFeaturesStore.getState().setEndpointCompatibility({
+    kafkaVersion: 'v26.2.0',
+    endpoints: [
+      {
+        endpoint: Feature.ShadowLinkSchemaRegistrySync.endpoint,
+        method: Feature.ShadowLinkSchemaRegistrySync.method,
+        isSupported,
+      },
+    ],
+  });
```
malinskibeniamin, line 33: This is clean!

frontend/src/components/pages/shadowlinks/edit/schema-registry-edit-section.tsx
```diff
@@ -0,0 +1,87 @@
...
+const ModeTransitionAlert = ({ originalMode }: { originalMode: SchemaRegistryMode }) => {
+  const { control } = useFormContext<FormValues>();
+  const mode = useWatch({ control, name: 'schemaRegistry.mode' });
+
+  if (originalMode === SCHEMA_REGISTRY_MODE.TOPIC && mode !== SCHEMA_REGISTRY_MODE.TOPIC) {
+    return (
+      <Alert testId="sr-mode-transition-topic-alert" variant="warning">
+        <TriangleAlert />
+        <AlertDescription>
+          Turning off _schemas topic shadowing does not remove the _schemas shadow topic if it was already added. To
```
malinskibeniamin: As discussed today during design review meeting, we may want to decide how to handle the internal `_schemas` topic mentions

frontend/src/components/pages/shadowlinks/mappers/schema-registry.ts
```diff
@@ -165,6 +173,124 @@ function mapSchemaRegistryApiOptions(api: SchemaRegistryApiOptionsMessage): Unif
...
+  const seconds = totalMs / 1000;
+  if (seconds % 60 !== 0) {
+    return `${seconds}s`;
+  }
+  const minutes = seconds / 60;
+  if (minutes % 60 !== 0) {
+    return `${minutes}m`;
+  }
+  return `${minutes / 60}h`;
+}
```
malinskibeniamin, line 217: I think proto has a utility function built for that, or maybe we use date-fns instead

frontend/src/components/pages/shadowlinks/mappers/schema-registry.ts
```diff
@@ -165,6 +173,124 @@ function mapSchemaRegistryApiOptions(api: SchemaRegistryApiOptionsMessage): Unif
...
+
+  // Every field is set explicitly (no base spread) so adding a schema field
+  // without deciding its hydration fails to compile.
+  return {
+    mode: SCHEMA_REGISTRY_MODE.API,
+    sourceUrl: api.sourceUrl,
+    authMethod: basic ? SR_AUTH_METHOD.BASIC : SR_AUTH_METHOD.NONE,
+    // The password is never returned; like the connection credentials it is
+    // re-entered on save.
+    basicCredentials: { username: basic?.username ?? '', password: '' },
```
malinskibeniamin, line 265: [P1 Major] Keep the stored Basic password on edit

What: API-mode hydration deliberately leaves this write-only password blank, but the edit page still uses `FormSchema`, whose Basic-auth rule rejects a blank password; the new page test explicitly confirms Save is blocked until the secret is re-entered.

Why: Operators cannot edit any API-mode setting unless they know or rotate the existing credential, even though the PR contract and Redpanda update converter both support blank password = keep existing.

Suggested fix: Hydrate `passwordSet` into form state and require a password only for create/new Basic auth; allow blank when an existing password is configured.

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, update `create/model.ts`, `mappers/schema-registry.ts`, and the API-mode edit tests to preserve a `passwordSet` marker and allow blank-on-edit, then run `bun run test:file:unit src/components/pages/shadowlinks/create/model.test.ts src/components/pages/shadowlinks/edit/get-update-values-for-schema-registry.test.ts && bun run test:file:integration src/components/pages/shadowlinks/edit/shadowlink-edit-page.test.tsx`.

frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/source-connection-section.tsx
```diff
@@ -115,7 +141,20 @@ export const SourceConnectionSection = () => {
                   <InfoTooltip content="For Confluent Cloud, this is the API secret." />
                 </FormLabel>
                 <FormControl>
-                  <Input {...field} testId="sr-basic-password-input" type="password" />
+                  {isEmbedded() ? (
+                    <SecretSelector
```
malinskibeniamin, line 145: [P1 Major] Allow valid short Basic passwords in Cloud

What: This new embedded path routes password creation through `SecretSelector`, whose shared `NewSecretFormSchema` hard-codes `.min(20)`, while Schema Registry Basic auth and this feature's own form accept any non-empty password.

Why: A valid 1–19 character source Schema Registry password cannot be created through this UI, so the normal Cloud setup flow is blocked.

Suggested fix: Make the selector's secret-value validation configurable and pass a minimum length of 1 for this Basic-auth field.

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, add configurable secret-value validation to `components/ui/secret/secret-selector.tsx`, use non-empty validation here, cover a short password in the selector/source-connection tests, then run `bun run test:file:integration src/components/ui/secret/secret-selector.test.tsx src/components/pages/shadowlinks/create/configuration/schema-registry/source-connection-section.test.tsx`.

frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/sr-tls-configuration.tsx
```diff
@@ -174,13 +205,152 @@ const SrCaConfiguredBadge = () => {
...
+
+  return (
+    <FormField
+      control={control}
+      name="schemaRegistry.mtls.clientKey"
+      render={({ field }) => (
+        <FormItem>
+          <FormLabel>Client private key secret</FormLabel>
+          <FormControl>
+            <SecretSelector
```
malinskibeniamin, line 288: [P1 Major] Preserve the form label on secret selectors

What: `FormControl` supplies its generated `id`/ARIA props to this composite, but `SecretSelector` drops unknown props instead of forwarding them to its inner `SelectTrigger`; the new password selector has the same issue.

Why: On keyboard or screen-reader focus, the combobox is announced by its current value/placeholder rather than “Client private key secret” or “Password,” making the two secret controls ambiguous.

Suggested fix: Let `SecretSelector` accept and forward control/ARIA props to `SelectTrigger`, and assert both comboboxes by accessible name.

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, forward `id`, `aria-describedby`, `aria-invalid`, and labeling props through `components/ui/secret/secret-selector.tsx` to `SelectTrigger`, pass the two form labels here and in `source-connection-section.tsx`, then run the two Schema Registry integration test files.

frontend/src/components/pages/shadowlinks/create/schema-registry-request.ts
```diff
@@ -58,19 +63,33 @@ export const durationFromString = (value: string): Duration | undefined => {
...
+ * connection step's buildTLSSettings there is no file-path variant in the UI.
+ * The raw PEM private key goes into the request as-is; an empty key with a
+ * cert present means "keep the key stored server-side" (edit flow). File-path
+ * settings hydrated from an rpk-created link are round-tripped verbatim.
  */
 const buildSrTlsSettings = (sr: SchemaRegistryFormValues): TLSSettings => {
+  if (sr.mtls.filePaths) {
+    return create(TLSSettingsSchema, {
+      doNotSetSniHostname: sr.mtls.doNotSetSniHostname,
+      enabled: true,
```
malinskibeniamin, line 137: [P1 Major] Do not clear disabled TLS material during an edit

What: The server can return `enabled: false` together with file paths/PEMs; hydration retains that material, but the outer builder emits no `tlsSettings` whenever `useTls` is false, so any Schema Registry edit replaces the parent message without the stored paths.

Why: This silently deletes read-only rpk/Admin API TLS configuration despite the UI promising to preserve it as-is.

Suggested fix: Build `TLSSettings` whenever hydrated material exists, set `enabled` from `sr.useTls`, and omit the message only when TLS is off and there is truly no preserved configuration.

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, update `schema-registry-request.ts` to round-trip disabled file/PEM TLS with `enabled: false`, add a mapper-to-update regression test, then run `bun run test:file:unit src/components/pages/shadowlinks/create/schema-registry-request.test.ts src/components/pages/shadowlinks/edit/get-update-values-for-schema-registry.test.ts`.

frontend/src/components/pages/shadowlinks/mappers/schema-registry.ts
```diff
@@ -165,6 +173,124 @@ function mapSchemaRegistryApiOptions(api: SchemaRegistryApiOptionsMessage): Unif
...
+/**
+ * Format a proto Duration as the largest single-unit string the sync-behavior
+ * inputs accept ("90s", "5m") so it round-trips through durationFromString.
+ * Zero or unset durations map to '' so the cluster default applies.
+ */
+export function formatDurationForInput(duration: Duration | undefined): string {
+  if (!duration) {
+    return '';
+  }
+  const totalMs = Number(duration.seconds) * 1000 + Math.round(duration.nanos / 1_000_000);
```
malinskibeniamin: [P1 Major] Round-trip sub-millisecond durations exactly

What: This rounds protobuf nanos to whole milliseconds; for example, `500000ns` hydrates as `1ms` and rebuilds as `1000000ns`. rpk accepts Go durations below 1ms and the API stores full protobuf precision.

Why: Editing any other Schema Registry field applies the parent mask and silently changes an existing interval.

Suggested fix: Support `us`/`ns` in formatting, parsing, and validation, or retain the raw hydrated `Duration` until that input is actually changed.

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, make `formatDurationForInput` and `durationFromString` preserve protobuf nanos exactly, add a `500000ns` mapper-to-builder regression, then run `bun run test:file:unit src/components/pages/shadowlinks/create/model.test.ts src/components/pages/shadowlinks/create/schema-registry-request.test.ts src/components/pages/shadowlinks/mappers/schema-registry.test.ts`.

frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/source-connection-section.tsx
```diff
@@ -41,6 +56,17 @@ export const SourceConnectionSection = () => {
   const { control, setValue } = useFormContext<FormValues>();
   const authMethod = useWatch({ control, name: 'schemaRegistry.authMethod' });
 
+  // Fetch secrets for SecretSelector (embedded mode only)
+  const { data: secretsData } = useListSecretsQuery({}, { enabled: isEmbedded() });
```
malinskibeniamin, line 60: [P2 Minor] Distinguish secret loading failures from an empty store

What: Only `data` is consumed, so both a pending query and an auth/network failure become `availableSecrets = []`; `SecretSelector` then renders “No secrets available” and offers creation. The client-key picker repeats this behavior.

Why: Operators can be misled into creating duplicate secrets or cannot tell that retrying is the right recovery.

Suggested fix: Carry pending/error state into a shared secret-picker wrapper and show loading plus retry/error UI before the true empty state.

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, add shared pending/error handling for both Schema Registry secret queries and cover loading, failure, and empty states in the two component tests, then run those integration tests.

frontend/src/components/pages/shadowlinks/edit/shadowing-tab.tsx
```diff
@@ -9,18 +9,23 @@
...
+  schemaRegistryOriginalMode?: SchemaRegistryMode;
+}) => (
   <div className="space-y-4">
     <TopicsStep />
     <AclsStep />
     <ConsumerOffsetStep />
-    {/* Always the legacy switch: the edit request builder can't express the
-        version-gated API sync mode for now, only the create wizard can. */}
-    <LegacySchemaRegistrySection readOnlyApiMode={schemaRegistryApiMode} />
+    <SchemaRegistryEditSection originalMode={schemaRegistryOriginalMode} readOnlyApiMode={schemaRegistryApiMode} />
```
malinskibeniamin: [P3 Patch] Derive the read-only state from the original mode

What: `schemaRegistryApiMode` and `schemaRegistryOriginalMode` now describe the same server oneof through separate prop paths, so they can disagree even though `originalMode === API` fully determines the gate-off card.

Why: Parallel state adds hook/prop/test surface and risks showing an editable legacy switch while locking tabs as API mode.

Suggested fix: Derive `readOnlyApiMode` inside `SchemaRegistryEditSection` and remove the redundant boolean plumbing (and hook return if it has no remaining consumer).

One-shot prompt: In redpanda-data/console on `new-sr-sl-edit`, derive gate-off read-only behavior solely from `originalMode` across the edit components, remove redundant `isSchemaRegistryApiMode` plumbing, then run `bun run test:file:integration src/components/pages/shadowlinks/edit/schema-registry-edit-section.test.tsx src/components/pages/shadowlinks/edit/shadowlink-edit-page.test.tsx`.

malinskibeniamin (changes requested):
Blocking on the P1 findings below; P2/P3 notes are included inline as requested.

malinskibeniamin (approved):

frontend/src/components/pages/shadowlinks/mappers/schema-registry.ts
r-vasquez: >Suggested fix: Hydrate passwordSet into form state and require a password only for create/new Basic auth; allow blank when an existing password is configured.

passwordSet is an output-only field; it cannot be hydrated.

frontend/src/components/pages/shadowlinks/mappers/schema-registry.ts
r-vasquez: 👍 Switched the ms conversion to durationMs from @bufbuild/protobuf/wkt for the totalMs. 

The unit selection has to stay handmade as we have right now: the value must round-trip without any change for the input (10s/5m), `date-fns` provides "5 minutes" which is not a valid input for example, also checked `pretty-ms` is but it is multi unit, e.g: 90000ms will be handled as 1m 30s which is not a valid input either.

frontend/src/components/pages/shadowlinks/create/configuration/schema-registry/source-connection-section.tsx
r-vasquez: Right now this happens in every secret selector so a fix here would have to be a bigger change to not diverge only SR. I'll add this as a follow-up

r-vasquez:
@malinskibeniamin After a call with Product, also changed what you suggested:

image

PR #2576, opened 2026-07-23
Console - Full-screen page mode for SQL and RPCN editors

SpicyPete:
## Summary

- Footer pins to the viewport bottom on short pages.
- Topics and security-tab pages drop ListLayout's forced min-h-screen (min-h-0 override), removing large dead whitespace at bottom.
- Embedded Console cancels the Cloud UI host gutters with measured negative margins and owns its page gutter (px-12) — deploy-order-safe with cloud-ui removing its p-10 later.
- New expanded-page mode: data-page-expanded on <html> (utils/page-expanded)
  + useExpandedPageMode hook release every shell's horizontal constraints via global CSS while the page stays in document flow, footer below. The SQL studio's fixed-overlay fullscreen is replaced by this in-flow mode, and the RPCN pipeline editor gains the same toggle; both place the shared ExpandedPageToggle at the top-right of their work surface, clear of Save.
- /sql becomes a normal route; new breadcrumbOnlyHeader staticData flag keeps the app header breadcrumb-only for pages with their own title bar.

##### Footer always bottom

Before -> After
Screenshot 2026-07-23 at 2 21 26 PM
Screenshot 2026-07-23 at 2 21 45 PM

##### RPCN full screen
Screenshot 2026-07-23 at 2 22 30 PM
Screenshot 2026-07-23 at 2 22 43 PM
Screenshot 2026-07-23 at 2 23 00 PM


SpicyPete:
@claude review

frontend/src/hooks/use-expanded-page-mode.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Fullscreen ("expanded") mode for a work-surface page, in normal document flow:
+ * while expanded and on screen, stamps `data-page-expanded` on `<html>` so the
+ * shells release their horizontal constraints. Persists under `storageKey`.
+ *
+ * Attach `ref` to the page root — the attribute is held only while the root is
+ * visible (embedded Cloud UI keeps Console mounted but hidden on host routes).
+ */
+export function useExpandedPageMode({ storageKey }: { storageKey: string }): {
```
eblairmckee, line 40: I'm missing where we unset this when the page is unmounted. iirc there's a tanstack util that can help with this when you change routes

frontend/src/federation/federated-routes.tsx
```diff
@@ -85,25 +86,79 @@ function FederatedRootLayout() {
   );
 }
 
+/**
+ * Cancels the host gutters around the federated Console outlet with equal negative
+ * margins. Measured, not hardcoded, so either project can deploy first. Top padding
+ * is left alone — cancelling it would pull Console under the host's header.
+ */
```
eblairmckee, line 106: is there a way to do this is classname specificity on the `body` element rather than in a use layout effect? has a weird code smell to it..

SpicyPete: Not in this case, as the padding is coming from CloudV2, so this is to account for when CloudV2 removes that padding, we don't want to double-add it in the meantime while the deploys differ, especially when there may be older version of console running.

frontend/src/index.scss
malinskibeniamin, line 1: I would not extend the SCSS file unless we can help it

frontend/src/routes/__root.tsx
```diff
@@ -75,7 +74,9 @@ function SelfHostedLayout() {
       <AnnouncementBar />
       <SidebarLayout>
         <SidebarInset>
-          <div className="container mx-auto max-w-[1500px] px-12">
+          {/* Centered page column; the gutter and the width cap are released while a
+              page is expanded (index.scss). */}
+          <div className="page-expanded-flush page-expanded-uncap container mx-auto flex max-w-[1500px] flex-1 flex-col px-12 transition-[max-width,padding] duration-300 ease-in-out">
```
malinskibeniamin, line 79: Can we stick to tailwind?

SpicyPete: I don't think so in this case, or at least not in a way which is as understandable, as we have to add a class which targets attributes at the body level, which is owned by the CloudV2 federating app most of the time

frontend/src/index.scss
SpicyPete: We will have to modify a SCSS file, but I moved it to globals instead

frontend/src/routes/__root.tsx
SpicyPete: 1. The variant CSS isn't reliably in the loaded stylesheets. Expressed as an arbitrary variant on the element — [html[data-page-expanded]_&]:px-0 on #mainLayout — the generated rule was present when loading /sql and absent when loading /rp-connect/create: same element, same class string, full page loads both times. Plain px-12 applied everywhere; only the variant rule was chunk-dependent in the federated build. A @custom-variant doesn't help — it's still a JIT-generated utility subject to the same chunking. Hand-written CSS in an entry stylesheet is always loaded.

2. Two apps have to release in lockstep. The attribute is a contract with Cloud UI: Console drops its gutter and width cap while cloud-ui's wrapper drops expandableWidth. Only one side of that lives in this repo's className strings; the signal has to be a document-level attribute either way.

3. The state isn't the element's own. Whether the page is expanded is set by a hook on <html>, not by the component rendering #mainLayout — so a utility on that element would have to be conditionally applied by a component that doesn't own the state, which is what the release-hook classes avoid.

SpicyPete:
Failure seems to be related to a console-enterprise change, will fix it there if it isn't fixed by Tuesday

frontend/src/components/layout/footer.tsx
```diff
@@ -54,6 +54,7 @@ export const VersionInfo = () => {
   );
 };
 
+// Bottom placement is CSS: `.footer` has `margin-top: auto` inside #mainLayout's flex column.
```
eblairmckee, line 56: this is so useful while we still have scss styles using selectors here

frontend/src/components/layout/header.tsx
```diff
@@ -102,8 +97,16 @@ function AppPageHeader({ breadcrumbOnly = false }: { breadcrumbOnly?: boolean })
...
+  // is no sidebar trigger — so with the title row hidden this would be a bare divider
+  // above a page that already has its own title bar.
+  if (hideTitleRow && isEmbedded()) {
+    return null;
+  }
+
   return (
-    <div>
+    // Expanded pages release #mainLayout's gutter; keep the header off the viewport edge.
+    <div className="page-expanded-inset transition-[padding] duration-300 ease-in-out">
```
eblairmckee, line 109: surprised you don't have to use a template string for `[padding]` is that some new tailwind compiler thing that maps to a var?

eblairmckee (approved):
i think this is fine for now. albeit over engineered in some sense because we're having to deal with scss, tailwind, and interpolated styles.. but long term i would LOVE to just kill chakra from layouts altogether in cloud and console. 
all said, I do feel good seeing the LOC diff being nearly 0 

Image

PR #2588, opened 2026-08-03
feat: new topic messages page behind enableNewTopicMessagesPage flag

jvorcak:
Redesigned messages UX (from the Console Messages UX prototype):

- Read scope popover (newest/oldest/offset/timestamp) with live tail in the menu, continuous pagination for newest/oldest only, and an illustrated "how reading starts" doc sheet
- Filter bar: free-text input with recognized `field:value` words pill-highlighted (parsed continuously as you type, no commit step), word-scoped autocomplete/ghost-completion, and JS predicate filters as a separate chip
- Field filters, partition, quick search and selected message persist in the URL; JS filters stay in sessionStorage
- Message detail as docked resizable panel or expanded sheet, with persisted view state (mode, widths, section expansion) and a metadata table, headers grid and troubleshoot reports
- View settings sidebar: row density, drag-reorderable columns with per-column config (timestamp format, deserializers, preview fields)
- Quick info stats rebuilt on the registry Stat component

> **Filter bar was redesigned separately and is reviewable on its own** — it was reworked after the
> initial commit into the free-text/pill-highlight design described above, and doesn't touch the
> rest of the page's data flow. Core files: `toolbar/filter-bar.tsx`, `toolbar/filter-suggestions.ts`,
> `utils/filter-token.ts`, `utils/filter-line.ts` (each with a `.test.ts(x)`).
>
> Its UI pieces were also extracted into generic, domain-agnostic registry components rather than
> living inline in the filter bar: `Chip` (removable/editable pill), `HighlightedInput` (a real
> `<input>` with a purely visual, non-interactive highlight overlay), and `Listbox`/`ListboxOption`
> (a stateless dropdown shell for a caller-driven autocomplete list) — all under
> `src/components/redpanda-ui/components/`.

Fixes surfaced while building it:

- assignDeep now replaces arrays wholesale; the index-wise merge mutated shared elements through the uiSettings sync and corrupted reordered arrays (duplicate/lost message columns after drag'n'drop); getMessageColumns heals already-corrupted persisted entries
- SidebarInset gets min-w-0 so wide content (messages table) scrolls inside its container instead of widening the page past the viewport

Before:
Screenshot 2026-08-13 at 16 41 05

After:
Screenshot 2026-08-13 at 16 41 48
Screenshot 2026-08-13 at 16 41 40


jvorcak:
The filter bar is still under developments, other things can be reviewed

frontend/src/components/pages/topics/messages/table/messages-footer.tsx
```diff
@@ -0,0 +1,166 @@
...
+        {continuousMode ? (
+          <div className="flex items-center gap-3.5">
+            <span className="flex items-center gap-2 text-muted-foreground text-xs">
+              <span className="inline-block size-2 rounded-full bg-green-600" />
+              {windowSize} in buffer · window {windowCap}
+            </span>
+            {canLoadMore && (
+              <Button
+                isLoading={isLoadingMore}
+                onClick={onLoadMore}
```
malinskibeniamin, line 148: [P1] Implement the promised scroll-triggered pagination

Continuous mode is described as "pages load as you scroll", but `loadMore` is reachable only from this button; there is no scroll handler or `IntersectionObserver` in the messages page. A user who enables the mode still stops after the first batch unless they manually click **Load more**, so the headline behavior is missing. Add a bottom sentinel that invokes `onLoadMore` (guarded by `canLoadMore`/`isLoadingMore`), or stop presenting this as continuous pagination.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,541 @@
...
+              </Button>
+            </div>
+            <MessagesTable
+              columnConfig={columnConfig}
+              density={density}
+              hasActiveFilter={
+                urlState.quickSearch.trim().length > 0 || fieldTokens.length > 0 || filterInterpreterCode !== ''
+              }
+              isLiveWaiting={urlState.liveTail && search.messages.length === 0}
+              isLoading={isSearching && !urlState.liveTail}
```
malinskibeniamin, line 506: [P1] Surface stream failures instead of the empty state

`useMessageSearch` stores non-abort failures in `search.error`, but this page never consumes that field. `runStream` then sets the phase to `done`, making `MessagesTable` render **No messages** when the request failed before returning rows. Network, auth, and timeout failures therefore look like a valid empty topic. Render an error state with retry and suppress empty/waiting UI while `search.error` is set.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,541 @@
...
+    search
+      .start({ ...searchParams, startOffset: PartitionOffsetOrigin.End, pageSize: undefined }, { live: true })
+      .catch(() => {
+        // errors are surfaced through search.error
+      });
+    return () => search.stop();
+    // searchParams is intentionally not a dependency: scope edits are disabled while live,
+    // and deserializer changes take effect on the next (re)start.
+    // biome-ignore lint/correctness/useExhaustiveDependencies: see above
+  }, [urlState.liveTail, search.start, search.stop]);
```
malinskibeniamin: [P1] Apply or disable live-tail setting changes

This effect deliberately ignores `searchParams`, yet while live tail is active the partition/max-results controls and View Settings deserializers remain editable (the latter says changes apply instantly). Those edits update URL/store state while the running stream keeps the old request, so the UI and data disagree. Either restart the live stream when relevant request parameters change, or disable and clearly label these controls until live tail stops.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,541 @@
...
+  });
+
+  return (
+    <div className="flex flex-col gap-4" data-testid="topic-messages-view">
+      <MessagesToolbar
+        filterSlot={
+          <FilterBar
+            canUseJsFilters={canUseJsFilters}
+            fieldTokens={fieldTokens}
+            jsFilters={jsFilters.filter((f) => f.isActive)}
```
malinskibeniamin: [P1] Don't show JavaScript filters as active after disabling them

`continuousActive` makes `canUseJsFilters` false and `filterInterpreterCode` empty, so the next search is unfiltered, but active `jsFilters` are still passed here and rendered as chips. Users see an active predicate while results ignore it. When continuous mode or permissions disable JavaScript filtering, deactivate/remove the chips or render an explicit disabled state and exclude them from the active-filter summary.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,541 @@
...
+            urlState.setStartOffset(PartitionOffsetOrigin.Timestamp);
+            urlState.setPageIndex(0);
+          },
+          maxResults: urlState.maxResults,
+          onMaxResultsChange: (maxResults) => {
+            urlState.setMaxResults(maxResults);
+            urlState.setPageIndex(0);
+          },
+          continuousMode: urlState.continuousMode,
+          onContinuousModeChange: urlState.setContinuousMode,
```
malinskibeniamin: [P1] Reset pagination when switching to continuous mode

If the user is on page 2 or later, this handler preserves `pageIndex` while `pagination.pageSize` becomes the entire loaded window. TanStack then slices from `pageIndex * windowSize`, returns no rows, and the continuous footer removes page controls—an apparent empty topic with no way forward inside this mode. Set `pageIndex(0)` here; do the same for the client-filter callbacks above because `autoResetPageIndex` is intentionally disabled.

frontend/src/components/pages/topics/messages/toolbar/filter-bar.tsx
```diff
@@ -0,0 +1,425 @@
...
+            key={filter.id}
+            onEdit={() => onEditJsFilter(filter)}
+            onRemove={() => onRemoveJsFilter(filter.id)}
+            removeLabel="Remove filter"
+            title={filter.name ? `${filter.name}: ${filter.code}` : 'Edit JavaScript filter'}
+          >
+            ƒ {filter.name || filter.code}
+          </Chip>
+        ))}
+        <HighlightedInput
```
malinskibeniamin, line 411: [P1] Add combobox/listbox semantics to this autocomplete

`HighlightedInput` is a plain input and the popup is a `div` of buttons: there is no `role="combobox"`, `aria-expanded`, `aria-controls`, `aria-activedescendant`, `role="listbox"`/`option`, or `aria-selected`. Screen readers cannot discover or track suggestions even though keyboard navigation exists. Add stable IDs and ARIA state, or use the registry Combobox while preserving the current behavior.

frontend/src/components/pages/topics/messages/toolbar/filter-bar.tsx
```diff
@@ -0,0 +1,425 @@
...
+            }
+            itemIdx += 1;
+            const isActive = itemIdx === activeIdx;
+            return (
+              <ListboxOption
+                active={isActive}
+                key={`${item.label}-${i}`}
+                onClick={() => applySuggestion(item.action)}
+                onMouseEnter={() => {
+                  setActiveIdx(itemIdx);
```
malinskibeniamin: [P2] Capture each option's index before creating this handler

`itemIdx` is a mutable render-scope variable. Every `onMouseEnter` closure reads its final value after `map` completes, so hovering any suggestion marks the last option active; pressing Enter then applies the wrong suggestion. Assign `const optionIndex = ++itemIdx` and close over that value.

frontend/src/components/pages/topics/messages/utils/filter-line.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  for (const word of words) {
+    const parsed = parseFilterInput(word.text);
+    if (!parsed) {
+      remainderParts.push(word.text);
+      continue;
+    }
+    tokenRanges.push({ start: word.start, end: word.end });
+    if (parsed.field === 'partition') {
+      partitionId = Number(parsed.value);
```
malinskibeniamin, line 119: [P2] Validate the partition token before updating request state

Every recognized partition operator/value is coerced with `Number`: `partition>2` silently becomes exact partition 2, `partition:abc` becomes `NaN`, and negatives/fractions pass through. This can serialize invalid URL/request state or query an unintended partition. Accept only equality with a nonnegative safe integer (and range-check against the partition count where available); reject or visibly retain other input.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,541 @@
...
+  useEffect(() => {
+    appGlobal.searchMessagesFunc = () => setRefreshCounter((c) => c + 1);
+    return () => {
+      appGlobal.searchMessagesFunc = undefined;
+    };
+  }, []);
+
+  // Keyboard nav follows the on-screen order: current sort, then current page
+  const visibleKeys = useMemo(() => {
+    const sorted = [...windowedMessages].sort((a, b) => {
```
malinskibeniamin: [P2] Derive keyboard order from the actual table order in continuous mode

Continuous mode disables table timestamp sorting to preserve server order, but `visibleKeys` always reapplies URL sorting. With **Oldest + continuous** and the default timestamp-desc sort, `j`/arrow keys therefore select rows in a different order from the displayed rows. Use `windowedMessages` order while sorting is disabled, or derive these IDs from the table row model.

malinskibeniamin (commented):
Inline findings: P1/P2. No P0/P3 identified.

frontend/.gitignore
```diff
@@ -8,6 +8,7 @@ public/static
 
 # misc
 .DS_Store
+.playwright-mcp/
```
malinskibeniamin: nit: would suggest using playwright CLI instead of MCP - more token efficient.

frontend/src/components/pages/topics/messages/detail/message-detail-panel.tsx
```diff
@@ -0,0 +1,230 @@
...
+    compression: msg.compression,
+    transactional: msg.isTransactional,
+  };
+  const blob = new Blob([toJson(record, 2)], { type: 'application/json' });
+  const url = URL.createObjectURL(blob);
+  const link = document.createElement('a');
+  link.href = url;
+  link.download = `record-p${msg.partitionID}-o${msg.offset}.json`;
+  link.click();
+  URL.revokeObjectURL(url);
```
malinskibeniamin: **Priority: P1**

**"Download Record" can silently do nothing.** The anchor is never inserted into the document and `URL.revokeObjectURL(url)` runs synchronously on the next line. This repo's existing export path does the opposite deliberately — `frontend/src/components/pages/topics/Tab.Messages/dialogs/save-messages-dialog.tsx:176` and `:184` both carry `document.body.appendChild(link); // required in firefox` — and `debug-bundle-link.tsx` appends too. Firefox ignores `click()` on a detached anchor, and revoking the object URL in the same tick races the download start in other engines as well.

Consequence: the primary action of the new detail panel produces no file and no error feedback for at least one supported browser.

Correction: mirror the existing helper — `document.body.appendChild(link); link.click(); link.remove();` and revoke on a later tick (`setTimeout(() => URL.revokeObjectURL(url), 0)`) rather than immediately.

Verify: open the detail panel in Firefox and Chromium, click Download Record, confirm `record-p{partition}-o{offset}.json` lands in Downloads with the expected body.

*Automated `/review`.*

frontend/src/components/pages/topics/messages/utils/message-order.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import type { SortingState } from '@tanstack/react-table';
+
+import { messageKey } from './message-key';
+import type { TopicMessage } from '../../../../../state/rest-interfaces';
+
+const compareBySorting = (a: TopicMessage, b: TopicMessage, sorting: SortingState): number => {
+  for (const sort of sorting) {
+    const direction = sort.desc ? -1 : 1;
+    const left = sort.id === 'timestamp' ? a.timestamp : a.offset;
```
malinskibeniamin: **Priority: P1**

**Keyboard row order can differ from the rendered row order.** `compareBySorting` collapses every non-`timestamp` sort id to `offset`, but the table it is supposed to mirror handles those ids differently (`table/messages-table.tsx`): `key` is registered with `accessorKey: 'keyJson'`, `value` with `'valueJson'`, and every column except `timestamp` has `enableSorting: false`.

The persisted sort comes from the shared store, and the legacy page allowed and persisted sorting on `key` and `value` (`Tab.Messages/index.tsx:1059`, `:1087` set `enableSorting` for both). For a returning user whose stored sorting is `[{id:'key'}]`, the table renders either keyJson order or unsorted stream order, while `visiblePageKeys` hands `useKeyboardNav` an offset-ordered key list — so `j`/`k`/arrow keys select a row other than the visually adjacent one, and `c` then copies the wrong record's value. Separately, `DEFAULT_SORTING` is `[timestamp desc, offset desc]`, yet the new table registers `offset` with `enableSorting: false`; if table-core drops non-sortable entries from the sorting state (it filters by `getCanSort()`), the offset tiebreaker applies here but not in the table, reproducing the same divergence on default settings whenever timestamps tie — which is routine for a produce burst.

Correction: derive the order from one source of truth instead of a second comparator — either read the rendered order off the table instance (`table.getRowModel().rows`) and lift it to the page, or normalize/filter the sorting state to exactly the ids the table marks sortable before comparing, and mirror each sortable id's accessor (`keyJson`/`valueJson`) rather than defaulting to `offset`.

Verify: extend `utils/message-order.test.ts` with a persisted `[{id:'key',desc:false}]` state and with two rows sharing a timestamp, asserting the returned keys equal the table's rendered row ids.

*Automated `/review`.*

frontend/src/components/pages/topics/messages/view-settings/column-list.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  >
+                    <div className="flex items-center gap-2 px-1.5 py-1.5">
+                      <span
+                        className="cursor-grab text-muted-foreground/60"
+                        {...dragProvided.dragHandleProps}
+                        title="Drag to reorder"
+                      >
+                        <GripVerticalIcon className="size-4" />
+                      </span>
+                      <Checkbox
```
malinskibeniamin, line 74: **Priority: P1**

**Column-visibility checkboxes have no accessible name.** `<Checkbox checked onCheckedChange testId>` gets no `aria-label`, no `id`/`htmlFor` pairing, and no `aria-labelledby`. The visible column name sits in a sibling `<span className="flex-1 text-[13.5px]">` (line 79), not a `<label>`, so nothing associates the two. The registry `Checkbox` (`components/redpanda-ui/components/checkbox.tsx:49`) renders a Base UI button-role control and does not synthesize a name from context.

Consequence: a screen-reader user hears seven identical unnamed checkboxes and cannot tell which column each one shows or hides — a WCAG 4.1.2 (name, role, value) failure on the only control for column visibility. The same pattern makes the drag-handle `<span>` (line 66) reachable only by pointer, though the checkbox naming is the blocking part.

Correction: give the label an id and wire it up — `<span id={`col-label-${column.id}`}>{COLUMN_LABELS[column.id]}</span>` with `aria-labelledby={`col-label-${column.id}`}` on the `Checkbox` — or pass `aria-label={`Show ${COLUMN_LABELS[column.id]} column`}`.

Verify: in `view-settings-panel.test.tsx`, assert `getByRole('checkbox', { name: /Timestamp/i })` resolves for each configured column.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 3 finding(s).

frontend/src/components/pages/topics/messages/detail/message-detail-panel.tsx
jvorcak: Fixed — appended the anchor to the DOM before `click()` (required in Firefox) and moved `revokeObjectURL` to a later tick instead of the same one, mirroring `save-messages-dialog.tsx`'s pattern. In a900a89e0.

frontend/src/components/pages/topics/messages/utils/message-order.ts
jvorcak: Fixed — introduced `isSortableColumnId` in `message-order.ts` as the single source of truth for which column the table actually sorts on (only `timestamp`, and only when not disabled), used by both the table's `enableSorting` and `visiblePageKeys`. A persisted `key`/`value` sort from the legacy page (or anything else) is now ignored exactly like tanstack's own `getSortedRowModel` ignores it, instead of silently falling back to an offset tiebreak the table itself never applies. Added regression tests for a persisted non-sortable sort id and for tied timestamps. In a900a89e0.

frontend/src/components/pages/topics/messages/view-settings/column-list.tsx
jvorcak: Fixed — wired each checkbox to its label via `aria-labelledby` pointing at an `id` on the label `<span>`. Added the suggested test asserting `getByRole('checkbox', { name })` resolves per column. In a900a89e0.

malinskibeniamin (commented):
Automated `/review`: 2 finding(s).

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
```diff
@@ -0,0 +1,365 @@
...
+      } finally {
+        if (abortControllerRef.current === abortController) {
+          abortControllerRef.current = null;
+        }
+        if (flushTimerRef.current) {
+          clearTimeout(flushTimerRef.current);
+        }
+        flush();
+        setIsLoadingMore(false);
+        setPhase('done');
```
malinskibeniamin: **Priority: P1**

**A superseded stream finalizes the state of the stream that replaced it.**

`runStream` starts by calling `stop()`, which aborts the previous controller, then installs its own. But the previous run's `finally` block still executes unconditionally — after the new run has already set `phase='connecting'`, cleared `bufferRef`, and reset stats. The identity check just above guards only `abortControllerRef`; every state write below it is ungated:

- `flush()` publishes whatever is in the now-shared `bufferRef`
- `setIsLoadingMore(false)`
- `setPhase('done')` ← this line
- `setBackendPhase(null)`

The abort rejection surfaces on a later microtask than the synchronous setup of the new run, so the stale `finally` reliably lands *after* it. Reachable on ordinary flows in `topic-messages-view.tsx`: toggling live tail (the live effect's `start` aborts the paged stream, whose `finally` then reports `done` while live tail is still connecting), changing partition/deserializer/read scope within the 100ms debounce window, and clicking "Load more" while an auto-search restart is in flight.

Consequence: the toolbar and table stop showing progress for a request that is still running — `isSearching` goes false, the backend phase label disappears, and a continuous-mode load can lose its loading state. With `flush()` publishing a buffer the new run just emptied, the table can also briefly render as "done, no messages" for a live search.

Correction: capture a generation counter (or compare `abortControllerRef.current === abortController`) and gate *all* of the `finally` writes plus every `setPhase`/`setStats`/`setBackendPhase` inside the `for await` loop on it, so only the current run may write state. Add a regression test that starts search A, starts search B before A settles, and asserts `phase` never becomes `'done'` from A.

Verify: `cd frontend && bun vitest run src/components/pages/topics/messages/hooks/use-message-search.test.tsx`

*Automated `/review`.*

frontend/src/components/pages/topics/messages/table/messages-table.tsx
```diff
@@ -0,0 +1,319 @@
...
+                const canSort = header.column.getCanSort();
+                const sortDir = header.column.getIsSorted();
+                return (
+                  <TableHead
+                    aria-sort={sortDir ? (sortDir === 'asc' ? 'ascending' : 'descending') : undefined}
+                    // The table-corner action buttons (save / view settings) float over the
+                    // header's top-right — keep the last column's title clear of them.
+                    className={cn(canSort && 'cursor-pointer select-none', 'last:pr-20')}
+                    key={header.id}
+                    onClick={canSort ? header.column.getToggleSortingHandler() : undefined}
```
malinskibeniamin: **Priority: P1**

**Column sorting is mouse-only — no keyboard path exists.**

The sort affordance is an `onClick` on `TableHead` (a plain `<th>`) with `cursor-pointer select-none` styling. There is no `role="button"`/`<button>` wrapper, no `tabIndex`, and no `onKeyDown`, so the header is not focusable and Enter/Space do nothing. `aria-sort` correctly announces the *current* state, which makes the omission worse: a screen-reader or keyboard-only user is told the table is sortable but has no way to sort it. The page's own `useKeyboardNav` covers row movement and copy, not sorting, so no alternative route exists.

This also conflicts with the repo standard for click handlers on non-interactive elements (role + tabIndex + keyboard handler).

Consequence: WCAG 2.1.1 (keyboard) failure on a primary table interaction of a new customer-facing page.

Correction: render the sortable header label as a real `<button type="button">` inside the `<th>` (keeping `aria-sort` on the `th`), move `onClick` onto it, and give it `focus-visible:ring-*`; keep the plain `<th>` for non-sortable columns. Add a test that tabs to the Timestamp header and presses Enter, asserting `onSortingChange` fired.

Verify: `cd frontend && bun vitest run src/components/pages/topics/messages/table/messages-table.test.tsx`

*Automated `/review`.*

SpicyPete:
Left some feedback with some issues via Slack.  Let me know when those are resolved and I can check this branch again.
I would also recommend running an agent on the branch with the goal of reducing comment length, and removing non-critical comments.    Overall these pages are looking much better now

frontend/src/components/pages/topics/messages/table/messages-table.tsx
jvorcak: I think this is more like a component library issue?

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
jvorcak: It should be fixed now

jvorcak:
@SpicyPete your feedback should be addressed now

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,623 @@
...
+    let cancelled = false;
+    // The backend bounds a "from newest" consume to maxResults per partition and then
+    // completes (list_messages.go) — it isn't an actually-endless stream. Restart
+    // transparently on every natural completion (append, so the table doesn't reset)
+    // so "Live tail" keeps tailing for as long as the toggle is on, instead of quietly
+    // going stale after ~maxResults records while the toolbar still claims it's streaming.
+    const startLive = (options: { live: true; append?: boolean }) =>
+      search
+        .start({ ...searchParams, startOffset: PartitionOffsetOrigin.End, pageSize: undefined }, options)
+        .then(() => {
```
malinskibeniamin: **Priority: P1**

**Live tail retries hard failures in an unbounded loop, and hides the error while doing it.**

**Evidence:** this `.then` restarts the tail after every resolution of `search.start()`. But `runStream` never rejects on a real failure — `hooks/use-message-search.ts:246-249` catches the stream error and only calls `setError(...)`, and a backend `error` control frame (line 1725-1727 of the diff) merely raises a toast. So `start()` resolves identically for "stream finished normally" and "stream failed outright", and the `.catch` below — commented "don't loop on a hard failure" — is unreachable for exactly the case it claims to guard. Each restart additionally runs `setError(null)` at the top of `runStream`, so the failure is wiped from the UI on every iteration.

**Consequence:** with live tail on and a persistently failing consume (403/permission, network down, broken deserializer, backend `error` frame), the page issues `listMessages` back-to-back at RTT speed for as long as the toggle stays on, while the error state flickers and the toolbar keeps claiming it is streaming. That is a client-driven request storm against the cluster with no visible cause.

**Correction:** make the outcome observable and gate the restart on success — e.g. have `runStream`/`start` resolve to a discriminated result (or rethrow non-abort failures) and only recurse on a clean completion, with a bounded backoff (and a retry ceiling) before re-arming.

**Verify:** `npx vitest run frontend/src/components/pages/topics/messages/hooks/use-message-search.test.tsx` plus a new case in `topic-messages-view` that makes the transport fail and asserts the tail issues one retry after a delay rather than N immediate ones.

*Automated `/review`.*

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,623 @@
...
+    urlState.pageSize,
+    urlState.pageIndex,
+    urlState.setPageIndex,
+  ]);
+
+  // In continuous or live mode only the newest DISPLAY_WINDOW_CAP rows stay rendered
+  const { rows: windowedMessages, trimmed } = useMemo(
+    () =>
+      continuousActive || urlState.liveTail
+        ? applyDisplayWindow(filteredMessages, DISPLAY_WINDOW_CAP)
```
malinskibeniamin: **Priority: P1**

**Continuous + "Newest" trims the newest rows, not the oldest.**

**Evidence:** `applyDisplayWindow` (`utils/live-window.ts`) trims the *front* of the array, documented as "rows arrive oldest→newest, so the front is trimmed". That assumption only holds for the non-paged path. Continuous mode sets `pageSize: continuousActive ? urlState.maxResults : undefined` (line 161), and on the backend `PageSize > 0` routes through `resolvePagedConsumeRequests` → `resolvePageToken`, which picks `directionDescending` for every start offset except `StartOffsetOldest` (`backend/pkg/console/list_messages.go:314-317`); `fetchMessages` then collects and reverses per partition (same file, 888-929). So in continuous + "Newest" the buffer accumulates newest→oldest — the front is the newest page — and `orderForContinuousNewest` right below only re-sorts whatever survived the trim. Continuous + "Oldest" (ascending) and live tail (`pageSize: undefined`, legacy ascending path) are unaffected.

**Consequence:** with `DISPLAY_WINDOW_CAP = 150` and the default `maxResults` of 50, roughly three "Load more"s — which the footer's IntersectionObserver sentinel triggers on scroll without any click — silently delete the most recent records from the table, keeping the oldest instead, and the footer reports them as "{n} older trimmed". Users watching the newest end of a topic lose exactly the rows they are there for, with the UI asserting the opposite.

**Correction:** window after ordering rather than before (apply `applyDisplayWindow` to `orderedMessages`), or make the trim direction explicit — pass the stream's arrival direction in and drop from the tail when the scope streams descending. Fix the footer's "older trimmed" wording to follow the same source of truth.

**Verify:** extend `utils/live-window.test.ts` with a descending-arrival case, and add a `topic-messages-view` test that loads 200 rows in continuous + "Newest" via `createRouterTransport` and asserts the first rendered row is the highest-offset record: `npx vitest run frontend/src/components/pages/topics/messages`.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 2 finding(s).

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,626 @@
...
+
+  const continuousNewest = continuousActive && urlState.readScopeMode === 'newest';
+
+  // In continuous or live mode only the newest DISPLAY_WINDOW_CAP rows stay rendered. Once
+  // ordered above, continuous+"Newest" holds its newest row first, so its window trims the
+  // tail instead of the front that every other (oldest→newest) scope trims.
+  const { rows: windowedMessages, trimmed } = useMemo(
+    () =>
+      continuousActive || urlState.liveTail
+        ? applyDisplayWindow(orderedMessages, DISPLAY_WINDOW_CAP, { newestFirst: continuousNewest })
```
malinskibeniamin, line 287: [P1 Major] Continuous “Newest” cannot advance past 150 rows

What: After three default 50-row pages, this window keeps the first/newest 150 rows. `loadMore` appends the next older page, but ordering descending and then taking `slice(0, 150)` discards every row just fetched. A four-page probe leaves offsets 200–51 visible both before and after offsets 50–1 arrive.

Why: Loading beyond the cap changes only the footer counters; users cannot inspect records older than the initial window, so continuous pagination is functionally stuck.

Suggested fix: Make the cap a sliding/virtualized viewport: show the newly loaded older page, preserve the scroll anchor, and keep the trimmed side reachable when scrolling back. Alternatively, stop offering `loadMore` once fetched rows cannot be displayed.

Verify: Load four 50-row pages in continuous + Newest; assert page-four rows become visible and the newest rows remain reachable.

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
```diff
@@ -0,0 +1,391 @@
...
+              toast.error('Backend error', { description: controlMessage.value.message });
+              streamError = new Error(controlMessage.value.message);
+              setError(streamError);
+              break;
+            case 'data': {
+              const message = convertListMessageData(controlMessage.value);
+              if (isLongLived) {
+                appendWithSlackCap(bufferRef.current, message, LIVE_BUFFER_MAX, LIVE_BUFFER_SLACK);
+              } else {
+                bufferRef.current.push(message);
```
malinskibeniamin, line 248: [P1 Major] Continuous pagination retains an unbounded message buffer

What: Paged Newest/Oldest requests are not `isLongLived`, so every `loadMore` appends here without a cap. The 150-row limit is only applied later while rendering; each flush still clones the entire buffer, and client filters/path hints scan it.

Why: Scrolling a normal large topic retains all decoded and raw payloads and performs increasingly large copies/scans. The page can become progressively slower or exhaust browser memory despite claiming a bounded window.

Suggested fix: Bound the continuous-mode backing buffer/page cache along with the rendered window, while keeping cumulative counts and the next-page token separately; or use a virtualized, explicitly bounded cache.

Verify: Append many continuous pages and assert retained hook state stays bounded while forward/back navigation still works; include a heap/profile check with representative payloads.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,626 @@
...
+              error={search.error?.message ?? null}
+              hasActiveFilter={
+                urlState.quickSearch.trim().length > 0 || fieldTokens.length > 0 || filterInterpreterCode !== ''
+              }
+              isLiveWaiting={urlState.liveTail && search.messages.length === 0}
+              isLoading={isSearching && !urlState.liveTail}
+              messages={tableData}
+              newKeys={search.newKeys}
+              onPaginationChange={handlePaginationChange}
+              onRetry={handleRefresh}
```
malinskibeniamin, line 510: [P1 Major] Live-tail Retry is a no-op after retries exhaust

What: After the loop gives up on the sixth failed start, `liveTail` remains true and `search.error` renders the Retry action. That action calls `handleRefresh`, which immediately returns in live mode; neither `active` nor `start` changes, so the loop cannot restart. The toolbar also keeps spinning and says “Streaming live…”.

Why: The visible recovery path does nothing after a transient/persistent outage, while the page falsely reports an inactive consumer as live.

Suggested fix: Expose exhausted/restart state from the live loop. Make Retry reset/restart the loop (or disable live and re-enable it), and derive the toolbar status from the actual loop phase rather than the URL flag.

Verify: Reject through `maxRetries`, advance timers, click Retry, and assert a new request starts and the status reflects the stopped/retrying state.

frontend/src/components/pages/topics/messages/hooks/use-keyboard-nav.ts
malinskibeniamin, line 1: This seems like a potential registry hook

frontend/src/globals.css
```diff
@@ -113,3 +113,17 @@ button[aria-label="Open TanStack Router Devtools"]:hover {
...
+  0% {
+    background-color: var(--color-green-100, #d1fadf);
+  }
+  100% {
+    background-color: transparent;
+  }
+}
+.animate-message-flash {
+  animation: message-flash 2.6s ease-out;
+}
```
malinskibeniamin, line 129: is there any way we can just use tailwind animations in that particular instance and avoid updating the globals.css file by hand

frontend/rsbuild.config.ts
```diff
@@ -174,9 +174,12 @@ export default defineConfig({
...
       config.output.publicPath = 'auto';
 
-      // Prevent rebuild loop by ignoring generated route tree file
+      // Prevent rebuild loop by ignoring generated route tree file and Playwright MCP's
+      // scratch output (screenshots/console logs it writes into the project root while
+      // driving the dev server) — rspack's incremental build panics on rapid rebuilds
+      // triggered by files that never touch the actual module graph.
       config.watchOptions = {
-        ignored: ['**/routeTree.gen.ts'],
+        ignored: ['**/routeTree.gen.ts', '**/.playwright-mcp/**'],
```
malinskibeniamin, line 182: nit: strongly recommend using playwright CLI over MCP, it's official guidance to save money on tokens

frontend/src/stores/topic-settings-store.ts
```diff
@@ -102,6 +128,13 @@ export type TopicSettingsStore = {
   setPreviewDisplayMode: (topicName: string, mode: 'single' | 'wrap' | 'rows') => void;
   getPreviewDisplayMode: (topicName: string) => 'single' | 'wrap' | 'rows';
 
+  // Actions for the messages-table view settings (new messages UX)
+  setRowDensity: (topicName: string, density: RowDensity) => void;
+  getRowDensity: (topicName: string) => RowDensity;
+  setMessageColumns: (topicName: string, columns: MessageColumnConfig[]) => void;
+  getMessageColumns: (topicName: string) => MessageColumnConfig[];
+  resetViewSettings: (topicName: string) => void;
```
malinskibeniamin, line 136: how many of these actions actually could have been handled inside the tanstack table internal store vs needing global zustand setup

frontend/src/components/redpanda-ui/components/chip.tsx
```diff
@@ -0,0 +1,79 @@
+import { XIcon } from 'lucide-react';
```
SpicyPete, line 1: How does a `Chip` differ from our existing `Badge` components.
https://redpanda-ui-registry.netlify.app/docs/badge

Also general comment on adding components, I think we should reserve the `redpanda-ui` folder for registry components, or components will plan to upstream in the near future.

frontend/src/components/redpanda-ui/components/highlighted-input.tsx
```diff
@@ -0,0 +1,88 @@
...
+          aria-hidden="true"
+          className={cn('pointer-events-none absolute inset-0 flex items-center whitespace-pre', textClassName)}
+          data-testid={overlayTestId}
+        >
+          {segments.map((seg, i) => (
+            <span className={cn('text-transparent', seg.highlighted && highlightClassName)} key={`${i}-${seg.text}`}>
+              {seg.text}
+            </span>
+          ))}
+          {ghostText ? <span className="text-muted-foreground/60">{ghostText}</span> : null}
```
SpicyPete, line 72: With the updated theme colours, we want to avoid `color/opacity` as much as possible.  
Still a valuable pattern, but with the updates we should see if we still need them added in places like this.

frontend/src/components/ui/date-time-input.tsx
```diff
@@ -159,21 +162,32 @@ const DateTimePickerPanel = ({
...
-          disabled={disabled}
-          onChange={(e) => setFromTime(e.target.value)}
-          step={1}
-          type="time"
-          value={timeInputValue}
-        />
-        <Button disabled={disabled} onClick={() => onChange(Date.now())} size="sm" type="button" variant="primary">
-          Now
-        </Button>
+      {/* Scoped override: the registry's `--color-selected` token (indigo-800) reads as a
```
SpicyPete, line 165: Many of these comments are outdated now with the updated theme.
Suggest we run a cleanup agent to remove comments which aren't too valuable after this branch merges, and be updated to be up to date.

frontend/src/components/ui/date-time-input.tsx
```diff
@@ -159,21 +162,32 @@ const DateTimePickerPanel = ({
...
+      disconnected dark navy next to this panel's other indigo-600 selection affordances —
+      repaint the selected day with the same shade instead of touching the shared Calendar
+      component. `p-0` strips Calendar's own built-in padding, which otherwise stacks on top
+      of this panel's own spacing and insets the grid from its siblings. */}
+      <div className="[--color-selected-foreground:var(--color-primary-foreground)] [--color-selected:var(--color-primary)]">
```
SpicyPete, line 170: Let's double check if we need this still after the registry updates.

frontend/src/components/pages/topics/messages/view-settings/view-settings-panel.tsx
```diff
@@ -0,0 +1,254 @@
...
+            <p className="text-muted-foreground text-xs leading-relaxed">
+              How key bytes are decoded for display. <span className="font-mono">Automatic</span> detects the format.
+            </p>
+          </div>
+        );
+      case 'value':
+        return (
+          <div className="flex flex-col gap-3.5">
+            <div className="flex flex-col gap-1.5">
+              <Label className="text-[10px] uppercase tracking-wide">Deserializer</Label>
```
SpicyPete, line 166: There are now text utilities to cover this font size and styling, perhaps lets do a pass on aiming to remove inline utility classes.

frontend/src/components/pages/topics/messages/view-settings/view-settings-panel.tsx
```diff
@@ -0,0 +1,254 @@
...
+      default:
+        return null;
+    }
+  };
+
+  return (
+    <div className="flex h-full min-h-0 flex-col bg-card" data-testid="view-settings-panel">
+      <div className="flex shrink-0 items-center gap-2 border-b px-4 py-3">
+        <div className="min-w-0 flex-1">
+          <div className="font-semibold text-[15px]">View settings</div>
```
SpicyPete, line 186: Same here for example, lets try aim to rely on our theme sizes as much as possible

frontend/src/components/pages/topics/messages/view-settings/view-settings-panel.tsx
```diff
@@ -0,0 +1,254 @@
...
+          <div className="text-muted-foreground text-xs">Changes apply to the table instantly</div>
+        </div>
+        <Button className="size-7" onClick={onClose} size="icon" testId="view-settings-close" variant="ghost">
```
SpicyPete, line 189: Do we need a larger button size?  if so we should upstream that to the Button itself as a fast follow

frontend/src/components/pages/topics/messages/view-settings/preview-fields-editor.tsx
```diff
@@ -0,0 +1,182 @@
...
+    setPreviewTags(
+      topicName,
+      tags.map((t) => (t.id === id ? { ...t, ...patch } : t))
+    );
+  };
+
+  return (
+    <div className="flex flex-col gap-3">
+      <div>
+        <div className="mb-1 font-semibold text-[10px] text-muted-foreground uppercase tracking-wide">
```
SpicyPete, line 85: Another spot we can likely use one of our text utilities for consistency.  
Won't comment on more in this PR, but a pass should catch most/all of these.

frontend/src/components/pages/topics/messages/toolbar/read-scope-popover.tsx
```diff
@@ -0,0 +1,332 @@
...
+};
+
+const ModeRow = ({ mode, selected, onSelect }: { mode: ReadScopeMode; selected: boolean; onSelect: () => void }) => {
+  const meta = READ_SCOPE_META[mode];
+  const Icon = meta.icon;
+  return (
+    <button
+      className={cn(
+        'flex w-full items-start gap-3 rounded-md px-3 py-2.5 text-left hover:bg-accent',
+        selected && 'bg-accent/60'
```
SpicyPete, line 85: With the new set of wash colours, we should avoid opacity mixins unless really needed

frontend/src/components/pages/topics/messages/toolbar/messages-toolbar.tsx
```diff
@@ -0,0 +1,54 @@
...
+  actionsSlot,
+  isRefreshing,
+  isLive,
+  onRefresh,
+}: MessagesToolbarProps) => (
+  <div className="flex items-center gap-2.5">
+    <ReadScopePopover {...scopeProps} />
+    <div className="min-w-0 flex-1">{filterSlot}</div>
+    {actionsSlot}
+    <Button
```
SpicyPete, line 43: https://ui-playground.netlify.app/button
Does the Lg button work well?  we want to avoid custom base level components when possible for consistency

SpicyPete (commented):
Handful of comments, mostly related to using the updated registry theme.
I think with a once-over to update to the new colours, use base components a bit more, and use the new set of text-utility components will resolve most/all of these.

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { AlertTriangleIcon, ChevronRightIcon } from 'lucide-react';
+import type { ReactNode } from 'react';
+
+import type { Payload, TopicMessage } from '../../../../../state/rest-interfaces';
+import { TimestampDisplay } from '../../../../../utils/tsx-utils';
+import { prettyBytes } from '../../../../../utils/utils';
+import { PayloadComponent } from '../../Tab.Messages/message-display/payload-component';
+
+/** Collapsible section with an uppercase label, optional right-side meta and copy action.
+ * Open state is controlled — it lives in the persisted detail view state. */
```
eblairmckee, line 25: this looks almost identical to an [Accordion](https://ui-playground.netlify.app/accordion)

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
```diff
@@ -0,0 +1,235 @@
...
+    return null;
+  }
+  return typeof value.payload === 'object' ? JSON.stringify(value.payload) : String(value.payload);
+};
+
+/** Compact key/value grid (the design mock's header list — no table chrome or pagination). */
+const HeaderGrid = ({ headers }: { headers: TopicMessage['headers'] }) => (
+  <div className="overflow-hidden rounded-md border bg-card">
+    <div className="grid grid-cols-[minmax(110px,42%)_1fr] border-b">
+      <div className="px-2.5 py-1 font-semibold text-caption text-muted-foreground uppercase tracking-wide">Key</div>
```
eblairmckee, line 166: can we use semantic html elements for text (like `label`)

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
```diff
@@ -0,0 +1,235 @@
...
+      </div>
+      <div className="flex flex-col gap-1">
+        {report.map((entry) => (
+          <div className="break-words font-mono text-body-sm leading-relaxed" key={entry.serdeName}>
+            <span className="font-semibold capitalize">{entry.serdeName}:</span> {entry.message}
+          </div>
+        ))}
+      </div>
+    </div>
+  );
```
eblairmckee, line 129: can we use [Alert](https://ui-playground.netlify.app/alert) instead

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
```diff
@@ -0,0 +1,235 @@
...
+      const text = headerValueText(header.value);
+      return (
+        <div
+          className="grid grid-cols-[minmax(110px,42%)_1fr] border-b font-mono text-body-sm last:border-b-0"
+          // biome-ignore lint/suspicious/noArrayIndexKey: header keys can repeat; order is stable
+          key={`${header.key}-${i}`}
+        >
+          <div className="min-w-0 break-all px-2.5 py-1.5 text-primary">{header.key}</div>
+          <div className="min-w-0 break-all border-l px-2.5 py-1.5">
+            {text === null ? <span className="text-muted-foreground italic">null</span> : text}
```
eblairmckee, line 181: I'd rather render nothing than `null`

frontend/src/components/pages/topics/messages/detail/message-detail-panel.tsx
```diff
@@ -0,0 +1,226 @@
...
+      window.removeEventListener('pointerup', onUp);
+      document.body.style.userSelect = '';
+      document.body.style.cursor = '';
+      patchDetailViewState({ sheetWidth: latestWidth });
+    };
+    window.addEventListener('pointermove', onMove);
+    window.addEventListener('pointerup', onUp);
+    document.body.style.userSelect = 'none';
+    document.body.style.cursor = 'col-resize';
+  };
```
eblairmckee, line 142: this seems overengineered, would [resizable](https://ui-playground.netlify.app/resizable) suffice?

frontend/src/components/pages/topics/messages/detail/message-detail-panel.tsx
```diff
@@ -0,0 +1,226 @@
...
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement;
+      if (e.key === 'Escape' && !expanded && !/^(input|textarea|select)$/i.test(target.tagName)) {
+        onClose();
+      }
+    };
+    window.addEventListener('keydown', onKeyDown);
+    return () => window.removeEventListener('keydown', onKeyDown);
+  }, [expanded, onClose]);
```
eblairmckee, line 154: iirc there's already a use-keyboard-shortcut (or something similar) in console

frontend/src/components/pages/topics/messages/hooks/use-keyboard-nav.ts
```diff
@@ -0,0 +1,76 @@
...
+      // OS/browser binding — not this row-copy shortcut; clobbering the user's real clipboard
+      // selection with the row's JSON would otherwise be a very unpleasant surprise.
+      if (e.key === 'c' && selectedKey && !(e.metaKey || e.ctrlKey || e.altKey)) {
+        const text = getCopyText(selectedKey);
+        if (text !== undefined) {
+          navigator.clipboard
+            .writeText(text)
+            .then(() => toast.success('Value copied to clipboard'))
+            .catch(() => toast.error('Could not copy to clipboard'));
+        }
```
eblairmckee, line 70: https://redpanda-ui-registry.netlify.app/docs/use-copy-to-clipboard

frontend/src/components/pages/topics/messages/table/messages-footer.tsx
```diff
@@ -0,0 +1,203 @@
...
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+import { Button } from 'components/redpanda-ui/components/button';
+import { ChevronDownIcon, ChevronLeftIcon, ChevronRightIcon, ClockIcon, DownloadIcon, Trash2Icon } from 'lucide-react';
+import { useEffect, useRef } from 'react';
+
+import { prettyBytes } from '../../../../../utils/utils';
```
eblairmckee, line 16:
```suggestion
import { prettyBytes } from 'utils/utils';
```

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
jvorcak: We can do that, I've used Collabpsible, because it looked a bit more consistent (paddings of actions on a right side etc). But I've changed that to Accordion now and I can always fine tune details later on.

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
jvorcak: I've ported this to table, which semantically makes the most sense I think.

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
jvorcak: Done

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
jvorcak: Screenshot 2026-09-08 at 14 49 37
  In current master, or in version w/o the feature flag, we commonly use `null` for this

frontend/src/components/ui/chip.tsx
```diff
@@ -0,0 +1,79 @@
+import { XIcon } from 'lucide-react';
```
jvorcak: Chip differs from Badge by being interactive — it carries an inline edit button (unwrap back into editable text) and a remove button, while Badge is a non-interactive rounded-full pill with overflow-hidden/truncate, so composing it would be more override than reuse. Agreed on the folder rule though: moved it out of the registry folder to `components/ui/chip.tsx` (and dropped its opacity mixins while there) in 1d59b27e8.

frontend/src/components/redpanda-ui/components/highlighted-input.tsx
jvorcak: Done in 1d59b27e8 — the ghost text now uses the `text-placeholder` ink token instead of `text-muted-foreground/60`.

frontend/src/components/ui/date-time-input.tsx
jvorcak: Tightened this one in 1d59b27e8 (dropped the stale shade name-dropping). Agreed on running a broader comment-cleanup pass after this merges.

frontend/src/components/ui/date-time-input.tsx
jvorcak: Checked against the current theme: `--color-selected` (#2c30a0) still differs from `--color-primary` (#444ce7), and the panel's "Now" button paints primary — so the override is still needed for the selected day to match. Kept it, updated the comment in 1d59b27e8.

frontend/src/components/pages/topics/messages/view-settings/view-settings-panel.tsx
jvorcak: Done in 1d59b27e8 — did a text-role pass over the view-settings and read-scope files: `text-caption` for the `text-[10px]/[11px]` uppercase micro-labels, `text-body-sm` for `text-xs`, `text-label` for the semibold `text-sm`/`text-[15px]` headings. One deliberate leftover: the table cells' `text-[13px]` mono (message-cells/messages-table) has no matching role (body is 14, body-sm is 12) — happy to move them whichever way you prefer.

frontend/src/components/pages/topics/messages/view-settings/view-settings-panel.tsx
jvorcak: Done in 1d59b27e8 — now `font-semibold text-label`.

frontend/src/components/pages/topics/messages/view-settings/view-settings-panel.tsx
jvorcak: Turned out we don't need a new size — the registry's `icon-sm` (size-8) covers it, so the custom `size-7` is gone in 1d59b27e8.

frontend/src/components/pages/topics/messages/view-settings/preview-fields-editor.tsx
jvorcak: Done in 1d59b27e8 — `text-overline`-style micro-labels are `text-caption` and the help text is `text-body-sm` (also swept column-list.tsx).

frontend/src/components/pages/topics/messages/toolbar/read-scope-popover.tsx
jvorcak: Done in 1d59b27e8 — the selected row uses `bg-selected-wash` now.

frontend/src/components/pages/topics/messages/toolbar/messages-toolbar.tsx
jvorcak: `icon-lg` works well here — it's exactly the filter bar's `min-h-10`, so the custom `size-10` is gone in 1d59b27e8. I kept one `[&_svg]:size-4` override so the glyph stays 16px like the toolbar's other icons rather than icon-lg's 24px default; if the 24px default looks right in the playground I can drop that too.

frontend/src/components/pages/topics/messages/detail/detail-sections.tsx
jvorcak: Update: swapped to the registry Accordion after all in ef7afb1d9 — controlled multi-open via the root's value array, persisted as before. Two workarounds to be aware of: the copy button is overlaid next to the trigger (a button can't nest inside the trigger button), and the expanded view's fill mode overrides the panel's measured-height class so the Value section can flex-stretch.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
jvorcak: Fixed in cf1c420f3 — the display window now keeps the tail of the ordered array, so it follows the loading frontier: "Load more" always reveals the page it just fetched, paging backward through ever-older rows (same shape as continuous+Oldest advancing forward). Once past the cap the trimmed side is the newest rows, and the footer says so ("{n} newer trimmed"). Regression tests in live-window.test.ts and message-order.test.ts assert the fetched page becomes visible. Full virtualization with the trimmed side scroll-reachable would be the deluxe version — happy to file it as a follow-up.

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
jvorcak: Fixed in cf1c420f3 — the paged/continuous path now appends through appendWithSlackCap with PAGED_BUFFER_MAX = 5000 (display cap is 150; the multiple feeds client filters and path hints). The trim drops the buffer's front, which in both continuous scopes is the end the display window has already discarded. Cumulative bytes/messages still come from the stream's stats, and the next-page token is tracked separately, so pagination is unaffected.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
jvorcak: Fixed in cf1c420f3 — useLiveTailLoop now returns { status, restart } (idle/running/retrying/exhausted). handleRefresh in live mode calls restart() (fresh failure budget) instead of returning early, so both the error state's Retry and the toolbar refresh actually reconnect; and the toolbar's "Streaming live…" spinner derives from status !== 'exhausted' rather than the URL toggle, so an exhausted loop no longer claims to be live. Tests cover exhaustion, restart re-arming, and status transitions.

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
jvorcak: Confirmed fixed: runStream's finally now compares abortControllerRef.current against its own controller and skips every finalizing write (flush, isLoadingMore, phase, backendPhase) when a newer run has taken over; the in-loop setters are guarded by the signal.aborted break at the top of each iteration.

frontend/src/components/pages/topics/messages/table/messages-table.tsx
jvorcak: This is fixed locally now — the sortable header label renders as a real <button type="button"> inside the <th> with a focus-visible ring, onClick moved onto it, and aria-sort stays on the <th>; non-sortable columns keep the plain span. (No component-library change needed after all.)

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
jvorcak: Fixed since: runStream tracks streamError (thrown stream errors and backend `error` control frames both) and rejects for real failures while resolving only clean completions, and useLiveTailLoop restarts only on success — failures back off exponentially (1s→30s cap) and give up after 5 consecutive failures. As of cf1c420f3 the exhausted state is also visible and recoverable (see the newer thread).

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
jvorcak: Fixed since: ordering now happens before windowing (orderForContinuousNewest → applyDisplayWindow in topic-messages-view), and as of cf1c420f3 the window keeps the loading frontier with the footer naming which side was trimmed — continuous+Newest reports "newer trimmed", every other scope "older trimmed". live-window.test.ts and message-order.test.ts cover the descending-arrival composition.

frontend/src/components/pages/topics/messages/hooks/use-keyboard-nav.ts
jvorcak: It stays messages-page-specific I think: the hook is row-key navigation over the visible (sorted, paginated) table order, a `/`-focuses-the-filter shortcut wired to this page's testid, and a value-copy toast — the generic parts it composes already come from shared code (`useCopyToClipboard` from the registry, `isTypingTarget`/`useHotKey` from hooks/). If a second table grows j/k navigation we can extract the movement core then.

frontend/src/globals.css
jvorcak: None of Tailwind's built-ins can express a background *flash* (pulse/ping animate opacity/transform), so custom keyframes are unavoidable — but agreed on not hand-rolling the class: as of cf1c420f3 it's a Tailwind v4 `@theme` token (`--animate-message-flash` + nested `@keyframes`), so the `animate-message-flash` utility is generated by Tailwind like any other animation, and globals.css is the app-owned theme entry it lives in.

frontend/rsbuild.config.ts
jvorcak: Ack — the e2e tooling here drives Playwright through the CLI; this watchOptions entry is just defense so anyone (or any agent) running the MCP server against the dev server can't wedge rspack's incremental build with its scratch output. Cheap to keep even if unused.

frontend/src/stores/topic-settings-store.ts
jvorcak: Sorting and pagination do live in tanstack (plus the URL for shareability) — the zustand store only holds the per-topic *persisted* view preferences (density, column set, preview tags/format), which need to survive unmounts, page switches, and reloads via localStorage and are read by more than one component (table, view-settings panel, preview cells). Tanstack table state is per-instance and ephemeral, so it can't cover those; nothing in the store duplicates what the table already manages.

SpicyPete (approved):
Seems good to me now 👍 
Small issue where the table settings draw underlaps the Ask AI box at the top, besides that didn't notice anything obvious

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,634 @@
...
+  useEffect(() => {
+    if (urlState.liveTail) {
+      lastSignatureRef.current = null;
+    }
+  }, [urlState.liveTail]);
+
+  const startLiveTail = useCallback(
+    (append: boolean) =>
+      search.start(
+        { ...searchParams, startOffset: PartitionOffsetOrigin.End, pageSize: undefined },
```
SpicyPete, line 226: Each live restart sends `startOffset=End` again, and the backend resolves that to the *current* high watermark. A bounded consume completes once `maxResults` records arrive (`isRequestSatisfied` in `list_messages.go`), the loop restarts immediately, and everything produced between that completion and the new HWM is skipped while the UI still says Live. With the default 50 that is a gap on every cycle of a busy topic.

Options: send a large `maxResults` for live runs so a single stream stays open, or restart from each partition's last seen offset + 1.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,634 @@
...
+        { live: true, append }
+      ),
+    // searchParams as a whole is intentionally not a dependency: scope edits (offset,
+    // timestamp, max results) are disabled in the UI while live, and deserializer changes
+    // take effect on the next (re)start. filterInterpreterCode and partitionId are singled
+    // out below because JS filters and the partition selection both stay editable while
+    // live — committing either should restart the tail with the new predicate/partition
+    // applied, not leave the UI showing a change that the running stream never picks up.
+    // biome-ignore lint/correctness/useExhaustiveDependencies: see above
+    [search.start, filterInterpreterCode, urlState.partitionId]
```
SpicyPete, line 236: `searchParams` is spread into the request but is not a dependency, so a live (re)start uses whatever params were current when this callback was last built. Deserializer or max-results changes made while live is *off* don't reach the next live start either, because the callback only rebuilds on `filterInterpreterCode` / `partitionId`. Refresh and Retry reuse the same stale closure.

Same underlying concern as the earlier live-tail settings thread; the "take effect on the next (re)start" note above doesn't hold today. Adding `searchParams` to the deps fixes it (the controls that would cause churn are already disabled while live). The `biome-ignore` also needs to sit on the hook call rather than the array so the diagnostic isn't silently hidden.

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
```diff
@@ -0,0 +1,402 @@
...
+            }
+            case 'done': {
+              const done = controlMessage.value;
+              const bytesConsumed = Number(done.bytesConsumed);
+              const elapsedMs = Number(done.elapsedMs) || Date.now() - startTime;
+              const nextPageToken = done.nextPageToken || null;
+              setStats((prev) => ({ ...prev, bytesConsumed, elapsedMs, nextPageToken }));
+              break;
+            }
+            case 'error':
```
SpicyPete, line 253: The backend emits a non-fatal `error` frame for offline partitions ("N of the requested partitions are offline…", `list_messages.go` ~L185) and then streams normally to `done`. Legacy only toasted it ("error doesn't necessarily mean the whole request is done"). Here it becomes `streamError`, so on any topic with one offline partition: paged mode renders the rows *and* the error state with Retry underneath, and live tail rejects on every restart, backs off, and exhausts after ~30s.

Suggest treating it as fatal only when no `done` frame follows (or no data arrived), and keeping the toast for the informational case.

frontend/src/components/pages/topics/messages/hooks/use-message-search.ts
```diff
@@ -0,0 +1,402 @@
...
+      // frame is as much a failure as a thrown/rejected stream: both make `start()` reject.
+      let streamError: Error | null = null;
+
+      // Live tail and push-down filters keep the stream open (backend semantics),
+      // so those runs get the long timeout — same rule as the legacy engine.
+      const isLongLived =
+        options.live === true ||
+        params.startOffset === PartitionOffsetOrigin.End ||
+        params.filterInterpreterCode !== '';
+      const timeoutMs = isLongLived ? LIVE_TIMEOUT_MS : DEFAULT_TIMEOUT_MS;
```
SpicyPete, line 224: For live runs the 30-minute client deadline is the only thing that ends a quiet stream: the `StartOffsetNewest` consume has `EndOffset = MaxInt64` and only completes on `maxResults`. connect-es aborts a *linked* controller on timeout, so `signal.aborted` stays false here, the run rejects with `deadline_exceeded`, the error state and toast appear, and `useLiveTailLoop` counts it as a failure. A low-volume topic exhausts live tail after ~6 idle windows with nothing actually wrong.

Treat a deadline on a live run as a benign restart (and reset the failure budget once data arrives), or drop the client deadline for live.

SpicyPete (commented):
Re-checked the branch at c250e82 against the legacy messages tab. The inline notes are the items that regress legacy behaviour or make live tail unreliable; smaller findings left out.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,634 @@
...
+  const lastSignatureRef = useRef<string | null>(null);
+  useEffect(() => {
+    if (urlState.liveTail) {
+      return; // live tail owns the stream (P5)
+    }
+    // Timestamp mode starts at -1 (unset) for one render until the date picker's
+    // own mount effect corrects it to "now" — searching on that sentinel would
+    // send an invalid startTimestamp to the backend instead of waiting the one
+    // tick for the real value.
+    if (searchParams.startOffset === PartitionOffsetOrigin.Timestamp && searchParams.startTimestamp < 0) {
```
SpicyPete, line 192: This guard waits for `StartOffsetDateTimePicker` to write `Date.now()`, but that picker only mounts inside the read-scope popover while it is open, and its own guard skips users with `startTimestampWasSetByUser` persisted. Two visible cases:

- A deep link with `?o=-4` and no `t` never searches and shows "No messages"; Refresh is a no-op.
- Anyone who used Timestamp on the legacy tab (which wrote `startTimestamp: null` on switching away) picks Timestamp here and nothing runs until they manually change the date.

Legacy set `Date.now()` in the mode-change handler; `setReadScopeMode('timestamp')` should do the same.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,634 @@
...
+              prev.some((f) => f.id === saved.id) ? prev.map((f) => (f.id === saved.id ? saved : f)) : [...prev, saved]
+            )
+          }
+          seedCode={jsDialog.seedCode}
+          seedName={jsDialog.seedName}
+        />
+      )}
+      {saveDialogOpen && (
+        <SaveMessagesDialog
+          messages={tableData}
```
SpicyPete: `tableData` is the 150-row display window, not the loaded set, but the dialog says it saves "all messages". In continuous or live mode a 500-row buffer exports 150 (for continuous + Newest, the tail). Legacy exported the full result set. Pass `filteredMessages` (or `search.messages`) here instead.

frontend/src/components/pages/topics/messages/topic-messages-view.tsx
```diff
@@ -0,0 +1,634 @@
...
+        onLoadMore={() =>
+          search.loadMore(urlState.maxResults).catch(() => {
+            // errors are surfaced through search.error
+          })
+        }
+        onPageChange={urlState.setPageIndex}
+        pageIndex={urlState.pageIndex}
+        pageSize={urlState.pageSize}
+        showStats={!(urlState.liveTail || isSearching)}
+        totalLoaded={filteredMessages.length}
```
SpicyPete, line 577: In live mode the pager renders (`continuousMode` is false) and pages over the full buffer, while the table only receives the 150-row window. With 900 rows and pageSize 50 the footer says "Page 1 of 18", but pages from 4 on render an empty table. The clamp effect above also uses `filteredMessages.length` and only runs while `phase === 'done'`, so it never corrects this. Either hide the pager in live mode as continuous does, or page over `windowedMessages.length`.

frontend/src/components/pages/topics/messages/hooks/use-messages-url-state.ts
```diff
@@ -0,0 +1,231 @@
...
+    'vd',
+    parseAsInteger.withDefault(PayloadEncoding.UNSPECIFIED)
+  );
+
+  const [pageIndex, setPageIndex] = useQueryState('page', parseAsInteger.withDefault(0));
+
+  const [pageSize, setPageSize] = useQueryStateWithCallback<number>(
+    {
+      onUpdate: (val) => setSearchParams(topicName, { pageSize: val }),
+      getDefaultValue: () => getSearchParams(topicName)?.pageSize ?? DEFAULT_MAX_RESULTS,
```
SpicyPete, line 152: This fallback reads the persisted legacy `pageSize`, which `createDefaultTopicSettings` seeds to 10, and this page never writes it (there is no rows-per-page control; the `setPageSize` branch in `handlePaginationChange` is unreachable). A fresh topic shows 50 rows on one page, but after any store-creating change (partition, sort, density, preview tag) a reload shows "Page 1 of 5" with no way to change it. Topics used on the legacy tab inherit its select value the same way.

A plain `useQueryState('pageSize', parseAsInteger.withDefault(50))` avoids the round-trip, or add a rows-per-page control.

frontend/src/components/pages/topics/messages/toolbar/read-scope-popover.tsx
```diff
@@ -0,0 +1,332 @@
...
+              <div className="flex flex-col gap-1.5">
+                <Label className="text-caption uppercase tracking-wide" htmlFor="read-scope-offset">
+                  Start offset
+                </Label>
+                <Input
+                  className="font-mono"
+                  id="read-scope-offset"
+                  inputMode="numeric"
+                  onChange={(e) => {
+                    const parsed = Number.parseInt(e.target.value, 10);
```
SpicyPete, line 232: The input is fully controlled by `customOffset` and NaN is dropped, so Backspace to empty snaps straight back to the previous value (select-all is the only way to replace it) and there is no invalid-state affordance. Legacy kept a local draft string and flagged invalid input. Each accepted digit also writes `startOffset` to the URL and triggers the debounced search, so typing `48210` can run five searches.

Keep a local draft string and commit on blur/Enter, or at minimum allow the empty state.

SpicyPete (commented):
Changing to comment, as recent review found some potential issues to confirm take a look at.  They may be existing logic gaps beyond the scope of the branch

PR #2593, opened 2026-08-05
RPCN - Pipelines listing page speed + UX improvements

SpicyPete:
## Rebuild the Redpanda Connect pipelines list for large clusters

The pipeline list was built for a handful of pipelines. On clusters with hundreds it was slow to
render, offered no way to narrow down what you were looking for, and buried broken pipelines
wherever the server happened to return them. This reworks the page around finding one pipeline in
a long list, and fixes the cost of getting that list in the first place.

### List page

- **Status tabs** — All / Running / Stopped / Error, each with a live count. Transitional states ride
  with their destination (starting counts as running, stopping as stopped). Counts come from the
  status column's faceted row model, so each tab shows what selecting it *would* yield under the
  current search and filters.
- **Search** by name or ID, debounced, matching case-insensitively against both.
- **Faceted filters** for input, output, and tag, with per-option counts. `Clear filters` appears
  only when something is actually filtered — the status tabs are views, not filters, so they aren't
  swept up by it — and it tracks the input directly rather than lagging 200ms behind the debounce.
- **Sortable Pipeline and Status columns.** Status sorts attention-first by default (errors and
  transitions above healthy pipelines, idle at the bottom), so a broken pipeline lands on page 1 of a
  large cluster instead of wherever the server put it.
- **Rows are clickable**, with the same guard the registry `DataTable` uses: clicks on portaled
  content (open menus, the delete-confirm backdrop) and on interactive descendants don't navigate,
  and neither does a click that ends a text selection — the pipeline ID is `select-all`, so one click
  grabs the whole thing for copying. ⌘/Ctrl/Shift-click and middle-click are left to the browser, so
  "open in a new tab" still means that.
- **Connector badges collapse duplicates** into `redpanda ×2` rather than repeating the same logo
  across the column.
- **Progressive rendering.** The list used to wait for every page before showing anything; it now
  renders as soon as the first page has rows and streams the rest in behind the table, with a
  distinct line for "still loading pages" vs "background refresh failed" — partial data and stale
  data read differently now.
- Search, facets and page survive a trip to the Kafka Connect tab and back (`keepMounted` — Base UI
  panels unmount by default).
- Dropped the Processors column and the old `DataTableFilter` stack from this page.

### Accessibility

- **The status tabs own a panel.** They sit outside the table (one table filtered per tab, not four
  panels), so without an `aria-controls` target a screen reader announced "tab, 1 of 4" with nowhere
  to move into. Each tab now points at the table region, which is labelled by the active tab.
- **Async status is announced.** "Loading more pipelines" and "Couldn't refresh pipelines" appear
  without user action, so each has an always-mounted `sr-only` live region (polite for the drain,
  `role="alert"` for the failure) — the visible lines animate in and out, and a live region only
  announces changes made while it's already in the DOM.
- Row click stays a pointer shortcut with no row tab stop: the name cell is already a real link, so a
  tab stop per row would just duplicate it.

### Fetch and render cost

- **Page size 500** instead of `MAX_PAGE_SIZE` (which is 25) — 20× fewer sequential round trips to
  drain. The server does the same work per call at any page size: it lists everything and slices.
- **Deduplicate the drain by pipeline ID.** The dataplane's keyset page token names the first ID of
  the next page; when that pipeline is deleted mid-drain, a server resolving the token by exact match
  restarts at page one and replays rows we already have.
- **Stop draining on any token the drain already requested.** Keyset tokens only move forward, so a
  repeat means the server sent us backwards. The first cut only caught an immediate `A → A` repeat,
  which a `A → B → A` cycle walks straight past — the drain then alternates forever, adding a page to
  the query cache every round. Now checked against every prior page param, with a test for each shape.
- **Memoize the YAML parse per config text.** The transform re-ran a full parse for every row on every
  drain step and poll tick; it's now O(new rows), with a bounded cache that evicts its oldest half.
- **Row identity keyed on pipeline ID** (not row index) and `autoResetPageIndex: false`, so streaming
  pages don't yank you back to page 1 or repaint a shifted window of rows. Filter and sort changes
  still reset to page 1, and a shrinking row set is clamped before paint.
- Cache the facet icon component per connector name, so logos in an open filter popover don't remount
  and flash on every poll; count the status tabs in a single pass; and memoize the per-row connector
  aggregation so cells don't re-derive it on every keystroke.
  
  ### Video
  
  
https://github.com/user-attachments/assets/33e50e14-d489-4039-85e9-935889cbc280


SpicyPete:
@claude review

frontend/src/components/misc/buttons/data-refresh/component.tsx
```diff
@@ -83,62 +81,63 @@ export const DataRefreshButton = () => {
...
-            size="xs"
-            variant="ghost"
+    <TooltipProvider>
+      <div className="flex items-center gap-1">
+        <Tooltip>
+          <TooltipTrigger
+            render={
+              <Button
+                aria-label={isActive ? 'Pause auto refresh' : 'Start auto refresh'}
+                className="size-7"
```
eblairmckee: iirc button has a child svg selector that specifies icon size, so if you're trying to make the svg bigger may want to `>svg` (can't remember syntax) here.

frontend/src/components/misc/buttons/data-refresh/component.tsx
```diff
@@ -83,62 +81,63 @@ export const DataRefreshButton = () => {
...
+              }
             />
-          </Popover>
+            <TooltipContent className="max-w-64">
+              <div className="flex flex-col gap-1">
+                <span className="font-medium">Force refresh</span>
+                <span>
+                  Refresh the data shown on this page. When switching pages, any data older than{' '}
+                  {prettyMilliseconds(REST_CACHE_DURATION_SEC * 1000)} is refreshed automatically.
+                </span>
```
eblairmckee, line 125: nice

frontend/src/components/pages/rp-connect/pipeline/list.tsx
```diff
@@ -79,44 +100,86 @@ type Pipeline = {
...
+        <List>
+          {connectors.slice(-overflow.length).map((c) => (
+            <ListItem key={c.name}>{c.count > 1 ? `${c.name} ×${c.count}` : c.name}</ListItem>
+          ))}
+        </List>
+      )}
+    >
+      {connectors.map((c) => (
+        <Badge key={c.name} variant="neutral-inverted">
+          <ConnectorLogo className="size-3.5" fallback={Box} name={c.name as ComponentName} />
```
eblairmckee, line 146: [shower thought]: i'd love to explore having classNames for icon sizes that at t-shirt sized, or add to our text utility classNames svg sizing so when we wrap an icon + label in a type util we know they will be similarly sized

frontend/src/components/pages/rp-connect/pipeline/list.tsx
```diff
@@ -128,75 +191,115 @@ const pipelineStateToStatusVariant: Record<Pipeline_State, StatusBadgeVariant> =
...
+  [Pipeline_State.STOPPED]: 5,
+  [Pipeline_State.UNSPECIFIED]: 6,
 };
 
 const PAGE_SIZE = 20;
 
+// Modified and non-primary clicks mean "open elsewhere", so a row leaves them to the browser —
+// the name cell is a real link, so ⌘-click and middle-click still work there.
+const isModifiedClick = (event: MouseEvent<HTMLElement>) =>
+  event.metaKey || event.ctrlKey || event.shiftKey || event.altKey || event.button !== 0;
```
eblairmckee: nit seems better fit for a keyboard util file

frontend/src/components/pages/rp-connect/pipeline/list.tsx
```diff
@@ -128,75 +191,115 @@ const pipelineStateToStatusVariant: Record<Pipeline_State, StatusBadgeVariant> =
...
+}) => (
+  <>
+    <div aria-live="polite" className="sr-only">
+      {isLoadingMorePages ? 'Loading more pipelines' : ''}
+    </div>
+    <div className="sr-only" role="alert">
+      {listErrorMessage ?? ''}
+    </div>
+  </>
+);
```
eblairmckee, line 208: [shower thought]: would love to upstream this to the pagination component in registry

frontend/src/react-query/api/pipeline.tsx
```diff
@@ -112,9 +111,12 @@ export const useListPipelinesQuery = (
...
         }
       : false,
-    getNextPageParam: (lastPage) => {
+    getNextPageParam: (lastPage, _allPages, _lastPageParam, allPageParams) => {
       const nextPageToken = lastPage?.response?.nextPageToken;
-      if (!nextPageToken) {
+      // Any token already requested, not just the last one: keyset tokens only move forward, so a
+      // repeat means the server sent us backwards (A→A or a longer A→B→A cycle) and the drain would
+      // loop forever, adding a page per round. O(pages) per step, ~20 for 10k pipelines.
+      if (!nextPageToken || allPageParams.some((param) => param?.pageToken === nextPageToken)) {
```
eblairmckee, line 119: [minor]: this makes me think we should fix pushdown filters for pipelines list... code smell when I see so much sorting/deduping on the frontend especially wrt pagination

frontend/src/components/ui/fade-presence.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { AnimatePresence, motion, useReducedMotion } from 'motion/react';
+import type { ReactNode } from 'react';
+
+/**
+ * Mount/unmount transition for conditional UI chrome — toolbar chips, status lines,
+ * inline hints. Exit is quicker than enter, `initial={false}` animates changes but not
+ * the first render, and reduced-motion collapses the movement to a pure fade.
+ */
+export function FadePresence({
```
eblairmckee, line 19: [shower thought]: would love to see this upstreamed to ui-registry and start to explore tokenized motion with that effort

frontend/src/components/ui/pipeline/constants.ts
```diff
@@ -31,18 +31,6 @@ export const PIPELINE_STATE_LABELS: Partial<Record<Pipeline_State, string>> = {
   [Pipeline_State.COMPLETED]: 'Completed',
 };
 
-/**
- * Pipeline state options for filtering.
- */
-export const PIPELINE_STATE_OPTIONS = [
```
eblairmckee, line 37: [follow]: I would be curious if this kind of tab/filter approach resonates with users. worth doing some usability testing with real customers

frontend/src/components/pages/rp-connect/pipeline/pipeline-throughput-card.tsx
```diff
@@ -56,23 +56,56 @@ type ThroughputContentProps = {
...
-      </Alert>
+      <ThroughputPlaceholder
+        action={
+          <Button className="mt-2" onClick={onRetry} size="sm" variant="outline">
+            Try again
+          </Button>
+        }
+        description="The metrics service didn't respond. Data will appear once it's reachable."
+        title="Throughput metrics aren't available right now"
+      />
```
eblairmckee, line 108: [nit]: could potentially use ui-registry Empty component here, although it may be overengineered for this case. iirc, we may also have a chart-placeholder component 🤔 but that may just be something I've seen in cloud-ui

frontend/src/components/pages/rp-connect/pipeline/list.tsx
```diff
@@ -79,44 +100,86 @@ type Pipeline = {
...
+      surplus -= 1;
+      if (surplus <= 0) {
+        break;
+      }
+    }
+  }
+  const parsed = parseConfigComponents(configYaml);
+  configComponentsCache.set(configYaml, parsed);
+  return parsed;
+};
```
eblairmckee: [nit]: would rather this live in a utils file rather than .tsx

eblairmckee (approved):
left some nits, but otherwise I love the pagination improvements, a11y considerations, and cleaner "list" presentation of pipelines

PR #2591, opened 2026-08-04
refactor(frontend): remove enableRpcnTiles flag and serverless get-started wizard

eblairmckee:
## Summary

- removes the `enableRpcnTiles` feature flag and every branch it gated, plus the `./connect-tiles` federated module exposure
- removes the serverless get-started API wizard (`/get-started/api` route, `APIConnectWizard`, `api-wizard-store`, onboarding code-snippet query) — its only entry point was Cloud UI's serverless onboarding wizard, deleted in the companion cloudv2 PR (redpanda-data/cloudv2#28773)
- removes the `/rp-connect/wizard` route entirely — pipeline creation always lands on `/rp-connect/create` now

## Why

`enableRpcnTiles` was the tiles-based create flow embedded in the legacy layout — it's been superseded since the new connect UI (`enablePipelineDiagrams` tier, #2285) shipped at the end of March, and that tier has proven stable since. The flag is now off in LaunchDarkly, so every `isFeatureFlagEnabled('enableRpcnTiles') && isEmbedded()` branch was dead code.

The `./connect-tiles` federated exposure goes with it: its only consumer was Cloud UI's serverless onboarding wizard, which is deleted in the companion cloudv2 PR (redpanda-data/cloudv2#28773).

The `/rp-connect/wizard` route is retired with it. Tier 1 already redirected wizard → create, and every "create pipeline" entry point now goes straight to `/rp-connect/create` — the wizard was an extra hop that no longer earned its place. That takes `ConnectOnboardingWizard`, `ConnectTiles`/`ConnectTile`, the wizard-only stepper constants, and the yaml-regeneration helpers with it. `AddTopicStep`/`AddUserStep` and `rpcn-wizard-store` stay — `PipelinePage` uses them.

## What changed

- `constants.ts` — dropped `enableRpcnTiles` from `FEATURE_FLAGS` (also swept the dead `enableServerlessOnboardingWizard` default, which nothing reads)
- `pipelines-create.tsx` / `pipelines-edit.tsx` / `pipelines-details.tsx` / `connect/overview.tsx` — removed the flag-gated `PipelinePage` / `PipelineListPage` early returns; tier-1 routing via `enablePipelineDiagrams` is unchanged
- `header.tsx` — `useShouldHideHeader` now checks `enablePipelineDiagrams` only
- `module-federation.config.ts` — removed the `./connect-tiles` expose
- `console-app.test.tsx` — flag passthrough tests now use `enablePipelineDiagrams`
[...]

PR #2594, opened 2026-08-06
Fix - Show the new pipeline list in Cloud without feature flag

SpicyPete:
The new Redpanda Connect list page (PipelineListPage) was built but never mounted — the Connect page always rendered the legacy tabs. Mount it on isEmbedded() rather than behind a feature flag: Cloud gets the new list, self-hosted keeps the legacy tabs and install intro.


eblairmckee:
fixes #2591 

regression on integration:
Screenshot 2026-08-06 at 12 05 21 PM

eblairmckee (approved):

PR #2612, opened 2026-08-23
chore(frontend): upgrade bun to 1.4.0

malinskibeniamin:
## Proven impact

- Pins all 7 live `setup-bun` steps to Bun 1.4.0 (0/7 before, 7/7 after).
- Keeps local development, CI, and the E2E workflow example on the same runtime.
- Regenerates the frontend lockfile with Bun 1.4.0 and removes the orphan root lockfile, which had no package manifest.

## Summary

- declare `bun@1.4.0` in `frontend/package.json`
- pin Bun 1.4.0 across frontend GitHub Actions and docs
- regenerate `frontend/bun.lock`; dedupe `semver` to 7.8.5
- delete the unusable root `bun.lock`

## Why

A single explicit Bun version prevents local/CI drift. Explicit action inputs also fail closed when the requested version does not exist.

## Commits

- `dd54465ab chore(frontend): upgrade bun to 1.4.0`

## Reviewer guide

1. Check the 7 `setup-bun` steps use the exact `1.4.0` pin.
2. Check `frontend/package.json` and regenerated lockfile agree.
3. Confirm the deleted root lockfile was orphaned (there is no root `package.json`).

## Dependency upgrade path

- **Toolchain:** Bun 1.4.0.
- **Call sites:** no Bun runtime API imports or `Bun.*` usage; Bun is used as package manager and script launcher.
- **Lockfile:** regenerated by Bun 1.4.0; the only package resolution change dedupes `semver` at 7.8.5.
- **Risk:** Bun minor release with documented breaking changes, bounded by frozen install, full test suite, build, and direct action execution.

## Test plan

- `bun install --frozen-lockfile --ignore-scripts`
- `bun run lint`
- `bun run type:check`
- `bun run test` — 177 files, 2,093 tests passed
- `bun run build`
- YAML parse for all changed workflows
- `git diff --check`

## Dogfood

**PASS**

- Ran `setup-bun@v2`'s real `dist/setup/index.js` entrypoint with `bun-version=1.4.0`; it resolved `1.4.0+34cbb9a40`.
- Ran the repository entrypoints: frozen install and production build; both passed.
- Break attempts failed closed: invalid Bun `999.999.999` exited 1; stale manifest against the frozen lockfile exited 1.
- Repaired the initial permissive version-file approach by switching every workflow to an explicit version pin, then replayed successfully.
- Limitation: no hosted GitHub runner was invoked locally; the actual action bundle was executed on macOS arm64.


frontend/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "frontend",
   "version": "0.1.0",
+  "packageManager": "bun@1.4.0",
```
malinskibeniamin, line 4: **Priority: P3**

`packageManager: "bun@1.4.0"` is added here, but the same literal is now hardcoded in six other places (`frontend-verify.yml` ×5, `frontend-react-doctor.yml`, `frontend-ui-audit.yml`, plus the `e2e-tester` reference doc and `CONTRIBUTING.md`). The next bump has to touch all of them, and a partial bump silently splits local and CI toolchains without any check failing.

Consequence: maintenance drift risk only — every call site is correct today.

Correction (optional): make this field the single source of truth. `oven-sh/setup-bun@v2` supports `bun-version-file: frontend/package.json` and reads the `packageManager` field, so the workflow steps can drop the literal; alternatively hoist it to one `env: BUN_VERSION` per workflow and reference `${{ env.BUN_VERSION }}`.

Verify: `rg -n 'bun-version|packageManager' .github frontend/package.json` should show one authoritative version after the change.

*Automated `/review`.*

CONTRIBUTING.md
```diff
@@ -28,7 +28,7 @@ The basics:
 
 Prerequisites that should be installed before proceeding:
 
-- Latest Node.js v22 & bun 1.2.4 or newer
+- Latest Node.js v22 and Bun 1.4.0
```
malinskibeniamin, line 31: **Priority: P3**

The wording moves from a floor (`bun 1.2.4 or newer`) to an exact version (`Bun 1.4.0`), so a contributor already on 1.4.1+ now reads the doc as telling them to downgrade. Nothing in the repo enforces an exact local version — `bun install` does not act on `packageManager`, and CI installs its own pinned toolchain — so the stricter phrasing overstates the requirement.

Consequence: minor contributor-facing confusion; no functional impact.

Correction (optional): `- Latest Node.js v22 and Bun 1.4.0 or newer`, matching the intent of the previous phrasing while naming the CI-pinned baseline.

Verify: read-only doc change; no command needed.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 2 finding(s).

SpicyPete (approved):

PR #2613, opened 2026-08-24
build(frontend): upgrade rsbuild to 2.2.1

malinskibeniamin:
## Proven impact

**Decision rule:** prove at least a 20% reduction in the largest emitted JavaScript asset while keeping total emitted JavaScript growth below 1% and preserving production runtime entrypoints.

| Lane | Metric | Merge-base `d146e9809` | PR head `4954d83e7` | Delta |
|---|---|---:|---:|---:|
| Codebase | Largest emitted JavaScript asset | 10,284,413 B | 5,954,958 B | -4,329,455 B (-42.1%) |
| Product guardrail | Total emitted JavaScript | 43,187,229 B | 42,988,490 B | -198,739 B (-0.46%) |
| Delivery surface | Emitted JavaScript files | 560 | 487 | -73 (-13.0%) |

**Selected-cap A/B at exact PR head:**

| Metric | No cap | 4 MiB cap | Delta |
|---|---:|---:|---:|
| Largest async JavaScript asset | 10,059,699 B | 5,650,771 B | -4,408,928 B (-43.8%) |
| Async JavaScript files | 468 | 481 | +13 (+2.8%) |
| Total emitted JavaScript | 42,935,714 B | 42,988,490 B | +52,776 B (+0.12%) |

A 512 KiB cap emitted 595 async JavaScript files (+127 versus no cap) but did not improve the 5,650,771 B maximum reached at 4 MiB, so 4 MiB is the measured Pareto choice.

**Value proven:** the primary metric exceeded the 20% threshold, total JavaScript improved rather than regressed, and the final cap captures the chunk-size benefit with 114 fewer async files than 512 KiB.

Method: `bun run build-local-test` with fixed build variables and independent frozen-lockfile installs for merge-base `d146e9809` and exact PR head `4954d83e7`; production asset bytes enumerated from clean `build/**/*.js` outputs. Chunk-cap A/B used `createRsbuild().build()` at exact head with identical source/dependencies and only `maxAsyncSize` varied. macOS 26.6.2, Apple M5 Max, AC power, Bun 1.4.0. Exact byte counts are deterministic; build timings are deliberately excluded because the earlier timing sample did not measure the final commit under the controlled paired protocol.

Runtime guardrail: whole-PR dogfood passed overview/topics/SQL/deep-link/invalid routes, 30/30 Module Federation sync assets, and 4/4 Monaco workers; all fetched artifacts returned HTTP 200 and were byte-identical, with zero chunk/worker browser errors. No live Redpanda backend was available.

## Summary

- Upgrade Rsbuild to the [2.2.0 GA release](https://github.com/web-infra-dev/rsbuild/releases/tag/v2.2.0) and update every official Rsbuild plugin to its latest compatible release.
- Replace the Babel React Compiler transform with Rsbuild's native Rust implementation and remove the React 19-unneeded compiler runtime.
- Enable every generally applicable Rspack experiment: async WebAssembly, future defaults, native watcher, production pure-function analysis, and Rsbuild-provided source imports.
- Keep feature-specific or incompatible experiments off: remote HTTP builds, unused deferred-import syntax, custom filesystem routing, and the unfinished ESM-only Rspack runtime mode.
- Remove `fasterModuleConcatenation` and `lazyBarrel`, which Rspack removed before GA.
- Parallelize SVGR and split asynchronous legacy UI and Monaco vendors into reusable chunks with a measured 4 MiB async cap.
- Exercise the fully resolved production config, including the real Module Federation plugin, in the integration suite.

## Why

Adopt Rsbuild 2.2's GA compiler, watcher, forward defaults, and bundling improvements while reducing the worst emitted chunk size. Application behavior and backend integrations are out of scope.

Surface review skipped: build-only change; no rendered UI changed.

## Experiment scope

- `asyncWebAssembly: true`: current Rspack default, pinned explicitly for the supported async Wasm pipeline.
- `futureDefaults: true`: opts into the next-major defaults now so incompatibilities fail during deliberate upgrades.
- `nativeWatcher: true`: uses Rust file watching for development rebuilds.
- `pureFunctions: isProd`: enables cross-module pure-function analysis only where production tree shaking applies.
- `sourceImport: true`: enabled by Rsbuild 2.2 and protected by the config contract.
- Excluded `buildHttp` because Console has no remote URL imports and enabling network-backed builds would expand the supply-chain surface.
- Excluded `deferImport` because Console uses no deferred-import proposal syntax.
- Excluded `runtimeMode: 'rspack'` because it remains under development and targets ESM `modern-module` output, while Console uses Module Federation's current runtime.
- Excluded `useInputFileSystem` because Console does not replace Rspack's input filesystem.

## Commits

- `8df6382bb` build(frontend): upgrade rsbuild to 2.2
- `b5e10815c` build(frontend): enable rspack experiments
- `4954d83e7` build(frontend): address rsbuild review feedback

## Reviewer guide

Start with `frontend/rsbuild.config.ts`, then review dependency intent in `frontend/package.json`, the fully resolved config contract in `frontend/tests/rsbuild-config.integration.test.ts`, suite routing in the two Vitest configs, and finally the generated lockfiles.

## Dogfood evidence

- **Verdict:** PASS
- **Entrypoints:** `bun run start -- --port 57086`, `bun x rsbuild preview --port 57087`, production browser routes, emitted worker files, and Module Federation HTTP artifacts.
- **Actions and break attempts:** Triggered native-watcher HMR; navigated and deep-reloaded overview, topics, SQL, and invalid routes; typed into the Monaco SQL editor; fetched all federation sync assets and the editor, JSON, TypeScript, and YAML workers; compared fetched bytes with emitted files; checked worker syntax and async-loader references.
- **Observations:** HMR rebuilt in 3.80 s without losing page state. Production routes and invalid-route recovery rendered. All federation artifacts and worker files returned HTTP 200 and matched emitted bytes. All worker files passed syntax checks and contained no async chunk-loader references. No chunk or worker errors appeared in the browser.
- **Repairs and replay:** Moved the production-config contract to the integration suite, removed the Module Federation mock and upstream-default assertion, selected the measured 4 MiB cap, rebuilt, and replayed every entrypoint successfully.
- **Limits:** No live Redpanda backend, so live cluster data workflows were not exercised; expected Connect/proxy errors rendered existing unavailable states.

## Dependency upgrade path

- **Upgrade evidence:** Production config contracts were developed RED → GREEN; frozen install, changed-file lint, type checking, unit tests, focused integration, production and RsDoctor builds, bundle analysis, and browser dogfood passed.
- **Packages:** `@rsbuild/core@2.2.0`; React `2.1.0`; Sass `2.0.1`; SVGR `2.0.5`; Tailwind CSS `2.0.3`; YAML `2.0.0`; Node polyfill remains latest at `1.4.6`. Removed `@rsbuild/plugin-babel`, `babel-plugin-react-compiler`, and `react-compiler-runtime`.
- **SemVer confidence:** Core is exactly pinned to GA `2.2.0`; `bun.lock` contains exactly one `@rspack/core@2.2.0` and one matching `@rspack/binding@2.2.0`. Official plugin upgrades use their current compatible releases.
- **Risk gate:** Exact core pinning contains API drift, the real resolved-config integration contract protects adopted APIs, and development, production, worker, and federation entrypoints were exercised.
- **Security notes:** No remote-build experiment was enabled; frozen install reproduced the reviewed lockfiles.

## Test plan

- [x] `bun install --frozen-lockfile`
- [x] `bun run lint:check:file rsbuild.config.ts tests/rsbuild-config.integration.test.ts vitest.config.integration.mts vitest.config.unit.mts`
- [x] `bun run type:check`
- [x] Unit suite: 57 files, 837 tests
- [x] Real Rsbuild/Module Federation config integration contract
- [x] Production build and `RSDOCTOR=1` production build
- [x] Async chunk-threshold sweep and emitted bundle analysis
- [x] Development HMR and production browser dogfood


frontend/rsbuild.config.ts
```diff
@@ -131,17 +111,39 @@ export default defineConfig({
...
+    cacheGroups: {
+      legacyUi: {
+        test: /[\\/]node_modules[\\/]@redpanda-data[\\/]ui[\\/]/,
+        name: 'lib-redpanda-ui',
+        priority: 40,
+        enforce: true,
+        reuseExistingChunk: true,
+      },
+      monaco: {
+        test: /[\\/]node_modules[\\/]monaco-editor[\\/]/,
```
malinskibeniamin, line 129: **Priority: P1**

**The Monaco `enforce` cache group can break Monaco's web workers.**

`MonacoWebpackPlugin` (configured at line 193 with `filename: 'static/js/[name].worker.js'`) emits *separate entrypoints* whose modules come from `node_modules/monaco-editor`. This cache group's `test` matches exactly those modules, and with the new top-level `chunks: 'all'` plus `enforce: true`, the worker entrypoints' own code becomes eligible to be hoisted into the shared `lib-monaco-editor` chunk.

A worker entry that depends on a second emitted file has no way to load it — the plugin points the worker URL at a single `*.worker.js`. The consequence is that yaml/json/typescript/protobuf language services fail at runtime (validation, completion, hover) while the editor itself still renders, so this fails quietly.

**Correction:** scope the cache group away from worker entries, e.g. add `chunks: (chunk) => !chunk.name?.endsWith('worker')` to the `monaco` group (and confirm the same for `legacyUi`), or drop `enforce` so size heuristics do not force the extraction.

**Verify:** `cd frontend && bun run build`, then check each `build/static/js/*.worker.js` is self-contained (no `__webpack_require__.e` / chunk-load calls) and open a topic's YAML/JSON editor in the built app with the console open — no worker load errors.

*Automated `/review`.*

frontend/rsbuild.config.ts
```diff
@@ -131,17 +111,39 @@ export default defineConfig({
...
     },
   },
   tools: {
-    rspack: (config, { appendPlugins }) => {
+    rspack: (config, { appendPlugins, isProd }) => {
       config.lazyCompilation = false;
       config.experiments = {
         ...config.experiments,
-        lazyBarrel: false,
+        ...(isProd ? { fasterModuleConcatenation: true } : {}),
```
malinskibeniamin: **Priority: P2**

**`lazyBarrel: false` was dropped here without justification.**

This line replaces the previous explicit `lazyBarrel: false` in `config.experiments`. Removing the opt-out re-enables Rspack's barrel-file optimization, which changes when re-export barrels' modules are evaluated and therefore the ordering of import side effects. This config depends on side-effect-sensitive setup (Monaco worker registration, the `react-onclickoutside` and `date-fns-tz` shims aliased just below, CSS imports), and the commit is titled purely as a version upgrade — so the behavior change is invisible to anyone reading the log.

**Correction:** restore `lazyBarrel: false`, or keep it removed and record in the commit body why the flag was originally set and what evidence shows enabling it is safe now.

**Verify:** `cd frontend && bun run build` and compare the emitted chunk graph against master, then smoke-test the app for missing side effects (Monaco workers, date formatting, datepicker shim).

*Automated `/review`.*

frontend/tests/rsbuild-config.test.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright 2026 Redpanda Data, Inc.
+
+import { createRsbuild } from '@rsbuild/core';
+import { describe, expect, it, vi } from 'vitest';
+
+import { fileURLToPath } from 'node:url';
+
+vi.mock('@module-federation/rsbuild-plugin', () => ({
```
malinskibeniamin: **Priority: P2**

**Mocking `pluginModuleFederation` means this test does not describe the production config.**

The real Module Federation plugin mutates `optimization` (including `splitChunks`) and injects the container/remote entries. Stubbing it out to a no-op `setup` means the asserted `optimization.splitChunks` shape and the `serializedConfig` string checks below are taken from a pipeline that never ships — the test can stay green while the actual production config differs on exactly the surface this PR changes.

**Correction:** either drop the mock and let the real plugin run during `inspectConfig` (it needs no network), or narrow the test's claims to the values this config file sets directly and assert the federation-influenced shape in the integration/build suite instead.

**Verify:** remove the `vi.mock` block and re-run the file; if the assertions still hold, the mock was unnecessary — if they change, the mock was hiding the real config.

*Automated `/review`.*

frontend/tests/rsbuild-config.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+    const { origin } = await rsbuild.inspectConfig({ mode: 'production' });
+    const [rspackConfig] = origin.bundlerConfigs;
+    const serializedConfig = JSON.stringify(rspackConfig);
+
+    expect(rspackConfig?.experiments).toMatchObject({
+      fasterModuleConcatenation: true,
+      nativeWatcher: true,
+    });
+    expect(rspackConfig?.experiments).not.toHaveProperty('lazyBarrel');
```
malinskibeniamin: **Priority: P3**

**`not.toHaveProperty('lazyBarrel')` is a change detector, not a behavior assertion.**

It asserts the absence of an upstream default rather than anything about the build output, so it will fail whenever Rspack begins materializing `lazyBarrel` in the resolved experiments — an unrelated upgrade breaks the test with no product defect. Prefer asserting the properties this config intentionally sets and leave upstream defaults unconstrained.

*Automated `/review`.*

frontend/tests/rsbuild-config.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  pluginModuleFederation: () => ({
+    name: 'test:module-federation',
+    setup: () => undefined,
+  }),
+}));
+
+import rsbuildConfig from '../rsbuild.config';
+
+describe('Rsbuild production config', () => {
+  it('uses the optimized compiler and chunking pipeline', async () => {
```
malinskibeniamin: **Priority: P3**

Repo convention is `test()` rather than `it()` for test cases (see the project test guidance and neighbouring files under `frontend/tests/`). Rename for consistency.

*Automated `/review`.*

frontend/tests/rsbuild-config.integration.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    name: 'test:module-federation',
+    setup: () => undefined,
+  }),
+}));
+
+import rsbuildConfig from '../rsbuild.config';
+
+describe('Rsbuild production config', () => {
+  it('uses the optimized compiler and chunking pipeline', async () => {
+    const rsbuild = await createRsbuild({
```
malinskibeniamin, line 9: **Priority: P3**

**A full `createRsbuild()` + `inspectConfig()` run now lives in the fast unit project.**

`vitest.config.unit.mts` includes `tests/**/*.test.ts` with `environment: 'node'`, so this case joins the pure unit suite while depending on `loadEnv`, plugin resolution, the TanStack Router plugin, and the real filesystem/cwd. That makes the unit suite slower and environment-sensitive.

**Correction:** move this to the integration config (or a dedicated build-config project) so unit runs stay hermetic and quick.

**Verify:** `cd frontend && bun run test:unit` and compare suite duration against master.

*Automated `/review`.*

frontend/rsbuild.config.ts
```diff
@@ -131,17 +111,39 @@ export default defineConfig({
     // production diagnostics.
     removeConsole: ['log', 'warn'],
   },
+  splitChunks: {
+    preset: 'default',
+    chunks: 'all',
+    // Cap asynchronous chunks without imposing a size limit on initial chunks.
+    maxAsyncSize: 512 * 1024,
```
malinskibeniamin: **Priority: P3**

`maxAsyncSize: 512 * 1024` has no supporting measurement. The comment explains what the cap does but not why 512 KB is the right threshold — capping every async chunk trades request count against payload size, and the effect depends on this app's route-split shape. Recording the before/after bundle report in the commit body (the config already supports `RSDOCTOR=1`) would make the number reviewable and safe to revisit later.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 9 finding(s).

These findings could not be anchored to a changed line:
- **P1** `frontend/bun.lock:400` — **Duplicate `@rspack/core` instance for module federation.**

`@rsbuild/core@2.2.0-beta.2` resolves `@rspack/core@2.2.0-rc.0`, but this lock now nests a second copy: `@module-federation/rspack/@rspack/core → 2.0.6` (and the same for `@rsdoctor/rspack-plugin` / `@rsdoctor/types`). Before this change everything deduped on `2.0.6`.

`@module-federation/rspack` imports `@rspack/core` directly, so its plugin classes and hook taps come from a different rspack instance — and a different native binding — than the compiler that runs them. Console ships `embedded.js` as a federated remote consumed by Cloud UI, so this sits on the critical output path, and it also pulls a second rspack native binary into the install.

**Correction:** align the versions — either hold `@rsbuild/core` on the 2.0.x line until `@module-federation/rsbuild-plugin` supports rspack 2.2, or add an override so a single `@rspack/core` resolves for all consumers.

**Verify:** `cd frontend && bun install && bun pm ls | grep rspack` (expect exactly one `@rspack/core` and one `@rspack/binding`), then `bun run build` and load `embedded.js` from a federation host.
- **P2** `frontend/package.json:788` — **A beta build toolchain is pinned for release builds, with a floating prerelease range underneath.**

`@rsbuild/core: 2.2.0-beta.2` declares `@rspack/core: ~2.2.0-beta.1`, which the lock satisfied with `@rspack/core@2.2.0-rc.0` — a *different* prerelease than the one named. That range keeps floating across prereleases (`beta.3`, `rc.1`, …) on any lock refresh, so the bundler that produces the shipped artifact can change without a reviewed diff, and prereleases carry no compatibility guarantee.

**Correction:** if the 2.2 features are needed now, state that in the commit body and add an exact override for `@rspack/core` so the prerelease cannot drift; otherwise wait for `@rsbuild/core@2.2.0` stable.

**Verify:** `cd frontend && bun install --frozen-lockfile && bun pm ls | grep '@rspack/core'` — the resolved version should equal the pinned one on a clean install.

frontend/rsbuild.config.ts
malinskibeniamin: Verified without changing the cache group: Rsbuild emits Monaco workers through separate child-compiler assets, not the main split-chunk graph. The editor, JSON, TypeScript, and YAML workers were fetched from the production preview, matched emitted bytes, passed syntax checks, and contained no async chunk-loader or lib-monaco references. The SQL Monaco route also loaded and accepted input with no worker/chunk console errors. These results show `enforce` did not hoist worker code in this build; the PR dogfood section records the evidence.

frontend/rsbuild.config.ts
malinskibeniamin: Addressed in the GA update: Rspack 2.2 removed the `lazyBarrel` experiment, so the obsolete opt-out cannot be restored. The PR now states this explicitly, and production build plus route/Monaco/date-dependent browser smoke coverage passed on GA.

frontend/tests/rsbuild-config.test.ts
malinskibeniamin: Fixed in 4954d83e7: removed the Module Federation mock. The renamed integration test now runs `createRsbuild()` with the real production plugin pipeline and asserts the fully resolved config.

frontend/tests/rsbuild-config.test.ts
malinskibeniamin: Fixed in 4954d83e7: removed the `lazyBarrel` absence assertion. The contract now asserts only experiments intentionally selected by this config.

frontend/tests/rsbuild-config.test.ts
malinskibeniamin: Fixed in 4954d83e7: renamed the case from `it()` to the repository-standard `test()`.

frontend/tests/rsbuild-config.integration.test.ts
malinskibeniamin: Fixed in 4954d83e7: renamed the file to `rsbuild-config.integration.test.ts`, included that pattern in the integration project, and excluded it from the unit project. The real config contract passes under the integration config; the unit suite remains 57 files / 837 tests.

frontend/rsbuild.config.ts
malinskibeniamin: Fixed in 4954d83e7 with a measured threshold sweep. No cap produced 468 async JS files and a 10.06 MB largest async chunk. The selected 4 MiB cap produced 481 files (+13 / +2.8%) and 5.65 MB (-43.8%) with +0.052 MB total JS. A 512 KiB cap produced 595 files (+127) with no further largest-chunk reduction, so the PR now uses and documents 4 MiB.

malinskibeniamin:
## Review feedback addressed

- **Monaco worker safety:** verified all emitted workers are standalone child-compiler assets; production fetches, byte comparisons, syntax checks, loader-reference checks, and SQL-editor browser dogfood passed with no chunk/worker errors.
- **Removed GA experiments:** documented that Rspack 2.2 GA removed `lazyBarrel` and `fasterModuleConcatenation`; removed the brittle upstream-default assertion.
- **Production config contract:** removed the Module Federation mock, moved the contract into the integration suite, and adopted repository-standard `test()`.
- **Chunk threshold:** replaced the unsupported 512 KiB cap with a measured 4 MiB cap: largest async chunk 10.06 → 5.65 MB while adding 13 rather than 127 async files.
- **Dependency resolution:** GA core is exactly pinned; the lockfile has one `@rspack/core@2.2.0` and one matching `@rspack/binding@2.2.0`.
- **PR evidence:** refreshed the description with current impact numbers, commit/read order, tests, dependency risk, and whole-PR dogfood receipt.

All review threads resolved. CI is green, including Enterprise E2E and Snyk.

SpicyPete (approved):

PR #2616, opened 2026-08-25
Upgrade tanstack react form

jvorcak:


jvorcak:
I've clicked through the frontend after an upgrade and found no critical issues so far.

r-vasquez (approved):

PR #2617, opened 2026-08-25
Point kafkaConnect test image at Docker Hub instead of Cloudsmith

PrzemekZglinicki:
connectors-unsupported moved off Cloudsmith to public Docker Hub (redpandadata/connectors-unsupported) in a companion cloudv2 change. Repointed here too since backend/pkg/testutil/images.go and frontend/tests/shared/test-images.mjs both read this file to spin up a real Kafka Connect container for tests.

Verified end-to-end, not just edited: ran the actual KafkaConnectImage() loader against this file and confirmed it resolves to redpandadata/connectors-unsupported:latest.

### DIFF
--- a/test-images.json
+++ b/test-images.json
@@ -9,7 +9,7 @@
       "tag": "v0.0.0-20260315git94f368c"
     },
     "kafkaConnect": {
-      "repository": "docker.cloudsmith.io/redpanda/connectors-unsupported/connectors",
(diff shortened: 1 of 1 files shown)


weeco (approved):

weeco:
Thanks @PrzemekZglinicki for being proactive with this! We may have more references e.g. in the labs or docs where we have docker compose samples to run console, redpanda, kafka connect etc. together

PR #2621, opened 2026-08-28
frontend/rpcn: render field descriptions as prose, link each field to its docs

SpicyPete:
## What

Field help text in the RPCN config form now renders the schema's prose properly instead of dumping raw AsciiDoc under the control, and every field links to its own heading in the docs.

- Prefer `FieldSpec.short_description` (a markup-free one-liner) when the dataplane serves it; fall back to the AsciiDoc `description`, which is still the majority path.
- New `FieldDescription` component: converts AsciiDoc to Markdown, and collapses long or multi-paragraph prose behind "Show more" so it doesn't bury the input. A single-paragraph description renders inline, so its docs link trails the sentence instead of claiming a row.
- AsciiDoc conversion moved out of the command-palette utils into `rp-connect/utils/asciidoc.ts`, shared by the palette, the form and the template gallery.
- Template gallery slots prefer the short description too, and flatten the AsciiDoc fallback.

## Per-field docs links

`getFieldDocsUrl(section, connector, path)` deep-links a field's own heading on its connector reference page. The docs generator anchors each field heading with its dotted path, list markers dropped — `batching.byte_size` → `#batching-byte_size`, `sasl[].aws.credentials.role` → `#sasl-aws-credentials-role` — which is exactly the path the form already carries, so no new plumbing beyond a context holding the component's identity.

Checked against the live docs rather than assumed: over 90 randomly sampled cloud component pages (all HTTP 200), 1,770 leaf-field headings resolve, with **3 (0.17%)** anchored `-2` because a prose section on the same page claimed the plain id first. Those land on that same-named section, and an anchor that misses at all leaves the reader at the top of the right page — never worse than the connector-level link we'd otherwise ship. Field headings sit in a flat `Fields` section after the common/advanced tabs, so the anchor isn't buried in a collapsed panel. Verified end-to-end in localdev: all 13 links the form generated for `input: redpanda`, nested `tls.*` included, exist verbatim on the published page.

Object-group headers (`batching`, `sasl`) deliberately get no link — they carry no description, and they're where the `-2` collisions cluster (11 of 231).

## AsciiDoc rendering fixes

Found by reviewing the new converter against the bundled schema (`rp-connect-schema-full.json`: 4,650 non-deprecated field descriptions + 439 component prose blocks), and each fix re-verified over the whole corpus:

| Issue | Before | After |
| --- | --- | --- |
| `<<anchor, label>>` cross-references were never converted | 19 fields + 17 component blocks leaked `A database <\<drivers, driver>> to use.` | `A database driver to use.` — 0 left |
| Markdown pipe tables unparsed (react-markdown is mounted without remark-gfm) | `sql_raw.query` and 4 others rendered `\| Driver \| Style \| \|---\|---\|` as literal text | flattened to `- Driver — Placeholder Style`; 0 pipe rows left in 5,089 outputs |
| `format=dsv` tables split on the wrong separator | `- CHAR, VARCHAR:string` | `- CHAR, VARCHAR — string`, separator read from the table's own attribute line |
| Collapsed preview kept raw Markdown when a link label nested brackets | 8 `dsn` fields showed ``[`clickhouse://[user[:pass]@]…`](https://…)`` verbatim, URL included | 0 residual links |
| `cleanText` applied to multi-paragraph prose in template slots | 9 fields leaked `==== Drivers :driver-support: mysql=certified…` into slot help | full conversion + plain-text pass; 0 leaks |
| Slot help used `shortDescription` untrimmed | a whitespace-only value from the wire blanked the slot, while the config form treats it as absent | both consumers agree |

**Known gap, not fixed here:** the line-based block transforms aren't fence-aware, so a `----` separator or a line like `["a","b"]` inside a ``` fenced example would be dropped. Zero triggers across all 49 fenced descriptions in the current schema, so it's latent rather than live — worth a follow-up rather than a restructure in this PR.

## Notes for review

- `node-config-form.tsx` is an 18-line change: the field docs lookup rides on `ResourceFieldContext`, which already carries the edited component's identity, rather than adding a provider (a second wrapper re-indented the whole form body for no benefit).
- `cleanText` is gone. `markdownToPlainText(asciidocToMarkdown(x))` is byte-identical on all 235 component summaries — its only render site — and strictly better on descriptions, so there is now one way to flatten schema prose.
- `text-body-sm` is the `text-xs` rung and computes to 10.5px here, which is why the docs-link icon is `size-3` and positioned with `vertical-align` rather than a flex box — a flex box baselines on the icon's bottom edge and drops the word below the prose.

## Testing

`bun run type:check`, `bun run lint` (clean tree), and the full rp-connect suites: 139 unit + 772 integration passing. New coverage: internal cross-references in both conversion paths, pipe-table and dsv flattening, bracketed link labels in both the rendered link and the collapsed preview, the field-level anchor contract (including nested paths through the form), and the two template-slot fallbacks.

### DIFF
--- a/frontend/src/components/pages/rp-connect/pipeline/field-description.tsx
+++ b/frontend/src/components/pages/rp-connect/pipeline/field-description.tsx
@@ -17,11 +17,9 @@ import ReactMarkdown, { type Components } from 'react-markdown';
 import type { RawFieldSpec } from '../types/schema';
 import { asciidocToMarkdown, markdownToPlainText } from '../utils/asciidoc';
 
-// Descriptions longer than this (or spanning paragraphs) collapse behind "Show more"; schema prose
-// runs to a few thousand characters and buries the control it belongs to.
 const CLAMP_OVER_CHARS = 200;
 
-// Field prose sits under its control, so headings collapse to the same compact label as the body.
+// Field prose sits under its control: headings get no more weight than the body.
 const MarkdownHeading = ({ children }: { children?: React.ReactNode }) => (
   <div className="mt-1 font-medium text-body-sm text-foreground">{children}</div>
 );
@@ -51,8 +49,7 @@ const MARKDOWN_COMPONENTS: Components = {
   strong: ({ children }) => <strong className="font-medium text-foreground">{children}</strong>,
 };
 
-// Paragraphs unwrapped, so a single-paragraph description can flow inline with the docs link that
-// trails it. Everything else (code spans, links, emphasis) renders the same as in a block.
+// Paragraphs unwrapped, so a one-paragraph description can flow inline with its trailing docs link.
 const INLINE_MARKDOWN_COMPONENTS: Components = {
   ...MARKDOWN_COMPONENTS,
   p: ({ children }) => <>{children}</>,
@@ -65,13 +62,9 @@ const MarkdownBody = ({ markdown }: { markdown: string }) => (
 );
 
 /**
- * Trailing link to the field's own heading on the connector's docs page. Muted and named after the
- * field, so a form of twenty of these reads as help text rather than twenty calls to action.
- *
- * Laid out inline, not inline-flex: a flex box takes its baseline from the icon's bottom edge, which
- * drops the word below the baseline of the sentence it follows. Inline keeps one shared line box, so
- * `Docs` sits on the prose baseline and the icon is placed against it by `vertical-align`. The
- * underline is on the word alone — an ancestor's decoration would rule through the icon too.
+ * Link to the field's own heading on the connector's docs page. Inline, not inline-flex: a flex box
+ * baselines on the icon's bottom edge, which drops the word below the prose it follows. The underline
+ * is on the word alone so it doesn't rule through the icon.
  */
 const FieldDocsLink = ({ href, fieldName }: { href: string; fieldName?: string }) => (
   <Link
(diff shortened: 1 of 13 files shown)


frontend/src/components/pages/rp-connect/pipeline/field-description.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const docsLink = docsUrl ? <FieldDocsLink fieldName={spec.name} href={docsUrl} /> : null;
+  const short = spec.shortDescription?.trim();
+  if (short) {
+    return <InlineHelp docsLink={docsLink}>{short}</InlineHelp>;
+  }
+  const description = spec.description?.trim();
+  if (!description) {
+    return docsLink;
+  }
+  return <LongDescription docsLink={docsLink} source={description} />;
```
malinskibeniamin, line 154: HUMAN REVIEW: I think for docs link we should use a docs badge similar to how ADP does it, it's also a common pattern in the case of cloudflare which what we originally adopted/got inspired from, just sprinkling docs where relevant so it's interweaved with the user journey.

malinskibeniamin (approved):

frontend/src/components/pages/rp-connect/pipeline/field-description.tsx
SpicyPete: I'm not seeing docs in ADP, can you attach a screenshot as an example?
Right now at he end of the descriptions we are showing the docs link + a docs icon

PR #2626, opened 2026-09-02
[Chakra exit · Phase 0 · PR 1/14] frontend: delete unreachable Chakra code

SpicyPete:
## Chakra exit · Phase 0 (clear the ground) · PR 1 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Phase 0 removes code no route can reach before anything is migrated. Branch `chakra-exit/01-delete-unreachable`, stacked on #2625; each PR on the ladder shows only its own diff and re-targets `master` as its base merges.

**Pure deletion. No migration in this PR.** Every file below is unreachable from the app's real entry points, so it needs no port. Deleting it first makes every per-area count on the ladder a count of live code and takes ~3,400 lines out of the later, riskier PRs.

### How unreachability was established
Full import graph over `src/**` (relative + `paths` alias + `import()` + `require()` + scss `@import/@use`), seeded from every real entry: `src/index.tsx` (rsbuild's default entry → `bootstrap.tsx` → `app.tsx`), `embedded-app.tsx`, `inject-app.tsx` (module-federation expose), `config.ts`, `src/federation/**`, `src/routes/**` and `routeTree.gen.ts` — 79 entries, 833 reachable modules. Test files do not count as reachability. None of the deleted files is reached; a second pass confirmed nothing else becomes orphaned except the one SVG listed below.

### Deleted (33 files)
| Area | Files | Why dead |
|---|---|---|
| `components/form/` | 13 | TanStack-Form field set; its only consumer was the superseded `topics/create-topic-modal` |
| `topics/create-topic-modal.tsx`, `topics/CreateTopicModal/` | 3 | replaced by `create-topic-dialog.tsx` (live, tested) |
| `topics/PublishMessagesModal/` | 2 | headers editor, no importer |
| `admin/admin-users`, `-roles`, `-role-bindings` | 3 | no admin route exists |
| `rp-connect/secrets/secrets-list.tsx` + `.test.tsx` | 2 | only its own test imported it |
| `misc/error401-page`, `misc/resource-in-use-alert` | 2 | no importer (the live `resource-in-use-alert` is under `pages/secrets-store`) |
| `pages/url-test-page.tsx`, `utils/pagination.ts` | 2 | no importer |
| `KowlJsonView`, `tabs/Tabs`, `ConnectorBoxCard`, `CreateConnector` `.module.scss` | 4 | their components no longer import them |
[...]

PR #2627, opened 2026-09-02
[Chakra exit · Phase 1 · PR 2/14] frontend: move showToast onto the Registry toast, drop Chakra's toast

SpicyPete:
## Chakra exit · Phase 1 (wrappers) · PR 2 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Phase 1 re-implements the shared wrappers behind their existing exports, so page code does not change. Branch `chakra-exit/02-toast`, stacked on #2626; each PR on the ladder shows only its own diff and re-targets `master` as its base merges.

Chakra's toast is gone. `showToast` in `utils/toast.utils.tsx` is the one app-level entry point, re-implemented on the Registry `toast` (Base UI) that #2625 vendored; the 14 files that called Chakra's `createStandaloneToast` / `useToast` directly now go through it. Nothing else in those files changes.

### Why the Registry toast and not sonner
Revision 2 of the plan sent `showToast` to sonner first and left the Registry toast as a follow-up, because the Registry had no `toast.tsx` yet. #2625 vendors it, and its module-level manager is callable from outside React (the `backend-api.ts` stream error fires from a websocket loop, not a component), so there is no reason for a throwaway sonner stop. Sonner remains the second live toast until its 53 callers are re-pointed at `showToast` — that follow-up is mechanical and not Chakra work. Both now sit top-right.

### The façade (`utils/toast.utils.tsx`)
```ts
showToast({ id?, resourceName?, status?, title?, description?, duration? }): string
updateToast(id, { status?, title?, description?, duration? })   // only the given fields change
closeToast(id)
```
Contract carried over from the Chakra version: non-errors dismiss after 5 s, **errors stay until closed**, `duration: null` is sticky, and `id` + `resourceName` still compose the toast id. A status change on `updateToast` re-applies that status's default unless a duration is given. One thing that only works now: the id-based dedupe. The old code created a fresh standalone instance per call, so `toast.isActive` could never see an earlier toast; Base UI upserts on a duplicate id. `status: 'loading'` never auto-dismisses (Base UI), which is what every loading→update flow here already assumed.

### Where the toasts moved
| Module | Before | After |
|---|---|---|
| `state/backend-api.ts` | own standalone instance + `redpandaTheme` | `showToast` |
| `utils/tsx-utils.tsx` `StatusIndicator` | standalone instance, rendered `<ToastContainer />`, `toast.update` on every render | `showToast` once, `updateToast` after; `render()` returns null. The progress body is spans only — Base UI renders the description as a `<p>` |
| `utils/tsx-utils.tsx` `navigatorClipboardErrorHandler` | module toast | `showToast` |
| `misc/error-boundary.tsx` | `useToast` | `showToast` |
| reassign-partitions ×2 | two standalone instances with `duration: 2000` defaults, `toast.update` / `toast.close` on refs | `showToast` / `updateToast` / `closeToast`; the 2 s instance default is written onto each non-error call so nothing gets faster or slower |
| transforms ×2, rp-connect ×6, connect ×2 | standalone instances / `useToast` | `showToast`, same strings, same explicit durations |
| `app.tsx`, `embedded-app.tsx`, `federated-providers.tsx` | `toastOptions={redpandaToastOptions}` on `ChakraProvider` | prop removed (`ChakraProvider` itself stays until PR 13) |
| `app.tsx`, `embedded-app.tsx`, `federated-providers.tsx` | `ChakraProvider toastOptions` | + the Registry `<Toaster testId="console-toasts">` viewport, mounted at the provider layer (above the router and its error boundary) and placed top-right by a scoped `index.scss` block |
| `tests/mocks/redpanda-ui.ts` | stubbed `createStandaloneToast`, `redpandaToastOptions`, `ToastId` | stubs removed |

Behaviour that changed on purpose: error toasts in reassign-partitions used to vanish after 2 s (the instance default); they now stay until closed, matching every other error toast in the app.

### Reading the diff
`reassign-partitions.tsx` shows ~470 changed lines; `git diff -w` shows 48. Removing the `<><ToastContainer />…</>` wrapper de-indented the whole `render()` body.

### Gates
`bun run type:check` · `bun run lint` (no dirty tree; remaining lint findings on touched files are pre-existing on master) · unit suite (59 files / 872 tests, incl. 7 new façade tests) · integration suite (121 files / 1270 tests). E2E specs that assert toast copy use plain `getByText`, and every string is preserved verbatim; CI's Playwright run is the check for those.

### Follow-ups (not in this PR)
- Re-point the 53 sonner callers at `showToast`, then drop sonner's `<Toaster>` and the `toasterTheme` plumbing (PR 8 drops the theme plumbing anyway).
- Registry: a `position` prop on `Toaster`, so the top-right placement in `index.scss` can go.
- Registry: a rich-body slot on the toast so progress toasts don't have to stay phrasing content.

### Review pass (2026-09-03)
- **Toasts were hidden behind Chakra modals.** The Registry viewport is `z-50`; Chakra's `ModalOverlay`/`ModalContent` sit at 1300/1400, so anything toasted while a modal was open (remove-throttle, quick-add secret, throttle dialog) was unreadable. A scoped `index.scss` block lifts Console's viewport to 1500 until Chakra leaves.
- **The viewport moved above the router.** It was mounted inside `AppContent`, below the root `ErrorBoundary`, so the crash screen's own "copied" toast and anything on `/login` went nowhere. It now mounts next to `ChakraProvider` in `app.tsx`, `embedded-app.tsx` and `federated-providers.tsx`, where the removed `toastOptions` lived.
- **`updateToast` re-applies the new status's default** unless a `duration` is given. Before, a `loading → error` update inherited the loading toast's 5 s timer and the error vanished. Tests cover both directions.
- **Two toasts that changed behaviour by accident keep their old 5 s**: the per-message "Backend Error" from the websocket loop (which could otherwise pile up as sticky toasts) and the clipboard-failure toast.
- **`StatusIndicator`**: the progress body no longer overflows the description column, reads as body text rather than muted small text, carries `role="progressbar"` semantics, and if the user closes the progress toast it stays closed for that search instead of silently going stale.
- **Smaller**: `priority: 'high'` dropped (it made Base UI render error copy twice, which trips `getByText`); `closeToast` ignores a missing id (Base UI would close every toast); the all-errors reassignment update says "Reassignment failed" instead of keeping the loading text; a `closeToast` + `showToast` pair became one `updateToast`; redundant `duration: null` on error toasts and dead durations on loading toasts removed; `resourceName` without an `id` no longer yields `undefined_…`; `no-legacy.md` now points at `showToast`.
- **Position.** The registry viewport is bottom-anchored and `Toaster` has no position prop, so Console's toasts would have moved corners. The same scoped `index.scss` block anchors the viewport and the stack at the **top right**, where Chakra's and sonner's toasts are, with the registry's stacking math mirrored (vertical signs flipped). It is keyed on the viewport's `console-toasts` test id so it cannot reach a host app's toasts, and it goes once the registry grows a position prop.
- **Accepted, noted for the follow-ups**: the Registry toast is `select-none` with swipe-to-dismiss, so long error text cannot be copied out of it — a registry-level change (swipe-ignore on the description) is the right fix. A pre-existing stale-closure bug in the throttle dialog's "Remove throttle" path was spotted and left for its own change.

### DIFF
--- a/frontend/.claude/skills/code-standards/rules/no-legacy.md
+++ b/frontend/.claude/skills/code-standards/rules/no-legacy.md
@@ -38,7 +38,7 @@ import { useDisclosure, useToast } from "@chakra-ui/react";
 import { Button } from "components/redpanda-ui/components/button";
 import { Dialog } from "components/redpanda-ui/components/dialog";
 // Use useState instead of useDisclosure
-// Use toast from sonner instead of useToast
+// Use showToast from utils/toast.utils instead of useToast
 ```
 
 **MobX:**
--- a/frontend/src/app.tsx
+++ b/frontend/src/app.tsx
@@ -39,7 +39,7 @@ import './globals.css';
 import { Content } from '@builder.io/sdk-react';
 import { TransportProvider } from '@connectrpc/connect-query';
 import { createConnectTransport } from '@connectrpc/connect-web';
-import { ChakraProvider, redpandaToastOptions } from '@redpanda-data/ui';
+import { ChakraProvider } from '@redpanda-data/ui';
 import { QueryClientProvider } from '@tanstack/react-query';
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 import { createRouter, RouterProvider } from '@tanstack/react-router';
@@ -57,6 +57,7 @@ import { patchedRedpandaTheme as redpandaTheme } from 'utils/redpanda-theme';
 import { applyOverrides as applyDebugFeatureFlagOverrides } from './components/debug-helper/feature-flag-overrides';
 import { NotFoundPage } from './components/misc/not-found-page';
 import { RoutePendingFallback } from './components/misc/route-pending-fallback';
+import { Toaster as BaseUiToaster } from './components/redpanda-ui/components/toast';
 import { addBearerTokenInterceptor, checkExpiredLicenseInterceptor, getGrpcBasePath, setup } from './config';
 import { routeTree } from './routeTree.gen';
 import { installUISettingsSideEffects } from './state/ui';
@@ -129,7 +130,9 @@ const App = () => {
   return (
     <CustomFeatureFlagProvider initialFlags={window.__E2E_FEATURE_FLAGS__ ?? {}}>
       <Content apiKey={BUILDER_API_KEY} content={null} customComponents={builderCustomComponents} model={''} />
-      <ChakraProvider resetCSS={false} theme={redpandaTheme} toastOptions={redpandaToastOptions}>
+      <ChakraProvider resetCSS={false} theme={redpandaTheme}>
+        {/* showToast viewport, above the router so the error boundary and login can toast */}
+        <BaseUiToaster />
         <TransportProvider transport={dataplaneTransport}>
           <QueryClientProvider client={queryClient}>
             <RouterProvider router={router} />
--- a/frontend/src/app.tsx
+++ b/frontend/src/app.tsx
@@ -132,7 +132,7 @@ const App = () => {
       <Content apiKey={BUILDER_API_KEY} content={null} customComponents={builderCustomComponents} model={''} />
       <ChakraProvider resetCSS={false} theme={redpandaTheme}>
         {/* showToast viewport, above the router so the error boundary and login can toast */}
-        <BaseUiToaster />
+        <BaseUiToaster testId="console-toasts" />
         <TransportProvider transport={dataplaneTransport}>
           <QueryClientProvider client={queryClient}>
             <RouterProvider router={router} />
--- a/frontend/src/components/misc/error-boundary.tsx
+++ b/frontend/src/components/misc/error-boundary.tsx
@@ -9,7 +9,7 @@
  * by the Apache License, Version 2.0
  */
 
-import { Box, Button, Flex, Icon, useToast } from '@redpanda-data/ui';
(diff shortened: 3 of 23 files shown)


eblairmckee:
@claude can we call out what a reviewer can visually verify when running this locally to ensure no regressions were introduced in this migration effort? ie: what routes and actions to take to ensure impacted critical user journeys aren't broken

SpicyPete:
Thanks for the initial feedback, moved the new Toast to the top right to match the previous toast + the ADP placement

eblairmckee:
Tested a bunch of toast instances, only found one that appears to be base-ui (confirmed it was dismissable). 
Screenshot 2026-09-04 at 8 41 37 AM

Confirmed sonner toast stacks. 
Screenshot 2026-09-04 at 8 48 05 AM


Sonner
Screenshot 2026-09-04 at 8 41 10 AM
Screenshot 2026-09-04 at 8 41 51 AM
Screenshot 2026-09-04 at 8 42 24 AM
Screenshot 2026-09-04 at 8 42 35 AM

eblairmckee:
was able to find one instance of chakra toast being replaced by base-ui toast. although you have to watch in slo-motion since it appears we aren't staggering the show/hide calls with enough delay in the middle.
https://github.com/user-attachments/assets/40dc047e-6f67-4591-bad9-12b630853650

eblairmckee (approved):

PR #2628, opened 2026-09-02
[Chakra exit · Phase 1 · PR 3/14] frontend: Registry internals for the shared display wrappers, drop Chakra type seams

SpicyPete:
## Chakra exit · Phase 1 (wrappers) · PR 3 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Phase 1 re-implements the shared wrappers behind their existing exports, so page code does not change. Branch `chakra-exit/03-display-primitives`, stacked on #2627; each PR on the ladder shows only its own diff and re-targets `master` as its base merges.

Five shared wrappers keep their exports and get Registry internals, so their importers do not change. `utils/tsx-utils.tsx` no longer imports `@redpanda-data/ui` at all. The one type seam in `state/ui.ts` moves to TanStack v9, and the `state/ui.ts → pages/connect/overview` import is gone.

### Wrappers (same export, new internals)
| Module | Was | Now | Importers touched |
|---|---|---|---|
| `tsx-utils` `DefaultSkeleton` | Chakra `Skeleton height={4} noOfLines={8}` | Registry `SkeletonText lines={8} width="full"`, same `motion.div` + 2 rem margin | 0 (27 importers) |
| `tsx-utils` `InfoText` | Chakra `Tooltip hasArrow label placement` | Registry `Tooltip` / `TooltipTrigger render` / `TooltipContent side`; `placement` narrows to `top\|bottom\|left\|right` (no caller passed one) | 0 |
| `tsx-utils` `OptionGroup` | Chakra `RadioGroup options=` (an attached `RadioCard` segment group) | Registry `ToggleGroup` single-select, segmented like the original | 0 |
| `misc/section` `Section` | Chakra `Section` (block `<section>`, white, border, `rounded-lg`, shadow, `p-6`) forwarding all `ChakraProps` | Registry `Card size="full" variant="elevated"` with `gap-0 px-6`; props are now `id`, `className`, `style` | **7 files, 17 call sites** — every `py=`, `my=`, `minWidth=`, `gridArea=`, `overflowY=`, `borderColor=` became a class or an inline `gridArea`; `flexDirection="column"` on a block Box was a no-op and is dropped |
| `misc/tabs` `Tabs` | Chakra `Tabs items=` (index-based) | Registry `Tabs` / `TabsList variant="underline"` / `TabsTrigger` / `TabsContent`, value-based; `isFitted` → `layout="full"`; inactive panels stay unmounted as before | 0 (5 importers); its 6 tests pass unchanged |
| `misc/statistic` `Statistic` | Chakra `Stat` + `StatNumber`/`StatLabel`, `SpaceProps` passthrough | Registry `Stat size="lg" label value`; props are `title`, `value`, `className`. No caller used `hint` or a space prop | 0 (4 importers, 24 sites) |
| `misc/small-stat` `SmallStat` | Chakra `Flex` + `Text` | a flex `div` + `span`; nothing here is a Stat | 0 |
| `.codeBox` (what `tsx-utils` `Code` and nine raw call sites render) | hardcoded light grey background, no text colour — unreadable in the dark theme | `--color-surface-subtle` / `--color-border` tokens, so it reads in both themes | 0 |

### Type seams
- `state/ui.ts`: `SortingState` now comes from `@tanstack/react-table` (v9). The shape is identical to the v8 one Chakra re-exported (`{ id: string; desc: boolean }[]`), and it was the last file importing it from `@redpanda-data/ui`.
- `ConnectTabKeys` is defined in `state/ui.ts` and no longer imported from a page component. `TopicTabId` still comes from `topics/topic-details` as an `import type` — erased at build, and that file belongs to the #2588 stream, so it moves when that lands.
- `utils/legacy-data-table.ts` is deliberately untouched: it derives column/row types from the legacy `DataTable` *because* the app's v9 `ColumnDef` is the wrong shape for that component. It deletes with the last legacy table (its own comment says so).

### What a reviewer should look at
The two hunks with judgement in them are `misc/section.tsx` (a `<section>` became a `Card` kept in block flow with a grid-area-only `style`) and `misc/tabs/tabs.tsx` (uncontrolled Registry Tabs seeded from `selectedTabKey`, which the old wrapper also read only once). Visually: `Statistic` now renders the Registry `Stat`, so the stat strips on Overview, Broker details, Kafka Connect cards and Reassign partitions switch from value-above-label to label-above-value — the same `Stat` the consumer-group page already uses. The old `status-bar` colour class on Overview's cluster-status stat is passed through, but its CSS was scoped under a dead `.ant-statistic` selector and has not painted for some time.

### Effect
Files importing `@redpanda-data/ui`: **95 → 89**.

### Gates
`bun run type:check` · `bun run lint` (no dirty tree; every remaining lint finding on touched files is pre-existing on master) · unit suite · integration suite (121 files / 1270 tests), including `misc/tabs/tabs.test.tsx`.

### Review pass (2026-09-03)
- **`Section` keeps block flow.** Card is a flex column, which stretched the "Create …" buttons in the security tabs to full width and stopped margins collapsing; `block` restores the old layout. Its `style` prop is typed to `gridArea` only, and Overview's `<hr>` keeps the margins the Chakra Section used to inject. The Card is a `div`, not a `<section>` — the registry Card has no `render` prop; nothing selects on the element.
- **`Tabs` is uncontrolled** (`defaultValue`), forwards `onValueChange`, and passes `activateOnFocus` so arrow keys still select a tab as Chakra's did. Function-valued content is rendered through a child component, so it runs only for the active panel. Panels keep Chakra's 1rem padding.
- **`Statistic` is a plain `Stat`**; spacing moved to the rows that lay stats out (`Flex gap={8}` / `flex gap-8`) instead of a margin baked into each stat.
- **`OptionGroup`** nests each radio in the registry `Label` (implicit association), replacing hand-built ids that could collide.
- `InfoText` builds one tooltip tree; `LogsTab` loses a `variant` no caller passed; the unit stub drops two dead type exports.

### Review pass 2 (2026-09-04, pre-merge)
Parity checks against the Chakra versions, applied where the old behaviour was silently lost:
- **`OptionGroup` is a segmented control again** (registry `ToggleGroup`), which is what Chakra's attached `RadioCard` group looked like on Broker details and the message preview settings; the radio-circle rendering was an undisclosed change.
- **`Statistic` renders `Stat size="lg"`**, so headline numbers keep their size and weight instead of dropping to body text.
- **`Tabs` forwards `onChange` only for user activation.** Base UI also reports its own initial or fallback selection, which the Chakra wrapper never surfaced.
- `.codeBox` paints with `surface-recess` and `border-subtle`, the tokens the theme documents for a code block; its dead `.noPadding` tint is gone. Overview drops a `gridArea` whose parent was a flex column.

Looked at and left alone: the legacy `Tabs`/`InfoText` prop surface (unused props are harmless on wrappers slated for removal), `ui.ts`'s dead `connectorsDetails`/`pipelinesDetails` settings (pre-existing, not Chakra), Overview's inert `status-bar` colour class (a `tone` on the cluster-status stat would be a feature for the Overview PR), and `tsx-utils` pulling `ToggleGroup` into the startup chunk (about 3 KB gzipped; splitting `OptionGroup` out is a follow-up).

### DIFF
--- a/frontend/src/components/misc/small-stat.tsx
+++ b/frontend/src/components/misc/small-stat.tsx
@@ -1,11 +1,10 @@
-import { Flex, Text } from '@redpanda-data/ui';
 import type { JSX } from 'react';
 
 export function SmallStat(p: { title: JSX.Element | string; children: JSX.Element | number | string }) {
   return (
-    <Flex color="var(--color-foreground)" fontFamily="Inter" fontWeight="400" gap="2">
-      <Text fontWeight="500">{p.title}: </Text>
+    <div className="flex gap-2 text-foreground">
+      <span className="font-medium">{p.title}: </span>
       {p.children}
-    </Flex>
+    </div>
   );
 }
--- a/frontend/src/components/misc/statistic.tsx
+++ b/frontend/src/components/misc/statistic.tsx
@@ -1,32 +1,7 @@
-import { Box, Flex, type SpaceProps, Stat, StatLabel, StatNumber, Tooltip } from '@redpanda-data/ui';
-import { InfoIcon } from 'components/icons';
+import { Stat } from 'components/redpanda-ui/components/stat';
+import type { ReactNode } from 'react';
 
-export function Statistic(
-  p: {
-    key?: string | number;
-    title: React.ReactNode;
-    value: React.ReactNode;
-    hint?: string;
-    className?: string;
-  } & SpaceProps
-) {
-  const { key, title, value, className, hint, ...rest } = p;
-
-  return (
-    <Stat className={className} flexBasis="auto" flexGrow={0} key={key} marginRight="2rem" {...rest}>
-      <StatNumber>{value}</StatNumber>
-      <StatLabel>
-        <Flex gap={1}>
-          {title}
-          {Boolean(hint) && (
-            <Tooltip hasArrow label={hint} placement="right">
-              <Box alignSelf="start">
-                <InfoIcon size={18} />
-              </Box>
-            </Tooltip>
-          )}
-        </Flex>
-      </StatLabel>
-    </Stat>
-  );
+// Legacy stat; the row that lays these out owns the spacing.
+export function Statistic(p: { title: string; value: ReactNode; className?: string }) {
+  return <Stat className={p.className} label={p.title} value={p.value} />;
 }
--- a/frontend/src/components/misc/statistic.tsx
+++ b/frontend/src/components/misc/statistic.tsx
@@ -3,5 +3,5 @@ import type { ReactNode } from 'react';
 
 // Legacy stat; the row that lays these out owns the spacing.
 export function Statistic(p: { title: string; value: ReactNode; className?: string }) {
-  return <Stat className={p.className} label={p.title} value={p.value} />;
+  return <Stat className={p.className} label={p.title} size="lg" value={p.value} />;
 }
--- a/frontend/src/components/pages/overview/broker-details.tsx
+++ b/frontend/src/components/pages/overview/broker-details.tsx
@@ -74,8 +74,8 @@ const BrokerDetailsContent: FC<{ brokerId: number }> = ({ brokerId }) => {
 
   return (
     <PageContent>
-      <Section py={4}>
-        <Flex>
+      <Section className="py-4">
+        <Flex gap={8}>
           <Statistic title="Broker ID" value={brokerId} />
           <Statistic title="Role" value={broker.isController ? 'Controller' : 'Follower'} />
           {/* biome-ignore lint/style/noNonNullAssertion: not touching MobX observables */}
@@ -84,7 +84,7 @@ const BrokerDetailsContent: FC<{ brokerId: number }> = ({ brokerId }) => {
           {Boolean(broker.rack) && <Statistic title="Rack" value={broker.rack} />}
         </Flex>
       </Section>
-      <Section py={4}>
+      <Section className="py-4">
         <BrokerConfigView entries={brokerConfigs} />
       </Section>
     </PageContent>
--- a/frontend/src/components/pages/overview/overview.tsx
+++ b/frontend/src/components/pages/overview/overview.tsx
@@ -139,20 +139,22 @@ class Overview extends PageComponent {
         </NullFallbackBoundary>
 
         <PageContent className="overviewGrid">
-          <Section my={4} py={5}>
-            <Flex>
-              <Statistic
-                className={`status-bar ${clusterStatus.className}`}
-                title="Cluster Status"
-                value={clusterStatus.displayText}
-              />
-              <Statistic title="Cluster Storage Size" value={brokerSize} />
-              <Statistic title="Cluster Version" value={version} />
-              <Statistic title="Brokers Online" value={brokersOnlineText} />
-              <Statistic title="Topics" value={overview.kafka?.topicsCount ?? NOT_AVAILABLE} />
-              <Statistic title="Replicas" value={overview.kafka?.replicasCount ?? NOT_AVAILABLE} />
-            </Flex>
-          </Section>
+          <div className="my-4">
+            <Section className="py-5">
+              <Flex gap={8}>
+                <Statistic
+                  className={`status-bar ${clusterStatus.className}`}
+                  title="Cluster Status"
+                  value={clusterStatus.displayText}
+                />
+                <Statistic title="Cluster Storage Size" value={brokerSize} />
+                <Statistic title="Cluster Version" value={version} />
+                <Statistic title="Brokers Online" value={brokersOnlineText} />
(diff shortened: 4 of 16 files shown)


frontend/src/components/misc/small-stat.tsx
```diff
@@ -1,11 +1,10 @@
...
   return (
-    <Flex color="var(--color-foreground)" fontFamily="Inter" fontWeight="400" gap="2">
-      <Text fontWeight="500">{p.title}: </Text>
+    <div className="flex gap-2 text-foreground">
+      <span className="font-medium">{p.title}: </span>
       {p.children}
-    </Flex>
+    </div>
   );
 }
```
eblairmckee, line 10: why not use the registry Stat?

eblairmckee: there is a small size https://ui-playground.netlify.app/stat

frontend/src/components/pages/overview/broker-details.tsx
```diff
@@ -74,8 +74,8 @@ const BrokerDetailsContent: FC<{ brokerId: number }> = ({ brokerId }) => {
 
   return (
     <PageContent>
-      <Section py={4}>
-        <Flex>
+      <Section className="py-4">
+        <Flex gap={8}>
```
eblairmckee, line 78: I thought we got rid of Flex since that was chakra, preferring plain divs + tailwind instead

frontend/src/components/pages/overview/overview.tsx
```diff
@@ -139,34 +139,36 @@ class Overview extends PageComponent {
...
+                <Statistic
+                  className={`status-bar ${clusterStatus.className}`}
+                  title="Cluster Status"
+                  value={clusterStatus.displayText}
+                />
+                <Statistic title="Cluster Storage Size" value={brokerSize} />
+                <Statistic title="Cluster Version" value={version} />
+                <Statistic title="Brokers Online" value={brokersOnlineText} />
+                <Statistic title="Topics" value={overview.kafka?.topicsCount ?? NOT_AVAILABLE} />
+                <Statistic title="Replicas" value={overview.kafka?.replicasCount ?? NOT_AVAILABLE} />
```
eblairmckee, line 154: i feel like we shouldn't have these thing wrappers of registry components... ie: replace Statistic with Stat, Section with Card. otherwise we're just perpetuating more tech debt we'll have to clean up in the future.

frontend/src/components/pages/rp-connect/pipelines-details.tsx
```diff
@@ -374,7 +374,7 @@ export const LogsTab = ({ pipeline, variant = 'card' }: { pipeline: Pipeline; va
     <>
       <Box my="1rem">The logs below are for the last five hours.</Box>
 
-      <Section borderColor={variant === 'ghost' ? 'transparent' : undefined} minWidth="800px" overflowY="auto">
```
eblairmckee, line 377: iiutc, the transparent border override is to prevent the global selector that adds a border to every div. I know we're removing that in an upcoming PR, but wanted to call it out in case this was missed

frontend/src/components/pages/overview/broker-details.tsx
SpicyPete: This PR is just to replace the components usage of Chakra, not at the page level.
Page level is coming in the Phase 2 work

frontend/src/components/misc/small-stat.tsx
SpicyPete: This PR is to remove the usage of Chakra within internal components + change no visuals ideally.  The Phase 2 work will be swapping to actual Registry components

frontend/src/components/pages/rp-connect/pipelines-details.tsx
SpicyPete: On the page I am not noticing any new borders being shown

frontend/src/components/pages/overview/overview.tsx
SpicyPete: Agreed, but for the migration plan, changing the UX/display is a separate effort from removing Chakra, so we should do this during the page migration, or as as separate ticket

frontend/src/components/misc/statistic.tsx
```diff
@@ -1,32 +1,7 @@
...
-              </Box>
-            </Tooltip>
-          )}
-        </Flex>
-      </StatLabel>
-    </Stat>
-  );
+// Legacy stat; the row that lays these out owns the spacing.
+export function Statistic(p: { title: string; value: ReactNode; className?: string }) {
+  return <Stat className={p.className} label={p.title} size="lg" value={p.value} />;
```
jvorcak, line 6: Does this wrapper exist because of a className? Should we make our UI registry components to accept className by default to avoid such wrappers?

jvorcak (approved):

PR #2629, opened 2026-09-02
[Chakra exit · Phase 1 · PR 4/14] frontend: SingleSelect and the error/not-found surfaces on Registry components, plus two guard tests

SpicyPete:
## Chakra exit · Phase 1 (wrappers) · PR 4 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Phase 1 re-implements the shared wrappers behind their existing exports, so page code does not change. Last PR of the phase. Branch `chakra-exit/04-interactive-surfaces`, stacked on #2628; each PR on the ladder shows only its own diff and re-targets `master` as its base merges.

The last Phase 1 PR. `SingleSelect` keeps its export and moves onto the Registry Select (and the Registry Combobox for its one `creatable` caller); the error, not-found and modal-container surfaces move onto Tailwind and Registry Dialog / Empty / Button. Two app-owned guard tests pin registry contracts the registry itself ships untested.

### `misc/select.tsx` — the hunk that carries judgement
`SingleSelect<T>` keeps `options` / `value` / `onChange` / `isDisabled` / `placeholder` / `creatable` / `data-testid`, so its 9 call sites in 7 files do not change — except `login.tsx`, which passed a `chakraStyles` object that returned `provided` unchanged; `type:check` flagged it and it is gone.

- **Values round-trip through a string key.** The Registry Select keys on strings; callers pass enums (`SASLMechanism`, `SCRAMAuth_Mechanism`), numbers (schema versions) and string unions. `toKey(value)` maps each option to `String(value)` and back, so `onChange` still receives the typed value. A `value` that matches no option (create-connector's `undefined` cluster) renders the placeholder instead of an empty label.
- **`creatable` goes to the Registry Combobox**, which is the searchable/creatable one; the plain Select has no text input. Only `preview-settings.tsx` uses it, with string values. `clearable={false}` keeps the old "no clear button" behaviour. Because `onChange` fires with the typed text on create, `T` must be `string` there — noted on the prop.
- **`data-testid` still wraps the control in a `div`**, as before. The e2e page object clicks that wrapper and then an option by exact text, which the Registry Select satisfies; the only live spec assertions on these three ids are `toBeVisible()`.
- `tests/mock-react-select.tsx` and its two setup-file imports are gone. Nothing rendered a react-select under test any more, so no test changes.

### Error and not-found surfaces (exports unchanged)
| Module | Was | Now |
|---|---|---|
| `misc/error-result` | `HStack`/`Stack`/`Heading`/`Text`, Chakra `CodeBlock`, Chakra `Avatars` asset | flex `div`s, `text-heading-lg` / `text-body-lg`, Registry `SimpleCodeBlock`, the local `ErrorBananaSlip.svg` the not-found page already uses |
| `misc/error-display` | Chakra `Result` + `Button`, Chakra `Section`, `List`/`ListItem`/`ListIcon` | Registry `Empty` (title, description, Retry `Button`), the app `Section` wrapper, a plain `ul` with the existing `WarningIcon` |
| `misc/error-modal` | Chakra `Modal` with `onCloseComplete` driving the modal queue | Registry `Dialog size="lg"`; Base UI's `onOpenChangeComplete(false)` is the same hook, so the queue's `afterClose` still fires after the exit transition. The Dialog adds its close button; it routes through `onClose` like OK does |
| `misc/error-boundary` | `Box`/`Flex`/`Icon`, Chakra `Button size="large" isLoading` | `div`s, Registry `Button size="lg"` with `icon` and `isLoading` |
| `misc/not-found-page`, `misc/not-found-content` | `Center`/`Stack`/`Image`/`Heading as="h1"` | flex `div`s, `img`, `h1.text-heading-xl` — the `h1` keeps the test's `getByRole('heading', { level: 1 })` contract; the test's hand-written Chakra mock is deleted |
| `utils/modal-container` | `Box` | `div` |

### Guard tests (new, app-owned, in `src/__tests__/` beside the existing consumer-side guard)
- **Expanded rows survive a data swap** (`data-table-expanded-rows.test.tsx`): expands a row, replaces the `data` array, flushes TanStack's scheduled reset, and expects the sub-row to stay. Verified negative: with `tableOptions={{ autoResetExpanded: true }}` the same test fails.
- **Alert paints one icon** (`alert-default-icon.test.tsx`): default icon, custom `icon` replacing it, `icon={null}` painting none. The Chakra pattern (an icon child) paints two, which is the case every `<AlertIcon />` swap in Phase 2 must avoid; 11 files still pass `<AlertIcon />` to Chakra's `Alert` and are listed in the plan.

### Effect
Files importing `@redpanda-data/ui`: **89 → 81**. `misc/` is down to the chrome leftovers PR 8 owns.

### Gates
`bun run type:check` · `bun run lint` (no dirty tree; remaining findings on touched files are pre-existing on master) · unit suite · integration suite (123 files / 1274 tests), including `not-found-page.test.tsx` and the two guards. To exercise by hand: the debug dialog's "Throw in render (ErrorBoundary)" button, and any failing backend call for `ErrorDisplay` / `showErrorModal`.

### Review pass (2026-09-03)
- **Base UI popups inside Chakra modals.** `SingleSelect` is the first Registry popup rendered inside a Chakra `Modal` (column settings, preview fields). Chakra's scroll lock swallows wheel events outside the modal and its focus lock pulls focus back, so both modals now provide a `PortalContainer` and the popups render inside them. Noted for the #2588 stream, which owns those files.
- **`SingleSelect`**: an unset value selects nothing instead of colliding with an empty-string option; the Combobox's clear signals (re-select, Escape, blur) are ignored so a preview-tag pattern can no longer be emptied by accident; `id` lands on the trigger and the login form's "SASL Mechanism" label points at it again; options are mapped once.
- **Behaviour that differs from react-select and is accepted**: the plain Select has no type-to-filter, and typed-but-uncommitted Combobox text stays visible after blur.
- Error boundary icons lead their labels again; the error dialog is `size="xl"` to keep its old width; the Alert guard drops the test that asserted the footgun itself, keeping the three contract tests.

### Review pass 2 (2026-09-04, pre-merge)
- **A value outside `options` still shows as text.** The first cut forced the Registry Select to `null` when the value wasn't in the list, which blanked a connector property whose stored value isn't among the suggestions (custom converter class, a region outside the curated list) and a schema version mid-navigation. The value now passes through and Base UI renders it, as react-select's fallback did.
- **The creatable field clears coherently.** Swallowing the Combobox's clear signals left the input blank while the pattern still applied. Clearing the field now clears the pattern (and the preview column with it), so what is shown is what applies; the clear button is back for the same reason. `SingleSelect` builds one option array for both Registry controls and labels are strings, which every caller passes.
- **Parity on the error and not-found surfaces**: headings keep their weight (`font-semibold`) and the error-result title stays at the 24 px rung; error-result keeps a text fallback when its illustration fails to load and the detail code block keeps the old geometry (no outer margin, no height cap); the error dialog sizes its icon directly instead of a 70 px wrapper; identical backend errors collapse so list keys are unique.
- The two registry guard tests moved to `src/__tests__/` beside the existing consumer-side guard, and drain TanStack's microtask reset with `act()` instead of a timer.

Accepted and documented: the Registry Select commits a value on a single keystroke while the closed trigger has focus (standard listbox behaviour, unlike react-select's type-to-filter); uncommitted Combobox text stays visible after blur; `id` reaches the Select trigger only (the Combobox exposes no input id). Known, latent: `showErrorModal` now renders at the Registry dialog's z-index, below a Chakra modal's overlay if one is open at that moment; the only live path is a sub-second window on the reassign page, and every Chakra modal host leaves in Phase 2 or 3. Cleanups left for later: a shared not-found hero (three near-identical shells), an `Alert`-based error list, and `CopyButton` for the crash screen.


frontend/src/components/misc/error-boundary.tsx
```diff
@@ -209,7 +209,7 @@ export class ErrorBoundary extends React.Component<{ children?: React.ReactNode
     }
 
     return (
-      <Box style={{ minHeight: '100vh', overflow: 'visible', padding: '2rem 4rem' }}>
+      <div style={{ minHeight: '100vh', overflow: 'visible', padding: '2rem 4rem' }}>
```
jvorcak, line 212: Since we're touching this, can we move the CSS to tailwind here?

jvorcak (approved):

PR #2633, opened 2026-09-05
[Chakra exit · Phase 2 · PR 5/14] frontend: transforms pages on Registry components

SpicyPete:
## Chakra exit · Phase 2 (route by route) · PR 5 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Phase 2 walks the routes one directory at a time. Branch `chakra-exit/05-transforms`, cut from #2629; re-targets `master` once that merges. Every Phase 2 branch is cut from the same base and touches a disjoint file set, so they merge in any order.

Four transforms files off Chakra, plus the shared search input, its test, a partition-sort test and a smoke spec. No API changes; behaviour follows the Chakra table contract (see the second round below).

### What moved
- **transforms-list** — Chakra `DataTable` → Registry `DataTable`, with `DataTableColumnHeader` on the two sortable columns; `SearchField` → the shared `misc/search-input` (`/` to focus, Escape to clear, same `search-field-*` test ids).
- **transform-details** — same table swap; drops the `LegacyColumnDef` import and the controlled `SortingState` (nothing else read it). The URL-derived pagination hook only fed an `initialState` TanStack reads once, so it is gone; log rows are keyed by partition+offset instead.
- **transforms-setup** — the Registry `Alert` paints its own icon, so `AlertIcon` goes; code blocks keep the terminal look: `SyncCodeBlock` for the `bash` command (highlighted synchronously, no skeleton flash) and `DynamicCodeBlock` for the plain directory listing.
- **modals** — Chakra `Modal` → Registry `Dialog`, opened through `utils/modal-container`.

### Coverage
`transforms.spec.ts` turned out to be **commented out end to end** — it needs transforms seeded through the backend API, which the UI cannot do. So transforms was an eighth uncovered route, not a covered one. The new spec covers what needs no seeding: the setup page, which is static and reachable in every variant.

### Worth a look
Two traps that recur across Phase 2, both fixed here in a follow-up commit after review:
- **Chakra's `DataTable` injected an expander column for `subComponent`; the Registry's renders no affordance at all** — the expanded log messages were unreachable.
- **Chakra painted the sort affordance on every header itself.** The Registry sorts only through `DataTableColumnHeader`, so `sorting` stayed enabled with no UI. `DataTableColumnHeader` also offers a "Hide" item, and these pages have no `DataTableViewOptions` to bring a column back — hence `enableHiding: false`.

### Effect
Files importing `@redpanda-data/ui`: **81 → 77**.

### Gates
`type:check` · `lint` · `lint:check` per changed file · unit (880) · integration (1304). Note `bun run lint` is `ultracite fix` and exits 0 with errors still present — `lint:check` is the gate.

- Tables paginate like the legacy DataTable again: 50 a page, pager only past that (the Registry defaults to 10 with a permanent pager; the partition table had lost paging entirely).
- Logs table: the Registry ignores column `size`, so the Value column had collapsed; it takes the slack again and timestamps stay on one line.
- Both search boxes are the shared `misc/search-input` (identical copy on #2636 and #2638): clear refocuses the field, the empty right edge is click-through, `/` focuses, Escape clears then blurs.
- Dialog title clears the absolute close button; the logs table shows no pager under its loading or empty row; the partition-sort test resets its store inside `act()`. The `?page`/`?pageSize` deep-link initial value was dropped deliberately (TanStack read it once at mount, with an empty list).

### Final review (2026-09-08)
- The bash snippet is back on `SyncCodeBlock` (terminal look, synchronous highlighting); the logs table shows no pager under its loading or empty row; `SearchInput` advertises its shortcuts (`aria-keyshortcuts`, a title on the clear button).

### DIFF
--- a/frontend/src/components/pages/transforms/transform-details.tsx
+++ b/frontend/src/components/pages/transforms/transform-details.tsx
@@ -9,11 +9,16 @@
  * by the Apache License, Version 2.0
  */
 
+import { ChevronDownIcon, ChevronRightIcon, CloseIcon } from 'components/icons';
 import { Button } from 'components/redpanda-ui/components/button';
-import { DataTable, type DataTableColumnDef } from 'components/redpanda-ui/components/data-table';
+import {
+  DataTable,
+  type DataTableColumnDef,
+  DataTableColumnHeader,
+} from 'components/redpanda-ui/components/data-table';
 import { Input, InputEnd, InputStart } from 'components/redpanda-ui/components/input';
-import { ChevronDown, ChevronRight, SearchIcon, XIcon } from 'lucide-react';
-import { Fragment, useEffect, useRef, useState } from 'react';
+import { SearchIcon } from 'lucide-react';
+import { Fragment, useEffect, useMemo, useRef, useState } from 'react';
 import { showToast } from 'utils/toast.utils';
 
 import { openDeleteModal } from './modals';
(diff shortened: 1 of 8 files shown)


frontend/src/components/pages/transforms/transform-details.tsx
```diff
@@ -162,21 +178,15 @@ const OverviewTab = (p: { transform: TransformMetadata }) => {
...
-            {
-              header: 'Status',
-              cell: ({ row: { original: r } }) => <PartitionStatus status={r.status} />,
-            },
-            { header: 'Lag', accessorKey: 'lag' },
-          ]}
+          columns={partitionStatusColumns}
           data={p.transform.statuses}
+          pagination={false}
+          sorting={false}
```
malinskibeniamin: [P2] Preserve sorting on the partition status table

The previous `@redpanda-data/ui@4.2.0` table defaults `sorting` to `true` and attaches sorting handlers to accessor-column headers. Setting this to `false` removes the existing Partition, Node, and Lag sorting, so users can no longer order partitions by lag to identify which are behind. Keep sorting enabled and render `DataTableColumnHeader` for those columns, with hiding disabled.

Verify: open a transform with different partition lag values and sort Lag ascending and descending.

malinskibeniamin: Fixed in f650d82798ee9291d8cf09aa7a2d99ea9bebca75. Restored Partition, Node, and Lag sort headers with column hiding disabled.

The Lag ascending/descending regression test passes after failing on the original head. Three pre-existing lint issues in unrelated LogsTab code remain unchanged. Source and focused test type checks pass. Chromium public-component consumer replay passes with seeded API stores and zero page errors; no live-cluster mutation was performed.

SpicyPete:
Enterprise is failing on all branches, seems to be unrelated and about a test container not starting in CI
`message: 'container 1e1cc5daa34baec0ad442ebd331d77e19727358db334df1df4882086af8a318b is not running'`

eblairmckee:
visually verified migration by route according to the [review guide](https://claude.ai/code/artifact/b2319202-c8c1-4947-8828-553314e9a7b0)
Screenshot 2026-09-08 at 1 41 33 PM

## transforms-setup
reviewed transform pages in cloud (top is local, bottom is production)
Screenshot 2026-09-08 at 1 26 28 PM
and standalone console
Screenshot 2026-09-08 at 1 26 34 PM

## transforms
Screenshot 2026-09-08 at 1 44 26 PM

### delete dialog
Screenshot 2026-09-08 at 1 46 23 PM


## transforms/$name
Screenshot 2026-09-08 at 1 44 58 PM
Screenshot 2026-09-08 at 1 45 06 PM

eblairmckee (approved):

PR #2634, opened 2026-09-05
[Chakra exit · Phase 2 · PR 6/14] frontend: admin, debug bundle and license pages on Registry components

SpicyPete:
## Chakra exit · Phase 2 (route by route) · PR 6 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Branch `chakra-exit/06-admin-license`, cut from #2629; re-targets `master` once that merges.

Nine files: the debug-bundle pages, the license notifications, the register modal and the upload/expired license pages. No API changes.

### What moved
Mostly Chakra layout props to Tailwind classes and Chakra `Alert` to the Registry one (which paints its own icon, so `AlertIcon` goes and the tone picks the glyph). Chakra's `FormField` becomes `Field` + `FieldLabel` + `FieldDescription` + `FieldError`; `Select isMulti` → `SimpleMultiSelect`; `PasswordInput` → `Input type="password"`, which already has the eye toggle; `ConfirmModal` → `AlertDialog`.

### Worth a look
- **Navigation dressed as a button stays an anchor.** The Registry `Button` forces `role="button"` even through `as={Link}`, which loses link semantics and breaks `getByRole('link')` — `debug-bundle-page.ts` reads "Bundle generation in progress" that way and `license-utils.test.tsx` reads both license CTAs as links. All three use `<Link className={buttonVariants(...)}>`, the shape `not-found-page` already uses.
- **Every control gets an `id`**, so `getByLabel(/cpu profiler/i)` and friends keep resolving. `SimpleMultiSelect` forwards neither `aria-label` nor `aria-labelledby` and its trigger hardcodes one, so the two labelled multi-selects compose the parts through a local `LabelledMultiSelect`.
- **Chakra's `Result` rendered its title as an `h1`** and `license.spec.ts` asserts on that, so the `h1` stays and the subtitle becomes body text — which is what it always was.
- The dev-panda SVG only existed as a data URI inside the Chakra bundle, so it is vendored to `assets/redpanda/`.
- The license-expired scrim was a literal `bg="white"`, unreadable in the dark theme; it takes `bg-background` now.
- The four "number + unit" fields were four hand-written copies of one control; they are one `NumberWithUnitField`.

`license-utils.test.tsx` asserted Chakra's `data-status` attribute; the Registry Alert carries its tone in the variant classes, so the assertion follows.

### Fixed while here (pre-existing)
`skipTlsVerification` and `metricsSamples` were collected but never sent; both are wired now. Label-selector rows were keyed by their own values, so every keystroke remounted the row and dropped focus; they carry a stable id.

### Effect
Files importing `@redpanda-data/ui`: **81 → 72**.

### Gates
`type:check` · `lint` · `lint:check` per changed file · unit (880) · integration (1304). To exercise by hand: `debug-bundle.spec.ts` and `license.spec.ts` (enterprise variant), plus the expiry and trial states.

- License banner: the enterprise-features sentence sat directly in the grid `AlertDescription`, so its `<strong>` became its own row; one block child. Overview banner: button-styled CTAs rendered as anchors were underlined by the description's anchor rule.
- Debug bundle: Cancel in the confirm dialog rendered outlined because `AlertDialogCancel` merges its own variant classes — use its `variant` prop; the number-with-unit rows had the unit select taking the row; the in-progress link derives from the RUNNING broker status; "Try again"/"Done" are Buttons again (review ask), with the radius `buttonVariants` alone does not give.
- Upload license: `Textarea rows` is inert under content sizing (`resize="vertical"` restores it); a rejected re-drop clears the stale file instead of submitting it under the error.

### Final review (2026-09-08)
- Comments trimmed to the constraints. Accepted as-is: the two multi-select labels name the control but do not focus it on click; a dropped license file can be replaced but not cleared; "Try again"/"Done" stay Buttons (role=button) as reviewed.


frontend/src/components/pages/admin/admin-debug-bundle-progress.tsx
```diff
@@ -121,18 +121,17 @@ const AdminPageDebugBundleProgressContent: FC = () => {
                 Stop
               </Button>
             ) : (
-              <Button
-                as={Link}
+              <Link
```
jvorcak: Image
Image

I don't like the inconsistency of a buttons / links. Why is does `Try again` button have no rounded borders, only because it's a link?

Also when it comes to a11y best practices, it's semantically a button, not a link. I think it should stay that way and we should only change the visual

jvorcak:
Screenshot 2026-09-08 at 10 51 02
 Should we migrate this component as well? I'm not sure if it's this PR or it's some other regression, but I can't type key/value pairs here.

frontend/src/components/pages/admin/admin-debug-bundle-progress.tsx
SpicyPete: Should be fixed now

eblairmckee (approved):
visually reviewed in both self-hosted and cloud https://redpanda.zoom.us/clips/share/GXDmxOkTSO-78fvrgSxOIQ

PR #2635, opened 2026-09-05
[Chakra exit · Phase 2 · PR 7/14] frontend: overview, broker and login surfaces on Registry components, plus two smoke specs

SpicyPete:
## Chakra exit · Phase 2 (route by route) · PR 7 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Branch `chakra-exit/07-entry-surfaces`, cut from #2629; re-targets `master` once that merges.

Eight files across the two first-impression surfaces — overview and login — plus the two `misc/` list components they use. Neither route had any Playwright coverage, so **two smoke specs land first, in their own commit**.

### What moved
Chakra layout props to Tailwind classes throughout. Chakra's `TextDivider` (login only) becomes the Registry `FieldSeparator`; `Skeleton noOfLines` becomes `SkeletonText lines`; the login error modal becomes a Registry `Dialog`.

### Worth a look
- **`auth.setup.ts` waited on `.chakra-modal__overlay`** — a class that no longer exists once the login modal is a Registry Dialog. It now waits on the dialog's own button disappearing (the Dialog unmounts on close), with no swallowing `.catch`.
- **"Log in with OIDC" navigates to a backend endpoint**, so it is an anchor with `buttonVariants()`, not a Button — same for the debug-bundle links in cluster-health-overview, which `debug-bundle-page.ts` reads via `getByRole('link')`.
- **Chakra's `DataTable` injected an expander column for `subComponent`; the Registry's renders none** — config-list's synonym sub-rows were unreachable. Tables keep the legacy paging contract — 50 a page, pager only past that; Chakra defaulted pagination on, the Registry to 10 with a permanent pager.
- **`whitespace-break-spaces` was inert on config values.** `.value` in `ConfigList.module.scss` is unlayered, so it beats any `@layer utilities` rule — where Chakra's emotion class, injected after the bundle, used to win. It was silently downgrading `break-spaces` to `pre-line`, collapsing runs of spaces. Fixed at the module rule, which is where it has to live.
- Two literal colours become tokens: the cluster-details hairline was `#ddd` (invisible in the dark theme) and `.nameText` had `#d1d1d1`/`#f5f5f5`. The login gradient and the Redpanda logo stay literal — both are fixed art.
- The waving-panda SVG only existed as a data URI inside the Chakra bundle, so it is vendored to `assets/redpanda/`.
- The login submit button is disabled in flight; the inputs were, the button was not, so a double click fired `loginSaslScram` twice.

Every `data-testid` is preserved, including `overview-license-name` and the four the login flow depends on.

### Effect
Files importing `@redpanda-data/ui`: **81 → 73**.

### Gates
`type:check` · `lint` · `lint:check` per changed file · unit (880) · integration (1304). Login also wants a manual pass per SSO provider — the OSS variant has no auth configured, so `/login` only exercises the redirect.

- Overview: the Cluster Status stripe never rendered (its selector still targeted `.ant-statistic`); the four section headings were a hard-coded near-black in dark mode and carried a dead text class; `Statistic` → Registry `Stat` here and on broker details; the broker table paginates like the legacy one; a 0-byte broker shows "0 B".
- Cluster health: link anchors drop the button height; unknown health reasons read "Unknown reason" instead of an enum number. `state/backend-api.ts`: the `debugBundleStatus` getter prefers the RUNNING broker status so the in-progress link gets its job id.
- Config list: the three remaining literal colours in `ConfigList.module.scss` are tokens. Login: Button `isLoading`, logo on the foreground token, the OR separator chip matches the page ground in dark, heading rung `text-heading-xl` (the original 1.5em), also on login-complete.

### Final review (2026-09-08)
- Broker table stays at ten a page (its legacy `defaultPageSize`), pager only past that. Cluster Details headings use `text-heading-sm` / `text-body-sm` instead of caption-sized rungs; the docs footer inherits its size.
- Config list drops dead column `size` hints, an inert chevron class, orphaned selectors in both stylesheets and history comments. The logo wordmark follows the theme (`currentColor`); the mark keeps its brand red.


eblairmckee:
@SpicyPete cluster status stripe is broken on self-hosted/when new overview page FF is off
should be full height of the card. but tbh, I think we should replace it with a StatusDot rather than a left border

## self-hosted
Screenshot 2026-09-08 at 2 01 12 PM

## cloud when FF off
Screenshot 2026-09-08 at 2 02 37 PM

eblairmckee:
toggle group on broker details page appear broken, i think they should use outline variant. 

Screenshot 2026-09-08 at 2 12 18 PM


difference between friendly/raw is unclear to me, just seems like more padding when friendly? not sure what old behavior/ui was, but worth investigating. 
Screenshot 2026-09-08 at 2 08 42 PM


Screenshot 2026-09-08 at 2 08 55 PM

eblairmckee:
Login LGTM. wasn't able to reproduce login options other than OIDC or Google Auth, though. Worth testing the "wrong password" UI in integration. 
Screenshot 2026-09-08 at 2 13 11 PM
Screenshot 2026-09-08 at 2 13 03 PM

eblairmckee (approved):

SpicyPete:
Thanks, updating some of the sections based on your feedback.

PR #2636, opened 2026-09-05
[Chakra exit · Phase 2 · PR 8/14] frontend: mount the Registry ThemeProvider and take the chrome off Chakra

SpicyPete:
## Chakra exit · Phase 2 (route by route) · PR 8 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Branch `chakra-exit/08-chrome-shell`, directly on `master`.

**Hard prerequisite for PR 13** — the cut cannot happen until the theme provider is mounted here.

### The load-bearing half
`data-theme` on `<html>` is what `theme.css` keys its dark palette on, and what `use-theme-appearance` watches to tell Monaco, CodeMirror, react-flow and sonner which ground they are painting on. Chakra's `ColorModeSwitch` was the only thing setting it.

- `components/layout/theme-mode-switch.tsx` replaces it on the Registry `ThemeProvider`'s `useTheme()`, keeping the `ColorModeSwitch` test id and the "Switch to *next* mode" label. It reads `resolvedTheme`, not `theme`, because the stored preference can be `system`.
- **The provider is mounted in `app.tsx` only**, not in embedded-app or federated-providers as the plan said. In those modes the Cloud UI host owns `data-theme`; a second writer inside Console would fight it.
- **`defaultTheme="light"`, deliberately.** Chakra pinned `initialColorMode: 'light'` with `useSystemColorMode: false`. The Registry default is `system`, which would flip every OS-dark user to the Registry's dark palette while the still-mounted Chakra half stayed light — and the toggle is dev-only, so no way back. Dark mode stays its own project.
- **The `theme={toasterTheme}` plumbing stays**, contrary to the plan. `sonner.tsx` reads `useTheme().theme`, which can be `system` and then resolves off `prefers-color-scheme` — disagreeing with an explicit toggle. `useThemeAppearance()` reads the ground actually painted, which is the answer sonner needs.

ChakraProvider's own ColorModeProvider is still a second `data-theme` writer one level down. It applies once on mount and this provider's effect lands after it, so the attribute is ours — but Chakra's internal colour mode no longer changes, so toggling in dev repaints the Registry surface and leaves Chakra components light. That resolves when PR 13 removes ChakraProvider; it is written down at the mount point.

### The misc leftovers
Each behind its existing export: `no-clipboard-popover` (hover Popover → Tooltip), `kowl-time-picker`, `removable-filter`, `search-bar` (`SearchField` → the shared `misc/search-input`, keeping the `search-field-*` test ids and the `/`/Escape shortcuts), `kowl-json-view`, `expandable-text`, `common` (update Modal → Dialog) and `wizard`.

`hooks/use-developer-view` loses Chakra's `useKey` + `useLocalStorage` for a keydown listener and localStorage directly — these non-visual hooks keep the dependency alive on their own. The replacement ignores `?` typed into a field or held with a modifier, which Chakra's `useKey` did not, and its test now asserts the toggle flips and persists rather than only that it does not crash.

### Worth a look
**The clipboard-unavailable tooltip could never open.** `render={children}` handed the trigger props to a component that forwards no rest props, and the child in that branch is a *disabled* button, which dispatches no pointer events. It wraps in a span now — the same shape the header's disabled buttons use. This is the only explanation users get for a copy button that silently fails over plain HTTP.

### Folded in: single-select tables (2026-09-08)
Thirteen Registry tables already on `master` (consumer groups list and details, the edit-offsets modal, quotas, schema registry, the new security tabs, the topic list and its partitions, consumers, ACL and messages tabs, the Connect log explorer) now pass `enableHiding: false` and `enableRowSelection: false`. The first removes the "Hide" item from the column header menus — none of these pages has a column-visibility control to bring a column back. The second removes the pager's "X of N row(s) selected." text, which the Registry gates on that option — none of them selects rows. The log explorer had blanked that text with a CSS hack; it uses the option now. The secrets store keeps both (it has row checkboxes and view options). Folded in here rather than opened as a fifteenth PR.

### Effect
Files importing `@redpanda-data/ui`: **81 → 71** — the largest single drop in Phase 2.

### Gates
`type:check` · `lint` · `lint:check` per changed file · unit (882) · integration (1311), including `header.test.tsx` and `use-developer-view.test.tsx`. By hand: the header on both embedded and self-hosted layouts, and that `data-theme` still lands on `<html>` with Monaco, CodeMirror, react-flow and sonner following the dev toggle.

- `index.scss`: the `p, dl, dd, ol, ul, h1–h5 { margin-bottom: 0 }` reset is in `@layer base` now — unlayered it beat every `mb-*` utility the migration put on a heading.
- Chakra's colour-mode manager is pinned to light so a stored `chakra-ui-color-mode=dark` cannot resurrect a dark Chakra half; `bootstrap.tsx` calls `initTheme` so the first paint already carries the theme.
- The `?` developer-view toggle ignores key repeat, IME composition, handled events and typing targets, and reads its stored flag as a boolean.
- Misc: clipboard notice spacing and dead `placement` prop; `ExpandableText` renders a span (it sits inside `<p>`) with a Registry link Button; UpdatePopup Cancel is plain outline; dead Wizard scss classes and an inert icon size removed; JSON view sizing on utilities; the error boundary is on Tailwind classes (asked on #2629 after it merged); the not-found anchor gets the radius `buttonVariants` alone does not give.

### Final review (2026-09-08)
- Escape in the search field is field-scoped (the Chakra SearchField cleared the filter from anywhere on the page) — deliberate, so Escape-to-close-a-dialog no longer wipes a filter. Comments trimmed.

### DIFF
--- a/frontend/src/components/misc/search-bar.tsx
+++ b/frontend/src/components/misc/search-bar.tsx
@@ -9,7 +9,9 @@
  * by the Apache License, Version 2.0
  */
 
-import { SearchField } from '@redpanda-data/ui';
+import { Button } from 'components/redpanda-ui/components/button';
+import { Input, InputEnd, InputStart } from 'components/redpanda-ui/components/input';
+import { SearchIcon, XIcon } from 'lucide-react';
 import { type ReactNode, useEffect, useMemo } from 'react';
 
 import { AnimatePresence, animProps_span_searchResult, MotionSpan } from '../../utils/animation-props';
@@ -72,12 +74,30 @@ function SearchBar<TItem>(props: SearchBarProps<TItem>) {
         alignItems: 'center',
       }}
     >
-      <SearchField
-        placeholderText={placeholderText}
-        searchText={filterText}
-        setSearchText={onQueryChanged}
-        width="350px"
-      />
+      <Input
+        containerClassName="max-w-[350px]"
+        onChange={(e) => onQueryChanged(e.target.value)}
+        placeholder={placeholderText ?? 'Search...'}
+        testId="search-field-input"
+        value={filterText}
+      >
+        <InputStart>
+          <SearchIcon className="size-4 text-muted-foreground" data-testid="search-field-search-icon" />
+        </InputStart>
+        {filterText !== '' && (
+          <InputEnd className="pointer-events-auto">
+            <Button
+              aria-label="Clear search"
+              data-testid="search-field-reset-icon"
+              onClick={() => onQueryChanged('')}
+              size="icon-xs"
+              variant="ghost"
+            >
+              <XIcon />
+            </Button>
+          </InputEnd>
+        )}
+      </Input>
 
       <AnimatePresence>
         {Boolean(filterSummary) && (
--- a/frontend/src/components/misc/search-bar.tsx
+++ b/frontend/src/components/misc/search-bar.tsx
@@ -9,9 +9,10 @@
  * by the Apache License, Version 2.0
  */
 
+import { CloseIcon } from 'components/icons';
 import { Button } from 'components/redpanda-ui/components/button';
 import { Input, InputEnd, InputStart } from 'components/redpanda-ui/components/input';
-import { SearchIcon, XIcon } from 'lucide-react';
+import { SearchIcon } from 'lucide-react';
 import { type ReactNode, useEffect, useMemo } from 'react';
 
 import { AnimatePresence, animProps_span_searchResult, MotionSpan } from '../../utils/animation-props';
@@ -75,7 +76,7 @@ function SearchBar<TItem>(props: SearchBarProps<TItem>) {
       }}
     >
       <Input
-        containerClassName="max-w-[350px]"
+        containerClassName="w-full max-w-[350px]"
         onChange={(e) => onQueryChanged(e.target.value)}
         placeholder={placeholderText ?? 'Search...'}
         testId="search-field-input"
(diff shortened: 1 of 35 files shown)


frontend/src/components/misc/search-bar.tsx
```diff
@@ -72,12 +75,32 @@ function SearchBar<TItem>(props: SearchBarProps<TItem>) {
...
-        searchText={filterText}
-        setSearchText={onQueryChanged}
-        width="350px"
-      />
+      <Input
+        containerClassName="w-full max-w-[350px]"
+        onChange={(e) => onQueryChanged(e.target.value)}
+        placeholder={placeholderText ?? 'Search...'}
+        testId="search-field-input"
+        value={filterText}
```
malinskibeniamin, line 73: [P2] Preserve the search field keyboard shortcuts

Replacing `SearchField` with `Input` drops the existing `/` focus and Escape clear/blur shortcuts, previously advertised by the search tooltip. This affects the Kafka Connect overview and cluster connector lists: Escape now leaves the active filter unchanged. Restore the keyboard handlers and input ref, with guards for other editable controls.

Verified in Chromium against pinned `@redpanda-data/ui@4.2.0`: `/` focuses and Escape clears `alpha` then blurs the empty field on the baseline; none works on this head. Add an interaction test covering that same sequence.

malinskibeniamin: Fixed in 6796d0ee1: restored slash focus and Escape clear/blur, guarded other editable controls and modified shortcuts, and cleaned up the document listener on unmount. Three regression tests pass (observed RED before implementation), focused lint and source/test type checks pass. Chromium baseline/head replay now agrees for all three shortcuts, with zero page errors. Full backend E2E not run locally; CI pending.

eblairmckee:
[theme switcher review](https://redpanda.zoom.us/clips/share/pPQcGPdGQkCVuxXudTprkA) 📺 
verified production issue where manually changing local storage chakra ui theme value introduced mixed themes in the UI was resolved by this branch. 
but I do have some questions:
1. I do want to make sure my assumptions are correct wrt cloud UI listening to the value of `chakra-ui-color-mode` localstorage item, since cloud ui theme seems to change whenever the local storage value is changed and the page reloaded. cloud UI theme toggler also seems to override the value of `chakra-ui-color-mode` as well. 
2. why does standalone console have a separate `redpanda-ui-theme` local storage item that embedded console does _not_ have? is the plan long-term to get rid of `chakra-ui-color-mode` in cloud-ui and rename to `redpanda-ui-theme`?

eblairmckee:
A few things I noticed (both standalone and embedded console) that were not using dark mode aware tokens. 
some uppercase labels in dialogs (mostly on topics page, which seems like we're not migrating in this effort?)
Screenshot 2026-09-08 at 3 04 34 PM
Screenshot 2026-09-08 at 3 04 42 PM

no code editors seemed to be theme aware
Screenshot 2026-09-08 at 3 03 07 PM

some tables were not theme aware
Screenshot 2026-09-08 at 3 03 43 PM

illustrations do not have transparent backgrounds, nor have enough contrast from dark mode background even if they did have a transparent background (can tackle this whenever we productionize dark mode)
Screenshot 2026-09-08 at 3 03 50 PM

SpicyPete:
I'll take a look at these, but are these pages specific to this branches migration?  Any page outside of this branch will still largely be on Chakra.  Anything which isn't adapting to dark mode looks like remaining Chakra in other PRs

eblairmckee (approved):
no blockers, just looking for clarity (see earlier comments)

SpicyPete:
Thanks, yes everything except the image background should be covered in following PR's.
Images (including brand logos) on dark background we'll likely need to do a dedicated branch for after the Chakra pass

PR #2637, opened 2026-09-05
[Chakra exit · Phase 2 · PR 10/14] frontend: the Kafka Connect dynamic-ui engine on Registry components

SpicyPete:
## Chakra exit · Phase 2 (route by route) · PR 10 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Branch `chakra-exit/10-connect-dynamic-ui`, based on `master` (rebased onto it 2026-09-09, 0 commits behind). **PR 11 (#2638) is stacked on this one.**

The schema-driven form renderer: a connector's config schema arrives from the backend and `PropertyComponent` picks an input per `definition.type`. **The schema contract does not change and neither does the value written back** — `updatePropertyValue` still writes `property.value` and calls `notifyChange()`.

### Old → new, by file
| File | Change |
|---|---|
| `forms/error-wrapper` | Chakra `FormField` → `Field` + `FieldLabel` + `FieldDescription` + `FieldError`. This is the shell **every** property renders through, so it is the widest hunk by reach. Its `onClick` stays on the field: a property can carry several errors and clicking is what advances `currentErrorIndex`. |
| `property-component` | `NumberInput` → `Input type="number"`; `RadioGroup options={…}` → composed `RadioGroupItem` + `Label` (Base UI has no options prop); `Switch isChecked/onChange` → `checked/onCheckedChange`. |
| `forms/secret-input` | The Registry `Input` already owns the reveal toggle for `type="password"` and disables it while read-only — which is the state an existing secret sits in before Edit — so the hand-rolled `InputGroup` + `InputRightElement` + `useBoolean` all go. |
| `forms/topic-input` | `Select isMulti` → `SimpleMultiSelect` (string-valued, and the property is a comma-joined string either way); FormControl/FormHelperText/FormErrorMessage → Field/FieldDescription/FieldError. |
| `list` | The rename hint Tooltip was `isOpen={hasFocus}` — focus-controlled, not hover — so it stays controlled via `open`. |
| `property-group` | Chakra `Accordion items={…}` → composed AccordionItem/Trigger/Content. **The sub-group heading rendered its name twice**; fixed rather than carried across. |
| `components`, `connector-step` | Layout props, and `Skeleton noOfLines` → `SkeletonText lines`. |

### The two hunks that carry real judgement
Both were caught in review and fixed in a follow-up commit; they are the reason this PR is worth reading closely.

1. **Numeric properties must stay strings.** Chakra's `NumberInput.onChange` is `(valueAsString, valueAsNumber)` and the old code took the *first* argument. `getConfigObject` sends `p.value` as-is **and** `===`-compares it against `default_value` to suppress untouched defaults — and `sanitizeDefaultValue` numbers only the INT/LONG/SHORT defaults. Storing a number changed which properties get sent, inconsistently across the numeric types; clearing a field stored `NaN`, which `JSON.stringify` turns into `null` in the validate and create requests.
2. **Chakra's `FormControl` generated an id and wired `FormLabel htmlFor`; the Registry `Field` does not.** Without it every property on the create/edit connector form was an unlabeled control. A derived `property-<name>` id now goes to both the control and the label. `FormField` also special-cased a `Switch` child into an inline row, and `Field`'s vertical variant carries `[&>*]:w-full`, which stretched the Switch pill across the row — BOOLEAN and RADIO_GROUP are `orientation="horizontal"` and the control is wrapped.

Also from review: the Registry `Input` **discards a caller's `style`** (it applies its own after the prop spread), so both list inputs had silently lost their flex sizing; the numeric steppers are back (`showStepControls`). A second review then caught `step="any"`: the Registry `Input` coerces `step` with `Number()`, so it became `NaN` and one stepper click wrote an empty value — `step={1}` now — and forwarded the property id into `SecretInput` so the password label associates.

### Effect
Files importing `@redpanda-data/ui`: **50 → 42**. (Earlier revisions of this body quoted 81 → 73, against the pre-Phase-2 master.)

### Gates
Re-run on the rebased branch (2026-09-09): `type:check` clean · unit **62 files / 882 tests** · integration **132 files / 1316 tests**, all passing · `lint:check` on the 10 changed files reports **1 error**, pre-existing (`useBlockStatements` in `dynamic-ui/list.tsx`, present on master). `connector.spec.ts` exercises this end to end.

### Final review (2026-09-08)
- Radio-group properties are named: a radiogroup div is not labelable, so the label carries an id and the group points at it with `aria-labelledby`.
- BOOLEAN and RADIO_GROUP fields are vertical again — Chakra's FormField never inlined them (its Switch check never matched at runtime), so the horizontal Field row was a layout change, not parity.
- The list ghost input's transparent border is a Tailwind class on the element: the unlayered `!important` in index.scss lost to the Registry's layered `!border-input`, so the field rendered bordered. The new-entry input fits its 29px row.
- Known: the Registry number stepper does not round, so stepping a DOUBLE like 2.3 can store 1.2999999999999998 (Chakra rounded to the value's precision).

- Required properties carry `required`/`aria-required` again — Chakra's `FormControl` stamped them and the Registry `Field` does not. Applied across STRING/PASSWORD/INT/BOOLEAN/RADIO_GROUP and the two remaining text branches.
- The comma-separated-list control takes the property id, so its new-entry input's label resolves.
- Heading/description gaps moved to `mt-*` on the following element. The `mb-*` they replaced was inert while the `index.scss` margin reset sat unlayered; **#2636 has since moved that reset into `@layer base`**, so either would work now — `mt-*` is kept because it is what was tested.
- The password test reveals the value before Undo, so it fails without the remount `key`.
- `.stringList` rows and buttons use theme tokens (`--color-surface-recess`) instead of literal greys, so they follow dark mode.

### DIFF
--- a/frontend/src/components/pages/connect/dynamic-ui/connector-step.tsx
+++ b/frontend/src/components/pages/connect/dynamic-ui/connector-step.tsx
@@ -9,8 +9,6 @@
  * by the Apache License, Version 2.0
  */
 
-import { Box, Heading, Text } from '@redpanda-data/ui';
-
 import type { ConfigPageProps } from './components';
 import { PropertyGroupComponent } from './property-group';
 import type { PropertyGroup } from '../../../../state/connect/state';
@@ -33,16 +31,10 @@ export const ConnectorStepComponent = (props: {
   }
 
   return (
-    <Box>
-      <Heading as="h3" mb="4" mt="8" size="md">
-        {step.name}
-      </Heading>
+    <div>
+      <h3 className="mt-8 mb-4 text-heading-md">{step.name}</h3>
 
-      {Boolean(step.description) && (
-        <Text mb="4" size="sm">
-          {step.description}
-        </Text>
-      )}
+      {Boolean(step.description) && <p className="mb-4 text-body-sm">{step.description}</p>}
 
       {groups.map((g) => (
         <PropertyGroupComponent
@@ -54,6 +46,6 @@ export const ConnectorStepComponent = (props: {
           showAdvancedOptions={props.showAdvancedOptions}
         />
       ))}
-    </Box>
+    </div>
   );
 };
--- a/frontend/src/components/pages/connect/dynamic-ui/connector-step.tsx
+++ b/frontend/src/components/pages/connect/dynamic-ui/connector-step.tsx
@@ -32,9 +32,9 @@ export const ConnectorStepComponent = (props: {
 
   return (
     <div>
-      <h3 className="mt-8 mb-4 text-heading-md">{step.name}</h3>
+      <h3 className="mt-8 text-heading-md">{step.name}</h3>
 
-      {Boolean(step.description) && <p className="mb-4 text-body-sm">{step.description}</p>}
+      {Boolean(step.description) && <p className="mt-4 text-body-sm">{step.description}</p>}
 
       {groups.map((g) => (
         <PropertyGroupComponent
--- a/frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
+++ b/frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
@@ -1,15 +1,6 @@
-import {
-  Button,
-  type ButtonProps,
-  Flex,
-  Icon,
-  Input,
-  InputGroup,
-  InputRightElement,
-  Tooltip,
-  useBoolean,
-} from '@redpanda-data/ui';
-import { EyeIcon, EyeOffIcon } from 'components/icons';
+import { Button } from 'components/redpanda-ui/components/button';
+import { Input } from 'components/redpanda-ui/components/input';
+import { Tooltip, TooltipContent, TooltipTrigger } from 'components/redpanda-ui/components/tooltip';
 import { useRef, useState } from 'react';
 
 export type SecretInputProps = {
@@ -18,22 +9,26 @@ export type SecretInputProps = {
   updating: boolean;
 };
 
-const EditButton = ({ onClick }: Pick<ButtonProps, 'onClick'>) => (
-  <Tooltip hasArrow={true} label="Edit secret value" placement="top">
-    <Button onClick={onClick} variant="link">
-      Edit
-    </Button>
+const EditButton = ({ onClick }: { onClick: () => void }) => (
+  <Tooltip>
+    <TooltipTrigger
+      render={
+        <Button onClick={onClick} variant="link">
+          Edit
+        </Button>
+      }
+    />
+    <TooltipContent side="top">Edit secret value</TooltipContent>
   </Tooltip>
 );
 
-const ClearButton = ({ onClick }: Pick<ButtonProps, 'onClick'>) => (
+const ClearButton = ({ onClick }: { onClick: () => void }) => (
   <Button onClick={onClick} variant="link">
     Undo
   </Button>
 );
 
 export const SecretInput = ({ value, onChange, updating = false }: SecretInputProps) => {
-  const [visible, setVisible] = useBoolean();
   const initialValueRef = useRef(value);
   const [canEdit, setCanEdit] = useState(!updating);
   // Intentionally seeded from prop: localValue is edited independently during user interaction
@@ -54,35 +49,32 @@ export const SecretInput = ({ value, onChange, updating = false }: SecretInputPr
         setCanEdit(false);
         setLocalValue(initialValueRef.current);
         onChange(initialValueRef.current);
-        setVisible.off();
       }}
     />
   );
 
   return (
-    <Flex flexDirection="row" gap={2}>
-      <InputGroup>
-        <Input
-          onChange={(e) => {
-            setLocalValue(e.target.value);
-            if (onChange) {
-              onChange(e.target.value);
-            }
-          }}
-          readOnly={!canEdit}
-          type={visible ? 'text' : 'password'}
-          value={localValue}
-        />
-        {Boolean(canEdit) && (
-          <InputRightElement>
-            <Button onClick={() => setVisible.toggle()} variant="ghost">
-              <Icon as={visible ? EyeIcon : EyeOffIcon} color="gray.500" />
-            </Button>
-          </InputRightElement>
-        )}
-      </InputGroup>
+    <div className="flex flex-row gap-2">
+      {/*
+        The Registry Input owns the reveal toggle for `type="password"`, and disables it while the
+        field is read-only — which is the state an existing secret sits in before Edit. `key` is
+        what re-masks on Undo: the toggle's state lives inside the Input, so the transition out of
+        editing has to remount it or Undo would leave the restored secret in plain text.
+      */}
+      <Input
+        key={canEdit ? 'editing' : 'masked'}
+        onChange={(e) => {
+          setLocalValue(e.target.value);
+          if (onChange) {
+            onChange(e.target.value);
+          }
+        }}
+        readOnly={!canEdit}
+        type="password"
+        value={localValue}
+      />
       {Boolean(updating) && (canEdit ? clearButton : editButton)}
-    </Flex>
+    </div>
   );
 };
 
--- a/frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
+++ b/frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
@@ -1,6 +1,6 @@
 import { Button } from 'components/redpanda-ui/components/button';
 import { Input } from 'components/redpanda-ui/components/input';
-import { Tooltip, TooltipContent, TooltipTrigger } from 'components/redpanda-ui/components/tooltip';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from 'components/redpanda-ui/components/tooltip';
 import { useRef, useState } from 'react';
 
 export type SecretInputProps = {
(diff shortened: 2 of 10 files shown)


frontend/src/components/pages/connect/dynamic-ui/property-component.tsx
```diff
@@ -118,21 +141,35 @@ export const PropertyComponent = (props: { property: Property }) => {
...
+            // would change which properties get sent — and `sanitizeDefaultValue` only numbers the
+            // INT/LONG/SHORT defaults, so it would not even be consistent across the numeric types.
+            updatePropertyValue(p, e.target.value);
           }}
-          value={Number(v)}
+          // Chakra's NumberInput shipped its steppers by default and the Registry Input needs
+          // asking. `step="any"` for the float types, which the Registry otherwise pins to 1 and
+          // the browser then marks non-integer values invalid.
+          showStepControls
+          step={def.type === 'DOUBLE' || def.type === 'FLOAT' ? 'any' : 1}
```
malinskibeniamin: [P1] Keep floating-point steppers from clearing the configuration value

Registry `Input` converts `props.step` with `Number(props.step)`, so `"any"` becomes `NaN` in both the native input and the stepper arithmetic. For a DOUBLE/FLOAT value such as `1.5`, either button calculates `"NaN"`; the number input sanitizes that to an empty string, which this change handler writes into the connector configuration. Preserve native `step="any"` while using a finite numeric increment for the buttons, or pass a supported finite step here.

Verify: open a connector with a DOUBLE/FLOAT property, enter `1.5`, and click each stepper; the value must remain numeric and the outgoing config must not become empty.

frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
```diff
@@ -54,35 +52,32 @@ export const SecretInput = ({ value, onChange, updating = false }: SecretInputPr
...
-        )}
-      </InputGroup>
+    <div className="flex flex-row gap-2">
+      {/*
+        The Registry Input owns the reveal toggle for `type="password"`, and disables it while the
+        field is read-only — which is the state an existing secret sits in before Edit. `key` is
+        what re-masks on Undo: the toggle's state lives inside the Input, so the transition out of
+        editing has to remount it or Undo would leave the restored secret in plain text.
+      */}
+      <Input
```
malinskibeniamin, line 69: [P2] Forward the property ID to the password input

`ErrorWrapper` now connects its label through `htmlFor={inputId}`, but the PASSWORD branch does not pass `fieldId` into `SecretInput`, and this input has no `id`. The previous Chakra FormControl supplied that association implicitly; Registry Field supplies validation state, not a control ID. Connector password fields therefore lose their accessible names, and clicking their labels no longer focuses them. Add an `id` prop, pass `fieldId` from PropertyComponent, and forward it here.

Verify: resolve the password input with `getByLabelText` and click its label, including after Edit/Undo.

frontend/src/components/pages/connect/dynamic-ui/property-component.tsx
malinskibeniamin: Fixed in 0df8ecfd71b638c370831cfbbad7655fd25f6d9b. Restored the finite default increment so DOUBLE/FLOAT steppers preserve fractional values.

Three PropertyComponent integration cases pass after failing on the original head. Source and focused test type checks pass. Chromium public-component consumer replay passes with seeded API stores and zero page errors; no live-cluster mutation was performed.

frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
malinskibeniamin: Fixed in 0df8ecfd71b638c370831cfbbad7655fd25f6d9b. Forwarded the property ID through SecretInput; its label stays associated through Edit/Undo.

Three PropertyComponent integration cases pass after failing on the original head. Source and focused test type checks pass. Chromium public-component consumer replay passes with seeded API stores and zero page errors; no live-cluster mutation was performed.

frontend/src/components/pages/connect/dynamic-ui/property-component.tsx
```diff
@@ -74,25 +96,28 @@ export const PropertyComponent = (props: { property: Property }) => {
...
               value={v}
             />
-          </Box>
+          </div>
         );
       } else {
         // Input
         inputComp = (
           <Input
             defaultValue={def.default_value ?? undefined}
```
eblairmckee, line 114: This `Input` receives both `defaultValue={def.default_value ?? undefined}` (this line) and `value={String(v)}` (L122). React logs:

> Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both)

Reproduces on the S3 sink form. The Registry `Input` spreads `{...props}` onto the DOM `<input>` (`redpanda-ui/components/input.tsx:152`), so both props land there.

The pair predates this PR, so not something you introduced — but since `value` + `onChange` already makes this controlled, `defaultValue` is dead and dropping it clears the warning. Same pattern in two other branches you're touching: the `LIST` branch (L192 + L198) and the `default:` branch (L207 + L213).

frontend/src/components/pages/connect/dynamic-ui/connector-step.tsx
```diff
@@ -54,6 +46,6 @@ export const ConnectorStepComponent = (props: {
           showAdvancedOptions={props.showAdvancedOptions}
```
eblairmckee, line 46: Re: `key={g.group.name}` just above (L45) — `ConnectorGroup.name` is optional (`state/rest-interfaces.ts:984`), so a group without a name yields `key={undefined}`, which React treats as no key at all:

> Each child in a list should have a unique "key" prop

Also reproduces on the S3 sink form. Pre-existing and unchanged by this PR, so purely FYI — raising it because `ConnectorPropertiesStore` carries a `fallbackGroupName`, which suggests unnamed groups are a real case rather than theoretical. An index-based fallback would settle it.

eblairmckee (commented):
Ran this branch against a live Kafka Connect cluster (Redpanda + `connectors-unsupported`, S3 sink) and hit two React console errors on the create-connector form. Both are latent issues on lines this PR doesn't change, so neither is a regression from your work — flagging them because the migration touches the same elements and they're cheap to clear while you're in here. Nothing blocking.

eblairmckee:
Moved to redpanda-data/console#2638 — `pages/connect/overview.tsx` is that PR's file, so the fix has a home there rather than here.

Original content: the Kafka Connect tab renders a permanent skeleton on self-hosted, because `checkRPCNSecretEnable()` throws on a 501 from `SecretService/ListSecretScopes` and blocks the `refreshConnectClusters()` call behind it, with a bare `catch {}` swallowing the error. Full write-up and suggested fix on #2638.

eblairmckee:
Small cross-repo note for anyone else trying to review this branch locally.

If you're testing through a local BYOC cluster and the Connect page shows *"You have no Redpanda Connect pipelines"*, that empty state may be lying to you — `PipelineService/ListPipelines` returns `503 {"code":"unavailable"}` and the UI renders it as empty rather than as an error. Root cause is in cloudv2 localdev, not here: `deploy-local-rpcn.sh` applied CRDs from a hardcoded filename list including one that was later deleted, so `set -euo pipefail` aborted the deploy before RBAC/Service/Deployment and the cluster came up with no `redpanda-connect-api` at all.

Fix: redpanda-data/cloudv2#29798. Re-run **Update RPCN (installpack)** in Tilt afterwards.

Worth noting separately that Kafka Connect — the surface this PR actually changes — can't be exercised through embedded cloud at all right now: those clusters have no `kafkaConnect` block in the Console configmap and no connectors workload, and `render()` returns `<PipelineListPage />` early when `isEmbedded()`, so the Kafka Connect tab never mounts there. Standalone Console against a real Kafka Connect cluster is the only path today. The dynamic-ui code is identical between the two apart from one runtime branch — `readOnly: isEmbedded() && p.context === 'EDIT'` in `components.tsx` — which is always `false` standalone, so the locked-JSON-editor state isn't reviewable locally.

eblairmckee (approved):
reviewed locally https://redpanda.zoom.us/clips/share/gg9HlMB7R6yzLgdy1V3-0Q
still trying to get a connection to save so i can test the configured connector page

frontend/src/components/pages/connect/dynamic-ui/forms/secret-input.tsx
```diff
@@ -54,35 +54,34 @@ export const SecretInput = ({ value, onChange, updating = false }: SecretInputPr
...
+        id={id}
+        key={canEdit ? 'editing' : 'masked'}
+        onChange={(e) => {
+          setLocalValue(e.target.value);
+          if (onChange) {
+            onChange(e.target.value);
+          }
+        }}
+        readOnly={!canEdit}
```
eblairmckee, line 78: **The reveal toggle now appears where it used to be absent, and it looks broken.**

Repro on an existing connector's Configuration tab (S3 sink, `aws.access.key.id`):

1. Field shows the masked value with an eye icon. Clicking it does nothing — `readOnly` disables it.
2. Click **Edit** → field clears, eye becomes enabled.
3. Click the eye → still empty, because there's nothing to reveal.

The clearing itself is unchanged from master — the handler is byte-identical (`setCanEdit(true); setLocalValue('')`), so revealing an empty field is correct behavior.

What changed is that master only rendered the eye **after** Edit:

```jsx
{Boolean(canEdit) && (
  <InputRightElement>
    <Button onClick={() => setVisible.toggle()} variant="ghost">…
  </InputRightElement>
)}
```

The Registry `Input` renders it unconditionally for `type="password"` (`isPasswordInput = type === 'password'`, `disabled={props.disabled || readOnly}`) — there's no prop to suppress it. So users now get a reveal affordance on the masked existing secret, and the only two things it can do are nothing (before Edit) or show an empty field (after). It reads as broken twice.

Worth knowing: the real value **is** in the form. Kafka Connect's `/connectors/{name}/config` returns secrets in plaintext, and the edit form seeds from it — the masks are 20 and 40 dots, matching the real key lengths, not the 8-char `[hidden]` that `config/validate` returns. So revealing is technically possible; master just never offered it.

Three ways out, in increasing order of scope:

- **`setLocalValue(value)` instead of `''`** — Edit unlocks the current value in place, reveal shows it, and the affordance means something in both states. One line. Changes Edit's semantics (clear manually to replace) and does expose the secret — though it's already in browser memory and the network log, so no new exposure.
- **Suppress the toggle while read-only** — restores master exactly, but needs a new `InputProps` flag on the shared registry `Input`.
- **Leave it** and accept a decorative eye.

I'd take the first, but it's a product call on whether an existing connector secret should be revealable, so flagging rather than asserting.

Separately: `setLocalValue('')` doesn't call `onChange`, so the store keeps the real secret while the field shows empty. Saving without typing preserves the old secret — benign today, but it means clicking Edit *intending to clear* a secret and saving silently restores it. Undo works correctly; the `key` remount re-masks as documented.

eblairmckee (commented):
One more from testing the stack (found while reviewing #2638, but it belongs here — `secret-input.tsx` is this PR's file, not PR 11's). Not a regression; it's an affordance change that reads as a broken button. Details inline.

PR #2638, opened 2026-09-05
[Chakra exit · Phase 2 · PR 11/14] frontend: the Kafka Connect pages on Registry components

SpicyPete:
## Chakra exit · Phase 2 (route by route) · PR 11 of 14

Part of the [Chakra exit plan and live tracker](https://claude.ai/code/artifact/1861e21c-624b-4270-98ad-9921b22c2498). Branch `chakra-exit/11-connect-pages`, **stacked on #2637** (the dynamic-ui engine these pages render); re-targets `master` after that merges. Re-stacked onto the rebased #2637 on 2026-09-09.

Six Connect files (~3,200 LOC — the largest area in the codebase). The shared `misc/search-input` this branch used to carry has since landed on `master` via #2633/#2636, so it is no longer part of this diff — these pages just consume it.

### Suggested reading order
`connector-box-card` and `cluster-details` are the shallow ones, then `overview`, then `helper` (shared building blocks all three pages import), then `connector-details` and `create-connector`.

### The hunks that carry judgement
- **`helper.tsx` `ConfirmModal`** — Chakra's AlertDialog took `leastDestructiveRef={cancelRef}` to put initial focus on "No". Base UI has no such prop, so the ref goes and Cancel is placed first in the footer, which is what gets it the focus. It is also an `AlertDialogDescription` with `min-h-0 overflow-y-auto`: `AlertDialogContent` is `overflow-hidden max-h-[85vh]`, so a long API error had been clipped with the No/Retry footer pushed out of view.
- **`connector-details` `NoEditPermissionTooltip`** — Chakra's Tooltip took `isDisabled` to suppress itself when the permission was present; Base UI has none, so the wrapper renders its child bare instead. It also wraps the trigger in a span: a disabled button fires no pointer events, so the tooltip would never open.
- **The Logs tab's disabled reason** — Chakra's `Tabs` took `isDisabled` as a *reason string*. The app's Tabs wrapper takes a boolean, so the reason moves to a tooltip on the trigger. The disabled trigger sets `aria-disabled:pointer-events-none`, which cascades, so the span re-enables them or the tooltip is unreachable — which is the whole point of moving the reason there.
- **`create-connector`'s filter tabs render empty panels**; the card grid below reacts to the tab key. Called out in a comment so it does not read as a mistake. Its "Creating connector..." dialog has no `onOpenChange` by design, and `showCloseButton={false}` — `DialogContent` renders one by default and it would do nothing.
- **The connector-details overview grid keeps `gridTemplateAreas` inline** — Tailwind has no arbitrary utility for it.

### Caught in review
- **The connector status stripe painted nothing, for every status.** `statusColors` held Chakra theme tokens (`green.500`), which only resolve through a Chakra style prop — as raw CSS the declaration is invalid and the browser drops it. They are `bg-*-strong` utilities now.
- **Four tables lost click-to-sort.** Chakra painted the affordance on every header itself; restored with `DataTableColumnHeader`, with `enableSorting: false` on the derived columns and `enableHiding: false` throughout (no `DataTableViewOptions` on these pages, so a hidden column could not come back).
- **"Create connector" stopped being a button.** `tests/shared/connector.utils.ts` clicks it with `getByRole('button')`, which the original `<Link><Button>` produced — so the nesting is back. Note this is the *opposite* contract to the debug-bundle and license links, which their own helpers read as links: the test decides, per call site.
- `PopoverContent` is a fixed `w-72` where Chakra's `size="stretch"` sized to content, so long fully-qualified connector classes spilled outside the popup; the connector-error dialog lost its JSON highlighting; `NotConfigured` now matches the already-migrated `SchemaNotConfiguredPage` shape.

### Effect
Files importing `@redpanda-data/ui`: **42 → 36** on top of #2637 (so 50 → 36 for the two together), and `pages/connect` is fully clear. (Earlier revisions of this body quoted 73 → 67, against the pre-Phase-2 master.)

### Gates
Re-run on the re-stacked branch (2026-09-09): `type:check` clean · unit **62 files / 882 tests** · integration **132 files / 1316 tests**, all passing · `lint:check` on the 16 changed files reports **12 errors, every one pre-existing** — each maps rule-for-rule and file-for-file onto master's copy of the same file (`useExhaustiveDependencies` ×2, `useBlockStatements` ×3, `useSemanticElements` ×2, `noUselessConstructor`, `noVoid`, `noShadow`). The branch also *clears* master's two `noRestrictedImports` errors on these files. `connector.spec.ts` + `connect-cluster.spec.ts` exercise these end to end.

- Logs tab: the columns array was rebuilt on every 200 ms poll tick, which re-created every header and cell (the sort menu could never stay open) — memoised; rows are keyed by partition+offset; the Registry ignores column `size`, so the Value column takes the slack explicitly. "View details" carries the click instead of the whole banner.
- Both search boxes are the shared `misc/search-input`, now imported from `master` (it landed with #2633/#2636; this branch no longer carries a copy).
- `helper.tsx`: the connector-class popover opens on hover as before (`openOnHover`); the confirm dialog's description was a flex column that stacked its inline sentence; the trace dialog title checks `undefined`, not `null`; the trace box is a `SimpleCodeBlock`; `statusColors` covers STOPPED and uses the disabled fill token; the docs CTA is a Registry Button; `Statistic` → `Stat`, `Code` → `InlineCode`.
- Alerts use `AlertTitle` (five places). The Tasks tab sorts on `connectorName` — the accessor pointed at a field that does not exist. The plugin list and the three overview tabs paginate like the legacy DataTable (50 a page, pager only past that).

### Final review (2026-09-08)
- The connector-error dialog highlights its JSON again (`SyncCodeBlock`). Heading spacing uses padding — written while `mb-*` on bare headings was still inert; **#2636 has since landed that layer fix**, so the padding is now belt-and-braces rather than required.
- Connectors list and tasks table stay at ten a page (their legacy `defaultPageSize`), pager only past that. Tasks tab columns are module-level so sort menus survive re-renders; the logs Value column is not sortable.
- Not counted as coverage: `connector.spec.ts` is skipped upstream and its logs locator predates the Registry table. The error banner opens through "View details" only (was: anywhere on the banner).

### Table polish (2026-09-08)
- Hiding is off at table level (`TABLE_OPTIONS.enableHiding: false`) instead of per column, so a column added later cannot bring the header menu's "Hide" item back.

- The four new tooltips open at the app's 150 ms `delayDuration`, not Base UI's 600.
- The confirm dialog's pending button keeps its accessible name (`aria-label`) while it shows a spinner.
- Heading rungs match the Chakra sizes they replaced: the create heading and status card go back to `xl`, "Connector Properties" to `lg`.
- The validation warning keeps the `my-4` its sibling alerts have.
- The logs table shows its pager only when it has rows, as the legacy table did.
- The not-configured docs CTA is an anchor again (`buttonVariants` + `rounded-md` — radius lives on `Button`, so a bare `buttonVariants()` anchor renders square).

### DIFF
--- a/frontend/src/components/pages/connect/cluster-details.tsx
+++ b/frontend/src/components/pages/connect/cluster-details.tsx
@@ -9,8 +9,9 @@
  * by the Apache License, Version 2.0
  */
 
-import { Box, Button, DataTable, Text } from '@redpanda-data/ui';
 import { Link } from '@tanstack/react-router';
+import { buttonVariants } from 'components/redpanda-ui/components/button';
+import { DataTable, type DataTableColumnDef } from 'components/redpanda-ui/components/data-table';
 import { useCallback, useState } from 'react';
 
 import { ClusterStatisticsCard, ConnectorClass, NotConfigured, TaskState, TasksColumn } from './helper';
@@ -78,23 +79,7 @@ class KafkaClusterDetails extends PageComponent<{ clusterName: string }> {
             <h3 style={{ marginLeft: '0.25em', marginBottom: '0.6em' }}>Connector Types</h3>
 
             <DataTable<ClusterAdditionalInfo['plugins'][0]>
-              columns={[
-                {
-                  header: 'Class',
-                  accessorKey: 'class',
-                  cell: ({ row: { original } }) => <ConnectorClass observable={original} />,
-                  size: 500,
-                },
-                {
-                  header: 'Version',
-                  accessorKey: 'version',
-                  size: 300,
-                },
-                {
-                  header: 'Type',
-                  accessorKey: 'type',
-                },
-              ]}
+              columns={pluginColumns}
               data={additionalInfo?.plugins ?? []}
               pagination
               sorting
@@ -105,6 +90,24 @@ class KafkaClusterDetails extends PageComponent<{ clusterName: string }> {
     );
   }
 }
+
+const pluginColumns: DataTableColumnDef<ClusterAdditionalInfo['plugins'][0]>[] = [
+  {
+    header: 'Class',
+    accessorKey: 'class',
+    cell: ({ row: { original } }) => <ConnectorClass observable={original} />,
+    size: 500,
+  },
+  {
+    header: 'Version',
+    accessorKey: 'version',
+    size: 300,
+  },
+  {
+    header: 'Type',
+    accessorKey: 'type',
+  },
+];
 
 const ConnectorsList = ({ clusterName, connectors }: { clusterName: string; connectors: ClusterConnectorInfo[] }) => {
   const [filteredResults, setFilteredResults] = useState<ClusterConnectorInfo[]>([]);
@@ -133,13 +136,17 @@ const ConnectorsList = ({ clusterName, connectors }: { clusterName: string; conn
 
   return (
     <div>
-      <div style={{ display: 'flex', marginBottom: '.5em' }}>
-        <Link params={{ clusterName }} to="/connect-clusters/$clusterName/create-connector">
-          <Button variant="solid">Create connector</Button>
+      <div className="mb-2 flex">
+        <Link
+          className={buttonVariants({ variant: 'primary' })}
+          params={{ clusterName }}
+          to="/connect-clusters/$clusterName/create-connector"
+        >
+          Create connector
         </Link>
       </div>
 
-      <Box my={5}>
+      <div className="my-5">
         <SearchBar<ClusterConnectorInfo>
           dataSource={dataSource}
           filterText={searchText}
@@ -148,7 +155,7 @@ const ConnectorsList = ({ clusterName, connectors }: { clusterName: string; conn
           onQueryChanged={onQueryChanged}
           placeholderText="Enter search term/regex"
         />
-      </Box>
+      </div>
 
       <DataTable<ClusterConnectorInfo>
         columns={[
@@ -164,9 +171,7 @@ const ConnectorsList = ({ clusterName, connectors }: { clusterName: string; conn
                 search={{} as never}
                 to="/connect-clusters/$clusterName/$connector"
               >
-                <Text whiteSpace="break-spaces" wordBreak="break-word">
-                  {original.name}
-                </Text>
+                <span className="whitespace-break-spaces break-words">{original.name}</span>
               </Link>
             ),
             size: Number.POSITIVE_INFINITY,
@@ -194,9 +199,9 @@ const ConnectorsList = ({ clusterName, connectors }: { clusterName: string; conn
           },
         ]}
         data={filteredResults}
-        defaultPageSize={10}
         pagination
         sorting
+        tableOptions={{ initialState: { pagination: { pageIndex: 0, pageSize: 10 } } }}
       />
     </div>
   );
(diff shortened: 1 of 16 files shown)


eblairmckee (approved):
Reviewed locally https://redpanda.zoom.us/clips/share/oQzPu56HQSez5NAJsStSMQ
Noticed one issue when viewing a secret, will share a more detailed comment in a moment once claude finishes investigating

eblairmckee:
### Kafka Connect tab never loads on self-hosted (moved here from #2637)

Raised this on #2637 first, but `pages/connect/overview.tsx` is *this* PR's file, so the fix has a home here. It's not a regression from your work either way — the affected lines are unchanged from `master`; the only edit this PR makes to them is dropping `Code` from the `tsx-utils` import.

**Symptom.** Self-hosted Console, `/connect-clusters` → **Kafka Connect** tab renders a skeleton forever. No error, nothing in the console. Reproduced against a live Kafka Connect cluster (Redpanda + `connectors-unsupported`, 19 plugins) where `/api/kafka-connect/connectors` returns `200` with `isConfigured: true`.

**Cause.** `overview.tsx:129`:

```js
private async initializeData(): Promise<void> {
  try {
    await this.checkRPCNSecretEnable();   // L131 — POSTs SecretService/ListSecretScopes
    await this.refreshData();             // L132 — api.refreshConnectClusters()
  } catch {
    // Error during initialization - component will show error state
  }
}
```

`ListSecretScopes` returns **501** on the OSS backend (no `SecretService`), so `checkRPCNSecretEnable()` throws, `refreshData()` never runs, `api.connectConnectors` stays `undefined`, and `TabKafkaConnect` hits `if (!api.connectConnectors) return DefaultSkeleton` (L508). The bare `catch {}` swallows the 501 — and contrary to its comment there is no error state, which is why this is silent.

The tab stays *visible* because visibility comes from `useKafkaConnectConnectorsQuery()` (React Query, succeeds) while the content reads the legacy Zustand store. Same resource, two independent data paths — one populated, one not. The 501 just exposes the split.

**Why it's gone unnoticed:**
- Cloud never hits it — `render()` returns `<PipelineListPage />` early when `isEmbedded()`, so these tabs don't render. That's the same `isEmbedded()` branch this PR's verdict asks reviewers to sanity-check in Cloud.
- The `kafka-connect` e2e variant asserts tab *visibility*, then navigates to `/connect-clusters/:cluster` and `/create-connector` — routes that call `api.refreshConnectClusters()` in their own `initPage`. Nothing asserts the overview tab's list content, so the suite stays green.

**Minimal fix** — let the optional check fail on its own:

```js
async checkRPCNSecretEnable() {
  if (!Features.pipelinesApi) return;
  try {
    await rpcnSecretManagerApi.checkScope(create(ListSecretScopesRequestSchema));
  } catch {
    // Secret store is optional; a 501 here must not block the Kafka Connect fetch.
  }
}
```

The deeper fix is collapsing the two data paths onto the React Query hook — more in the spirit of this migration series, but clearly bigger.

Unaffected routes, if you want to exercise the rest of this PR meanwhile:

```
/connect-clusters/<cluster>
/connect-clusters/<cluster>/create-connector
/connect-clusters/<cluster>/<connector>
```

Happy for this to become its own issue/PR instead if you'd rather keep this one scoped to the Registry migration — it predates the branch, so there's no argument that it has to ship here.

PR #2641, opened 2026-09-09
UX-1469: fix breadcrumb navigation while topic messages are loading

jvorcak:
## Summary
- The current topic's own breadcrumb was a real link to its bare path (no search/hash), so clicking it dropped query params (filters, pagination, active tab). It now renders as non-navigable text, like a standard breadcrumb "current page" crumb.
- Clicking the parent "Topics" breadcrumb while a message search is still in flight could race a queued `nuqs` URL update: the update flushes ~50-90ms later using a now-stale relative route match, and TanStack Router silently interpolates the literal string `"undefined"` into the path, landing on `/topics/undefined` (404). The topic details route now guards against `topicName === 'undefined'` and redirects back to `/topics`.
- Hardened error-message handling in the topic message search (`Tab.Messages/index.tsx`) and the default route error boundary (`route-error.tsx`) so a non-`Error` rejection never renders the literal text `undefined` to the user.

## Test plan
- [x] `bun run type:check`
- [x] `bun run lint`
- [x] `bun run test:integration -- components/layout/header`
- [x] `bun run test:integration -- route-error`
- [x] `bun run test:integration -- topics/\$topicName`
- [x] `bun run test:integration -- Tab.Messages`

### DIFF
--- a/frontend/src/components/layout/header.test.tsx
+++ b/frontend/src/components/layout/header.test.tsx
@@ -13,12 +13,13 @@ import { describe, expect, it, rs } from '@rstest/core';
 import { render, screen } from '@testing-library/react';
 import type { PropsWithChildren, ReactNode } from 'react';
 
-const { getMatchedRoutes } = rs.hoisted(() => ({
+const { getMatchedRoutes, mockPageBreadcrumbs } = rs.hoisted(() => ({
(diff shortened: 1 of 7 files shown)


SpicyPete (approved):
I think this looks good, Claude is flagging some issues, but they are mostly around the old security page.  

Can the older [security page](https://github.com/redpanda-data/console/pull/2604) be merged soonish to simplify things?

PR #2646, opened 2026-09-10
feat(console): adopt react 19.3 transitions and fragment refs

malinskibeniamin:
## Summary
- Upgrade React and React DOM from locked 19.2.7 to 19.3.0, their types to 19.3.0, and React DOM's scheduler to 0.28.0. Bun and generated Yarn audit mirror updated together.
- Animate lazy editor/diff/YAML reveals through a shared, wrapper-free ViewTransition + Suspense boundary. Initial fallbacks and cached content remain immediate. Reduced-motion CSS disables both named and browser-root snapshot animations.
- Focus expanded message controls through a Fragment ref, without an extra DOM wrapper or stealing focus when selecting another message. Add accessible names and regression tests.

## Impact
Keyboard users land on the expanded message's controls instead of losing focus to the document. Browser checks confirm animated loading reveals normally, zero CSS transition animations under reduced motion, and working close/reopen behavior. No performance claim.

## React 19.3 feature coverage
| Feature | Decision |
| --- | --- |
| ViewTransition / Suspense reveals | Adopted in KowlEditor, KowlDiffEditor, PipelinesYamlEditor. |
| Fragment refs | Adopted for expanded message action focus. |
| addTransitionType | No directional/carousel interaction in this slice; do not introduce artificial transitions. |
| Image/font suspension | Available through the boundary; no new image/font loading infrastructure needed for these editors. |
| browser() / direct RSC Context | Not applicable to this client-rendered Console; no SSR/RSC introduced. |
| Trusted Types | Runtime support upgraded; HTML sinks are CLI-managed registry files and cannot be edited locally. No global CSP policy enabled. |
| Other fixes/events | Runtime fixes inherited. No actual Fullscreen API or Server Actions call sites to migrate; expanded panels use CSS, not browser fullscreen. |

## Dependency upgrade path
Stable minor: React/DOM 19.2.7 → 19.3.0; @types/react 19.2.17 → 19.3.0; @types/react-dom 19.2.3 → 19.3.0; scheduler 0.27.0 → 0.28.0.

Requested exact release is less than the repository's 72-hour cooling window: used a one-shot `--minimum-release-age=0` for the four requested React packages. Repository bunfig policy unchanged; lifecycle scripts disabled. Clean frozen install passed. No arbitrary coupled package churn; existing wrapper peers accept React 19. @monaco-editor/react 4.7.0 remains the latest stable release.

React 18 Cloud host isolation is unchanged (`shared: {}` and separate React roots).

Sources: https://react.dev/blog/2026/09/09/react-19-3, https://react.dev/reference/react/ViewTransition, https://react.dev/reference/react/Fragment

## Verification
- Passed: `bun run type:check`, `bun run build`, `bun run test:ci` (993 unit + 1,429 integration + 1 federation tests).
- RED→GREEN: expanded message focus; loading-boundary tests. Browser reduced-motion failure caught browser-root animations and fixed them.
- Passed: fresh `bun install --frozen-lockfile --ignore-scripts`; lockfile diff limited to React ecosystem.
- Passed: focused lint on changed files covered by the project's lint configuration.
- Global lint remains failing: 2,180 diagnostics versus 2,179 baseline; additional reported group is an unchanged deprecated Router useBlocker import, not a changed call site. `lint:fix` script does not exist, so ran the project's `bun run lint` equivalent.
- Audit unchanged: 62 advisories across 21 packages; BSR registry does not support npm audit.
- Real Chrome component consumer: loading/cached/reduced-motion/focus/close/reopen passed. Editor surfaces mount, accept input and remount, but diff disposal emits the same upstream error on both React 19.2.7 baseline and 19.3.0. https://github.com/suren-atoyan/monaco-react/issues/647
- Full-stack E2E blocked: backend image built, but Docker Desktop became unavailable during Redpanda startup. Playwright-managed Chromium download also failed (timeout/DNS); component checks used installed Chrome.

## Draft publication limitations
User approved draft publication after disclosure of incomplete full-stack dogfood, the known upstream diff error, and missing uploaded before/after screenshot evidence. These remain verification limitations, not passing checks. Local evidence and scripts are under `.context/react-evidence/` and `.context/react-*.log`; these are not reviewer-accessible URLs.

Cleanup caveat: test network `1b3544115c222b3a307a8ff082a3bfa22b993a6ac6af66750845d2c76d112c54` could not be checked/removed while Docker Desktop was unavailable. The interrupted test process is stopped.

## Rebase verification (September 17)
Rebased onto current `origin/master`; resolved only lockfile conflicts, preserving master dependency removals and React 19.3 versions. Frozen install, type-check, and all 4 focused loading-boundary/message-detail tests passed. Earlier full-suite/build/browser evidence above predates this rebase and was not rerun. `lint:fix` remains unavailable.

## Review follow-up
- Fixed expanded-message focus restoration across separate docked/sheet mounts; added Escape, Collapse, Close, and persisted-reopen coverage.
- Disabled document-root snapshot capture and overlay hit testing. Real Chromium tests hold editor animations active and click the background; reduced motion has zero animations. Run `bunx playwright test --config tests/browser/playwright.config.ts` from frontend.
- Verified the original selector syntax and keyed-remount regression test were valid; replied with spec and mutation evidence.
- Fixed CI test compatibility by replacing Promise.withResolvers with a standard Promise executor.
- Current verification: full test suite, focused tests after the CI compatibility change, type-check, browser CSS checks, and real-component Chrome smoke pass. Focused lint passes except pre-existing findings in topic-messages-view.tsx. Earlier full-stack/screenshot limitations remain.


frontend/src/components/ui/loading-boundary.css
```diff
@@ -0,0 +1,11 @@
+@media (prefers-reduced-motion: reduce) {
+  /* The browser also creates a root snapshot alongside React's named boundary. */
+  ::view-transition-group(root),
+  ::view-transition-old(root),
+  ::view-transition-new(root),
+  ::view-transition-group(.console-content-reveal),
```
malinskibeniamin, line 6: **Priority: P1**

**Reduced-motion opt-out never applies — invalid view-transition selector**

`::view-transition-group(.console-content-reveal)` is not valid CSS. The argument grammar for these pseudo-elements is `<pt-name-selector>` (`*` or a `<custom-ident>`) optionally followed by a class selector, so a view-transition *class* must be written as `*.console-content-reveal`. A bare `.console-content-reveal` fails to parse, and because an invalid selector invalidates the **entire selector list**, the `root` opt-outs on lines 3-5 are dropped too.

Consequence: under `prefers-reduced-motion: reduce` the whole rule is discarded, so users who asked for reduced motion still get the full-page cross-fade every time a lazy editor resolves — the exact case this file exists to suppress.

Correction:
```css
@media (prefers-reduced-motion: reduce) {
  ::view-transition-group(*.console-content-reveal),
  ::view-transition-old(*.console-content-reveal),
  ::view-transition-new(*.console-content-reveal) {
    animation: none;
  }
  ::view-transition-group(root),
  ::view-transition-old(root),
  ::view-transition-new(root) {
    animation: none;
  }
}
```
Splitting the class rule from the `root` rule also keeps one bad selector from taking the other down in future edits.

Verify: load a page with a Monaco editor in Chrome with reduced motion forced (DevTools → Rendering → Emulate CSS `prefers-reduced-motion: reduce`) and confirm no cross-fade; DevTools Styles should show the rule matching rather than greyed out as invalid.

*Automated `/review`.*

frontend/src/components/pages/topics/messages/detail/message-detail-panel.tsx
```diff
@@ -56,6 +56,38 @@ export const downloadRecord = (msg: TopicMessage) => {
 
 const SECTION_KEYS: readonly DetailSectionKey[] = ['metadata', 'key', 'headers', 'value'];
 
+function ExpandedMessageActions({ onCollapse, onClose }: { onCollapse: () => void; onClose: () => void }) {
+  const actionsRef = useRef<FragmentInstance>(null);
+  useEffect(function focusExpandedActions() {
+    actionsRef.current?.focus({ preventScroll: true });
```
malinskibeniamin, line 62: **Priority: P2**

**Focus is moved into the sheet but never restored when it unmounts**

`focusExpandedActions` pulls focus onto the collapse button on mount, but every exit path unmounts this subtree: Escape and the collapse button flip `expanded` to `false` (`topic-messages-view.tsx:529/551` swap presentations), and Close unmounts the panel entirely. Focus then falls back to `document.body`, so keyboard users lose their row position and restart the tab order from the top of the page. Because `detailExpanded` is restored from persisted state (`topic-messages-view.tsx:79`), simply clicking a row can also yank focus out of the table with no way back.

Correction: capture `document.activeElement` before focusing (or have the parent own the trigger element) and restore it in the effect cleanup, guarding that the saved element is still connected:
```tsx
useEffect(function focusExpandedActions() {
  const previous = document.activeElement as HTMLElement | null;
  actionsRef.current?.focus({ preventScroll: true });
  return () => {
    if (previous?.isConnected) previous.focus({ preventScroll: true });
  };
}, []);
```

Verify: extend the new test to assert focus destination after Escape, Collapse, and Close (render the parent so the docked panel/table remains mounted), then tab-navigate the messages page manually to confirm focus returns to the originating control.

*Automated `/review`.*

frontend/src/components/ui/loading-boundary.tsx
```diff
@@ -0,0 +1,11 @@
+import { type ReactNode, Suspense, ViewTransition } from 'react';
+import './loading-boundary.css';
+
+/** Reveal lazy content without animating its initial fallback or cached content. */
+export function LoadingBoundary({ children, fallback }: { children: ReactNode; fallback: ReactNode }) {
+  return (
+    <ViewTransition default="none" update="console-content-reveal">
```
malinskibeniamin, line 7: **Priority: P2**

**A page-wide view transition fires whenever any lazy editor resolves**

`<ViewTransition>` drives a *document-level* transition, so the browser snapshots and cross-fades the `root` group as well — the CSS comment on `loading-boundary.css:2` acknowledges this. Every Monaco/YAML editor reveal (`kowl-editor.tsx:65`, `kowl-editor.tsx:82`, `pipelines-yaml-editor.tsx:109`) therefore animates the entire viewport, not just the editor slot, and the page is non-interactive for the duration. Outside `prefers-reduced-motion` there is no opt-out at all, and with the selector defect above there is none inside it either.

Correction: scope the reveal so the root snapshot does not animate for all users — add an unconditional `::view-transition-group(root) { animation: none; }` (or `animation-duration: 0s`) alongside the named class, keeping the reduced-motion block only for the named group. Alternatively gate the wrapper behind a `matchMedia('(prefers-reduced-motion: reduce)')` check and render a plain `<Suspense>` when motion is unwanted.

Verify: open the topic message detail value section in Chrome, record a performance trace during editor load, and confirm no full-viewport cross-fade and no input delay while the transition runs.

*Automated `/review`.*

frontend/src/components/pages/topics/messages/detail/message-detail-panel.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    expanded: true,
+    onClose: rs.fn(),
+    onExpandedChange: rs.fn(),
+    loadLargeMessage: rs.fn().mockResolvedValue(undefined),
+  };
+  const { rerender } = render(<MessageDetailPanel {...props} />);
+  expect(screen.getByRole('button', { name: 'Collapse back to panel' })).toHaveFocus();
+
+  await user.tab();
+  const close = screen.getByRole('button', { name: 'Close' });
```
malinskibeniamin, line 36: **Priority: P3**

**The "does not steal focus when the message changes" assertion cannot fail**

`focusExpandedActions` has an empty dependency array and `ExpandedMessageActions` is not keyed by message, so a `rerender` with a new `offset` can never re-run the effect — this assertion is green regardless of the behaviour it claims to protect. The regression it is meant to guard (re-focusing when the selected record changes) would only appear if the subtree remounted, for example if the header were keyed on `msg.offset` or the panel were remounted by the parent.

Correction: drive the change through the real entrypoint instead — render the parent view (or toggle `expanded` off and on) and assert focus placement across that cycle, so a keyed remount or a dependency added to the effect is caught.

Verify: temporarily add `key={msg.offset}` to `ExpandedMessageActions` locally; the test should go RED. It currently stays green.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 4 finding(s).

eblairmckee (approved):
lgtm. will test more on integration when embedded in cloud.

frontend/src/components/ui/loading-boundary.css
malinskibeniamin: Checked against the CSS View Transitions Level 2 grammar (https://drafts.csswg.org/css-view-transitions-2/#pseudo-element-class-additions): a bare class selector is valid. The original reduced-motion case also passed in Chrome. Still made the wildcard explicit and separated root handling in e53edd638. New browser regression checks confirm zero animations under reduced motion.

frontend/src/components/pages/topics/messages/detail/message-detail-panel.tsx
malinskibeniamin: Fixed in fc1fe37ce. Cleanup restores the newly mounted docked Expand button, otherwise the still-connected previous control, with a parent-owned table focus fallback when the original trigger was removed. Added a controlled presentation harness that actually unmounts/remounts the two presentations and covers Escape, Collapse, Close, and reopening with persisted expansion. Integration and real-component Chrome checks pass.

frontend/src/components/ui/loading-boundary.tsx
malinskibeniamin: Fixed in e53edd638. Opted the document root out of snapshot capture and disabled the transition overlay hit testing; only the named editor participates. Merely removing root animations still blocked pointer input in the new browser test, so the fix removes root capture instead. Tests pause the live editor animation and successfully click a background control before it finishes; reduced-motion creates no animations. Also verified the actual React LoadingBoundary consumer in Chrome.

frontend/src/components/pages/topics/messages/detail/message-detail-panel.test.tsx
malinskibeniamin: Verified the proposed mutation on the original test: adding key={msg.offset} to ExpandedMessageActions makes the existing assertion fail at line 39, as intended. Removed the mutation. The test therefore already guards keyed remounts, contrary to this finding. fc1fe37ce additionally covers actual expanded/docked unmount-remount cycles and focus after Escape, Collapse, Close, and reopening.

PR #2647, opened 2026-09-10
[Chakra exit · Phase 2 · PR 15] frontend: the topics pages on Registry components, plus a settings-dialog spec

SpicyPete:
Takes `pages/topics` off Chakra — all nine remaining files. **Rebased onto `master`, which is at 25, so this takes it to 16.** With #2645 (rp-connect) and #2649 (the security ACLs tab) that leaves **five** — all of them PR 13's own, which are deleted rather than migrated.

**This is also what stops three finished areas painting Chakra at runtime**: `MessagePreview` renders in the logs tables of transforms, connector-details and pipelines-details, so until this merges those pages are clean by the import metric and still Chakra on screen.

Stacked conceptually after #2645 (rp-connect) but independent of it: disjoint files, opened off `master`, merges in either order.

## Why this is not the throwaway work the plan assumed

The tracker has said for a week that these nine "go with the legacy view once `enableNewTopicMessagesPage` flips". **A reverse-dependency check says otherwise.** `Tab.Messages` is not a self-contained legacy view — it is a shared library, with eight modules imported from outside the directory, including by the new messages page that replaces it.

**Five of the nine survive the flag flip, and three are rendered by page areas already marked done:**

| File | Reached from |
|---|---|
| `tab-docu.tsx` | `topic-details.tsx` — the Documentation tab, never behind the flag |
| `message-display/message-preview.tsx` | **transforms (PR 5), connector-details (PR 11), pipelines-details (PR 9)** |
| `common/empty-badge.tsx` | via `message-preview` |
| `common/empty-icon.tsx` | via `expanded-message → message-headers`, same three pages |
| `preview-settings.tsx` | the new page's `messages/table/message-cells.tsx` imports `getPreviewTags` |

So `MessagePreview` has been painting Chakra `<Flex>`/`<Text>` — with `red.600`, `blue.500`, `gray.500` — in the Value column of three logs tables since their own migrations merged. **The importer metric could not see it**: the imports live under `topics/`, so they were booked against the topics bucket while those pages counted as clean. Worth stating once: *"0 importers" is the right gate for removing the dependency, and the wrong one for "is this page migrated".*

The remaining four (`message-key-preview`, `modals/column-settings`, `modals/preview-fields-modal`, `preview-settings/pattern-help-drawer`) really are legacy-only. **I migrated them anyway** — PR 13 cannot merge while anything imports the package, and the flag flip has no date. That work is deliberately spent so the flip stops being a gate.

## Reviewer's guide

| Change | What to look at |
|---|---|
| Colour props → tokens | `red.600` → `text-destructive`, `blue.500` → `text-informative`, `gray.500` → `text-muted-foreground`. These were `color=` props on Chakra components, so they worked; in Registry they need semantic classes. The `RenderPreview` catch in both preview components also moves off a hard-coded `style={{ color: 'red' }}`, so both error paths in the same function now agree and follow the theme. |
| Icon sizes | `fontSize` was **inert** on these lucide icons, which size from `size` (default 24) — so master rendered them at 24 and they stay at 24. A swap to `size={16}` would have been a silent third smaller across four surfaces. |
| `Tooltip` → compound | `empty-icon` keeps its instant open — Chakra had `openDelay={1}`, and the Registry would otherwise inherit Base UI's 600 ms. |
| `opacity-[0.66]` | Not `opacity-66`. A bare number depends on v4's dynamic scale; an arbitrary value is guaranteed to emit, and a silently-inert opacity is the exact trap this series keeps hitting. |
| Two `Modal` → `Dialog` | Both keep the `PortalContainerProvider` from the Chakra original, so the nested select and popover still portal *inside* the dialog's focus and scroll lock rather than to the body. |
| `Drawer` → `Sheet` | **`container={document.body}`**, because `SheetContent` honours the enclosing dialog's `PortalContainerProvider` and `DialogContent` is transformed and `overflow-hidden` — it would be both the containing block and the clipping ancestor for a fixed panel. **The width is set explicitly**: Chakra's drawer scale maps `xl` to 56rem, the Sheet's to 36rem, so `size="full"` drops the variant cap and `sm:max-w-4xl` restores the original. `SheetContent` is not a flex column of its own, so it is made one with a scrolling body div and a `flex-row justify-end` footer. |
| `Checkbox` | **Chakra's took its label as a child and wired it; the Registry's does not.** Every box now has an explicit `id` + `Label`, or an `aria-label` where the original had no visible text at all — the pattern-enable box in `preview-settings` had none. |
| `Empty` | `Empty description={title}` → `Empty` + `EmptyHeader` + `EmptyTitle`. No width is added: the Registry root already ships `w-full flex-1 min-w-0` — what it lacks is a border *width*, which is why `border-dashed` paints nothing. |

## Removed rather than ported

- **A `Popover` with no trigger child** in `preview-settings` (`content={globHelp} trigger="click"`, self-closing). It rendered nothing clickable and had no way to open — the adjacent `PatternHelpDrawer` provides the real glob-pattern link. Porting it faithfully would have meant a Registry Popover with an empty trigger.
- **The dead `rs.mock('@redpanda-data/ui')` block** in `message-display.test.tsx`. With these components off Chakra it mocked nothing. Dropping it means those four tests now exercise the **real** components instead of plain `div`s — same four passing, strictly better coverage.

## Coverage

Four specs already drive the legacy messages view (filtering, production, timestamps, navigation), so `MessagePreview` and `MessageKeyPreview` have real regression cover. None of them opened the three settings surfaces, which are the riskiest part of this diff — two dialog conversions and a drawer.

**`topic-messages-settings-dialogs.spec.ts`** covers them: the column-settings dialog opens, its checkboxes resolve *by name* (the contract Chakra used to provide), it closes; and the preview-fields dialog opens the glob-pattern sheet nested inside it. **Verified green against a live container.** It caught one real thing while being written: `DialogContent` renders its own close X with `aria-label="Close"`, so a role-by-name query matches two buttons.

## Gates

`type:check` clean · per-file `lint:check` with the colour codes stripped, net-new zero (two files carry pre-existing diagnostics) · **993** unit · **1427** integration · the new e2e spec passes in the `console` variant.

## Flagged, not fixed

- **`preview-settings.tsx` is still imported by the new page for one function.** `getPreviewTags` is a pure helper sitting in a 543-line component module. Extracting it would let the new page stop importing a component module entirely; left alone here because the module is now Registry-based either way, so it is hygiene rather than a dependency problem.
- **The four legacy-only files should be deleted, not kept, when the flag flips.** They are migrated so they cannot block PR 13, not because they are worth keeping.

### DIFF
--- a/frontend/src/components/pages/topics/tab-docu.tsx
+++ b/frontend/src/components/pages/topics/tab-docu.tsx
@@ -14,7 +14,8 @@ import { docsLinks } from 'utils/docs-links';
 
 import type { Topic } from '../../../state/rest-interfaces';
 import '../../../utils/array-extensions';
-import { Button, Empty, VStack } from '@redpanda-data/ui';
+import { Button } from 'components/redpanda-ui/components/button';
+import { Empty, EmptyHeader, EmptyTitle } from 'components/redpanda-ui/components/empty';
 import { motion } from 'motion/react';
 import ReactMarkdown, { defaultUrlTransform as baseUriTransformer } from 'react-markdown';
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
@@ -159,13 +160,18 @@ const errorEmpty = renderDocuError(
 function renderDocuError(title: string, body: JSX.Element) {
   return (
     <motion.div {...animProps} key={'b'} style={{ margin: '2rem 1rem' }}>
-      <VStack gap={4}>
-        <Empty description={title} />
+      {/* Chakra's VStack centred its children; the Registry Empty root has no width of its own. */}
+      <div className="flex flex-col items-center gap-4">
+        <Empty className="w-full">
+          <EmptyHeader>
+            <EmptyTitle>{title}</EmptyTitle>
+          </EmptyHeader>
+        </Empty>
         {body}
         <a href={docsLinks.selfManaged.console} rel="noopener noreferrer" target="_blank">
-          <Button variant="solid">Redpanda Console Documentation</Button>
+          <Button>Redpanda Console Documentation</Button>
         </a>
-      </VStack>
+      </div>
     </motion.div>
   );
 }
--- a/frontend/src/components/pages/topics/tab-docu.tsx
+++ b/frontend/src/components/pages/topics/tab-docu.tsx
@@ -15,7 +15,13 @@ import { docsLinks } from 'utils/docs-links';
 import type { Topic } from '../../../state/rest-interfaces';
 import '../../../utils/array-extensions';
 import { Button } from 'components/redpanda-ui/components/button';
-import { Empty, EmptyHeader, EmptyTitle } from 'components/redpanda-ui/components/empty';
+import {
+  Empty,
+  EmptyContent,
+  EmptyDescription,
+  EmptyHeader,
+  EmptyTitle,
+} from 'components/redpanda-ui/components/empty';
 import { motion } from 'motion/react';
 import ReactMarkdown, { defaultUrlTransform as baseUriTransformer } from 'react-markdown';
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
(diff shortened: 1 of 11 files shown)


r-vasquez (approved):
🚀 

Tested locally and all loogs good.
Image

Image

frontend/src/components/pages/topics/tab-docu.tsx
```diff
@@ -159,13 +166,17 @@ const errorEmpty = renderDocuError(
...
+        <EmptyHeader>
+          <EmptyTitle>{title}</EmptyTitle>
+          <EmptyDescription>{body}</EmptyDescription>
+        </EmptyHeader>
+        <EmptyContent>
+          <a href={docsLinks.selfManaged.console} rel="noopener noreferrer" target="_blank">
+            <Button variant="primary">Redpanda Console Documentation</Button>
+          </a>
+        </EmptyContent>
+      </Empty>
```
eblairmckee, line 179: tab is disabled on all topics tested, but this LGTM
Image

eblairmckee (approved):

PR #2649, opened 2026-09-11
[Chakra exit · Phase 2 · PR 16] frontend: the security ACLs tab on Registry components, plus a smoke spec

SpicyPete:
Takes the **last Chakra page file** off the package: `security/tabs/acls-tab.tsx`. **`master` goes 25 → 24**, and once #2645 and #2647 merge that leaves **five** — all of them PR 13's own, which are deleted rather than migrated.

#2604 replaced the users, roles and permissions tabs with new designs and left ACLs on the old one. There is no `acls-tab-new` and the route is live at `/security/acls` (enterprise `users-authorization.spec.ts` navigates straight to it), so this is a parity swap, not throwaway work. **Whether ACLs also gets the new security design is a separate product call** — this only removes the dependency.

## Reviewer's guide

| Change | What to look at |
|---|---|
| `SearchField` → `SearchInput` | The shared one, so the ACLs filter gains `/`-to-focus and Escape-to-clear like every other list. |
| `DataTable` → the Registry's | Both sortable columns get a `DataTableColumnHeader`: the Registry sorts **only** through that component, where Chakra painted an affordance on every header, so a naive swap leaves sorting unreachable with nothing failing. The action column gets `enableSorting: false`, and the table passes `enableHiding: false` — there is no column-visibility toolbar to undo a hide with. |
| Pagination | Legacy parity: 50 a page (`DEFAULT_TABLE_PAGE_SIZE`) with the pager only past that, matching `transforms-list` and `connect/overview`. Chakra's own default was `defaultPageSize = 50` and it rendered the footer only when `data.length > defaultPageSize`. |
| The two `size:` props | Dropped. The Registry `DataTable` never reads `getSize()`, so `Number.POSITIVE_INFINITY` and `size: 60` were inert either way. |
| `AclRowActions` | The row menu moves into its own component that owns its queries, which leaves the columns array closing over nothing but a stable state setter — so it is built once. `DataTableColumnHeader` is a dropdown trigger, and a re-created header tears an open sort menu down mid-interaction. |

## Also in here

**The dead `rs.mock('@redpanda-data/ui')` in `roles-tab.test.tsx` is deleted** — 156 lines stubbing ~30 components for a file whose tree no longer touches the package, since `roles-tab.tsx` is now just `export { RolesTabNew as RolesTab }` and `roles-tab-new.tsx` has zero Chakra imports. Both suites stay at 993 / 1427 with no failures, so those tests now exercise the real components instead of stub `div`s. This also clears one of PR 13's eleven bare-grep mop-up files, leaving ten.

## Coverage

The ACLs table had no spec of its own — only an enterprise *authorization* test visits the route. **`acls-tab.spec.ts`** covers what the swap actually risked: the filter renders, both headers expose a sort trigger, and the trigger opens a menu rather than sorting on one click. Verified against a live container, and **verified to fail** when the `DataTableColumnHeader` is swapped back for a plain string header. A fresh cluster has no ACLs, so it asserts the chrome rather than rows.

## Gates

`type:check` clean · `lint:check` **0 diagnostics** on `acls-tab.tsx` (down from 2 on `master` — the Chakra import was itself a `noRestrictedImports` error, and the extraction removed an empty `onClick`), parsed with the colour codes stripped · **993** unit · **1427** integration, both identical to `master` · the new spec passes 2/2 in the `console` variant.

## Flagged, not fixed

- **`acl-delete-multi-match.spec.ts` is `test.skip` end to end**, both cases, pending UX-1216. A spec file is not evidence of coverage.
- **The three re-export shims** (`users-tab.tsx`, `roles-tab.tsx`, `permissions-list-tab.tsx`) now only forward to their `-new` counterparts. Collapsing them so the routes import `-new` directly is #2604 follow-up hygiene, not Chakra work.
- **`permissions-list-tab.test.tsx:121` still mocks `enableNewSecurityPage`**, a flag that no longer exists in source. Vestigial after #2604.


frontend/src/components/pages/security/tabs/roles-tab.test.tsx
```diff
@@ -31,163 +31,6 @@ const { historyPushMock, refreshRoleMembersMock, refreshRolesMock, deleteRoleMut
   deleteRoleMutationMock: rs.fn().mockResolvedValue(undefined),
 }));
 
-rs.mock('@redpanda-data/ui', () => {
```
eblairmckee, line 34: i presume chakra is gone from the roles-tab then?

eblairmckee: ah. it just [reexports](https://github.com/redpanda-data/console/blob/master/frontend/src/components/pages/security/tabs/roles-tab.tsx) `./roles-tab-new'` 😑

eblairmckee (approved):

